PR #7, opened 2026-01-29
Ensure DLC graph I/O name same as ONNX for QDQ models

qti-yuduo:
Fixed DLC I/O tensor name mismatch original ONNX model when offload_graph_io_quantization=1.

When offload_graph_io_quantization=1, Q/DQ nodes at graph boundaries are
filtered out, causing fused node I/O names to differ from original ONNX
graph I/O names. This change maps them back so DLC files use user-expected
tensor names.


qti-shubham (approved):

qti-yuduo:
CC @qti-chuteng -- hope you don't mind I'll notify you every functional (non-infra) change so that we can port during the migration/rewrite.

PR #16, opened 2026-01-30
[AISW-166656] C++ model test runner uses plugin EP

qti-jkilpatrick:
### Description

* Cherry pick https://github.com/microsoft/onnxruntime/pull/25942 onto main.
* Fix segfaults and disable other failing tests.

### Motivation and Context

While it is already tested by unit tests, this adds additional coverage to the new ABI-stable EP plugin path. Specifically, the same model tests that have been run upstream will be run on Qualcomm devices in CI.


cmake/CMakeLists.txt
qti-jkilpatrick, line 1: It might be easiest to review this change one commit at a time. The first one is just a cherry-pick of https://github.com/microsoft/onnxruntime/pull/25942. It required a little rebasing and I fixed one obvious typo, but it's 99.9% the original author's work.

The second change is all mine and probably where your attention should be focused.

qti-kromero (approved):

PR #22, opened 2026-02-02
Introduce ORT Home in CI and Build System

qti-hungjuiw:
### Description
**New ORT_HOME Build System Features:**
- `onnxruntime_ORT_HOME` cmake build system
- `--ort_home` parameter in `build_args.py` 
- Flexible ORT prebuilt selection mechanism
- Enhanced build configuration options across platforms

### Motivation
**Build Flexibility and Complete Decoupling**
- **Complete decoupling**: QNN EP development becomes completely independent of specific ORT versions
    - **Flexibility to build against any ORT release**: Developers can specify any ORT version using `--ort_home` parameter
    - **Flexibility to build against any QAIRT release**: Independent selection of QAIRT versions without ORT constraints
- **Version Matrix Testing**: Ability to test QNN EP against multiple ORT/QAIRT combinations

**Developer Productivity and Workflow Enhancement**
- **Rapid Prototyping**: Quickly test against different ORT versions without rebuilding entire toolchain
- **Version Compatibility Testing**: Easy validation against multiple ORT releases for compatibility assurance

### DIFF
--- a/qcom/build_and_test.py
+++ b/qcom/build_and_test.py
@@ -58,6 +58,7 @@ from ep_build.util import (
 )
 
 DOCKER_CCACHE_ROOT_ENV_VAR = "ORT_BUILD_DOCKER_CCACHE_ROOT"
+ORT_PREBUILT_ROOT_ENV_VAR = "ORT_PREBUILT_ROOT"
 QAIRT_SDK_ROOT_ENV_VAR = "QAIRT_SDK_ROOT"
 QNN_SDK_ROOT_ENV_VAR = "QNN_SDK_ROOT"
 SNPE_ROOT_ENV_VAR = "SNPE_ROOT"
@@ -93,6 +94,9 @@ Environment variables
 
   ORT_NIGHTLY_BUILD
     If set to 1, instruct the ORT build to use a more verbose version string.
+
+  ORT_PREBUILT_ROOT
+    If specified, this will be used for the path of ORT prebuilt.
 
   QAIRT_SDK_ROOT
   QNN_SDK_ROOT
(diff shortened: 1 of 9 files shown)


qcom/scripts/windows/build.ps1
```diff
@@ -71,6 +75,22 @@ if (-not (Test-Path $BuildDir)) {
...
+if ($OrtPrebuiltRoot -eq "") {
+    if ($Arch -eq "x86_64") {
+        $OrtPrebuiltRoot = (Get-OrtX64PrebuiltRoot)
+    }
+    elseif ($Arch -eq "ARM64") {
+        $OrtPrebuiltRoot = (Get-OrtARM64PrebuiltRoot)
+    }
+    else {
+        # TODO: Consider Prebuilt for other Arch
+        $OrtPrebuiltRoot = (Get-OrtARM64PrebuiltRoot)
```
qti-kromero: should this throw an error or ortherwise not set this variable for other architectures? this will download the ARM64 prebuilt even if it might not be necessary for the build?

qcom/build_and_test.py
```diff
@@ -1012,6 +1046,22 @@ def get_docker_ccache_root(root_from_args: Path | None) -> Path | None:
...
 
 
+def get_ort_prebuilt_root(root_from_args: Path | None) -> Path | None:
+    ort_prebuilt: Path | None = None
+    if root_from_args is not None:
+        ort_prebuilt = root_from_args
+    elif ORT_PREBUILT_ROOT_ENV_VAR in os.environ:
+        ort_prebuilt = Path(os.environ[ORT_PREBUILT_ROOT_ENV_VAR])
+    else:
+        # Let the build fetch QAIRT
```
qti-kromero: QAIRT or ORT?

also, why is this fetching ORT? in my mind if prebuilt is not specified, then we should be building some onnxruntime that is checked in as a submodule on this repo (default behavior) IMO

maybe we need to go through the usecases for this in some kind of document?

qti-hungjuiw: Thanks for pointing this out — you're right, it should be `"ORT"` instead of `"QAIRT"`.

The `get_ort_prebuilt_root` function is intended to mirror the behavior of `get_qairt_sdk_root`. In our current scenario, the goal is to support the combination of Microsoft's ORT Core release with Qualcomm’s QNN EP plugin libraries, so we only download the official release version.

Supporting ORT Core built as a `submodule` or via `ExternalProject` is certainly a valid use case as well, but I’d prefer to address that in a separate PR.

qcom/scripts/windows/build.ps1
qti-hungjuiw: Thanks, that’s a good point. I’ve updated the change to avoid setting the variable and emit a warning. Please let me know if this is feasible.

PR #25, opened 2026-02-02
 Preserve ONNX I/O order in QNN graph (DLC) serialization

qti-yuduo:
Fix QNN graph input/output ordering to match ONNX declaration order.

Problem: When ONNX Runtime partitions a graph and creates a fused node for the QNN EP, the partitioning process can **reorder the inputs and outputs** relative to the original ONNX model declaration order. The QNN backend (specifically DLC serialization) determines tensor ordering by the order in which `tensorCreateGraphTensor()` is called. If the fused node's I/O order differs from the original ONNX model, the compiled QNN graph will have mismatched I/O positions — leading to **silent data corruption** at inference time (wrong tensors bound to wrong positions).

Solution:
- Capture ONNX I/O order in GetCapability (before graph partitioning)
- Pass order to ComposeGraph via ComposeGraphContext
- Register graph I/O with QNN in ONNX order before processing ops
- Add RegisterGraphInputOutputInOrder() to pre-register tensors

### DIFF
--- a/onnxruntime/core/providers/qnn-abi/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn-abi/builder/qnn_model.cc
@@ -22,6 +22,44 @@
 namespace onnxruntime {
 namespace qnn {
 
+namespace {
+
+// Resolves the correct I/O order, preferring ONNX declaration order when available.
+// When names don't match directly (e.g., with offload_graph_io_quantization), uses
+// tensor_name_overrides_reversed to translate ONNX names to internal tensor names.
+std::vector<std::string> ResolveGraphInputOutputOrder(
+    const std::vector<std::string>* onnx_names,
+    const std::unordered_set<std::string>& fused_names,
+    const std::vector<std::string>& fused_order,
+    const std::unordered_map<std::string, std::string>& tensor_name_overrides_reversed) {
+  if (!onnx_names) {
+    return fused_order;
+  }
+
+  bool names_match = std::all_of(onnx_names->begin(), onnx_names->end(),
+                                 [&](const std::string& n) { return fused_names.count(n) > 0; });
+  if (names_match) {
+    return *onnx_names;
+  }
+
+  if (!tensor_name_overrides_reversed.empty()) {
+    std::vector<std::string> mapped_order;
+    for (const auto& onnx_name : *onnx_names) {
+      auto it = tensor_name_overrides_reversed.find(onnx_name);
+      if (it != tensor_name_overrides_reversed.end() && fused_names.count(it->second)) {
+        mapped_order.push_back(it->second);
+      }
+    }
+    if (mapped_order.size() == onnx_names->size()) {
+      return mapped_order;
+    }
+  }
+
+  return fused_order;
+}
+
+}  // namespace
+
 bool QnnModel::GetGraphInfoFromModel(QnnModelWrapper& model_wrapper, const Ort::Logger& /* logger */) {
   bool rt = true;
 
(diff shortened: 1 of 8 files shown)


qti-yuduo:
@minfhong-qti @qti-chuteng would be great if you could also review this one. Thank you!!

onnxruntime/core/providers/qnn-abi/builder/qnn_model.cc
```diff
@@ -34,53 +72,110 @@ bool QnnModel::GetGraphInfoFromModel(QnnModelWrapper& model_wrapper, const Ort::
...
+    ORT_CXX_LOG(context.logger,
+                ORT_LOGGING_LEVEL_VERBOSE,
+                ("input " + std::to_string(idx) + " " + name).c_str());
+    graph_inputs_.indices.emplace(name, idx);
+    graph_inputs_.names.push_back(name);
+  }
+
+  for (size_t i = 0; i < input_defs.size(); ++i) {
+    const char* name = nullptr;
+    ORT_CXX_RETURN_ON_API_FAIL(api_ptrs_.ort_api.GetValueInfoName(input_defs[i], &name));
```
minfhong-qti: After my implementation using all these C APIs to get desire attributes, MSFT introduces more convenient C++  APIs to achieve the same thing. You may refer to `onnxruntime_cxx_api.h` and `onnxruntime_cxx_inline.h` and search for `Ort::ConstValueinfo`, `Ort::ConstTypeInfo`, `Ort::ConstTensorTypeAndShapeInfo`, etc. The code will be a lot cleaner with these APIs.

onnxruntime/core/providers/qnn-abi/builder/qnn_model_wrapper.h
```diff
@@ -111,15 +112,13 @@ class QnnModelWrapper {
 
   Qnn_ContextHandle_t GetQnnGraphContext() const { return graph_context_; }
 
-  // Move input tensor wrappers to GraphInfo, QnnModelWrapper end of live
   std::vector<QnnTensorWrapper>&& GetGraphInputTensorWrappers() {
-    GetGraphInputOutputTensorWrapper(model_input_names_, model_input_tensor_wrappers_);
+    GetGraphInputOutputTensorWrapper(SortNamesByIndices(input_index_map_), model_input_tensor_wrappers_);
```
minfhong-qti: Since `input_index_map_`/`output_index_map_` are passed in during construction as `unordered_map`, can they be constructed as ordered `map` at the beginning (probably in `qnn_execution_provider.cc`) and thus there is no need to sort it every time accessing them?

onnxruntime/core/providers/qnn-abi/qnn_execution_provider.cc
```diff
@@ -1340,6 +1340,54 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
...
+  // Store original graph input order for use in Compile.
+  // The partitioning process may reorder inputs, so we capture the original order here.
+  {
+    std::basic_string<ORTCHAR_T> model_path = GetModelPathString(graph, ep->ort_api);
+
+    std::lock_guard<std::mutex> lock(ep->onnx_graph_names_mutex_);
+    if (ep->onnx_graph_input_names_.find(model_path) == ep->onnx_graph_input_names_.end()) {
+      std::vector<std::string> input_order;
+      size_t num_inputs = 0;
+      if (ep->ort_api.Graph_GetNumInputs(graph, &num_inputs) == nullptr) {
```
minfhong-qti: Same for here. Use C++ APIs instead.

onnxruntime/core/providers/qnn-abi/builder/qnn_model.h
```diff
@@ -21,6 +21,37 @@ struct QnnTensorInfo {
...
+};
+
+// Configuration for QnnModel::ComposeGraph and QnnModel::SetGraphInputOutputInfo.
+struct QnnModelContext {
+  const OrtGraph& ort_graph;
+  const OrtNode& fused_node;
+  const Ort::Logger& logger;
+
+  // Names in ONNX declaration order, absent when loading from cached context.
+  const std::vector<std::string>* onnx_input_names = nullptr;
```
minfhong-qti, line 31: Could use `std::optional` instead of by pointer?

onnxruntime/core/providers/qnn-abi/builder/qnn_model_wrapper.h
qti-yuduo: We'll need sort by value not key -- map here not helpful. But good point on the cost saving -- I updated to avoid sorting during access.

onnxruntime/core/providers/qnn-abi/builder/qnn_model.h
qti-yuduo: `std::optional` would copy the vector data unnecessarily -- borrow semantics sounds more natural fit?

onnxruntime/core/providers/qnn-abi/builder/qnn_model.cc
```diff
@@ -34,74 +72,100 @@ bool QnnModel::GetGraphInfoFromModel(QnnModelWrapper& model_wrapper, const Ort::
...
+  std::vector<Ort::ConstValueInfo> input_defs = fused_node.GetInputs();
 
-  return Ort::Status();
-}
+  // Collect non-initializer inputs
+  std::unordered_set<std::string> fused_input_names;
+  std::vector<std::string> fused_input_order;
+  for (const auto& input : input_defs) {
+    std::string name = input.GetName();
+    if (!IsConstantInitializer(ort_graph, name)) {
```
minfhong-qti, line 93:
```suggestion
    if (!input.IsConstantInitializer()) {
```
Nit. And as well for other similar ones.

minfhong-qti (commented):
LGTM. Minor comments.

onnxruntime/core/providers/qnn-abi/builder/qnn_model.cc
qti-yuduo: We need query out_graph inputs here, the fused_node input (if using `input.` isn't correct).

onnxruntime/core/providers/qnn-abi/builder/qnn_model.cc
```diff
@@ -22,6 +22,44 @@
...
+      if (it != tensor_name_overrides_reversed.end() && fused_names.count(it->second)) {
+        mapped_order.push_back(it->second);
+      }
+    }
+    if (mapped_order.size() == onnx_names->size()) {
+      return mapped_order;
+    }
+  }
+
+  return fused_order;
```
qti-shubham: Yuduo, just a thought:
Let say any ONNX-declared I/O name is absent from the fused node then as per line 53, fused_order is returned.

eg: onnx_names = {i1, i2, i3} and mapped_order={i3, i1}, then we will return {i3, i1}.
shouldn't we align fused_order to matching onnx_names order and return {i1, i3} ?

qti-shubham (approved):

onnxruntime/core/providers/qnn-abi/builder/qnn_model.cc
qti-yuduo: Your confusion is valid... I think it's because of the default return.. lemme refactor to something less confusing. Essentially we have 2 paths: 1) fresh compilation so that we have ONNX graph, we align with original orders 2) cached, no onnx_names -- just use fused_nodes.

qti-yuduo: Updated caller to be more explicit. Also fail on mismatch.

PR #29, opened 2026-02-03
[QNN EP] ORT ABI library Zip Archive release

qti-chuteng:
- Add zip build pipeline


.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -147,6 +150,12 @@ jobs:
           retention-days: 1
           path: |
             ${{ github.workspace }}/build/${{inputs.target_os}}-${{ inputs.target_arch }}/${{ env.BUILD_CONFIG_DIR }}/dist/onnxruntime_qnn*.whl
+      - uses: actions/upload-artifact@v4
+        if: ${{ inputs.upload_zip }}
+        with:
+          name: ORT ${{inputs.target_os}} ${{ inputs.target_arch }} Python ${{ inputs.target_py_vsn }} Zip Asset
+          path: |
```
qti-kromero, line 171: can we add "retention-days: 1" here as above? or otherwise somehow limit it?

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -41,6 +41,9 @@ on:
       upload_wheel:
         default: false
         type: boolean
+      upload_zip:
+        default: true
```
qti-kromero: can we have this false by default? we've turned off all other uploading due to issues with artifact space limits

long-term I think this should still generally be false unless there is utility in having this zip on every single PR? I had envisioned a new workflow invoking this workflow which has the proper configuration settings for a release - perhaps this is still coming

onnxruntime/test/onnx/TestCase.cc
```diff
@@ -982,6 +982,7 @@ std::unique_ptr<std::set<BrokenTest>> GetBrokenTests(const std::string& provider
       {"mxnet_arcface", "Model is an invalid ONNX model"},
       {"unique_not_sorted_without_axis", "Expected data for 'Y' is incorrect and in sorted order."},
       {"cumsum_1d_reverse_exclusive", "only failing linux GPU CI. Likely build error."},
+      {"resize_downsample_scales_cubic", "results mismatch with onnx tests"},
```
qti-kromero: interesting - where did this come from/what is justification for this? related to issues we are seeing in the chat

tools/ci_build/build.py
```diff
@@ -30,6 +30,7 @@ def version_to_tuple(version: str) -> tuple:
 sys.path.insert(0, os.path.join(REPO_DIR, "tools", "python"))
 import util.android as android  # noqa: E402
 from build_args import parse_arguments  # noqa: E402
+from pkg_assets import build_zip_asset  # noqa: E402
```
qti-kromero, line 36: why is this something we are doing only from Windows? wouldn't we want to also do this on Linux?

I also feel like we probably want this functionality integrated into the wider build_and_test.py system with some kind of task. as-is we would have no easy way of reproducing this zip functionality if we turn args.build_zip_asset to false

thinking we should just call pkg_assets.py inside of a build task there? then we wouldn't have to modify the onnxruntime build files at all

tools/ci_build/pkg_assets.py
```diff
@@ -0,0 +1,269 @@
...
+import argparse
+import logging
+import os
+import platform
+import re
+import sys
+import zipfile
+from pathlib import Path
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "python"))
```
qti-kromero, line 14: this is a bit hacky and not sure it's the best way to get this file imported

tools/ci_build/pkg_assets.py
```diff
@@ -0,0 +1,269 @@
...
+                return version_match.group(1)
+            else:
+                log.warning(f"Could not parse version from VERSION_NUMBER: {version_line}")
+                return None
+    except (OSError, FileNotFoundError) as e:
+        log.warning(f"Could not read VERSION_NUMBER file: {e}")
+        return None
+
+
+def get_qnn_asset_file_list(is_windows_platform=None):
```
qti-kromero: this function feels like it could be simplified to two dictionaries that are stored in this file

tools/ci_build/pkg_assets.py
```diff
@@ -0,0 +1,269 @@
...
+
+        # Parse version from VERSION_NUMBER file
+        version = parse_version_number(source_dir)
+
+        zip_name = f"onnxruntime-qnn-{platform_name}-{arch}"
+        if zip_name_suffix:
+            zip_name += f"-{zip_name_suffix}"
+        if version:
+            zip_name += f"-{version}"
+        zip_name += f"-{config}.zip"
```
qti-kromero, line 152: does this zip naming align with what MSFT was previously doing on their repository?

tools/ci_build/pkg_assets.py
```diff
@@ -0,0 +1,269 @@
...
+            log.warning("Continuing with available files...")
+
+        if not found_files:
+            raise FileNotFoundError(f"No asset files found in {cwd}. Expected files: {asset_files}")
+
+        # Create zip file
+        log.info(f"Creating zip: {zip_path}")
+        with zipfile.ZipFile(zip_path, "w", zipfile.ZIP_DEFLATED) as zipf:
+            for filename, file_path in found_files:
+                if "onnxruntime_providers_qnn_abi" in filename:
```
qti-kromero, line 186: nit: maybe we flag this as something to cleanup later with either a TODO or just a comment

tools/ci_build/pkg_assets.py
qti-chuteng: We will keep the config suffix in our internal repo, but we will remove that suffix before we release it

onnxruntime/test/onnx/TestCase.cc
qti-chuteng: removed

tools/ci_build/pkg_assets.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
qti-kromero: Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
SPDX-License-Identifier: MIT

confirmed with legal today - should have above copyright on new files created by QCOM

kuanyul-qti: Done

tools/ci_build/build.py
kuanyul-qti: We can also do this on Linux. This is not a platform-specific file. We'll integrate it into build.sh when we have ABI support on Linux.

It has been integrated into build_and_test.py from the change https://github.com/onnxruntime/onnxruntime-qnn/pull/29/changes#diff-62adaf109c822186c1181ac81a7dc1dce2800b1c8d8b0eee625a0cad9d98fd0b. You can directly call build_and_test.py to reproduce the zip functionality, e.g., `python qcom/build_and_test.py --target-py-version 3.12 build_ort_windows_arm64`. The command will be updated to `python qcom/build_and_test.py --target-py-version 3.12 --build-zip build_ort_windows_arm64` in the deployment PR.

PR #32, opened 2026-02-03
Op Package registration changes for GPU Backend

qti-mattsinc:
WIP - do not review yet

### Description
Making some changes to custom OpPkg registration to support a GPU OpPkg for GenAI models. See commit descriptions.

### Motivation and Context
Some contrib ops seen in GenAI models will be handled through a contrib-op OpPkg when using the GPU backend.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -36,6 +36,7 @@
 #include "core/providers/qnn/builder/qnn_node_group/qnn_node_group.h"
 #include "core/providers/qnn/builder/qnn_profile_serializer.h"
 #include "core/providers/qnn/ort_api.h"
+#include "core/providers/qnn/qnn_ep_utils.h"
 
 #ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
 #include "core/providers/qnn/builder/qnn_file_mapping_interface.h"
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -12,6 +12,7 @@
 #include <dlfcn.h>
 #endif
 
+#include <filesystem>
 #include <functional>
 #include <memory>
 #include <mutex>
@@ -405,6 +406,10 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
                                       Qnn_Version_t req_version,
                                       T** interface_provider);
 
+  Ort::Status GetQnnOpPackageInfo(const char* lib_path,
+                                  const char* interface_name,
+                                  const QnnOpPackage_Info_t** op_package_info);
+
   bool IsDevicePropertySupported();
 
   std::string GetBackendBuildId() {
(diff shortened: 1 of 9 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -394,80 +399,115 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
       std::string suffix = "InterfaceProvider";
       if (op_package_for_registration.size() >= suffix.size() &&
           op_package_for_registration.compare(op_package_for_registration.size() - suffix.size(), suffix.size(), suffix) == 0) {
         op_package_for_registration.erase(op_package_for_registration.size() - suffix.size());
       }
+
       registerUDO(op_package.op_type, op_package_for_registration);
     }
 
+    if (GetQnnBackendType() == QnnBackendType::GPU) {
```
qti-mattsinc: nit.
```suggestion
    if (IsGpuBackend(GetQnnBackendType())) {
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -394,80 +399,115 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+      // Load the GPU OpPackage for extended ONNX support, and register all its ops as UDOs to QNN EP.
+      const auto op_package_path = std::filesystem::path(GetBackendLibDir()) /
+                                   onnxruntime::utils::MakeSharedLibraryPath("QnnGpuOpPkgOnnx");
+
+      const QnnOpPackage_Info_t* op_package_info{nullptr};
+      if (RegisterOpPackage(op_package_path.string(), "QnnOpPackage_interfaceProvider", "", op_package_info).IsOK() &&
+          op_package_info != nullptr) {
+        for (size_t i = 0; i < op_package_info->numOperations; i++) {
+          registerUDO(op_package_info->operationNames[i], op_package_info->packageName);
```
qti-mattsinc, line 554: note:
`registerUDO` makes use of the `UDOQDQFusion` class when handling UDOs. This was evidently done with HTP QDQ use-cases in mind, but works as-is for the new GPU UDO use-case (unrelated to QDQ). This could be a bit confusing for anyone poking into verbose logs, as they might see lines like `Validation PASSED for 1 nodes in UDOQDQFusion` in a graph without QDQ patterns.

In a follow up PR, we can decouple QDQ from the UDO handling so that logs/class names/etc are more accurate for all cases.

qti-chuteng:
**[M-1]** `onnxruntime/test/providers/qnn/gqa_test.cc:1`

New file uses Microsoft copyright. QNN EP convention requires Qualcomm copyright on every new file.

```cpp
```

---

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:280`

`GetQnnOpPackageInfo` calls `LoadLib(lib_path, ...)` but never `UnloadLib`s the handle, and the handle is not returned to the caller (unlike `GetQnnInterfaceProvider`). Every successful call leaks one shared-library handle for the rest of the process lifetime. Either return the handle through an out-param and track it in a `QnnBackendManager` member (similar to `backend_lib_handle_`), or document and unload in `~QnnBackendManager` / `ReleaseResources`.

---

**[M-3]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h:556`

```cpp
} else if constexpr (QNN_API_VERSION_MAJOR >= 2 && QNN_API_VERSION_MINOR >= 34) {
```

Field-independent AND compare. When `QNN_API_VERSION_MAJOR` bumps to 3 with `MINOR=0`, the warning silently disappears. Use proper lexicographic semver:

```cpp
} else if constexpr ((QNN_API_VERSION_MAJOR > 2) ||
                     (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 34)) {
```

The same idiom already lives in this file at line 43-44 (`static_assert`).

---

## Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/udo_fusion.cc:131-141, 172-178`

The "missing optional → emit QNN null tensor" block is duplicated across the input and output loops; error messages are already inconsistent. Extract a file-static helper `AppendNullTensor(...)`.

---

**[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/udo_fusion.cc:88-89`

Allman bracing inconsistent with the rest of the file:

```cpp
if (output_value_info != nullptr)
{    // ← move to previous line
```

---

**[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h:475-507`

The `RegisterOpPackage` lambda demotes every QNN error code from ERROR to WARNING. This is correct for the new GPU fallback path, but the same lambda also runs for user-supplied `op_packages_` — meaning a misconfigured user UDO now logs only at WARNING. Suggest splitting the two paths or threading a `bool optional_failure` flag.

---

**[N-4]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h:509-512`

Inside the `RegisterOpPackage` lambda, if `GetQnnOpPackageInfo(...)` fails after `backendRegisterOpPackage(...)` succeeded, the package is already registered with the backend. The lambda returns an error, callers think nothing happened, and a retry will see `QNN_BACKEND_ERROR_OP_PACKAGE_DUPLICATE`. Either unregister on failure or comment the partial-success behavior.

---

**[N-5]** `onnxruntime/test/providers/qnn/gqa_test.cc`

281 lines of helpers plumb 12 inputs and 11 attributes, but only one test runs (`GroupQueryAttention_Basic_FP32`, fixed config). Please add at least: `do_rotary=0` case, prompt-only (no past_key) case, and one negative case asserting `ExpectedEPNodeAssignment::None` when GPU is unavailable.

PR #33, opened 2026-02-03
Fix Cumsum GetOnnxAxis size validation bug. 

quic-muchhsu:
### Description

`axis should be a single element` check has bug when initializer is int64 which has 2*uint8 size.
Update rule when initalizer is int 64.

### DIFF
--- a/onnxruntime/core/providers/qnn-abi/builder/opbuilder/cumsum_op_builder.cc
+++ b/onnxruntime/core/providers/qnn-abi/builder/opbuilder/cumsum_op_builder.cc
@@ -17,7 +17,7 @@ Ort::Status GetOnnxAxis(QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnit& n
   RETURN_IF_NOT(axis_input_info.is_initializer, "axis must be initializers");
   std::vector<uint8_t> axis_unpacked_tensor;
   RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(axis_input_info.initializer_tensor, axis_unpacked_tensor));
-  RETURN_IF_NOT((axis_input_info.qnn_data_type == QNN_DATATYPE_INT_64 ? 2 : 1) == static_cast<uint32_t>(axis_unpacked_tensor.size() / sizeof(axis_input_info.qnn_data_type)),
+  RETURN_IF_NOT((axis_input_info.qnn_data_type == QNN_DATATYPE_INT_64 ? 2u : 1u) == static_cast<uint32_t>(axis_unpacked_tensor.size() / sizeof(axis_input_info.qnn_data_type)),
                 "axis should be a single element");
 
   int32_t axis = 0;
--- a/onnxruntime/core/providers/qnn-abi/builder/opbuilder/cumsum_op_builder.cc
+++ b/onnxruntime/core/providers/qnn-abi/builder/opbuilder/cumsum_op_builder.cc
@@ -17,7 +17,7 @@ Ort::Status GetOnnxAxis(QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnit& n
   RETURN_IF_NOT(axis_input_info.is_initializer, "axis must be initializers");
   std::vector<uint8_t> axis_unpacked_tensor;
   RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(axis_input_info.initializer_tensor, axis_unpacked_tensor));
-  RETURN_IF_NOT(1 == static_cast<uint32_t>(axis_unpacked_tensor.size() / sizeof(axis_input_info.qnn_data_type)),
+  RETURN_IF_NOT((axis_input_info.qnn_data_type == QNN_DATATYPE_INT_64 ? 2 : 1) == static_cast<uint32_t>(axis_unpacked_tensor.size() / sizeof(axis_input_info.qnn_data_type)),
                 "axis should be a single element");
 
   int32_t axis = 0;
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/qnn-abi/builder/opbuilder/cumsum_op_builder.cc
```diff
@@ -17,7 +17,8 @@ Ort::Status GetOnnxAxis(QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnit& n
   RETURN_IF_NOT(axis_input_info.is_initializer, "axis must be initializers");
   std::vector<uint8_t> axis_unpacked_tensor;
   RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(axis_input_info.initializer_tensor, axis_unpacked_tensor));
-  RETURN_IF_NOT(1 == static_cast<uint32_t>(axis_unpacked_tensor.size() / sizeof(axis_input_info.qnn_data_type)),
+  // Int 64 has double size.
+  RETURN_IF_NOT((axis_input_info.qnn_data_type == QNN_DATATYPE_INT_64 ? 2 : 1) == static_cast<uint32_t>(axis_unpacked_tensor.size() / sizeof(axis_input_info.qnn_data_type)),
```
qti-yuduo: Adding a unit test with `int64_t` would be great.

quic-muchhsu: Yes. Adding it.

quic-muchhsu: Added int 64 axis test and enabled few originally disabled test. Looks like it was failed because the default soc_model doesn't support fp Cumsum kernel. Use soc_model=87 for testing, and it passed.

qti-yuduo: CI failures looks related? Likely you need to set SoC just for this test.

onnxruntime/core/providers/qnn-abi/builder/opbuilder/cumsum_op_builder.cc
```diff
@@ -17,7 +17,8 @@ Ort::Status GetOnnxAxis(QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnit& n
   RETURN_IF_NOT(axis_input_info.is_initializer, "axis must be initializers");
   std::vector<uint8_t> axis_unpacked_tensor;
   RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(axis_input_info.initializer_tensor, axis_unpacked_tensor));
-  RETURN_IF_NOT(1 == static_cast<uint32_t>(axis_unpacked_tensor.size() / sizeof(axis_input_info.qnn_data_type)),
+  // Int 64 has double size.
```
qti-yuduo: nit: remove comments (it's obvious with the code bellow).

onnxruntime/core/providers/qnn-abi/builder/opbuilder/cumsum_op_builder.cc
quic-muchhsu: Yes, because the default soc is 30 (Lahaina, SM8350, snapdragon 888), which doesn't have a fp kernel for Cumsum. So we have to set to another soc that have fp kernel for Cumsum to test.

quic-muchhsu: This jira for example: https://jira-dc.qualcomm.com/jira/browse/AISW-158444
soc=35 doesn't have fp kernel for ScatterElement but HTP actually support it, so we have to use another soc_model when testing to pass the validation in FinalizeGraphs stage (which check node spec for given soc_model) 

Meanwhile in aihub's perspective, we're using the dlc which is generated before FinalizeGraphs stage. So we're good.

qti-yuduo (approved):

quic-muchhsu:
The inverse error is very interesting. Locally, running single test with default soc (soc_model=30) it passed, but I can reproduce the failure by setting soc_model = 87. Which somehow implies setting soc_model=87 in cumsum test will also propogates to later tests (e.g. inverse op test).

Update: It just failed with or without explicit setting soc_model=0 in inverse op test in CI, but locally it always pass...

image

quic-muchhsu:
@qti-yuduo  I temporarily disabled inverse op test, and create an issue on me to fix it in another PR: https://github.com/qcom-ai-hub/tetracode/issues/17522
It's too flicky.

minfhong-qti:
> The inverse error is very interesting. Locally, running single test with default soc (soc_model=30) it passed, but I can reproduce the failure by setting soc_model = 87. Which somehow implies setting soc_model=87 in cumsum test will also propogates to later tests (e.g. inverse op test).
> 
> Update: It just failed with or without explicit setting soc_model=0 in inverse op test in CI, but locally it always pass...
> 
> image

This phenomenon was first observed during the ABI enablement few months ago (https://github.com/onnxruntime/onnxruntime-qnn/blob/main/onnxruntime/test/providers/qnn-abi/qnn_test_utils.cc#L513) when running non-ABI UT after ABI UT. Furthermore, there are in fact several test cases (in both ABI and non-ABI) depending on previously set `soc_model` to pass and they would fail if executing alone. BTW, I believe the issue mostly occur for the HTP FP16/FP32 usage.
To solve this issue, my workaround is to set the `soc_model` again back to 30 in `TearDownTestSuite`. I suppose it could as well work for each test case by implementing `TearDown` but I think it will introduce huge overhead.
I'm afraid that proper solution requires to revisit all test cases and set proper expected `soc_model` individually.

PR #37, opened 2026-02-04
[AISW-162653] Update documentation for 64bit udma feature

qti-monumeen:
### Description
Document change to add new qnn ep option "extended_udma" and its benefit

### Motivation and Context
Exposing 64bit udma feature to customer


qti-yuduo (approved):

qti-yuduo:
@qti-monumeen I just realized this is targeting `gh-pages`. Not sure this branch would be used for doc release/publishes. You might need to put up PR on the actual ONNXRuntime repo that being used for doc deployment.

PR #38, opened 2026-02-05
[QNN EP] Added sustained high performance mode implementation

qti-monumeen:
### Description
Added sustained high performance mode


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1781,9 +1790,25 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
...
+
+    if (ep->GetHtpPowerConfigId(htp_power_config_id)) {
+      RETURN_IF_ERROR(ep->qnn_backend_manager_->SetState(onnxruntime::qnn::GraphState::INIT_DONE, htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_));
+    }
+
+    if (!finalize_status.IsOK()) {
+      return finalize_status.release();
+    }
+
+    RETURN_IF_ERROR(qnn_model->SetupQnnInputOutput(ep->logger_));
```
```suggestion
    RETURN_IF_NOT_OK(qnn_model->SetupQnnInputOutput(ep->logger_));
```
Please check for others.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1782,7 +1791,23 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
     }
 
     RETURN_IF_NOT_OK(qnn_model->ComposeGraph(context));
-    RETURN_IF_NOT_OK(qnn_model->FinalizeGraphs(ep->logger_));
+
+    uint32_t htp_power_config_id = 0;
+
+    if (ep->GetHtpPowerConfigId(htp_power_config_id)) {
+      RETURN_IF_ERROR(ep->qnn_backend_manager_->SetState(onnxruntime::qnn::GraphState::INIT_START, htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_));
```
```suggestion
      RETURN_IF_NOT_OK(ep->qnn_backend_manager_->SetState(onnxruntime::qnn::GraphState::INIT_START, htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_));
```
This function is returning `OrtStatue*` instead of `Ort::Status`. As well for LN1804.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -849,6 +849,63 @@ TEST_F(QnnHTPBackendTests, MultithreadDefaultHtpPowerCfgFromEpOption) {
...
+  status = session_obj.Initialize();
+  ASSERT_TRUE(status.IsOK());
+
+  std::vector<std::thread> threads;
+  constexpr int num_threads = 5;
+  constexpr int loop_count = 10;
+
+  for (int i = 0; i < num_threads; i++) {
+    threads.push_back(std::thread(RunSessionAndVerify, std::ref(session_obj), run_opts,
+                                  model->builder.feeds_, model->builder.output_names_,
```
minfhong-qti: Since we recently refactor UT infra, some usage may not be the same as before. The build error seems to be that `model->builder` does not have `output_names_` now.

qti-monumeen: Okkay, i will make the changes accordingly

qti-jkilpatrick:
It looks like this needs to be rebased on `main` and CI isn't clean yet. Please consider converting this to a draft until it's ready so reviewers can see they should wait. Thanks!

minfhong-qti:
Below review is generated from Claude. Please check and apply suggestions if applicable.

## Critical

**[C-1]** `qnn_backend_manager.cc` — Potential deadlock: `state_mutex_` → `perf_mutex_` → `AbortTimer()` vs. `TimerCallback()` → `SetState()` → `state_mutex_`

`SetState()` acquires `state_mutex_`, then calls `SetSustainedPerformance()` or `SetPerformance()`, both of which acquire `perf_mutex_` and then call `timer_->AbortTimer()`. `AbortTimer()` waits (via `cv_.wait`) for the timer thread to reach IDLE.

Meanwhile, the timer thread in `BkgTimer()` unlocks `timer_->mtx_` (line 57) before calling `TimerCallback()`, which calls `SetState()` — which tries to acquire `state_mutex_`. If the calling thread holds `state_mutex_` while waiting for the timer thread to go IDLE, and the timer thread is blocked waiting for `state_mutex_`, this is a deadlock.

**Suggested fix**: Redesign the lock hierarchy. One approach: update `graph_state_` under `state_mutex_`, release the lock, then call `AbortTimer()`/`LaunchTimer()` outside the lock scope. Alternatively, unify all state transitions under a single mutex.

---

## Major

**[M-1]** `timer.cc`, `timer.h` — Wrong copyright header (Microsoft instead of Qualcomm)

Both new files use:
```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License
```

New files must use the Qualcomm copyright:
```
```

---

**[M-2]** `timer.h`, `qnn_backend_manager.h` — `Timer` class is in the global namespace; `TimerCallbackArg`/`TimerResource` nested structs should be reorganized

`Timer` should be placed inside `namespace onnxruntime::qnn` to be consistent with all other EP classes. `TimerCallbackArg` and `TimerResource` have enough complexity to warrant being defined outside the `QnnBackendManager` class body (per [C31] pattern).

---

**[M-3]** `qnn_backend_manager.cc:2158–2166` — Unchecked optional dereference of `rpc_polling_time` / `rpc_control_latency`

```cpp
RETURN_IF_ERROR(SetState(..., *htp_power_configs.rpc_polling_time, *htp_power_configs.rpc_control_latency));
```

The branch only checks `pre_run_perf_mode.has_value()` — it does not check whether `rpc_polling_time`/`rpc_control_latency` have values. Currently safe because `GetPerThreadHtpPowerConfigs()` always sets them, but this is a fragile implicit precondition.

**Suggested fix**: Use `.value_or(0)` or change these fields to non-optional `uint32_t` with default values to make the invariant explicit.

---

## Minor

**[N-1]** `qnn_backend_manager.cc` — `SetSustainedPerformance()`/`SetPerformance()` read `graph_state_` while holding `perf_mutex_` only (not `state_mutex_`), creating unclear ownership

Minimum fix: add a comment on both functions stating "Caller must hold `state_mutex_`."

---

**[N-2]** `qnn_backend_manager.cc:329` — Null check after `std::make_unique` is dead code

`std::make_unique` throws `std::bad_alloc` on failure; it never returns `nullptr`. The `if (timer_callback_arg_ == nullptr)` check (and similar `new Timer()` null check) will never be true. Remove or replace with a try/catch.

---

**[N-3]** `qnn_backend_manager.cc:350` — `ReleaseTimerThread()` calls `timer_->DeInitialize()` (which joins the thread) while holding `state_mutex_`

If the timer thread is currently in `TimerCallback()` → `SetState()` waiting for `state_mutex_`, `join()` will never complete. Release `state_mutex_` before calling `DeInitialize()`.

---

**[N-4]** `qnn_execution_provider.cc:2380` — `SetDynamicOptionsImpl()` overwrites `default_rpc_polling_time_` permanently

When switching to Burst mode, `default_rpc_polling_time_ = 9999` is written. Switching back to non-Burst sets it to `0`, silently losing the user's original provider-option value.

**Suggested fix**: Introduce `dynamic_rpc_polling_time_` alongside `dynamic_htp_performance_mode_`; do not mutate `default_rpc_polling_time_`.

---

## Nit

**[T-1]** PR title typo: "performace" → "performance" (also in first commit message "perfomance").

**[T-2]** `qnn_def.h` — `kSleepHigherLatency = 65535` and `kDefaultTimerTimeoutUs = 300000` should have inline comments clarifying their human-readable values (~65.5ms and 300ms respectively).

**[T-3]** `timer.cc:22–28` — The `try { cv_.notify_all(); } catch (std::system_error)` block is dead code; `notify_all()` is `noexcept`. Remove it.

**[T-4]** `qnn_execution_provider.cc` — Two unrelated blank-line deletions (in `GetCapabilityImpl` and `~QnnEp`) add diff noise. Please revert.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -175,6 +175,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
 
   bool disable_cpu_ep_fallback_ = false;  // True if CPU EP fallback has been disabled for this session.
   bool qnn_context_embed_mode_ = true;
+
```
minfhong-qti: If no specific reason, please remove.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1609,8 +1609,8 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
 
   // Analyze nodes for QNN support
   std::vector<const OrtNode*> supported_nodes;
-  ep->GetSupportedNodes(graph, node_unit_map, node_unit_holder.size(), supported_nodes);
 
+  ep->GetSupportedNodes(graph, node_unit_map, node_unit_holder.size(), supported_nodes);
```
minfhong-qti: If no specific reason, please revert back.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1968,9 +1968,25 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
   QnnEp* ep = static_cast<QnnEp*>(this_ptr);
 
   if (qnn::IsOrtGraphHasCtxNode(graphs, count, ep->ort_api)) {
-    return ep->CompileContextModel(graphs, fused_nodes, count, node_compute_infos);
+    uint32_t htp_power_config_id = 0;
+    if (ep->GetHtpPowerConfigId(htp_power_config_id)) {
+      RETURN_IF_NOT_OK(ep->qnn_backend_manager_->SetState(onnxruntime::qnn::GraphState::INIT_START, htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_));
```
```suggestion
      RETURN_IF_NOT_OK(ep->qnn_backend_manager_->SetState(qnn::GraphState::INIT_START, htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_));
```
Please check for others as well.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: It seems `configs_set` will always be `true` at the end of function. Probably remove the check in caller?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1968,9 +1968,25 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
   QnnEp* ep = static_cast<QnnEp*>(this_ptr);
 
   if (qnn::IsOrtGraphHasCtxNode(graphs, count, ep->ort_api)) {
-    return ep->CompileContextModel(graphs, fused_nodes, count, node_compute_infos);
+    uint32_t htp_power_config_id = 0;
```
minfhong-qti, line 2806:
```C++
class ScopedStateInit {
 public:
  ScopedStateInit(...) {
    if (ep->GetHtpPowerConfigId(htp_power_config_id)) {
      RETURN_IF_NOT_OK(ep->qnn_backend_manager_->SetState(onnxruntime::qnn::GraphState::INIT_START, htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_));
    }
  }

  ~ScopedStateInit(...) {
    if (ep->GetHtpPowerConfigId(htp_power_config_id)) {
      RETURN_IF_NOT_OK(ep->qnn_backend_manager_->SetState(onnxruntime::qnn::GraphState::INIT_DONE, htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_));
    }
  }

 private:
  uint32_t htp_power_config_id = 0;
};

Ort::Status status;
{
  auto scoped_state_init = std::make_unique<ScopedStateInit>();
  status = ep->CompileContextModel(graphs, fused_nodes, count, node_compute_infos);
}
return status;
```
Define an RAII class to avoid duplicate code.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -158,6 +158,208 @@ Ort::Status ReadBinaryFromFile(const std::string& file_path, uint8_t* buffer, si
   return Ort::Status();
 }
 
+bool QnnBackendManager::IsTimerThreadRunning() {
```
minfhong-qti: A general comment. I think it is possible and reasonable to move most of the modifications into `HtpPowerConfigManager` and make `QnnBackendManager` as clean as possible.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.h
```diff
@@ -20,49 +21,105 @@ namespace power {
 // updates power configurations for the HTP backend
 class HtpPowerConfigManager {
  public:
-  HtpPowerConfigManager();
+  HtpPowerConfigManager(const Ort::Logger& logger);
```
minfhong-qti: Strongly prefer to pass logger when ever needed instead of cache in here.
If any place not accessible (e.g., some callbacks?), please use default logger in ort_api.h instead.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -229,4 +231,70 @@ class QnnEp : public OrtEp, public ApiPtrs {
   mutable std::atomic<uint64_t> genie_kv_cache_rewind_{1};
 };
 
+namespace qnn {
```
minfhong-qti: Move inside `qnn_htp_power_config_manager.h`?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: Seem missing this comment.

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -82,11 +82,42 @@ enum class HtpPerformanceMode : uint8_t {
   kHtpExtremePowerSaver,
 };
 
+enum class DcvsState {
```
minfhong-qti: Prefer to move below three new classes into `qnn_htp_power_mananger.h`.

minfhong-qti (commented):
Below is the review generated by AI Agent.

---

**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc` — `SetPerThreadHtpPowerConfigs`

Unconditional dereference of `*htp_power_configs.rpc_polling_time` and `*htp_power_configs.rpc_control_latency` when building `HtpPerfConfig_t`. In `GetPerThreadHtpPowerConfigs`, `rpc_polling_time` is only set when `dynamic_htp_performance_mode_` or `default_htp_performance_mode_` is not `kHtpDefault`. If both remain `kHtpDefault`, `rpc_polling_time` is `std::nullopt` and the dereference is undefined behavior (crash).

Suggested fix:
```cpp
uint32_t rpc_polling = htp_power_configs.rpc_polling_time.value_or(0);
uint32_t rpc_latency = htp_power_configs.rpc_control_latency.value_or(0);
HtpPerfConfig_t config{htp_power_config_id, *htp_power_configs.pre_run_perf_mode, rpc_polling, rpc_latency};
```
Or ensure `rpc_polling_time` is always populated in `GetPerThreadHtpPowerConfigs` (e.g., fall back to `default_rpc_polling_time_`).

---

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc` — `SetState` / `SetSustainedPerformance`

`SetState` sets `graph_state_` under `state_mutex_`, releases the lock, then calls `SetSustainedPerformance` which acquires `perf_mutex_` and re-reads `graph_state_`. Since these are different mutexes, another thread (e.g., `TimerCallback` calling `SetState(TIMEOUT, ...)`) can modify `graph_state_` between the two lock acquisitions, causing `SetSustainedPerformance` to operate on stale state.

Suggested fix: Pass the graph state as a parameter into `SetSustainedPerformance`/`SetPerformance` instead of re-reading `graph_state_`:
```cpp
// In SetState, after setting graph_state_:
return SetSustainedPerformance(state, config);  // pass state as arg
// In SetSustainedPerformance, use the passed state parameter
```

---

**[M-3]** `onnxruntime/core/providers/qnn/qnn_execution_provider.h` — `HtpPowerStateGuard`

`HtpPowerStateGuard` (~70 lines, custom ctor/dtor, deleted copy, two public methods) is defined at the bottom of the EP header. Per codebase conventions ([C31b]), non-trivial classes should live in their own `.h` file rather than be appended to the EP header.

Suggested fix: Move `HtpPowerStateGuard` to a new header file (e.g., `qnn_htp_power_state_guard.h` or into `qnn_htp_power_config_manager.h`) and `#include` it from `qnn_execution_provider.h`.

---

**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.h` — `AddHtpPerformanceMode` doc comment

The doc comment still says "If the value is the same as the last previously set, then there will be no new performance mode staged" but `last_set_htp_performance_mode_` and the dedup guard were removed by this PR. Please update the comment to reflect the new behavior.

---

**[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.h`

`SetRelaxedPerfPowerConfig`, `SetReleasedPerfPowerConfig`, and `SetExtremeLowPerfPowerConfig` are internal helpers used only within `HtpPowerConfigManager`. They should be in the `private` section, not `public`.

---

**[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc` — `SetHtpPowerConfigs` / `SetHtpPowerCustomConfigs`

`qnn_interface_` is `nullptr` until `Init()` is called. `SetHtpPowerConfigs` dereferences it via `*qnn_interface_` with no null-check. Consider adding a guard:
```cpp
RETURN_IF(qnn_interface_ == nullptr, "Init() must be called before SetState()");
```

---

**[N-4]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc` — `MultithreadSustainedHighPowerCfgFromEpOption`

Similar tests in the same fixture have an ARM64 platform guard (timer behavior requires real HTP hardware). This new test is missing one. Please add a `SKIP_IF_NOT_HTP_DEVICE()` or equivalent guard consistent with `MultithreadDefaultHtpPowerCfgFromEpOption`.

---

**[T-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc`

Comment `// add in htp_power_configs the default power config id also so to run when we execute` is grammatically unclear. Suggest: `// Use default_perf_mode as fallback if no per-run perf mode override is set`

---

**[T-2]** `onnxruntime/core/providers/qnn/builder/qnn_def.h` — `DcvsState`

`DCVS_NUM_STATES` is a C-style count sentinel, non-idiomatic in `enum class` and unused. Remove it.

---

**[T-3]** `onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.h` — `SetHtpPowerCustomConfigs`

`QnnHtpPerfInfrastructure_PowerConfig_t power_config` is passed by value. Change to `const QnnHtpPerfInfrastructure_PowerConfig_t& power_config` to avoid an unnecessary copy.

onnxruntime/core/providers/qnn/builder/qnn_htp_power_state_guard.h
```diff
@@ -0,0 +1,74 @@
...
+  }
+  HtpPowerStateGuard(const HtpPowerStateGuard&) = delete;
+  HtpPowerStateGuard& operator=(const HtpPowerStateGuard&) = delete;
+
+ private:
+  power::HtpPowerConfigManager* power_manager_;
+  bool valid_power_config_id_;
+  power::GraphState done_state_;
+  power::HtpPerfConfig_t config_;
+  const Ort::Logger* logger_ptr_;
```
minfhong-qti:
```suggestion
  const Ort::Logger& logger_;
```
Use reference is enough.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -24,6 +24,7 @@
 #include "core/providers/qnn/genie/genie_api_loader.h"
 #include "core/providers/qnn/genie/genie_node.h"
 #include "core/providers/qnn/genie/genie_node_compute_info.h"
+#include "core/providers/qnn/builder/qnn_htp_power_state_guard.h"
```
minfhong-qti, line 33: Please sort it.

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -127,17 +134,22 @@ bool IsGpuBackend(QnnBackendType backend_type);
...
+constexpr const int kSleepHigherLatency = 65535;
+// constexpr config values
 constexpr const int kDcvsDisable = 0;
 constexpr const int kDcvsEnable = 1;
 constexpr const uint32_t kDisableRpcPolling = 0;
 constexpr const uint32_t kDisableRpcControlLatency = 0;
 constexpr const uint32_t kMaxRpcPolling = 9999;
 
+// performance timer timeout value is in microseconds
+static constexpr uint64_t kDefaultTimerTimeoutUs = 300000;
```
minfhong-qti:
```suggestion
constexpr const uint64_t kDefaultTimerTimeoutUs = 300000;
```
Align with above?

minfhong-qti (commented):
---

**[C-1]** `qnn_htp_power_config_manager.cc` — `SetSustainedPerformance`

**Critical: Deadlock — `AbortTimer()` called while `state_mutex_` is held.**

`SetState()` holds `state_mutex_`, then calls `SetSustainedPerformance()`, which calls the blocking `AbortTimer()` when `RUN_START` or `INIT_START` fires while the timer is running. `AbortTimer()` waits for the timer thread to reach `IDLE`. But if the timer thread is in `CALLING` state (running `TimerCallback`), and `TimerCallback` calls `SetState()` which tries to re-acquire `state_mutex_` — the process deadlocks.

Interestingly, you already identified and fixed the exact same pattern in `ReleaseTimerThread` with the comment: _"Deinitialize outside the lock to avoid deadlock: an in-flight TimerCallback calls SetState() which acquires state_mutex_."_ The same fix needs to be applied to the `AbortTimer()` call paths in `SetSustainedPerformance`.

---

**[M-1]** `qnn_htp_power_config_manager.h` — mutex declarations

The lock acquisition order (`state_mutex_` → `perf_mutex_`) is undocumented. Please add a comment near the mutex declarations: _"Lock order: state_mutex_ must always be acquired before perf_mutex_."_ This prevents future maintainers from accidentally introducing a deadlock by acquiring in the reverse order.

---

**[M-2]** `qnn_htp_power_state_guard.h` — copyright header

New file uses Microsoft copyright. Per repo convention, new files must use:
```
```

**[N-1]** `qnn_htp_power_config_manager.cc` — `graph_state_` double-write

`SetState()` sets `graph_state_` to the new state, but `SetSustainedPerformance()` immediately overwrites it with `GraphState::NONE` in every branch. The early-exit condition in `SetState` (`state == graph_state_`) only works correctly because of this reset, but the behavior is undocumented. Consider centralizing `graph_state_` updates in one place and adding a comment explaining the lifecycle.

---

**[N-2]** `qnn_htp_power_config_manager.cc` — `AddHtpPerformanceConfig` parameter

`AddHtpPerformanceConfig(QnnHtpPerfInfrastructure_PowerConfig_t htp_performance_cfg)` takes by value then `std::move`s into the vector — this incurs one unnecessary copy. Prefer `const QnnHtpPerfInfrastructure_PowerConfig_t&` or `QnnHtpPerfInfrastructure_PowerConfig_t&&`.

---

**[T-2]** `qnn_def.h` — comment ordering

`kSleepHigherLatency` is now placed after the `// latency values are in microseconds` comment but before the `// constexpr config values` comment — yet it's a latency value, not a config value. The comment placement makes it look like a config value. Suggest keeping `kSleepHigherLatency` within the latency block.

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -94,34 +94,29 @@ static std::string_view PerformanceModeToString(HtpPerformanceMode htp_performan
...
+  power_configs_.emplace_back(std::move(htp_performance_cfg));
+  htp_performance_mode_set_ = true;
+  return Ort::Status();
+}
+
 Ort::Status HtpPowerConfigManager::AddHtpPerformanceMode(HtpPerformanceMode htp_performance_mode,
                                                          uint32_t htp_power_config_client_id,
                                                          const Ort::Logger& logger) {
   RETURN_IF(htp_performance_mode_set_, "There is already a pending HTP performance mode config");
-  if (htp_performance_mode == last_set_htp_performance_mode_) {
```
huaychou, line 101: Why do we remove this check?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_state_guard.h
```diff
@@ -0,0 +1,77 @@
...
+  }
+  ~HtpPowerStateGuard() {
+    if (!finalized_ && power_manager_ && valid_power_config_id_) {
+      // Error cannot be propagated from a destructor; silently ignore.
+      power_manager_->SetState(done_state_, config_, logger_);
+    }
+  }
+  // Returns (by move) the status of setting HTP performance before work begins.
+  // Should be checked immediately after construction.
+  Ort::Status SetPreRunHtpPerfStatus() { return std::move(start_status_); }
```
huaychou: This requires the developer to call SetPreRunHtpPerfStatus immediately after construction. Would it be possible to set the pre-run state and return the status instead, to allow for more flexibility?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  dcvs_v3.setBusParams = 1;
+  dcvs_v3.coreVoltageCornerMin = DCVS_VOLTAGE_VCORNER_MIN_VOLTAGE_CORNER;
+  dcvs_v3.coreVoltageCornerTarget = DCVS_VOLTAGE_VCORNER_MIN_VOLTAGE_CORNER;
+  dcvs_v3.coreVoltageCornerMax = DCVS_VOLTAGE_VCORNER_MIN_VOLTAGE_CORNER;
+  dcvs_v3.setCoreParams = 1;
+}
+
+void HtpPowerConfigManager::CreateTimerThread(uint32_t htp_power_config_client_id) {
+  std::lock_guard<std::mutex> lk(state_mutex_);
+  const Ort::Logger& logger = OrtLoggingManager::GetDefaultLogger();
```
huaychou, line 370: Can we pass the logger instead of GetDefaultLogger?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2051,29 +2041,23 @@ Ort::Status QnnBackendManager::SetPerThreadHtpPowerConfigs(const std::thread::id
...
     if (htp_power_configs.pre_run_perf_mode.has_value()) {
-      RETURN_IF_ERROR(htp_power_config_manager_.AddHtpPerformanceMode(*htp_power_configs.pre_run_perf_mode,
-                                                                      htp_power_config_id,
-                                                                      *logger_ptr_));
-    }
-
-    if (htp_power_configs.rpc_control_latency.has_value()) {
-      RETURN_IF_ERROR(htp_power_config_manager_.AddRpcControlLatency(*htp_power_configs.rpc_control_latency,
-                                                                     *logger_ptr_));
+      power::HtpPerfConfig_t config{htp_power_config_id, *htp_power_configs.pre_run_perf_mode, *htp_power_configs.rpc_polling_time, *htp_power_configs.rpc_control_latency};
```
huaychou, line 2256: Do we need to use HtpPowerStateGuard for this config?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2051,29 +2041,23 @@ Ort::Status QnnBackendManager::SetPerThreadHtpPowerConfigs(const std::thread::id
...
+      RETURN_IF_ERROR(htp_power_config_manager_.SetState(power::GraphState::RUN_START, config, *logger_ptr_));
+    } else if (htp_power_configs.default_perf_mode.has_value()) {
```
huaychou, line 2258: Can this be else?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  // Holding state_mutex_ across AbortTimer() would therefore deadlock.
+  // The same pattern is already applied in ReleaseTimerThread().
+  Ort::Status status;
+  if (config.perf_mode == qnn::HtpPerformanceMode::kHtpSustainedHighPerformance || config.perf_mode == qnn::HtpPerformanceMode::kHtpBurst) {
+    status = SetSustainedPerformance(state, config, logger);
+  } else if (config.perf_mode == qnn::HtpPerformanceMode::kHtpDefault) {
+    if (timer_ && timer_->TimerInUse()) {
+      timer_->AbortTimer();
+    }
+    status = Ort::Status();
```
huaychou, line 644: Do we need to SetPerformance for qnn::HtpPerformanceMode::kHtpDefault?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  }
+
+  // Dispatch to performance setters outside state_mutex_ to avoid deadlock:
+  // AbortTimer() blocks until the timer thread is idle, but the timer thread
+  // (inside TimerCallback) calls SetState() which acquires state_mutex_.
+  // Holding state_mutex_ across AbortTimer() would therefore deadlock.
+  // The same pattern is already applied in ReleaseTimerThread().
+  Ort::Status status;
+  if (config.perf_mode == qnn::HtpPerformanceMode::kHtpSustainedHighPerformance || config.perf_mode == qnn::HtpPerformanceMode::kHtpBurst) {
+    status = SetSustainedPerformance(state, config, logger);
```
huaychou, line 637: Just wondering, who holds the state_mutex_  lock?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: actually as we will be adjusting performance configuration of particular Perf type before and after the run the perf type won't be changing but configuration will change and with this check the congiguration will not apply

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: actually previously we were passing logger but due to a review comment we shifted to GetDefaultLogger

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-monumeen: I don't think its needed as it will be populated before run starts or end

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: No there is no as such requirement for default but if someone sets it by mistake we are handling it here

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: state_mutex_ lock is hold while we are updating the graph_state_ variable and we can't keep it holding all time due to timer callback function

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-monumeen: Actually here we populate default_perf_mode in the structure only when it is other then Kdefault perf type and as we don't need any perf setting for this perf type i have kept else if and it also take care of condition where we are not accessing something which was not populated

onnxruntime/core/providers/qnn/builder/qnn_htp_power_state_guard.h
qti-monumeen: yeah, i can make this change

onnxruntime/core/providers/qnn/builder/timer.h
```diff
@@ -0,0 +1,81 @@
...
+#include <condition_variable>
+#include <atomic>
+#include "core/providers/qnn/ort_api.h"
+
+namespace onnxruntime {
+namespace qnn {
+namespace power {
+
+class Timer {
+ public:
```
huaychou, line 18: Could we add ORT_DISALLOW_COPY_AND_ASSIGNMENT(Timer); to avoid copy of Timer?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.h
```diff
@@ -11,13 +11,42 @@
...
 // Manages staging of any new power configurations and
-// updates power configurations for the HTP backend
+// updates power configurations for the HTP backend.
+//
+// IMPORTANT: Init() must be called before any other methods that access
+// the QNN interface (SetState, SetPowerConfig, etc.), typically during
+// backend initialization. Failure to call Init() will result in errors
+// when attempting to set power configurations.
 class HtpPowerConfigManager {
  public:
```
huaychou, line 46: Can we add ORT_DISALLOW_COPY_AND_ASSIGNMENT(HtpPowerConfigManager); since we have mutex locks in HtpPowerConfigManager now?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.h
```diff
@@ -34,35 +63,90 @@ class HtpPowerConfigManager {
...
+  std::mutex perf_mutex_;
+  std::mutex state_mutex_;
+  std::unique_ptr<Timer> timer_;
+  struct TimerResource {
+    static constexpr uint64_t sustained_timer_duration_ = kDefaultTimerTimeoutUs;  // in microseconds
+    std::atomic<bool> caller_busy_ = false;
+    std::atomic<bool> timer_active_ = false;
+  };
+  TimerResource timer_resource_;
+  std::atomic<GraphState> graph_state_ = GraphState::NONE;
```
huaychou: Since graph_state_ can only be updated when holding state_mutex_, I think we don't need to define it as std::atomic.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -394,6 +395,7 @@ Ort::Status QnnBackendManager::LoadQnnSerializerBackend() {
        QNN_API_VERSION_PATCH},
       &serializer_interface_provider)));
   qnn_interface_ = serializer_interface_provider->QNN_INTERFACE_VER_NAME;  // NOTE: QnnSaver/Ir will provide the interfaces
+  htp_power_config_manager_.Init(qnn_interface_);
```
huaychou, line 414: I think we don't need this for QnnSerializerBackend.

onnxruntime/core/providers/qnn/builder/qnn_def.h
huaychou: Please also fix this

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -158,7 +159,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
   } GraphFinalizationInfo_t;
 
   // Will return true if any power config options need to be updated
```
huaychou: Should update the comment.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-monumeen: In this function we will have qnn interface with us so we need to update the qnn interface member variable of power config mananger class. Hence, we are calling this

onnxruntime/core/providers/qnn/builder/timer.cc
```diff
@@ -0,0 +1,98 @@
...
+
+Timer::~Timer() { this->DeInitialize(); }
+
+void Timer::BkgTimer() {
+  {
+    std::unique_lock<std::mutex> lk(mtx_);
+    try {
+      thread_status_ = threadState::IDLE;
+      cv_.notify_all();
+    } catch (const std::system_error& e) {
```
huaychou: **[N-2] (Minor)** try/catch in BkgTimer() is dead code: cv_.notify_all() is noexcept

In the initialization block of `BkgTimer()` (timer.cc lines 25–32), `cv_.notify_all()` is wrapped in a `try { ... } catch (const std::system_error& e)`. The C++11 standard specifies that `std::condition_variable::notify_all()` is `noexcept` and can never throw `std::system_error`. The catch block is therefore dead code, and the `thread_status_ = FAILED` path can never execute.

**Impact:** The `FAILED` error-handling path in `Initialize()` can never be triggered, making the error handling misleading.

**Suggested fix:**

Remove the try/catch and write directly: `thread_status_ = threadState::IDLE; cv_.notify_all();`. To catch thread-startup errors, handle `std::system_error` around the `std::thread` constructor call instead.

onnxruntime/core/providers/qnn/builder/qnn_htp_power_state_guard.h
```diff
@@ -0,0 +1,86 @@
...
+    pre_run_called_ = true;
+    if (power_manager_ && valid_power_config_id_) {
+      return power_manager_->SetState(start_state_, config_, logger_);
+    }
+    return Ort::Status();
+  }
+  // Explicitly sets HTP performance after work is done and returns its status.
+  // After this call the destructor will not invoke SetState again.
+  Ort::Status SetPostRunHtpPerf() {
+    finalized_ = true;
```
huaychou: **[N-3] (Minor)** SetPostRunHtpPerf() sets finalized_=true before calling SetState — destructor fallback skipped on failure

In `HtpPowerStateGuard::SetPostRunHtpPerf()` (lines 65–70), `finalized_ = true` is set *before* calling `power_manager_->SetState(done_state_, ...)`.

If `SetState(done_state_)` fails and returns an error, `finalized_` is already `true`, so the destructor's `if (pre_run_called_ && !finalized_)` condition is false and the destructor will not retry the done-state transition. The HTP power config is left in start_state (high-performance) until the next call.

**Impact:** In the edge case where `SetPostRunHtpPerf()` fails (power-config API error), HTP stays in high-performance mode instead of being relaxed.

**Suggested fix:**

Move `finalized_ = true` to after the `SetState(done_state_)` call, setting it only on success. Alternatively, keep the current behavior (always set `finalized_`) but log a WARNING on failure to indicate that the HTP power state may not have been correctly relaxed.

huaychou (commented):
Just wondering, what is the benefit of this PR? Should it reduce the init/deinit time of each execution?

qti-monumeen:
Refer to this page for reason to add this change: 
https://qualcomm-confluence.atlassian.net/wiki/spaces/ML/pages/434806935/Sustained+High+Performance+enablement+on+ORT

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  dcvs_v3.busVoltageCornerMax = DCVS_VOLTAGE_VCORNER_MIN_VOLTAGE_CORNER;
+  dcvs_v3.setBusParams = 1;
+  dcvs_v3.coreVoltageCornerMin = DCVS_VOLTAGE_VCORNER_MIN_VOLTAGE_CORNER;
+  dcvs_v3.coreVoltageCornerTarget = DCVS_VOLTAGE_VCORNER_MIN_VOLTAGE_CORNER;
+  dcvs_v3.coreVoltageCornerMax = DCVS_VOLTAGE_VCORNER_MIN_VOLTAGE_CORNER;
+  dcvs_v3.setCoreParams = 1;
+}
+
+void HtpPowerConfigManager::CreateTimerThread(uint32_t htp_power_config_client_id) {
+  std::lock_guard<std::mutex> lk(state_mutex_);
```
huaychou: Do we need state lock here?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+      status = SetHtpPowerConfigs(config, logger);
+      break;
+    default:
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, "Invalid graph state");
+      break;
+  }
+  return status;
+}
+
+Ort::Status HtpPowerConfigManager::SetState(GraphState state, const HtpPerfConfig_t& config, const Ort::Logger& logger) {
```
huaychou, line 585: Is it possible to make SetState a pure function for update GraphState?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
   return Ort::Status();
 }
 
+void HtpPowerConfigManager::SetRelaxedPerfPowerConfig(QnnHtpPerfInfrastructure_PowerConfig_t& power_config, uint32_t htp_power_config_client_id, DcvsState dcvsState) {
```
huaychou: I think this is duplicated with SetHtpPerformancePowerConfig. Could you update the SetHtpPerformancePowerConfig instead of creating a new function?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  dcvs_v3.busVoltageCornerTarget = DCVS_VOLTAGE_VCORNER_SVS;
+  dcvs_v3.busVoltageCornerMax = DCVS_VOLTAGE_VCORNER_SVS;
+  dcvs_v3.setBusParams = 1;
+  dcvs_v3.coreVoltageCornerMin = DCVS_VOLTAGE_VCORNER_SVS2;
+  dcvs_v3.coreVoltageCornerTarget = DCVS_VOLTAGE_VCORNER_SVS;
+  dcvs_v3.coreVoltageCornerMax = DCVS_VOLTAGE_VCORNER_SVS;
+  dcvs_v3.setCoreParams = 1;
+}
+
+void HtpPowerConfigManager::SetExtremeLowPerfPowerConfig(QnnHtpPerfInfrastructure_PowerConfig_t& power_config, uint32_t htp_power_config_client_id) {
```
huaychou, line 325: ditto

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  dcvs_v3.busVoltageCornerTarget = DCVS_VOLTAGE_CORNER_DISABLE;
+  dcvs_v3.busVoltageCornerMax = DCVS_VOLTAGE_CORNER_DISABLE;
+  dcvs_v3.setBusParams = 1;
+  dcvs_v3.coreVoltageCornerMin = DCVS_VOLTAGE_CORNER_DISABLE;
+  dcvs_v3.coreVoltageCornerTarget = DCVS_VOLTAGE_CORNER_DISABLE;
+  dcvs_v3.coreVoltageCornerMax = DCVS_VOLTAGE_CORNER_DISABLE;
+  dcvs_v3.setCoreParams = 1;
+}
+
+void HtpPowerConfigManager::SetReleasedPerfPowerConfig(QnnHtpPerfInfrastructure_PowerConfig_t& power_config, uint32_t htp_power_config_client_id, DcvsState dcvsState) {
```
huaychou: ditto

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+}
+
+void HtpPowerConfigManager::ReleaseTimerThread() {
+  std::unique_ptr<Timer> local_timer;
+  std::unique_ptr<TimerCallbackArg> local_callback_arg;
+  {
+    std::lock_guard<std::mutex> lk(state_mutex_);
+    if (timer_ != nullptr) {
+      timer_resource_.timer_active_ = false;
+      graph_state_ = GraphState::NONE;
```
huaychou: Just wondering, why do we need to update the graph_state_ when ReleaseTimerThread?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  {
+    std::lock_guard<std::mutex> lk(state_mutex_);
+    if (state != graph_state_) {
+      graph_state_ = state;
+    } else {
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, "State is the same as current. Ignoring request.");
+      return Ort::Status();
+    }
+    if (config.perf_mode == qnn::HtpPerformanceMode::kHtpSustainedHighPerformance || config.perf_mode == qnn::HtpPerformanceMode::kHtpBurst) {
+      RETURN_IF(timer_resource_.timer_active_ == false, "Timer is not active. Cannot set state.");
```
huaychou: I think the error message is quite misleading. We already set state in L510. How can we return "Cannot set state" in L516?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2068,9 +2068,33 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
...
+    bool power_config_valid = ep->GetHtpPowerConfigId(htp_power_config_id);
+    qnn::power::HtpPerfConfig_t perf_config{htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_};
+    qnn::HtpPowerStateGuard power_guard(
+        &ep->qnn_backend_manager_->GetHtpPowerConfigManager(),
+        power_config_valid,
+        qnn::power::GraphState::INIT_START, qnn::power::GraphState::INIT_DONE,
+        perf_config,
+        ep->logger_);
```
huaychou, line 2814: Can we move this outside or create another function to create HtpPowerStateGuard to avoid ducplicated?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2068,9 +2068,33 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
...
-    return ep->CompileDlcContextModel(this_ptr, graphs, fused_nodes, count, node_compute_infos);
+    uint32_t htp_power_config_id = 0;
+    bool power_config_valid = ep->GetHtpPowerConfigId(htp_power_config_id);
+    qnn::power::HtpPerfConfig_t perf_config{htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_};
+    qnn::HtpPowerStateGuard power_guard(
+        &ep->qnn_backend_manager_->GetHtpPowerConfigManager(),
+        power_config_valid,
+        qnn::power::GraphState::INIT_START, qnn::power::GraphState::INIT_DONE,
+        perf_config,
+        ep->logger_);
```
huaychou, line 2828: ditto

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2182,7 +2206,21 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
...
-      RETURN_IF_NOT_OK(qnn_model->FinalizeGraphs(ep->logger_));
+      uint32_t htp_power_config_id = 0;
+      bool valid_power_config_id = ep->GetHtpPowerConfigId(htp_power_config_id);
+      qnn::power::HtpPerfConfig_t perf_config{htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_};
+      qnn::HtpPowerStateGuard power_guard(
+          &ep->qnn_backend_manager_->GetHtpPowerConfigManager(),
+          valid_power_config_id,
+          qnn::power::GraphState::INIT_START, qnn::power::GraphState::INIT_DONE,
+          perf_config,
+          ep->logger_);
```
huaychou: ditto

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2207,6 +2245,16 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
...
     tp.Start();
+    uint32_t htp_power_config_id = 0;
+    bool valid_power_config_id = ep->GetHtpPowerConfigId(htp_power_config_id);
+    qnn::power::HtpPerfConfig_t perf_config{htp_power_config_id, ep->default_htp_performance_mode_, ep->default_rpc_polling_time_, ep->default_rpc_control_latency_};
+    qnn::HtpPowerStateGuard power_guard(
+        &ep->qnn_backend_manager_->GetHtpPowerConfigManager(),
+        valid_power_config_id,
+        qnn::power::GraphState::INIT_START, qnn::power::GraphState::INIT_DONE,
+        perf_config,
+        ep->logger_);
```
huaychou: ditto

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+      if (IsTimerThreadRunning()) {
+        timer_->AbortTimer();
+      } else {
+        status = SetHtpPowerConfigs(config, logger);
+      }
+      timer_resource_.caller_busy_ = true;
+      break;
+    case GraphState::INIT_DONE: {
+      QnnHtpPerfInfrastructure_PowerConfig_t init_done_htp_performance_cfg{};
+      SetRelaxedPerfPowerConfig(init_done_htp_performance_cfg, config.htp_power_config_client_id, DcvsState::DCVS_DEFAULT);
```
huaychou: Why do we need to set relaxed power config for sustained performance mode when INIT_DONE?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+        timer_->AbortTimer();
+      } else {
+        status = SetHtpPowerConfigs(config, logger);
+      }
+      timer_resource_.caller_busy_ = true;
+      break;
+    case GraphState::TIMEOUT: {
+      if (!timer_resource_.caller_busy_) {
+        QnnHtpPerfInfrastructure_PowerConfig_t timeout_htp_performance_cfg{};
+        SetRelaxedPerfPowerConfig(timeout_htp_performance_cfg, config.htp_power_config_client_id, DcvsState::DCVS_DEFAULT);
```
huaychou: Ditto

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -289,6 +284,323 @@ Ort::Status HtpPowerConfigManager::SetHtpPerformancePowerConfig(QnnHtpPerfInfras
...
+  return status;
+}
+
+Ort::Status HtpPowerConfigManager::SetPerformance(GraphState state, const HtpPerfConfig_t& config, const Ort::Logger& logger) {
+  std::lock_guard<std::mutex> lk(perf_mutex_);
+  Ort::Status status = Ort::Status();
+  switch (state) {
+    case GraphState::RUN_DONE:
+    case GraphState::INIT_DONE:
+      switch (config.perf_mode) {
```
huaychou, line 546: Do we need to consider kHtpDefault?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: I don't think making its work only updating graphState is correct ele we need to make some new function which will do rest of work, max to max i think we can rename it so that no one gets confused from name

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: okkay let me re phrase it

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: There is no need as in caller of SetPerformance we are handling it and here it will be redirected if someone by mistake give it as performance mode to defualt case in switch

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: The lock serializes timer creation against teardown and use, so each sees a consistent state

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: ReleaseTimerThread is teardown for the timer, and a new timer can be created afterward via
 CreateTimerThread. So if graph_state_ is not set to NONE it may get carry forward for next Timer creation so its better to update it

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: It is to reduce the npu load when we are not doing any work on it as such, as after init done we will do input processing and all, so just before execute we can make it up and use npu efficiently

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
qti-monumeen: A performance mode has two aspects: 1) perf mode related high clocks which are set before we run anything on npu 2) perf mode related low clocks which are set after we are done with running on npu and npu is not in work
So here the SetHtpPerformancePowerConfig deals with the first scenario which take care of high clocks related to perf mode that's why we need these function SetRelaxedPerfPowerConfig, SetExtremeLowPerfPowerConfig

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
```diff
@@ -94,34 +94,29 @@ static std::string_view PerformanceModeToString(HtpPerformanceMode htp_performan
...
 
-    power_configs_.emplace_back(std::move(htp_performance_cfg));
-
-    last_set_htp_performance_mode_ = htp_performance_mode;
-    htp_performance_mode_set_ = true;
-  }
+  power_configs_.emplace_back(std::move(htp_performance_cfg));
+  htp_performance_mode_set_ = true;
 
   return Ort::Status();
```
huaychou, line 131:
```suggestion
  return AddHtpPerformanceConfig(htp_performance_cfg);
```

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Acknowledge. Could you point-out the document related to the HTP performance mode setting?

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Acknowledge

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Acknowledge

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Acknowledge

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Acknowledge

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Acknowledge

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Acknowledge

onnxruntime/core/providers/qnn/builder/qnn_htp_power_config_manager.cc
huaychou: Sure, I think we can rename this function.

huaychou (approved):
Minor issues:
1. Use AddHtpPerformanceConfig to reduce duplicated code
2. Rename HtpPowerConfigManager::SetState
3. Clean QnnEp::CompileImpl to reduce duplicated code
Follow ups:
1. Point-out HTP performance mode setting related document.

qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🟠 Concern` `[Arch]` `qnn_htp_power_config_manager.cc` (`SetState`, L506-547)
The GraphState "state machine" holds no state: `SetState()` sets `graph_state_ = state` under the lock, dispatches, then **unconditionally** resets `graph_state_ = GraphState::NONE` (L538-541). So the entry-time `if (state != graph_state_) ... else ignore` dedup compares against `NONE` on essentially every call and never dedups consecutive different states; the member is inert between calls. It also races the TIMEOUT callback — in the gap between the two locked sections, `SetState(TIMEOUT)` can set `TIMEOUT`, which the original call then overwrites with `NONE`.
→ *Better design*: decide whether `graph_state_` should survive across calls. If it is a persistent state machine (LWD intent), don't reset it unconditionally and define the transitions explicitly; if it's a transient marker, demote/rename it and stop calling it a state machine. Make the dedup + TIMEOUT-race behavior observable by a test.

**[D-4]** `🟠 Concern` `[Arch]` `qnn_htp_power_state_guard.h` + `qnn_execution_provider.cc` (CompileImpl, 4 sites)
The RAII guard's correctness depends on the caller remembering to call `SetPreRunHtpPerfStatus()` after construction. If they construct it but forget, the destructor **silently no-ops** (`pre_run_called_` is false), setting no power state and reporting no error. The interface doesn't encapsulate its own correctness — the exact PR #38 anti-pattern in the design knowledge base (C1.2/C2.2). The 6-line "construct + SetPreRun + work + SetPostRun" boilerplate repeats at all 4 Compile paths.
→ *Better design*: have the guard enter the start state at construction (factory returning status) or take the protected work as a lambda, so "enter start ⇒ reach done" is type-guaranteed. If the header's claimed "setup between construction and state setting" flexibility is real, cite a site that uses it — none of the 4 do.

---

### 🟩 Maintainability

**[D-8]** `🔵 Improvement` `[Maint]` `qnn_htp_power_config_manager.cc` (`SetRelaxed/Released/ExtremeLowPerfPowerConfig`)
`shrink`: three ~20-line presets share a near-identical DCVS-V3 scaffold (differing only in voltage corners + sleepLatency); with the existing `SetHtpPerformancePowerConfig` switch, "how to assemble a DCVS-V3 config" now lives in 4 places (C1.1/C3.1). Fold the presets into the existing switch or a `(voltageCorner, sleepLatency, powerMode)`-parameterized helper.

---

### 🟩 Simplicity

**[D-6]** `🔵 Improvement` `[Simplicity]` `qnn_htp_power_config_manager.{h,cc}` (`SetRelaxed/ReleasedPerfPowerConfig`)
`yagni`: `DcvsState` is a 3-value enum but every one of the 4 call sites passes `DCVS_DEFAULT`, so the `if (dcvsState == DCVS_ENABLE) powerMode = ADJUST_UP_DOWN` branch is dead. The LWD itself says dcvs_state is not tracked. Drop the param + enum, inline `POWER_SAVER_MODE`.

**[D-7]** `🔵 Improvement` `[Simplicity]` `qnn_htp_power_config_manager.cc` (`AddHtpPerformanceMode`)
`AddHtpPerformanceMode` lost its "same as last set → skip" dedup (and the `last_set_htp_performance_mode_` member). For sustained mode, which calls SetState every run, this may re-issue redundant `setPowerConfig` DSP RPCs each inference (C8.3 unexplained removal). It looks replaced by SetState's `state != graph_state_` guard, but D-1 shows that guard is inert. Either justify the removal or keep an effective dedup.

---

### 🔵 Completeness

**[D-2]** `🟠 Concern` `[Completeness]` `qnn_htp_power_config_manager.{h,cc}` + `qnn_backend_manager.cc` (`SetPerThreadHtpPowerConfigs`)
The manager owns exactly one `timer_`, one `caller_busy_`, one `graph_state_`, but the EP runs multi-threaded inference with a per-thread power-config map, each thread calling `SetState(RUN_START/RUN_DONE)` on this shared manager. `caller_busy_` is a single bool, not a counter: A `RUN_START` (busy=true), B `RUN_START` (busy=true), B `RUN_DONE` clears busy and launches the timer — while A is still running, TIMEOUT hits `if(!caller_busy_)` and relaxes the DSP, dropping A out of high clocks. This is the PR's headline scenario and its most likely failure.
→ *Better design*: make `caller_busy_` an active-inference counter (++ on RUN_START, -- on RUN_DONE), launch the relax timer only at zero; or explicitly define + document the "one timer per session" semantics. Add a test that can observe the transitions (D-3).

**[D-3]** `🟠 Concern` `[Completeness]` `qnn_basic_test.cc` (`MultithreadSustainedHighPowerCfgFromEpOption`, L986-1035)
The one added test runs a QDQ Add across 5 threads x 10 loops and checks output equality. Mutation test: delete every line of the timer / sustained / GraphState code and this test still passes — zero coverage of the new feature. The E2E clock check is hard to automate, but the timer state transitions, `IsTimerThreadRunning()`, and the busy-flag logic (D-2) are pure-ish logic that can be unit-tested directly. No coverage of burst parity, the TIMEOUT relax path, or dynamic-mode-during-sustained.
→ *Better design*: unit-test the pure logic (drive SetState sequences, assert timer launch/abort, `caller_busy_` count, TIMEOUT suppressed while in-flight); add burst + dynamic-during-sustained cases; keep E2E clock verification manual with a tracking note.

**[D-5]** `🟠 Concern` `[Completeness]` `qnn_execution_provider.cc` (`SetDynamicOptionsImpl`, L2543-2557; `GetPerThreadHtpPowerConfigs`)
Old code applied the dynamic perf mode immediately via `SetHtpPowerConfigs`. New code only stores `dynamic_htp_performance_mode_` and defers to the next `OnRunStart` — a behavior change for callers expecting "set takes effect now" (C8.3). Also, the assignment is wrapped in `if (htp_performance_mode != kHtpDefault)`, so once a non-default mode is set the dynamic mode **can never be reset back to `kHtpDefault`** — a one-way latch.
→ *Better design*: document/acknowledge the deferred-apply change (if deferral is more correct for a plugin EP, say so); and handle the `kHtpDefault` path so a user can actually clear the dynamic mode back to default.

qti-chuteng (approved):

qti-chuteng:
### 🟠 [Design/Completeness] Run path can leak `inflight_run_count_` and pin the HTP boosted forever

**Problem**

In `qnn_model.cc` (`BindAndExecuteGraph`), the per-inference boost/relax is a manual two-call pair straddling `graphExecute`:

```cpp
RETURN_IF_ERROR(qnn_backend_manager_->SetPerThreadHtpPowerConfigs(thread_id, true));   // SetState(RUN_START): inflight_run_count_ += 1
execute_status = qnn_interface.graphExecute(...);
RETURN_IF_ERROR(qnn_backend_manager_->SetPerThreadHtpPowerConfigs(thread_id, false));  // SetState(RUN_DONE):  inflight_run_count_ -= 1
```

Inside `SetState(RUN_START)`, `inflight_run_count_` is incremented under `state_mutex_` **before** dispatching to the downstream setter (`SetSustainedPerformance`). So there is a leak window:

- If the pre-run call increments the count and then its downstream setter returns a non-OK status, the outer `RETURN_IF_ERROR` returns **before** the `RUN_DONE` line, so the decrement is skipped.
- Same outcome if anything between the two calls throws a C++ exception.

This matters because the TIMEOUT relax path is gated on `inflight_run_count_.load() == 0` (`qnn_htp_power_config_manager.cc:540`). A single leaked count means the release timer fires but **never relaxes** — the HTP stays pinned at max voltage/turbo for the rest of the session. That is a silent thermal/power regression (no crash, no error log) that only manifests in sustained/burst mode — exactly the mode this feature exists for, i.e. it inverts the PR's own success criterion ("relax after idle timeout").

The root cause is an asymmetry: the once-per-session **compile** path already guarantees the DONE transition via RAII (`HtpPowerStateGuard`), but the per-inference **hot path** relies on both manual calls happening to run.

> Note: `graphExecute` failure itself does **not** trigger this — its result is returned via the `execute_status` out-param and handled by the caller at `:746`, not early-returned here. The leak window is specifically (a) the pre-run call failing after the increment, or (b) an exception between the two calls.

**Suggested change (minimal, matches existing idiom)**

`qnn_model.cc` already includes `<gsl/gsl>` and the EP uses `gsl::finally` widely. Wrap the relax in a `finally` so the `RUN_DONE` decrement runs on every exit path (early return or exception):

```cpp
auto thread_id = std::this_thread::get_id();
RETURN_IF_ERROR(qnn_backend_manager_->SetPerThreadHtpPowerConfigs(thread_id, true));

// RUN_START incremented inflight_run_count_; guarantee the matching RUN_DONE runs on
// every exit path (early return / exception). Otherwise a leaked count keeps the
// release timer from ever relaxing, pinning the HTP boosted for the rest of the session.
auto power_relax = gsl::finally([&]() {
  auto st = qnn_backend_manager_->SetPerThreadHtpPowerConfigs(thread_id, false);
  if (!st.IsOK()) {
                ("Post-run HTP power relax failed: " + st.GetErrorMessage()).c_str());
  }
});

execute_status = qnn_interface.graphExecute(...);

// ... profiling stop block ...

// (removed the manual SetPerThreadHtpPowerConfigs(thread_id, false) call here — now owned by the finally)
RETURN_IF_ERROR(qnn_backend_manager_->ExtractBackendProfilingInfo(profiling_info));
return Ort::Status();
```

Notes:

- **Delete the original manual `SetPerThreadHtpPowerConfigs(thread_id, false)` line** so the decrement lives in exactly one place. (`SetState(RUN_DONE)` already clamps at 0, so even a double-call couldn't go negative, but keeping one owner is cleaner.)
- The `finally` intentionally warns instead of `RETURN_IF_ERROR` — a scope-exit callback must not throw/return. This is actually *safer* than today: the current `RETURN_IF_ERROR` on the relax line early-returns on failure and skips `ExtractBackendProfilingInfo`.
- One behavioral nuance: `ExtractBackendProfilingInfo` now runs *before* the relax (the finally fires at return). If you need to preserve the original "relax → extract profiling" order, keep the relax on the normal path and use the finally only as a fallback guarded by a `bool run_done_called` flag.

**Please also add a regression test** (ties to the test-coverage gap): unit-test `SetState` directly — drive `RUN_START`, simulate a downstream-setter failure, and assert `inflight_run_count_` returns to 0. This is pure host-side C++ with no device dependency, so it can live in the existing test suite and will actually catch this class of leak.

qti-chuteng (approved):

PR #39, opened 2026-02-05
Deployment of ABI artifacts

kuanyul-qti:
### Description
- Deploy artifacts to Artifactory or external
- Check permission for triggering the workflow
- Combine building artifacts jobs into one file
- Only build nuget and zip once
- Add Python 3.14
- Add is_installer in packages.yml
- Add version_suffix build argument
- Minor fix nightly_build tag

### Motivation and Context
Set up GitHub workflows to publish packages including building and pushing.


.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -23,6 +23,10 @@ on:
         description: "If true, include a verbose version string in wheel file names."
         type: boolean
         default: true
+      version_suffix:
+        description: "Version suffix."
```
qti-jkilpatrick: Could this be a little more descriptive?

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
```diff
@@ -0,0 +1,154 @@
...
+        type: boolean
+        default: true
+      version_suffix:
+        description: "Version suffix for python wheel, nuget and zip."
+        type: string
+        default: ""
+      # Cross-compiling from X64 to ARM64 is faster with our runners, but cannot produce a wheel.
+      windows_arm64_runner_arch:
+        type: string
+        default: "X64"
```
qti-jkilpatrick: This does not appear to be used.

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
```diff
@@ -0,0 +1,154 @@
...
+      nightly_build: ${{ inputs.nightly_build }}
+      version_suffix: ${{ inputs.version_suffix }}
+      target_os: windows
+      target_arch: arm64
+      target_py_vsn: "3.11"
+      build_runner_arch: ARM64
+      test_runner_arch: ARM64
+      run_tests: false
+      # Python wheel ARM64 testing is currently not supported.
+      # test_wheel: true
```
qti-jkilpatrick: Why isn't this supported?

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
```diff
@@ -0,0 +1,154 @@
...
+      upload_test_archive: false
+      upload_wheel: true
+
+  # Disabled due to access denied error when executing the installation command.
+  # Error message: returned non-zero exit status 2147942405
```
qti-jkilpatrick: Installers can't be run from the GitHub Actions service (which is good) so Python needs to be installed manually. I see you reached out to Kyle to do this. Can we uncomment this code now?

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,224 @@
...
+      - name: Download Artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: ORT windows arm64ec Python 3.11 Wheel
+          path: ${{ github.workspace }}/build/dist
+
+      # The wheel built by default is marked as py3-none, but it may be confused with
+      # other wheels and will be replaced by other wheels after downloading all wheels.
+      # Therefore, we need to rename it first to match the target Python version
+      # so that it won't be replaced by the wheel file downloaded afterwards.
```
qti-jkilpatrick: I don't have context on this problem so this is an honest question: why not just fix the filename in the first place instead of renaming it here?

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,224 @@
...
+          twine check ${{ github.workspace }}/build/dist/*
+
+      - name: Publish
+        if: ${{ !inputs.skip_publish }}
+        run: |
+          source uplevel_venv/bin/activate
+          twine upload -r re-artifactory-pypi-test-project \
+            --cert ./qcom/scripts/upleveling/certs/artifactory-ca.pem \
+            --config-file=./qcom/scripts/upleveling/.pypirc \
+            -u ortqnnepci \
```
qti-jkilpatrick: Please make this a secret

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,224 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: ORT windows arm64 Python 3.12 QNN NuGet Package
+          path: ${{ github.workspace }}/build/nuget
+
+      - name: Publish
+        if: ${{ !inputs.skip_publish }}
+        shell: powershell
+        run: |
+          python -m venv uplevel_venv
```
qti-jkilpatrick: On Windows, please use the `py` launcher rather than `python`.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,224 @@
...
+          .\uplevel_venv\Scripts\activate
+          
+          # Install required packages for package_manager.py
+          python -m pip install pip --upgrade
+          python -m pip install certifi pyyaml tqdm requests
```
qti-jkilpatrick: * Please use `uv pip install` so that the packages are cached.
* Please use `qcom/requirements.txt` so this doesn't break if `package_manager.py`'s requirements change.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,224 @@
...
+              -Source $sourceUrl `
+              -username ortqnnepci `
+              -password "${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }}"
+            
+            nuget setApiKey "ortqnnepci:${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }}" -Source $sourceName
+            
+            nuget push $file.FullName -Source $sourceName
+
+            nuget sources Remove -Name $sourceName
+          }
```
qti-jkilpatrick: Maybe all of this should be in a separate file?

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,224 @@
...
+      - name: Publish
+        if: ${{ !inputs.skip_publish }}
+        run: |
+          for file in $(find ${{ github.workspace }}/build/zip -name "*.zip" -type f); do
+            file_basename=$(basename "$file")
+            filename_no_ext="${file_basename%.zip}"
+            version=$(echo $filename_no_ext | cut -d'-' -f3 )
+            curl -T $file \
+              --cacert ./qcom/scripts/upleveling/certs/artifactory-ca.pem \
+              -u ortqnnepci:${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }} \
```
qti-jkilpatrick: Use secret for username here, too.

.github/workflows/qualcomm-internal-upload-artifactory-main.yml
```diff
@@ -0,0 +1,63 @@
...
+  workflow_dispatch:
+    inputs:
+      version_suffix:
+        description: "Version suffix for released package."
+        type: string
+        default: "rc1"
+  # TODO: Remove this before PR merge to mainline. This is only for testing purposes.
+  push:
+      branches:
+        - dev/kuanyul-qti/deployment
```
qti-jkilpatrick: You could also remove this now and simply trigger the workflow manually.

.github/workflows/qualcomm-internal-upload-artifactory-main.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  check-admin-permission:
+    name: Check if user is admin
+    runs-on: ubuntu-latest
```
qti-jkilpatrick: Let's run this on one of our self-hosted runners.

.github/workflows/qualcomm-internal-upload-artifactory-main.yml
```diff
@@ -0,0 +1,63 @@
...
+    name: Publish artifacts
+    needs: check-admin-permission
+    uses: ./.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
+    secrets: inherit
+    with:
+      nightly_build: false
+      skip_publish: false
+      # version_suffix: ${{ inputs.version_suffix }}
+      # TODO: Remove this before PR merge to mainline. This is only for testing purposes.
+      version_suffix: "rc1"
```
qti-jkilpatrick: This is another reason to consider triggering this manually during testing.

.github/workflows/qualcomm-internal-upload-artifactory.yml
```diff
@@ -0,0 +1,202 @@
...
+        description: |
+          To which server to store the artifact. Choose one of
+          ["pypi, testpypi, nuget, testnuget, test-users, test-project, project, public]
+        type: string
+        default: "test-users"
+
+jobs:
+  check-admin-permission:
+    name: Check if user is admin
+    runs-on: ubuntu-latest
```
qti-jkilpatrick: Run on self-hosted runner. (Please always use self-hosted. It's possible I missed one or more of these).

.github/workflows/qualcomm-internal-upload-artifactory.yml
```diff
@@ -0,0 +1,202 @@
...
+
+      - name: Publish
+        shell: powershell
+        run: |
+          python -m venv uplevel_venv
+          .\uplevel_venv\Scripts\activate
+
+          # Install required packages for package_manager.py
+          python -m pip install pip --upgrade
+          python -m pip install certifi pyyaml tqdm requests
```
qti-jkilpatrick: * use uv
* use requirements.txt

.github/workflows/qualcomm-internal-upload-artifactory.yml
```diff
@@ -0,0 +1,202 @@
...
+
+          . ".\qcom\scripts\windows\tools.ps1"
+          . ".\qcom\scripts\windows\utils.ps1"
+          $TargetNugetDir = (Get-NugetBinDir)
+          $env:Path = "$TargetNugetDir;" + $env:Path
```
qti-jkilpatrick: If the goal is to get `nuget.exe` into `$env:Path`, why not have `qnn_ep_uplevel.py` (or whatever is invoking `nuget.exe` use `package_manager.py`? I think there's a Python version of `tools.ps1` available somewhere in this repo.

csharp/OnnxRuntime.CSharp.proj
```diff
@@ -17,11 +17,13 @@ CMake creates a target to this project
     <TargetArchitecture Condition=" '$(TargetArchitecture)' == '' ">x64</TargetArchitecture>
     <IsReleaseBuild Condition=" '$(IsReleaseBuild)' == '' ">false</IsReleaseBuild>
     <ReleaseVersionSuffix></ReleaseVersionSuffix>
+    <VersionSuffix></VersionSuffix>
     <IsWindowsBuild Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</IsWindowsBuild>
     <IsLinuxBuild Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'true'">true</IsLinuxBuild>
     <ExecutionProvider Condition=" '$(ExecutionProvider)' == '' ">None</ExecutionProvider>
     <IsPlatformSpecificSubPackage Condition=" '$(IsPlatformSpecificSubPackage)' == '' ">false</IsPlatformSpecificSubPackage>
     <NuSpecName Condition=" '$(NuSpecName)' == '' ">NativeNuget.nuspec</NuSpecName>
+    <UseNinja Condition=" '$(UseNinja)' == '' ">false</UseNinja>
```
qti-jkilpatrick, line 26: Why?

qcom/scripts/all/package_manager.py
```diff
@@ -230,15 +230,18 @@ def install(self) -> None:
...
-            tools_path = nuget_tool_dir / package_path.name
-            shutil.copyfile(package_path, tools_path)
-        elif package_path.suffix == ".exe":
-            self.__run_installer(package_path)
+        if package_path.suffix == ".exe":
+            is_installer = True
+            if "is_installer" in self.__config:
+                is_installer = self.__config["is_installer"]
+            if is_installer:
+                self.__run_installer(package_path)
```
qti-jkilpatrick: Can this be simplified to something like this?

```python
if self.__config.get("is_installer", True):
    self.__run_installer(package_path)
```

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,224 @@
...
+        run: |
+          source uplevel_venv/bin/activate
+          twine check ${{ github.workspace }}/build/dist/*
+
+      - name: Publish
+        if: ${{ !inputs.skip_publish }}
+        run: |
+          source uplevel_venv/bin/activate
+          twine upload -r re-artifactory-pypi-test-project \
+            --cert ./qcom/scripts/upleveling/certs/artifactory-ca.pem \
```
qti-jkilpatrick, line 108: Isn't the cert file also defined in pypirc?

qcom/scripts/upleveling/create_uplevel_venv.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+python3 -m venv uplevel_venv
+source uplevel_venv/bin/activate
+pip install requests==2.32.3 twine==6.0.1 wheel==0.45.1
```
qti-jkilpatrick: Please use `uv`

qcom/scripts/upleveling/create_uplevel_venv.sh
```diff
@@ -0,0 +1,8 @@
...
+deactivate
```
qti-jkilpatrick: No need to `deactivate` since the environment this script runs in does not persist when it exits.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+from pathlib import Path
+from typing import ClassVar
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+logger = logging.getLogger(__name__)
+
+# Constants
+ARTIFACTORY_USER = "ortqnnepci"
```
qti-jkilpatrick: Get from environment

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+    "zip": "re-artifactory-zip",
+}
+
+ARTIFACT_SUFFIXES = {"wheel": ".whl", "nuget": ".nupkg", "zip": ".zip"}
+
+
+class ConfigManager:
+    """Manages configuration file reading and URL generation."""
+
+    def __init__(self, pypi_rc_file: str, ini_file: str):
```
qti-jkilpatrick, line 49: `Path` instead of `str`?

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+        if not artifact_list:
+            raise RuntimeError(
+                f"Expected to find at least one artifact_file for version {self.args.version_from} "
+                f" to uplevel, got none"
+            )
+
+        # Download each artifact
+        for artifact_file in artifact_list:
+            url_path = f"{url}/{artifact_file}"
+            download_path = os.path.join(download_dir, artifact_file)
```
qti-jkilpatrick, line 171: Use` pathlib.Path` here and elsewhere. (Don't mix Path and traditional file functions in the same file if you can avoid it)

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+
+            logger.info(f"Downloading {artifact_file}")
+            response = requests.get(url_path, auth=auth_credentials, verify=verify)
+
+            if response.status_code != 200:
+                raise RuntimeError(f"Unable to fetch {artifact_file} from {url_path}")
+
+            with open(download_path, "wb") as f:
+                f.write(response.content)
```
qti-jkilpatrick, line 180: If `download_path` becomes a `Path`, this turns into a one-liner.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+    def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir: str) -> None:
+        """Update artifact versions. Must be implemented by subclasses."""
+
+    @abstractmethod
+    def upload_artifacts(self, distribution_dir: str) -> None:
+        """Upload artifacts to repository. Must be implemented by subclasses."""
+
+    def run(self) -> None:
+        """Execute the complete upleveling workflow."""
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
qti-jkilpatrick: Always add a `prefix=` or `suffix=` arg to `TemporaryDirectory()` so that it's obvious where the problem is when leaks happen.

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-jkilpatrick: Actually, it looks like you have a strong preference for `os.path`, etc. Why is `pathlib.Path` imported at all?

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+        return "wheel"
+
+    def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir: str) -> None:
+        """Update wheel package versions."""
+        for wheel_file in artifact_list:
+            logger.info(
+                f"Updating version from {self.args.version_from} to {self.args.version_to} for wheel {wheel_file}"
+            )
+
+            with tempfile.TemporaryDirectory() as tmp_dir:
```
qti-jkilpatrick: Add prefix or suffix

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+
+                # Rename dist-info directory
+                new_dist_info = dist_info_file.replace(self.args.version_from, self.args.version_to)
+                os.rename(
+                    os.path.join(tmp_dir, dist_info_file),
+                    os.path.join(tmp_dir, new_dist_info),
+                )
+
+                # Repack wheel
+                subprocess.run(f"wheel pack {tmp_dir} -d {output_dir}", shell=True, check=True)
```
qti-jkilpatrick: Security: whenever possible, avoid `shell=True` and pass arguments as a list rather than a string.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+        cert_file = "" if is_pypi else f"--cert {ARTIFACTORY_CERTS_FILE}"
+        config_file = "" if is_pypi else f"--config-file {PYPI_RC_FILE}"
+
+        cmd = (
+            f"twine upload {distribution_dir}/* --repository {upload_repository} "
+            f"{cert_file} {config_file} -u {username} -p {password} "
+            f"--verbose --disable-progress-bar"
+        )
+
+        subprocess.run(cmd, shell=True, check=True)
```
qti-jkilpatrick: Pass list; avoid shell=true

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+            subprocess.run(["nuget", "sources", "Remove", "-Name", source_name], check=True, shell=True)
+
+    def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir: str) -> None:
+        """Update NuGet package versions."""
+        for nuget_file in artifact_list:
+            logger.info(
+                f"Updating version from {self.args.version_from} to {self.args.version_to} for nuget {nuget_file}"
+            )
+
+            with tempfile.TemporaryDirectory() as tmp_dir:
```
qti-jkilpatrick: prefix/suffix

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+                output_dir,
+                os.path.basename(zip_path.replace(self.args.version_from, self.args.version_to)),
+            )
+
+            subprocess.run(
+                f"cp {zip_path} {updated_zip_path}",
+                shell=True,
+                check=True,
+                stdout=subprocess.DEVNULL,
+            )
```
qti-jkilpatrick: Use `shutil` or similar? Maybe I'm missing something, but I don't understand why this needs to be a shell call to `cp`.

qti-jkilpatrick: This is also a really great illustration of why it's important to pass commands as lists rather than strings: if there are spaces in either path, this will behave unexpectedly.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+            for f in os.listdir(distribution_dir)
+            if os.path.isfile(os.path.join(distribution_dir, f)) and f.endswith(self.artifact_suffix)
+        ]
+
+        # Use url_to_display for ZIP uploads
+        for zip_file in zip_files:
+            zip_path = os.path.join(distribution_dir, zip_file)
+            upload_url = os.path.join(self.url_to_display, zip_file)
+            cmd = f"curl -T {zip_path} {cert_file} -u {username}:{password} {upload_url}"
+            subprocess.run(cmd, shell=True, check=True)
```
qti-jkilpatrick: * Pass args as a list
* Does curl have an option to read the password? If so, let's use it so that it can't be discovered by running `ps` at the right moment.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,673 @@
...
+import zipfile
+from abc import ABC, abstractmethod
+from configparser import ConfigParser
+from pathlib import Path
+from typing import ClassVar
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+logger = logging.getLogger(__name__)
```
qti-jkilpatrick: I see you use this for logging throughout. Why not just use `logging.info` etc.?

qcom/scripts/upleveling/update_wheel_py_version.py
```diff
@@ -0,0 +1,109 @@
...
+        py_none_str = "py3-none"
+        py_version_without_dot = py_version.replace(".", "")
+        py_version_str = f"cp{py_version_without_dot}-cp{py_version_without_dot}"
+
+        if py_none_str not in wheel_path:
+            continue
+
+        logger.info(f"Updating python version from {py_none_str} to {py_version_str} for wheel {wheel_path}")
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
qti-jkilpatrick: prefix/suffix

qcom/scripts/upleveling/update_wheel_py_version.py
```diff
@@ -0,0 +1,109 @@
...
+import tempfile
+import zipfile
+
+logger = logging.getLogger(__name__)
+
+
+def parse_arguments() -> argparse.Namespace:
+    """Parse and validate command line arguments."""
+    parser = argparse.ArgumentParser(
+        description="Repack wheels with python version updating support.",
```
qti-jkilpatrick: Sorry, I don't quite understand this. What support is being updated?

qcom/scripts/upleveling/update_wheel_py_version.py
```diff
@@ -0,0 +1,109 @@
...
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
+
+    parser.add_argument(
+        "--input_dir",
```
qti-jkilpatrick: In other parts of the build, we always use dashes in arguments rather than underscores.

qcom/scripts/upleveling/update_wheel_py_version.py
```diff
@@ -0,0 +1,109 @@
...
+            wheel_file = os.path.join(tmp_dir, dist_info_file, "WHEEL")
+
+            # Update METADATA
+            with open(wheel_file) as f:
+                wheel = f.read()
+            with open(wheel_file, "w") as f:
+                f.write(wheel.replace(py_none_str, py_version_str))
+
+            # Repack wheel
+            subprocess.run(f"wheel pack {tmp_dir} -d {input_dir}", shell=True, check=True)
```
qti-jkilpatrick: list, not string; no shell=true

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
kuanyul-qti: Removed.

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
kuanyul-qti: Added.

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
kuanyul-qti: I haven't investigated it. It is currently broken as shown in https://github.com/onnxruntime/onnxruntime-qnn/blob/main/.github/workflows/qualcomm-internal-build-and-test.yml#L97
There is your message on https://github.com/onnxruntime/onnxruntime-qnn/pull/75#discussion_r2855107178

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
kuanyul-qti: I see, thanks for the context. Yes, I've reached out to Kyle, but I didn't get the feedback on this. It seems that the issue is not resolved as shown in the latest run - https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/22426980762/job/64937216308
Hi @qti-jkilpatrick , could you help install python 3.14 ARM and X86 manually on device? Thanks.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: The problem is that
(Take arm64 as an example)
- For py311 wheel, the wheel name is "onnxruntime_qnn-2.0.0rc1-py3-none-win_arm64.whl".
- For py312 wheel, the wheel name is "onnxruntime_qnn-2.0.0rc1-py3-none-win_arm64.whl".

If we download these two wheels directly without changing their names, the one downloaded earlier will be replaced by the one downloaded later.
After calling the script, the wheel will be changed to "onnxruntime_qnn-2.0.0rc1-cp311-cp311-win_arm64.whl" for py311 wheel.

Move the fix to .github/workflows/qualcomm-internal-build-and-test-single-os.yml

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Yes, but if `--cert` is not specified here, the upload fails due to [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1007)')

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done. Added as ARTIFACTORY_USERNAME.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: We need to get nuget files from `${{ github.workspace }}\build\nuget`. Not sure if we can do this in a separate file.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory-main.yml
kuanyul-qti: Temporarily enable this as a test workaround for the issue that we can't manually trigger the workflow until the file is merged into mainline. I'll remove this after testing.
For your reference, we can manually trigger this https://github.com/onnxruntime/onnxruntime-qnn/actions/workflows/qualcomm-internal-release.yml
But there is no "Run workflow" button in https://github.com/onnxruntime/onnxruntime-qnn/actions/workflows/qualcomm-internal-upload-artifactory-main.yml

.github/workflows/qualcomm-internal-upload-artifactory-main.yml
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory-main.yml
kuanyul-qti: Temporarily enable this as a test workaround for the issue that we can't manually trigger the workflow until the file is merged into mainline. I'll remove this after testing.

.github/workflows/qualcomm-internal-upload-artifactory.yml
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory.yml
kuanyul-qti: Done

csharp/OnnxRuntime.CSharp.proj
kuanyul-qti: The `NativeBuildOutputDir` is `$(OnnxRuntimeBuildDirectory)\$(Configuration)\$(Configuration)` with Visual Studio. Currently, however, CI arm device builds ORT using ninja, and the `NativeBuildOutputDir` is `$(OnnxRuntimeBuildDirectory)\$(Configuration)` when using ninja, so we need to add an argument to identify whether ninja is used.

qcom/scripts/all/package_manager.py
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory.yml
kuanyul-qti: We'd like to make `qnn_ep_uplevel.py` a general-purpose script, which means it can be used for other products such as qairt-visualier. For this purpose, I prefer not to use `package_manager.py` directly within `qnn_ep_uplevel.py`; instead, the required environment should be set up before calling the script.

qcom/scripts/upleveling/create_uplevel_venv.sh
kuanyul-qti: Done

qcom/scripts/upleveling/create_uplevel_venv.sh
kuanyul-qti: Removed.

qcom/scripts/upleveling/update_wheel_py_version.py
kuanyul-qti: It aims to update the py version in wheel name. For example, from "onnxruntime_qnn-2.0.0rc1-py3-none-win_arm64.whl" to "onnxruntime_qnn-2.0.0rc1-cp311-cp311-win_arm64.whl"

qcom/scripts/upleveling/update_wheel_py_version.py
kuanyul-qti: Changed to dash.

qcom/scripts/upleveling/update_wheel_py_version.py
kuanyul-qti: Done

qcom/scripts/upleveling/update_wheel_py_version.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Using named logger has the following advantages:

- Granular control: You can set different log levels for different modules
- Better debugging: Log messages show which module they came from

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: It is str since we use os.path.

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: - Done
- No, the password should be given by -u user:password. When running the command, the password will not be displayed; instead, it appears as ***.

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Get rid of pathlib.Path and use os.path for consistency.

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Get rid of pathlib.Path and use os.path for consistency.

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
qti-jkilpatrick: Is the intention to eventually run tests here? If so, what is the plan to get them enabled?

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
qti-jkilpatrick: Done

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
qti-jkilpatrick: But why doesn't the build produce `onnxruntime_qnn-2.0.0rc1-cp311-cp311-win_arm64.whl` in the first place? Doesn't `twine` do this for us?

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
qti-jkilpatrick: Pass them as parameters to the script

.github/workflows/qualcomm-internal-upload-artifactory-main.yml
qti-jkilpatrick: Ah good point.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,674 @@
...
+        is_pypi = self.credential_manager.is_pypi_index(upload_repository)
+        cert_file = "" if is_pypi else f"--cert {ARTIFACTORY_CERTS_FILE}"
+        config_file = "" if is_pypi else f"--config-file {PYPI_RC_FILE}"
+
+        cmd_str = (
+            f"twine upload {distribution_dir}/* --repository {upload_repository} "
+            f"{cert_file} {config_file} -u {username} -p {password} "
+            f"--verbose --disable-progress-bar"
+        )
+        cmd = cmd_str.split()
```
qti-jkilpatrick: This replicates the problem of the shell potentially splitting arguments on whitespace. Please build the command as a list, don't just split the string. I understand this means you might need to use `glob` to identify files to upload.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,674 @@
...
+        ]
+
+        # Use the source name that was registered in _add_nuget_sources
+        source_name = self.upload_source_name
+        if not source_name:
+            raise RuntimeError("Upload source name not set. Ensure _add_nuget_sources was called first.")
+
+        for nuget_file in nuget_files:
+            cmd_str = f"nuget push {nuget_file} -Source {source_name}"
+            cmd = cmd_str.split()
```
qti-jkilpatrick: Construct the command as a list. Do not use split.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,674 @@
...
+            if os.path.isfile(os.path.join(distribution_dir, f)) and f.endswith(self.artifact_suffix)
+        ]
+
+        # Use url_to_display for ZIP uploads
+        for zip_file in zip_files:
+            zip_path = os.path.join(distribution_dir, zip_file)
+            upload_url = os.path.join(self.url_to_display, zip_file)
+            cmd_str = f"curl -T {zip_path} {cert_file} -u {username}:{password} {upload_url}"
+            cmd = cmd_str.split()
+            subprocess.run(cmd, check=True)
```
qti-jkilpatrick: Construct the command using a list.

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-jkilpatrick: I am not concerned about the terminal output while running the script, since there is not interactive terminal. Log into a Linux box and run `ps x`. Observe that command arguments are visible. To avoid this, curl provides a few ways to read passwords from places other than the command line as discussed [here](https://everything.curl.dev/cmdline/passwords.html) and [here](https://stackoverflow.com/questions/2594880/using-curl-with-a-username-and-password).

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-jkilpatrick: Sure, but this is a just a standalone script (the factory function even takes the output of argparse, suggesting it's not meant to be embedded into something bigger). I'm not going to demand you change this, but it seems more complicated than it needs to be.

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
qti-jkilpatrick: As I think about this more, it seems like we really should be testing the wheels here since this is the release path. Enabling this should probably be tracked by a Jira so we don't forget.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,194 @@
...
+            --config-file=./qcom/scripts/upleveling/.pypirc \
+            -u ${{ secrets.ARTIFACTORY_USERNAME }} \
+            -p ${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }} \
```
qti-jkilpatrick: This is another place where we should not be passing a password on the command line. Fortunately, it looks like twine [has some alternatives](https://twine.readthedocs.io/en/stable/).

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,194 @@
...
+            Write-Host "${fileBasename}: $version"
+
+            $sourceName = "re-artifactory-nuget-test-project-$version"
+            $sourceUrl = "https://re-artifactory.qualcomm.com/artifactory/api/nuget/aisw-nuget-test-project/onnxruntime-qnn/$version"
+
+            nuget sources Add `
+              -Name $sourceName `
+              -Source $sourceUrl `
+              -username ${{ secrets.ARTIFACTORY_USERNAME }} `
+              -password "${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }}"
```
qti-jkilpatrick: Another spot to not pass passwords as command line arguments. Some pages on the topic:

* https://learn.microsoft.com/en-us/nuget/consume-packages/consuming-packages-authenticated-feeds#credentials-in-environment-variables
* https://stackoverflow.com/questions/77262438/hide-clear-text-password-in-nuget-config-from-source-control#:~:text=Option%203:%20Use%20NuGet's%20convention,=username;Password=password%20.
* https://www.reddit.com/r/dotnet/comments/1ixhxsi/private_nuget_feed_passing_credentials/#:~:text=nuget%20source%20add%20is%20the,these%20in%20the%20command%20line.&text=For%20build%20environment%2C%20I%20create,text%20and%20as%20environment%20variables.&text=Thanks%2C%20will%20give%20this%20a%20go!&text=Be%20very%20careful%20with%20this,own%20credentials%20to%20the%20codebase.&text=Yup%2C%20that%20is%20why%20the,config.&text=EDIT:%20Ok%2C%20so%20took%20a%20little%20longer.&text=In%20your%20pipeline%20you%20need,library%20with%20key%20vault%20backing.&text=Thanks%20m8%2C%20I%20owe%20you%20a%20beer.&text=I%20think%20the%20solution%20by%20u/revanchist11%20is%20smarter%20btw%20:D

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -0,0 +1,194 @@
...
+      - name: Publish
+        if: ${{ !inputs.skip_publish }}
+        run: |
+          for file in $(find ${{ github.workspace }}/build/zip -name "*.zip" -type f); do
+            file_basename=$(basename "$file")
+            filename_no_ext="${file_basename%.zip}"
+            version=$(echo $filename_no_ext | cut -d'-' -f3 )
+            curl -T $file \
+              --cacert ./qcom/scripts/upleveling/certs/artifactory-ca.pem \
+              -u ${{ secrets.ARTIFACTORY_USERNAME }}:${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }} \
```
qti-jkilpatrick: Don't pass the password as a command line argument

.github/workflows/qualcomm-internal-upload-artifactory.yml
qti-jkilpatrick: Okay, then what if you made something like `qnn_ep_uplevel.ps1` that handled this?

I really don't like seeing scripts in GitHub workflows because they're so hard to debug. It's one of the main reasons we wrote `build_and_test.py` on my last team and we took the time to recreate on this one. As someone contributing to the build, life is better when you can reproduce as much of it as possible on your local machine.

qcom/scripts/upleveling/create_uplevel_venv.ps1
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+py -m venv uplevel_venv
+.\uplevel_venv\Scripts\activate
+# Use the Python from the activated virtual environment instead of the system Python.
```
qti-jkilpatrick: I don't think this comment is necessary.

qcom/scripts/upleveling/create_uplevel_venv.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+python3 -m venv uplevel_venv
+source uplevel_venv/bin/activate
+pip install uv
+uv pip install requests==2.32.3 twine==6.0.1 wheel==0.45.1
```
qti-jkilpatrick: Rather than repeating the same list of specifically versioned requirements, create a `requirements.txt` and use it both places.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,674 @@
...
+            actual_source_url = "https://api.nuget.org/v3/index.json"
+            is_public_nuget = True
+
+        # Clean up first
+        subprocess.run(["nuget", "sources", "Remove", "-Name", source_name], check=False)
+
+        # Add the source
+        add_source_args = ["nuget", "sources", "Add", "-Name", source_name, "-Source", actual_source_url]
+        if not is_public_nuget:
+            add_source_args.extend(["-username", username, "-password", password])
```
qti-jkilpatrick: don't pass the password on the command line

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,674 @@
...
+
+        logger.info(f"Adding source {actual_source_url} with name {source_name}")
+        subprocess.run(add_source_args, check=True)
+
+        # Set API Key
+        logger.info(f"Setting API Key for {source_name}")
+        api_key_value = password if is_public_nuget else f"{username}:{password}"
+        subprocess.run(["nuget", "setApiKey", api_key_value, "-Source", source_name], check=True)
```
qti-jkilpatrick: Can we avoid passing the password via the command line? I didn't dig into this particular usage, but I suspect Microsoft has thought about this one, too.

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-jkilpatrick: don't pass the password on the command line

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -0,0 +1,674 @@
...
+        "test-users, test-project, project, public]",
+    )
+    parser.add_argument(
+        "--artifactory_user",
+        type=str,
+        default="",
+        help="Artifactory username.",
+    )
+    parser.add_argument(
+        "--artifactory_api_key",
```
qti-jkilpatrick: Related to passing passwords on the command line, maybe we shouldn't be passing sensitive API keys as command line arguments either.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
qti-jkilpatrick: The wheel's filename being incorrect in the build appears to be what is breaking the python wheel tests (or at least that's one of the issues). Please, let's create the wheel with the correct name so we can fix both problems and simplify deployment.

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
qti-jkilpatrick: I'm fixing smoke tests in #73.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
qti-jkilpatrick: I'm fixing this in https://github.com/onnxruntime/onnxruntime-qnn/pull/73/changes#r2861985182.

kuanyul-qti: Removed this part as the python version of wheel is fixed in mainline.

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory.yml
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done by passing `--netrc-file`

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done by removing `logging.getLogger(__name__)`

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
kuanyul-qti: Done

qcom/scripts/upleveling/create_uplevel_venv.ps1
kuanyul-qti: Removed.

qcom/scripts/upleveling/create_uplevel_venv.sh
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: Done

.github/workflows/qualcomm-internal-build-wheels-nugets-zips.yml
kuanyul-qti: Open test_wheel

qcom/scripts/upleveling/upload_zip_to_artifactory.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+# Script to upload zip files to Artifactory using netrc authentication
+# Usage: upload_zip_to_artifactory.sh <input_dir> <netrc_file>
+
+set -e
```
qti-jkilpatrick: [Consider using](http://redsymbol.net/articles/unofficial-bash-strict-mode/)

```bash
set -euo pipefail
```

qcom/scripts/upleveling/upload_zip_to_artifactory.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ ! -f "$NETRC_FILE" ]; then
+    echo "Error: Netrc file '$NETRC_FILE' does not exist"
+    exit 1
+fi
+
+# Get the repository root directory
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
+# Upload files using --netrc-file
+for file in $(find "$INPUT_DIR" -name "*.zip" -type f); do
```
qti-jkilpatrick: Handle files with spaces in the path

qcom/scripts/upleveling/upload_nupkg_to_artifactory.ps1
```diff
@@ -0,0 +1,49 @@
...
+    $sourceName = "re-artifactory-nuget-test-project-$version"
+    $sourceUrl = "https://re-artifactory.qualcomm.com/artifactory/api/nuget/aisw-nuget-test-project/onnxruntime-qnn/$version"
+
+    # Use PackageSourceCredentials environment variables for secure authentication
+    # NuGet automatically reads credentials from environment variables in the format:
+    # NuGetPackageSourceCredentials_<SourceName> with value "Username={Username};Password={Password}"
+    # Sanitize source name to make it valid for environment variable names (alphanumeric and underscores only)
+    $sourceNameSafe = $sourceName -replace '[^a-zA-Z0-9]', '_'
+    $sourceCredentialsVar = "NuGetPackageSourceCredentials_${sourceNameSafe}"
+    Set-Item -Path "env:$sourceCredentialsVar" -Value "Username=$env:ARTIFACTORY_USERNAME;Password=$env:ARTIFACTORY_PASSWORD"
```
qti-jkilpatrick, line 36: Does this need to be cleaned up?

qcom/scripts/upleveling/upload_nupkg_to_artifactory.ps1
```diff
@@ -0,0 +1,49 @@
...
+        -Name $sourceName `
+        -Source $sourceUrl
+
+    nuget push $file.FullName -Source $sourceName
+
+    nuget sources Remove -Name $sourceName
+
+    # Clean up environment variables
+    Remove-Item -Path "env:$usernameVarName" -ErrorAction SilentlyContinue
+    Remove-Item -Path "env:$passwordVarName" -ErrorAction SilentlyContinue
```
qti-jkilpatrick: How/where are `$usernameVarName` and `$passwordVarName` defined?

qti-jkilpatrick (commented):
Excellent revision!

qcom/scripts/upleveling/upload_nupkg_to_artifactory.ps1
kuanyul-qti: Done

qcom/scripts/upleveling/upload_nupkg_to_artifactory.ps1
kuanyul-qti: Removed.

qcom/scripts/upleveling/upload_zip_to_artifactory.sh
kuanyul-qti: Done

qcom/scripts/upleveling/upload_zip_to_artifactory.sh
kuanyul-qti: Done

qti-jkilpatrick (approved):
I don't know if Kyle had any additional comments, but this looks good to me. Thank you for your patience.

PR #52, opened 2026-02-17
Fix HTP Pad to support fp16 when pad_constant_value=0.

quic-muchhsu:
### Description

Add support on HTP for fp16 pad_constant_value.
Also fix some unit tests that use pad_constant_value but `has_constant_value_input` was set to false.

### Motivation and Context

fp16 pad_constant_value wasn't supported at all and will fall back to cpu.
Though HTP doesn't support fp16 pad_constant_value, but it use default 0. 
This change (when pad_constant_value is fp16) will reject pad_constant_value != 0 and let pad_constant_value == 0 cases to pass.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
@@ -289,17 +289,16 @@ Ort::Status PadOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model
     TensorInfo input_info = {};
     RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[2], input_info));
     // Pad doesn't support QNN_DATATYPE_FLOAT_16 pad_constant_value.
-    if (input_info.qnn_data_type != QNN_DATATYPE_FLOAT_16){
+    if (input_info.qnn_data_type != QNN_DATATYPE_FLOAT_16) {
       // Process optional input constant_value
       RETURN_IF_ERROR(ProcessConstantValue(qnn_model_wrapper, param_tensor_names, node_unit, inputs[2]));
-    }else{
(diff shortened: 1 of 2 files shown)


qti-shubham (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
```diff
@@ -286,8 +286,19 @@ Ort::Status PadOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model
...
-    // Process optional input constant_value
-    RETURN_IF_ERROR(ProcessConstantValue(qnn_model_wrapper, param_tensor_names, node_unit, inputs[2]));
+    TensorInfo input_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[2], input_info));
+    // Pad doesn't support QNN_DATATYPE_FLOAT_16 pad_constant_value.
+    if (input_info.qnn_data_type != QNN_DATATYPE_FLOAT_16) {
+      // Process optional input constant_value
+      RETURN_IF_ERROR(ProcessConstantValue(qnn_model_wrapper, param_tensor_names, node_unit, inputs[2]));
+    } else {
+      // when dtype == QNN_DATATYPE_FLOAT_16, HTP doesn't support specify pad_constant_value. Use default 0.
```
qti-mattsinc: If this is an HTP-specific constraint, please restrict the added logic behind an `is_npu_backend` check. As it stands it looks like this would affect other backends even if they do support fp16 `pad_constant_value`.

qti-mattsinc (changes requested):

onnxruntime/core/providers/qnn/builder/opbuilder/pad_op_builder.cc
quic-muchhsu: Good point. I just checked doc for `pad_constant_value`, looks like none of the processers supports fp16. This logic should be safe for backends. I will update the comment.

image
image
image
image

qti-mattsinc (approved):

PR #4, opened 2026-01-28
[QNN EP] Add 'mode:cubic' attribute support to Resize op

qti-shubham:
### Description
* Add 'mode:cubic' attribute support to Resize op
* Add corresponding unit tests

### Motivation and Context
- Currently DLC generation fails with error : `resize_op_builder.cc:130 IsOpSupported QNN EP: Resize does not support mode cubic`
- Adds support for `efficientnet_l2_seg` model (https://dev.aihub.qualcomm.com/jobs/jpyrxz845/)
image

PR #53, opened 2026-02-18
[QNN EP] Resize 'mode:cubic' attribute support with Scale check

qti-shubham:
### Description
Resubmission of [PR 4](https://github.com/onnxruntime/onnxruntime-qnn/pull/4) with following constraints:
If Resize op contains `scales` as input and then check if it matches QNN auto generated scales.

### Motivation and Context
https://github.com/onnxruntime/onnxruntime-qnn/pull/4 failed as QNN Resize only ingest 'input[0]' and auto-generates other inputs based on output shape. 

`onnxruntime_provider_test --gtest_filter=ResizeOpTest.ResizeOpCubicDownSampleTest`
```
https://docs.qualcomm.com/doc/80-63442-10/topic/MasterOpDef.html#resize
The QNN Resize op can only take in[0]
There is no provision to pass [ROI, Scales, Sizes].
 
QNN Scales are auto-calculated as = out[0]/in[0] for each axis.
 
For the failing test case, QNN computes Scales = [1, 1, 0.75, 0.75],
however actual Scales passed = [1, 1, 0.8, 0.8]
 
This is resulting in accuracy difference between ORT and QNN.
```

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
@@ -3,6 +3,7 @@
 
 #include <array>
 #include <cassert>
+#include <cmath>
 #include <unordered_map>
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
@@ -12,6 +13,11 @@
 
 namespace onnxruntime {
 namespace qnn {
+
+namespace {
+constexpr const char* kNpuBF16InterpolationLimitations =
+    "QNN EP: Resize support BF16 inputs with linear mode only on the NPU.";
+}
 
 class ResizeOpBuilder : public BaseOpBuilder {
  public:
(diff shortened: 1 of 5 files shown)


qti-yuduo (approved):
💯

onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
```diff
@@ -218,6 +227,53 @@ Ort::Status ResizeOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
   RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(output_0.shape, output_shape),
                 "QNN EP: Cannot get shape for Resize output");
 
+  if (interp_mode == "cubic") {
```
qti-shubham, line 230: additional check added here.
rest of the code is same as previous PR.

PR #59, opened 2026-02-19
[QNN EP] Add Support for GroupNorm Operator in QNN EP

qti-ashwshan:
### Description
- Add support for GroupNorm Op
- Add unit tests to run on HTP

### Motivation and Context
- QNN EP does not support GroupNorm Op


qti-yuduo (approved):
Would be good to add a JIRA to trace follow up work on QDQ support
.

PR #62, opened 2026-02-19
[NFC] Make filename consistent

qti-yuduo:
## Motivation:

- Make op test filename consistent: removed `_htp_`, `_cpu_` from filename as they always contain both.
- Align naming consistent with ONNX ops.


qti-ashwshan (approved):
Nice!

PR #63, opened 2026-02-19
Add comment trigger validation and remove push trigger

qti-kromero:
### Description
* Add admin ability to comment '/test' and have CI start
* Add workflow_dispatch ability to emulate above issue_comment trigger
* Remove automatic running of CI on push
* Note: Breaks old branch checks - need to remake after merging this
* Note: Checks tab on PRs is broken for the time being, change includes debugging code for this

### Motivation and Context
* See above


quic-muchhsu (approved):
Approving this PR as it doesn’t appear to touch any functional codebase and in case you want to merge it.

qti-yuduo:
> Add admin ability to comment '/test' and have CI start

Is `/test` only effective for admin? Other users would have to wait for admin manual trigger? 

It make sense if we have many coming in PRs, but would really delay for us usually. Anyway we could have a set of users that still auto-trigger?

qti-kromero:
> > Add admin ability to comment '/test' and have CI start
> 
> Is `/test` only effective for admin? Other users would have to wait for admin manual trigger?
> 
> It make sense if we have many coming in PRs, but would really delay for us usually. Anyway we could have a set of users that still auto-trigger?

our options are unfortunately fairly limited for who we can restrict this to - it would actually be only people who are members of the onnxruntime organization as I understand it currently. which is basically just Jeff and I. however, for the time being anyone with write permissions should also be able to invoke CI on PRs from the workflow_dispatch location on the actions tab.

we will definitely have a period of adjustment associated with this change where we need to figure out the best balance between security and efficiency. the policy I'm opting for right now is to err on more secure policies that we can relax over time

qti-kromero:
> > Add admin ability to comment '/test' and have CI start
> 
> Is `/test` only effective for admin? Other users would have to wait for admin manual trigger?
> 
> It make sense if we have many coming in PRs, but would really delay for us usually. Anyway we could have a set of users that still auto-trigger?



> > Add admin ability to comment '/test' and have CI start
> 
> Is `/test` only effective for admin? Other users would have to wait for admin manual trigger?
> 
> It make sense if we have many coming in PRs, but would really delay for us usually. Anyway we could have a set of users that still auto-trigger?

this comment may have actually triggered the workflow so this might be a bug

qti-kromero:
jeff wants to make repo public and then merge this for only external collaborator use case

qti-jkilpatrick:
I think we're abandoning this in favor of the `Approve workflows to run` button.

qti-kromero:
closed as tested it's not needed after making repo public

PR #64, opened 2026-02-20
Avoid CAST(FP16/32) -> Q -> DQ fusion as QNN Convert op

qti-shubham:
### Description
CAST -> Q -> DQ fusion as CONVERT op should be avoided for Float16/32 Cast outputs, 
instead CAST -> QUANTIZE should be used.

### Motivation and Context
Fixes https://github.com/qcom-ai-hub/tetracode/issues/16559 

Model has following topology:
INT32 -> Cast -> Fp32 -> Q -> DQ -> UFXP16

This was resulting in : INT32 -> Convert -> UFXP16 , but QNN Convert op don't support INT32 to UFXP16

Added check in `cast_lone_q_fusion.cc` to avoid fusion if Casting to FP16/32 type as it should be fold subsequent QDQ to Quantize op

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/cast_lone_q_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/cast_lone_q_fusion.cc
@@ -13,6 +13,22 @@ namespace qnn {
 constexpr char kOpCast[] = "Cast";
 constexpr char kOpConvert[] = "Convert";
 
+static bool IsQnnConvertInputSupported(Qnn_DataType_t data_type) {
+  switch (data_type) {
+    case QNN_DATATYPE_FLOAT_16:
+    case QNN_DATATYPE_FLOAT_32:
+    case QNN_DATATYPE_BOOL_8:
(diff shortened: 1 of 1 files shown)


onnxruntime/core/providers/qnn/builder/qnn_node_group/cast_lone_q_fusion.cc
```diff
@@ -82,6 +82,15 @@ std::unique_ptr<IQnnNodeGroup> CastLoneQFusion::TryFusion(
   if (qnn_model_wrapper.IsConstantInput(cast_node_unit.Inputs()[0].name)) {
     return nullptr;
   }
+
+  TensorInfo cast_input_info = {};
+  if (!qnn_model_wrapper.GetTensorInfo(cast_node_unit.Inputs()[0], cast_input_info).IsOK()) {
```
quic-ykota, line 103: Should this be querying the outputs of cast_node_unit instead? I guess the Inputs()[0] would be of type INT32 and Outputs()[0] FP32 in your example.

qti-yuduo (approved):

PR #66, opened 2026-02-21
Fix negative-index shared Gather indices naming

qti-shubham:
### Description
Make negative-index Gather initializers use deterministic, axis-specific tensor names so the positive converted QNN indices are not reused across different dimension sizes Gather ops.

### Motivation and Context
Fixes https://github.com/qcom-ai-hub/tetracode/issues/17329

- In the given customer model, some Gathers failed QNN validation because they shared indices initializer with -1 which was converted to Positive based on Axis Dimensions, leading to out-of-range indices.

- scenario:
- - Indices tensor with value -1 used by Gather input0 shapes (16,1,384), (8,1,384) and (4,1,384) should map to 15, 7 and 3 respectively
- - However, due to common QNN name, it generate 15, 15, 15 for all Gathers resulting in QNN validation failure ( probably due to out-of-bound )

image

- The change ensures each Gather uses the correct neg-to-pos conversion while still reusing tensors across passes.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
@@ -72,7 +72,8 @@ Ort::Status GatherOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
 template <typename SrcType, typename DstType>
 static bool FixStaticIndices(const std::vector<uint8_t>& onnx_bytes,
                              int64_t input0_axis_dim,
-                             /*out*/ std::vector<uint8_t>& qnn_bytes) {
+                             /*out*/ std::vector<uint8_t>& qnn_bytes,
+                             /*out*/ bool* had_negative_indices) {
   const size_t num_elems = onnx_bytes.size() / sizeof(SrcType);
   gsl::span<const SrcType> onnx_indices{reinterpret_cast<const SrcType*>(onnx_bytes.data()), num_elems};
 
@@ -84,6 +85,9 @@ static bool FixStaticIndices(const std::vector<uint8_t>& onnx_bytes,
 
     // Try to make a negative index positive by adding rank.
     if (onnx_index < 0) {
+      if (had_negative_indices != nullptr) {
+        *had_negative_indices = true;
+      }
       onnx_index += static_cast<SrcType>(input0_axis_dim);
     }
 
@@ -130,6 +134,7 @@ static Ort::Status GetInput0AxisDimValue(const QnnModelWrapper& qnn_model_wrappe
 // The HTP backend only supports dynamic int64 indices if they are a graph input.
 static Ort::Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
                                        const OrtNodeUnitIODef& indices_input,
+                                       const std::string& gather_input0_name,
                                        int64_t input0_axis_dim,
                                        const Ort::Logger& logger,
                                        std::vector<std::string>& input_names,
@@ -140,6 +145,7 @@ static Ort::Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
   RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(indices_input, indices_info));
 
   std::vector<uint8_t> qnn_indices_bytes;
+  bool had_negative_indices = false;
 
   // A. Get raw bytes for static indices.
   //    If indices are int64, convert them to int32 and update indices_info.qnn_data_type.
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
```diff
@@ -148,33 +154,48 @@ static Ort::Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
...
   std::vector<uint32_t> cast_output_shape(indices_info.shape);
-  if (qnn_model_wrapper.IsQnnTensorWrapperExist(indices_tensor_name)) {
-    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, ("Tensor already added, skip it: " + indices_tensor_name).c_str());
+  std::string indices_tensor_name_to_use = indices_tensor_name;
+  // If static indices have negative values, use a deterministic neg-to-pos tensor name
+  // that is stable across graph passes and unique per Gather: input0 + axis size.
+  // Note: Using 'GetUniqueName' here causes replication of same indices across passes.
+  if (indices_info.is_initializer && had_negative_indices) {
+    indices_tensor_name_to_use = indices_tensor_name + "_neg2pos_axis" + std::to_string(input0_axis_dim) +
+                                 "_" + gather_input0_name;
```
qti-yuduo: Looks like using `node_unit.Name()` would be sufficient -- that'd be deterministic across passes and still unique to a node? And thus, no need to pass in `gather_input0_name`?

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
```diff
@@ -148,33 +154,48 @@ static Ort::Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
...
                     "QNN does not support negative index values for Gather* ops");
     } else {
       qnn_indices_bytes = std::move(onnx_indices_bytes);
     }
   }
 
   std::vector<uint32_t> cast_output_shape(indices_info.shape);
-  if (qnn_model_wrapper.IsQnnTensorWrapperExist(indices_tensor_name)) {
-    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, ("Tensor already added, skip it: " + indices_tensor_name).c_str());
+  std::string indices_tensor_name_to_use = indices_tensor_name;
```
qti-yuduo: nit: do we have to rename this variable? Just override it if possible.

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
```diff
@@ -140,6 +145,7 @@ static Ort::Status ProcessIndicesInput(QnnModelWrapper& qnn_model_wrapper,
   RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(indices_input, indices_info));
 
   std::vector<uint8_t> qnn_indices_bytes;
+  bool had_negative_indices = false;
```
qti-yuduo: nit: `has_negative_indices` is more common.

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
```diff
@@ -72,7 +72,8 @@ Ort::Status GatherOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
 template <typename SrcType, typename DstType>
 static bool FixStaticIndices(const std::vector<uint8_t>& onnx_bytes,
```
qti-yuduo: Not yours, but `FixStaticIndices` doesn't indicate what was the fix.. if you do another change, perhaps think about better / more explicit naming.

qti-yuduo:
Looks reasonable, if it's not too complicated, let's add some unit tests covering these corner cases. Claude should be good at that..

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
qti-shubham: Yuduo, I feel `node_unit.Name()` is not a safe unique key, as ONNX specs allows empty / non-unique node names, while Tensor names must be unique within the graph, so I chose `gather_input0_name`

onnxruntime/core/providers/qnn/builder/opbuilder/gather_op_builder.cc
qti-shubham: yes, done. copied`const auto&` into mutable string.

qti-yuduo (approved):

PR #68, opened 2026-02-23
Fix order-dependent tests; Enforce SoC when opt-in FP16

qti-yuduo:
## Motivation

Test case now is order-dependent, it would fail if order isn't preserved. This is because SDK has cache and not stateless. 

## Changes:

- Enforce SoC (can't be unknown) when `enable_htp_fp16_precision` is set; This allow fail fast. Less confusing. (this is consistent with BF16 support).
- Add missing Windows V68 skip guard. Make it explicit. Hidden in macro is confusing.
- TearDown soc_model reset no longer useful -- removed.
- SetUp, remove early return so we can actually query on Linux.
- Test name consistency, `_FP32_as_FP16` for all relevant tests.


quic-muchhsu (approved):

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -509,15 +502,7 @@ void QnnHTPBackendTests::SetUp() {
...
-// There is an unknown behavior that "soc_model" config somehow remains in HTP backend throughout different testcases
-// within the same process. Once the option "soc_model=60" is set, all following testcases would be implicitly applied
-// (which can be checked in verbose logging). This problem causes QnnHTPBackendTests.Inverse_2d/3d/4d on Linux failed
-// with accuracy issue 0.259040415 vs. 0.256103545. Although this precision mismatch may be expected as HTP fp16/fp32
-// behaviors could differ on different platforms, here adopts workaround to avoid modifying non-ABI parts. Concretely,
-// "soc_model=30", which is the default setting, is set to HTP backend again after QnnHTPBackendTests testsuite is
-// completed. Note that this is not an ABI-specific issue and exists in non-ABI UT as well but it is not observed
-// previously due to the execution order of testcases.
+// QNN caches soc_model at process level; reset it after the suite to prevent cross-test contamination.
 void QnnHTPBackendTests::TearDownTestSuite() {
```
minfhong-qti: Previously this `TearDownTestSuite` (note that it is not `TearDown` which executes after each testcase) is added to avoid ABI testcases interfering non-ABI ones. Since non-ABI codes are removed, I think this is in fact useless.
Additionally, `soc_model` setting to UNKNOWN in fact will not be executed (refer to https://github.com/onnxruntime/onnxruntime-qnn/blob/main/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc#L655), and that's why I set it to 30 previously.
IMO, there is an explicit check in EP, and it should be safe enough to guard all invalid usecases.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -471,13 +471,8 @@ static BackendSupport GetHTPSupport(const onnxruntime::logging::Logger& logger)
 }
 
 void QnnHTPBackendTests::SetUp() {
-  if (cached_htp_support_ == BackendSupport::SUPPORTED) {
```
minfhong-qti, line 474: Could you explain more on why removing this check? Based on PR description `remove early return so we can actually query on Linux`, I did not get what to be queried on Linux.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
quic-muchhsu: Sorry to chime in, I have 2 questions.

In my understanding, UNKNOWN is the default setting in ort-qnn, and qnn will assign "soc_model=30" when seeing UNKNOWN?

Looks like UNKNOWN will not be executed from the line you pointed out, does that mean the previous "soc_model" setting will pass down to next test? Or if you mean ort-qnn will reset to soc="30" everytime, can you point me to the code base?

Thanks.

minfhong-qti: > UNKNOWN is the default setting in ort-qnn, and qnn will assign "soc_model=30" when seeing UNKNOWN?

I've not traced the QNN codebase but based on my previous experiments through QNN EP, QNN SDK will assign 30 to `soc_model` if the user did not set this option. I never tried deliberately setting to UNKNOWN and pass through QNN API.

> Looks like UNKNOWN will not be executed from the line you pointed out, does that mean the previous "soc_model" setting will pass down to next test? Or if you mean ort-qnn will reset to soc="30" everytime, can you point me to the code base?

From my previous experiments, in the same process, "the previous "soc_model" setting will pass down to next test" if user did not explicitly set again.

qti-yuduo: > From my previous experiments, in the same process, "the previous "soc_model" setting will pass down to next test" if user did not explicitly set again.

this is exactly why it's bad. You can swap testing orders, and some would fail and others would change numerics. I'll dig a bit deeper into QNN behavior than make the test deterministic.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
qti-yuduo: On Linux, `cached_htp_support_` defaults to `SUPPORTED`, so the early return meant `QnnHTPBackendTests::SetUp()` did nothing. That'd skip the platform‑attribute query and logging, and left cached_platform_attrs_ unset

onnxruntime/test/providers/qnn/qnn_test_utils.cc
qti-yuduo: > SDK defaults to `socModel 30`

Thanks for point this out @minfhong-qti, have it verified -- I think we can just remove this entirely then.

onnxruntime/test/dlc_output/QNNExecutionProvider_QNNExecutionProvider_11142297670459488311_0_0.dlc
minfhong-qti, line 1: Accidentally commit this file?

onnxruntime/test/providers/qnn/qnn_test_utils.cc
minfhong-qti: Sorry to open this thread again.
I think querying platform attribute on non-Qualcomm device (e.g., x86 Linux) does not make sense as there is no NPU to be queried. IMO, instead of removing the early return here, we probably should revise the macro of setting default `cached_htp_support_=SUPPORTED` to only non-Qualcomm platforms.

qti-yuduo: Reverted this change.

PR #70, opened 2026-02-24
[QnnNodeGroup] Handle 6D Reshape-Einsum-Reshape pattern.

minfhong-qti:
### Description

Transform into functionally equivalent sequence of Reshape-DepthToSpace-Transpose through node group mechanism.
```
(1024,32) -> Reshape -> (1,32,32,32) -> DepthToSpace -> (1,64,64,8) -> Transpose -> (1,8, 64,64)
```

### Motivation and Context

The target pattern has 6D shapes in between, which is not supported by HTP backend.
```
(1024,32) -> Reshape -> (1,32,32,2,2,8) -> Einsum -> (1,8,32,2,32,2) -> Reshape -> (1,8,64,64)
```

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
@@ -1,0 +1,288 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.h"
+
+#include <array>
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include "core/providers/qnn/builder/qnn_def.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_node_group/qnn_node_group.h"
+#include "core/providers/qnn/builder/qnn_node_group/utils.h"
+#include "core/providers/qnn/ort_api.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+constexpr size_t kEinsumRank6 = 6;
+constexpr std::array<uint32_t, 6> kEinsumPerm6{0, 5, 1, 3, 2, 4};
+constexpr auto RearrangeShape6 = [](const std::vector<uint32_t>& shape) {
+  return std::vector<uint32_t>{shape[0], shape[5], shape[1] * shape[3], shape[2] * shape[4]};
+};
+
+namespace {
+
+bool IsEquationEqualToTranspose(std::string equation, std::vector<uint32_t>& perm) {
+  equation.erase(std::remove(equation.begin(), equation.end(), ' '), equation.end());
+  if (equation.empty()) {
+    // Empty equation.
+    return false;
+  }
+
+  size_t idx_arrow = equation.find("->");
+  if (idx_arrow == std::string::npos) {
+    // Unexpected equation format without "->".
+    return false;
+  }
+
+  const std::string lhs = equation.substr(0, idx_arrow);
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
```diff
@@ -0,0 +1,282 @@
...
+  // Combine the last 3 dimensions to serve as the input depth for the following DepthToSpace.
+  std::vector<uint32_t> pre_reshape_output_shape = {
+      pre_reshape_output_info.shape[0],
+      pre_reshape_output_info.shape[1],
+      pre_reshape_output_info.shape[2],
+      pre_reshape_output_info.shape[3] * pre_reshape_output_info.shape[4] * pre_reshape_output_info.shape[5]};
+  QnnTensorWrapper pre_reshape_output_wrapper(pre_reshape_output.name,
+                                              QNN_TENSOR_TYPE_NATIVE,
+                                              pre_reshape_output_info.qnn_data_type,
+                                              std::move(pre_reshape_output_info.quant_param),  // Should be empty.
```
kuanyul-qti: Just curious, is it possible that the tensor contains quantization information?

minfhong-qti: In current method to match the pattern, QDQ nodes are not taken into consideration, and thus if any QDQ nodes are in between the target pattern, it will not match.
I will revise here to explicitly pass empty quant param to avoid confusion.

minfhong-qti: Add assert to check this and explicitly pass empty quant param.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
```diff
@@ -0,0 +1,288 @@
...
+  if (lhs.empty() || rhs.empty()) {
+    // Unexpected equation format without at least one input term and one output term.
+    return false;
+  }
+
+  size_t idx_comma = lhs.find(",");
+  if (idx_comma != std::string::npos) {
+    // Transpose-equivalent equation should have only one input term.
+    return false;
+  }
```
kuanyul-qti: Should we also check `rhs`?

kuanyul-qti, line 54: Should we also check `rhs`?

minfhong-qti: `rhs` should not have comma by spec. And if it really has one, I believe it will fail below when calculating Transpose perm.

kuanyul-qti (approved):

onnxruntime/test/providers/qnn/qnn_node_group/reshape_einsum_reshape_test.cc
```diff
@@ -0,0 +1,63 @@
...
+#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
+
+TEST_F(QnnHTPBackendTests, ReshapeEinsumReshape_Float) {
+  ProviderOptions provider_options = GetProviderOptions();
+  RunQnnModelTest(BuildReshapeEinsumReshapeFloatTestCase(),
+                  provider_options,
+                  /*opset_version=*/13,
+                  /*expected_ep_assignment=*/ExpectedEPNodeAssignment::All,
+                  /*fp32_abs_err=*/1e-2f);
+}
```
kuanyul-qti, line 76: Please add AssertOpInQnnGraph() added in the commit - https://github.com/onnxruntime/onnxruntime-qnn/commit/1bc89564e93851ffd8ae6165b54b25904637a3eb

minfhong-qti: Added.

huaychou (approved):
LGTM

PR #73, opened 2026-02-24
Re-enable Python smoke tests on Windows ARM

qti-jkilpatrick:


qti-yuduo (approved):

setup.py
```diff
@@ -102,17 +102,6 @@ def finalize_options(self):
...
-            # Override to use py3 tag instead of specific Python versions
-            # This makes the wheel work with any Python 3.10+
-            # Since there are no .pyd files, only native DLLs/SOs
-            impl, abi, plat = super().get_tag()
-            # Use 'py3' instead of specific version like 'cp310', 'cp311', etc.
-            impl = "py3"
-            # Use 'none' for ABI since we have no Python C extensions
-            abi = "none"
-            return impl, abi, plat
-
```
qti-jkilpatrick, line 115: I understand the motivation here and agree that nothing in our wheel is specific to a python version, but this makes life hard for the build and release processes, which both assume specific python versions.

.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -94,7 +92,7 @@ jobs:
       nightly_build: ${{ inputs.nightly_build }}
       target_arch: arm64ec
       upload_wheel: true
-      test_wheel: false
+      test_wheel: true
```
qti-jkilpatrick, line 95: When I looked at this earlier in the week, I figured the best course would be to test ARM64ec wheels at night, but I've changed my mind:

* This is what we were doing previously.
* Downloading the wheel does not appear to suffer from the super slow download problem.
* Rolling wheel tests into the arm64 build is definitely possible, but would further complicate `qualcomm-internal-build-and-test-single-os.yml`.

Given that, I'm inclined to continue building the ARM64ec wheel on x64 and testing it in a separate job.

qcom/model_test/model_test.py
```diff
@@ -240,6 +243,49 @@ def run(self) -> None:
...
+def get_qnn_ep_device(backend_type: BackendT) -> tuple[onnxruntime.OrtEpDevice, Path]:
+    """
+    Get the QNN EP device.
+
+    :return: A tuple of OrtEpDevice and a dict of backend names to QNN backend paths.
+    """
+    global _ep_plugin_loaded  # noqa: PLW0603
+    if not _ep_plugin_loaded:
+        onnxruntime.register_execution_provider_library("QNNExecutionProvider", onnxruntime_qnn.get_library_path())
+        _ep_plugin_loaded = True
```
qti-jkilpatrick, line 276: It's true, I never bother unregistering this library.

qcom/scripts/linux/run_tests.sh
```diff
@@ -52,32 +52,6 @@ function run_model_test() {
     fi
 }
 
-#
-# Run a model test with onnx_test_runner.
-#
-function run_legacy_model_test() {
```
qti-jkilpatrick, line 58: This should have been deleted previously.

qcom/scripts/windows/run_tests.ps1
```diff
@@ -154,29 +154,6 @@ $TestModelsViaEpPlugin = {
     return $false
 }
 
-$TestModelsViaLegacy = {
```
qti-jkilpatrick, line 157: Also dead code that should have been deleted before.

qcom/model_test/smoke_test.py
```diff
@@ -4,12 +4,11 @@
 import os
 from pathlib import Path
 
+import onnxruntime_qnn
```
qti-kromero, line 7: totally unrelated to this PR, possibly something we can't change, and I totally approved the PR that made this change but didn't realize we are using an underscore until I saw this import

idk if we can change it or if others are bothered by it but it's explicitly discouraged by PEP8 - could be something to discuss

qti-kromero (approved):

PR #49, opened 2026-02-13
Fix wheel name suffix issue

qti-kromero:
### Description
Fix bug where wheel name was not getting wheel name suffix added to it
Undoes some of compromises introduced by https://github.com/onnxruntime/onnxruntime-qnn/pull/1 and https://github.com/onnxruntime/onnxruntime-qnn/pull/14

### Motivation and Context
Bug introduced in churn of plugin-EP refactor

PR #75, opened 2026-02-25
Reduce usage of artifacts

qti-jkilpatrick:
* GitHub artifacts can be quite slow. Use them less.

### DIFF
--- a/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
+++ b/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
@@ -28,8 +28,8 @@ on:
         default: "X64"
         type: string
       build_test_same_runner:
-        description: "If true, run tests on the same runner that build them."
-        default: false
+        description: "If true, run tests on the same runner that built them."
+        default: true
         type: boolean
       test_runner_arch:
         default: "X64"
@@ -41,7 +41,7 @@ on:
         default: false
         type: boolean
       upload_test_archive:
-        default: true
+        default: false
         type: boolean
       upload_wheel:
         default: true
@@ -78,8 +78,8 @@ on:
         default: "X64"
         type: string
       build_test_same_runner:
-        description: "If true, run tests on the same runner that build them."
-        default: false
+        description: "If true, run tests on the same runner that built them."
+        default: true
         type: boolean
       test_runner_arch:
         description: "The architecture of the test machine to use: X64 or (Windows only) ARM64."
@@ -95,11 +95,11 @@ on:
         type: boolean
       upload_test_archive:
         description: "If true, build a test archive and include is as an artifact."
-        default: true
+        default: false
         type: boolean
       upload_wheel:
         description: "If true, include the wheel as a workflow artifact."
-        default: true
+        default: false
         type: boolean
 
 env:
@@ -147,6 +147,7 @@ jobs:
         if: ${{ inputs.upload_test_archive }}
         with:
           name: ORT ${{inputs.target_os}} ${{inputs.target_arch}} py${{ inputs.target_py_vsn != 'None' && inputs.target_py_vsn || 'none'}} Tests
+          retention-days: 7
           path: |
             ${{ github.workspace }}/build/onnxruntime-tests-${{inputs.target_os}}-${{ inputs.target_arch }}.*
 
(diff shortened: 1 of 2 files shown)


.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       nightly_build: ${{ inputs.nightly_build }}
       target_arch: arm64ec
       upload_wheel: true
-      # test_wheel: true  # Smoke tests are broken with new plugin EP
+      test_wheel: false
```
qti-jkilpatrick, line 97: The issue is that the wheel's name has changed. I plan to fix this [when I enable](https://jira-dc.qualcomm.com/jira/browse/AISW-170995) ARM64ec wheel tests in nightly runs.

.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -80,7 +80,7 @@ jobs:
     secrets: inherit
     with:
       nightly_build: ${{ inputs.nightly_build }}
-      build_runner_arch: ${{ inputs.windows_arm64_runner_arch }}
+      build_runner_arch: ARM64
```
qti-kromero: why are we just freezing this to ARM64 now?

qti-jkilpatrick: So that we can build and test on a single machine. If we build on x64, we'll frequently wait 25 minutes on the ARM64 runner to download the artifacts. Might as well build at that point.

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -95,11 +95,11 @@ on:
         type: boolean
       upload_test_archive:
         description: "If true, build a test archive and include is as an artifact."
-        default: true
+        default: false
         type: boolean
       upload_wheel:
         description: "If true, include the wheel as a workflow artifact."
-        default: true
+        default: false
```
qti-kromero, line 102: does this change any of the behavior of the nightly job? am unsure if I explicitly listed upload_wheel:true for the x86 wheel we are getting currently

qti-jkilpatrick: The x86_64 Linux wheel does have `upload_wheel` set to `true` in the nightly job.

.github/workflows/qualcomm-internal-build-and-test.yml
qti-kromero: aren't build_runner_arch and inputs.windows_arm64_runner_arch inconsistent with the changes as-is? i.e. you could have inputs.windows_arm64_runner_arch set to X64 by default but build_runner_arch is forced to ARM64

would changing the default of inputs.windows_arm64_runner_arch from X64 to ARM64 also accomplish this?

.github/workflows/qualcomm-internal-build-and-test.yml
qti-kromero: depending on which name change you're referring to I may have already fixed this here but not realized that this also fixed the testing issue: https://github.com/onnxruntime/onnxruntime-qnn/pull/49/changes#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R302

qti-jkilpatrick: That's great! I'll definitely take advantage of that when I enable wheel testing in the nightly run.

.github/workflows/qualcomm-internal-build-and-test.yml
qti-jkilpatrick: It would, but I don't want to accidentally increase the load on our only WoS machine.

Yes, it is internally a little confusing, but the contract remains the same: if someone (who?!?) requests an ARM64 build on an x64 machine, it'll actually happen on ARM64, but it'll be functionally equivalent -- if we were going the other direction, we'd potentially lose functionality, but that's not the case here. If I'm a casual user (who?!?) who doesn't read the logs, I'll never know that my request for x64 wasn't honored.

qti-jkilpatrick: Discussed offline. Kyle is 100% correct. I have vacation brain.

qti-kromero (approved):

PR #77, opened 2026-02-25
[OpBuilder] Handle no-op Split.

minfhong-qti:
### Description

- Define a utility function to add a Reshape node with identical in/out shapes to represent no-op node.
- Revise Split builder to take no-op Split into consideration.

### Motivation and Context

Current Split builder failed to correctly handle cases that generate only one output (i.e., no Split perform).
Since current builder and wrapper structure are unfriendly to support reroute if facing no-op nodes, use no-op Reshape node to represent no-op instead which should be optimized in the backend.


kuanyul-qti (dismissed):

kuanyul-qti (approved):

huaychou (approved):
LGTM

PR #82, opened 2026-02-26
[QNN EP] Add Support for SimplifiedLayerNormalization

qti-ashwshan:
- Map SimplifiedLayerNorm to RMSNorm Op in QNN EP
- Add unit tests for SimplifiedLayerNorm to run on NPU

### Description
- Map SimplifiedLayerNorm to RMSNorm Op in QNN EP
- Add unit tests for SimplifiedLayerNorm to run on NPU

### Motivation and Context
- SimplifiedLayerNorm Operator support does not exist in QNN and falls back to CPU.


onnxruntime/test/providers/qnn/simplifiedlayernormalization_test.cc
```diff
@@ -0,0 +1,288 @@
...
+                                          const std::vector<ONNX_NAMESPACE::AttributeProto>& attrs,
+                                          ExpectedEPNodeAssignment expected_ep_assignment) {
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = "cpu";
+  provider_options["offload_graph_io_quantization"] = "0";
+
+  RunQnnModelTest(
+      BuildOpTestCase<float>("SimplifiedLayerNormalization", {input_def, scale_def}, {}, attrs),
+      provider_options,
+      17,  // opset version
```
qti-yuduo, line 31: Let's use `1` per https://xadupre.github.io/draft/inference/operators/onnx__SimplifiedLayerNormalization.html. 17 should work but 1 remains max compatibility.

qti-yuduo (approved):

qti-yuduo:
At some point -- we'd better distinguish contrib op vs. core I think.. Leaving that as future work

onnxruntime/test/providers/qnn/simplifiedlayernormalization_test.cc
qti-ashwshan: We wont be able to change the opset version to 1 because the Quantize and Dequantize also become opset 1 and that makes the test fail. We would have to keep it 17 for now

qti-ashwshan:
> Let's use `1` per https://xadupre.github.io/draft/inference/operators/onnx__SimplifiedLayerNormalization.html. 17 should work but 1 remains max compatibility.

Yes, will follow up on that

PR #84, opened 2026-02-26
[QnnNodeGroup] Fix bug in Rank6ToRank5Fusion.

minfhong-qti:
### Description

Fix by manually adding tensor and node wrappers instead of incorrectly using `QnnModelWrapper.AddReshapeNode`.

### Motivation and Context

Current implementation did not consider the case that the pattern is connected to graph IO, and thus adding tensor wrapper in wrong type.

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_node_group/reshape_transpose_rank5_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_node_group/reshape_transpose_rank5_test.cc
@@ -71,6 +71,38 @@ GetTestModelFn BuildRank6ToRank5FloatTestCase() {
   };
 }
 
+// Build float test: Input -> Reshape(rank-6) -> Transpose -> Reshape
+//                         -> Add
+GetTestModelFn BuildRank6ToRank5FloatMultiConsumerTestCase() {
+  return [](ModelTestBuilder& builder) -> void {
+    builder.graph_->set_name("rank6_to_rank5_fusion_float_multi_consumer_graph");
+
+    auto input_def = TestInputDef<float>({256, 64}, false, -10.0f, 10.0f);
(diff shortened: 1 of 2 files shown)


qti-ashwshan (approved):

onnxruntime/test/providers/qnn/qnn_node_group/reshape_transpose_rank5_test.cc
```diff
@@ -66,6 +94,15 @@ TEST_F(QnnHTPBackendTests, Rank6ToRank5Fusion_Float) {
...
 }
 
+TEST_F(QnnHTPBackendTests, Rank6ToRank5Fusion_Float_MultiConsumer) {
+  ProviderOptions provider_options = GetProviderOptions();
+  RunQnnModelTest(BuildRank6ToRank5FloatMultiConsumerTestCase(),
+                  provider_options,
+                  /*opset_version=*/13,
+                  /*expected_ep_assignment=*/ExpectedEPNodeAssignment::All,
+                  /*fp32_abs_err=*/1e-2f);
+}
```
kuanyul-qti, line 130: Please add AssertOpInQnnGraph() added in the commit - https://github.com/onnxruntime/onnxruntime-qnn/commit/1bc89564e93851ffd8ae6165b54b25904637a3eb

minfhong-qti: I believe the new commit is to more thoroughly verify fusions/replacement node group since assignment check is not enough. However, this node group is to avoid 6D shape which originally would fail to partition onto QNN EP, and thus I believe the new check is not necessary.
In addition, the target node group in this PR was not added the check in the commit as well.

kuanyul-qti (approved):

PR #89, opened 2026-02-26
[ONNX2EP] Add Roialign Op to HTP.

quic-muchhsu:
### Description

Add  op support for roi align.
Notice that the unit test uses hard coded model and a golden input/output to have mininal test for roi align node.
The reason is we can't use ort-cpu output as reference like other unit test, because the ort-cpu only support "aligned=Flase", while HTP only support "aligned=True".

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
+++ b/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
@@ -253,7 +253,7 @@ OpBuilderRegistrations::OpBuilderRegistrations() {
     CreateTanOpBuilder("Tan", *this);
   }
 
-  {}
+  {
     CreateRoiAlignOpBuilder("RoiAlign", *this);
   }
 }
--- a/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
+++ b/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
@@ -252,6 +252,10 @@ OpBuilderRegistrations::OpBuilderRegistrations() {
   {
     CreateTanOpBuilder("Tan", *this);
   }
+
+  {}
+    CreateRoiAlignOpBuilder("RoiAlign", *this);
+  }
 }
 
 const IOpBuilder* GetOpBuilder(const std::string& onnx_op_type) {
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
@@ -1,0 +1,130 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "QnnOpDef.h"
+
+#include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/op_builder_factory.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/providers/qnn/ort_api.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+class RoiAlignOpBuilder : public BaseOpBuilder {
+ public:
+  RoiAlignOpBuilder() : BaseOpBuilder("RoiAlignOpBuilder") {}
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(RoiAlignOpBuilder);
+
+  Ort::Status IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                            const OrtNodeUnit& node_unit,
+                            const Ort::Logger& logger) const final ORT_MUST_USE_RESULT;
+
+ protected:
+  Ort::Status ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
+                            const OrtNodeUnit& node_unit,
+                            const Ort::Logger& logger,
+                            std::vector<std::string>& input_names,
+                            bool do_op_validation) const override ORT_MUST_USE_RESULT;
+
+  Ort::Status ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                          const OrtNodeUnit& node_unit,
+                                          std::vector<std::string>&& input_names,
+                                          const Ort::Logger& logger,
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+};
+
+Ort::Status RoiAlignOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                             const OrtNodeUnit& node_unit,
+                                             const Ort::Logger& logger) const {
+  OrtNodeAttrHelper node_helper(node_unit);
+
+  // Sanity checks on RoiAlign onnx attrs
+  // HTP supports only align=False which corresponds to coordinate_transformation_mode="output_half_pixel"
+  std::string coordinate_transformation_mode = node_helper.Get("coordinate_transformation_mode", "half_pixel");  // ONNX default "half_pixel"
+  RETURN_IF_NOT(coordinate_transformation_mode == "output_half_pixel", "HTP only supports coordinate_transformation_mode=output_half_pixel.");
+
+  // HTP supports only average pooling
+  std::string mode = node_helper.Get("mode", "avg");  // ONNX default "avg"
+  RETURN_IF_NOT(mode == "avg", "HTP only supports avg pooling mode.");
+
+  // HTP doesn't support sampling_ratio = 0(adaptive mode)
+  int sampling_ratio = node_helper.Get("sampling_ratio", 0);
+  RETURN_IF_NOT(sampling_ratio != 0, "HTP doesn't support sampling ratio = 0.");
+
+  // Sanity check on output_width output_height
+  // Output tensor size [N, C, H, W]
+  TensorInfo output_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Outputs()[0], output_info));
+  RETURN_IF_NOT(4 == static_cast<int>(output_info.shape.size()), "Roialign requires 4d output (num_rois, C, output_height, output_width).");
+  int output_height = node_helper.Get("output_height", 1);
+  RETURN_IF_NOT(output_height == static_cast<int>(output_info.shape[2]), "Expect output_height == output_tensor.shape[2]");
+  int output_width = node_helper.Get("output_width", 1);
+  RETURN_IF_NOT(output_width == static_cast<int>(output_info.shape[3]), "Expect output_width == output_tensor.shape[3]");
+
+  // RoiAlignOp are sensitive with data layout, requires NHWC data layout.
+  // Continue RoiAlign if NHWC format.
+  if (node_unit.Domain() == kMSInternalNHWCDomain) {
+    return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
+  }
+
+  return Ort::Status();
+}
+
+Ort::Status RoiAlignOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
+                                             const OrtNodeUnit& node_unit,
+                                             const Ort::Logger& logger,
+                                             std::vector<std::string>& input_names,
+                                             bool do_op_validation) const {
+  ORT_UNUSED_PARAMETER(do_op_validation);
+
+  const auto& inputs = node_unit.Inputs();
+  const auto input_count = GetInputCountQnnRequired(node_unit);
+  for (size_t input_i = 0; input_i < input_count; ++input_i) {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[input_i], logger, input_names));
+  }
+
+  return Ort::Status();
+}
+
+Ort::Status RoiAlignOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                                           const OrtNodeUnit& node_unit,
+                                                           std::vector<std::string>&& input_names,
+                                                           const Ort::Logger& logger,
+                                                           bool do_op_validation) const {
+  ORT_UNUSED_PARAMETER(logger);
+
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[0], input_info));
+
+  OrtNodeAttrHelper node_helper(node_unit);
+  std::vector<std::string> param_tensor_names;
+  std::vector<uint32_t> img_size_ratio_shape{static_cast<uint32_t>(2)};
+
+  // Convert onnx::spatial_scale to htp::img_size_ratio
+  float spatial_scale = node_helper.Get("spatial_scale", 1.0f);
+  RETURN_IF(spatial_scale == 0, "Roialign got invalid spatial_scale=0");
+  std::vector<float> img_size_ratio{1 / spatial_scale, 1 / spatial_scale};
+  QnnParamWrapper roi_align_param_img_size_ratio = createQnnParamWrapper<float>(
+      node_unit.Index(),
+      node_unit.Name(),
+      QNN_OP_ROI_ALIGN_PARAM_IMG_SIZE_RATIO,
+      std::move(img_size_ratio_shape),
+      std::move(img_size_ratio));
+  param_tensor_names.push_back(roi_align_param_img_size_ratio.GetParamTensorName());
+  qnn_model_wrapper.AddParamWrapper(std::move(roi_align_param_img_size_ratio));
+
+  return ProcessOutputs(qnn_model_wrapper,
+                        node_unit,
+                        std::move(input_names),
+                        std::move(param_tensor_names),
+                        logger, do_op_validation, QNN_OP_ROI_ALIGN);
+}
+
+void CreateRoiAlignOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations) {
+  op_registrations.AddOpBuilder(op_type, std::make_unique<RoiAlignOpBuilder>());
+}
+
+}  // namespace qnn
+}  // namespace onnxruntime
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
@@ -38,6 +38,12 @@ class RoiAlignOpBuilder : public BaseOpBuilder {
 Ort::Status RoiAlignOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
                                              const OrtNodeUnit& node_unit,
                                              const Ort::Logger& logger) const {
+  // RoiAlignOp are sensitive with data layout, requires NHWC data layout.
+  // Continue RoiAlign if NHWC format.
+  if (node_unit.Domain() == kMSInternalNHWCDomain) {
+    return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
+  }
+
   OrtNodeAttrHelper node_helper(node_unit);
 
   // Sanity checks on RoiAlign onnx attrs
@@ -62,12 +68,6 @@ Ort::Status RoiAlignOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
   RETURN_IF_NOT(output_height == static_cast<int>(output_info.shape[2]), "Expect output_height == output_tensor.shape[2]");
   int output_width = node_helper.Get("output_width", 1);
   RETURN_IF_NOT(output_width == static_cast<int>(output_info.shape[3]), "Expect output_width == output_tensor.shape[3]");
-
-  // RoiAlignOp are sensitive with data layout, requires NHWC data layout.
-  // Continue RoiAlign if NHWC format.
-  if (node_unit.Domain() == kMSInternalNHWCDomain) {
-    return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
-  }
 
   return Ort::Status();
 }
@@ -94,9 +94,6 @@ Ort::Status RoiAlignOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_
                                                            const Ort::Logger& logger,
                                                            bool do_op_validation) const {
   ORT_UNUSED_PARAMETER(logger);
-
-  TensorInfo input_info = {};
-  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[0], input_info));
 
   OrtNodeAttrHelper node_helper(node_unit);
   std::vector<std::string> param_tensor_names;
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -1838,6 +1838,11 @@ OrtStatus* ORT_API_CALL QnnEp::ShouldConvertDataLayoutForOpImpl(_In_ OrtEp* this
     *should_convert = 1;
   }
 
+  if (std::string(domain) == kOnnxDomain && std::string(op_type) == "RoiAlign") {
+    // RoiAlign is translated to QNN's RoiAlign, which requires the NHWC layout for processing.
+    *should_convert = 1;
+  }
+
   return nullptr;
 }
 
--- a/onnxruntime/test/onnx/main.cc
+++ b/onnxruntime/test/onnx/main.cc
@@ -915,7 +915,12 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
         ORT_TSTR("mod_mixed_sign_int16"),
         ORT_TSTR("mod_mixed_sign_int8"),
         ORT_TSTR("mod_uint16"),
-        ORT_TSTR("mod_uint64")};
+        ORT_TSTR("mod_uint64"),
+        // QNN only support aligned = False and op version 16.
+        // [ONNXRuntimeError] : 9 : NOT_IMPLEMENTED : Could not find an implementation for RoiAlign(22) node with name ''
+        ORT_TSTR("roialign_aligned_false"),
+        ORT_TSTR("roialign_aligned_true"),
+        ORT_TSTR("roialign_mode_max")};
 
     std::unordered_set<std::basic_string<ORTCHAR_T>> all_disabled_tests(std::begin(immutable_broken_tests), std::end(immutable_broken_tests));
 
--- a/onnxruntime/test/providers/qnn/roialign_op_test.cc
+++ b/onnxruntime/test/providers/qnn/roialign_op_test.cc
@@ -1,0 +1,265 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#if !defined(ORT_MINIMAL_BUILD)
+
+#include <cassert>
+#include <string>
+
+#include "test/providers/qnn/qnn_test_utils.h"
+#include "core/graph/node_attr_utils.h"
+
+#include "core/graph/onnx_protobuf.h"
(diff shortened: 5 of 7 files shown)


onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,244 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
qti-yuduo: Update to Qualcomm header.

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,244 @@
...
+                  ExpectedEPNodeAssignment::All,
+                  1e-5f);
+}
+
+#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
+
+//
+// HTP tests:
+//
+TEST_F(QnnHTPBackendTests, TestRoialignOp) {
```
qti-yuduo:
```suggestion
TEST_F(QnnHTPBackendTests, TestRoiAlign) {
```

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,244 @@
...
+  ASSERT_EQ(1u, count);
+  const float* raw = out.GetTensorData<float>();
+  EXPECT_NEAR(raw[0], 2.5f, fp32_abs_err);
+}
+
+//
+// CPU tests:
+//
+
+TEST_F(QnnCPUBackendTests, TestRoialignOp) {
```
qti-yuduo:
```suggestion
TEST_F(QnnCPUBackendTests, TestRoiAlign) {
```

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,151 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
qti-yuduo: Update to Qualcomm header.

qti-yuduo:
```
```

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,151 @@
...
+  ORT_UNUSED_PARAMETER(logger);
+
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[0], input_info));
+
+  OrtNodeAttrHelper node_helper(node_unit);
+  std::vector<std::string> param_tensor_names;
+  std::vector<uint32_t> img_size_ratio_shape{static_cast<uint32_t>(2)};
+  float spatial_scale = node_helper.Get("spatial_scale", static_cast<float>(0));
+  std::vector<float> img_size_ratio{1 / spatial_scale, 1 / spatial_scale};
```
qti-yuduo, line 105: Here `spatial_scale` defaults to `0`, won't we hit division-by-zero issue? ONNX op seems to have default of `1.0`. 

```
float spatial_scale = node_helper.Get("spatial_scale", 1.0f);
```

And if we want additional safety, we can add a guard on `spatial_scale != 0`.

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,151 @@
...
+  const auto& inputs = node_unit.Inputs();
+  const auto input_count = GetInputCountQnnRequired(node_unit);
+  for (size_t input_i = 0; input_i < input_count; ++input_i) {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[input_i], logger, input_names));
+  }
+
+  return Ort::Status();
+}
+
+Ort::Status RoiAlignOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
```
qti-yuduo, line 91: Inputs like `output_height`, `output_width` are ignored -- I guess it still works because it's not strictly required as it can be inferred -- but let's maybe add some test case that has non-default value 1 to gain some confidence?

qti-yuduo: Same as `sampling_ratio`, `mode`, `coordinate_transformation_mode`, etc. We should have much wider testing coverages. Properly accept what's QNN supported and reject what QNN cannot handle.

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,151 @@
...
+  const std::string& roialign_output_name = utils::GetUniqueName(node_unit, "roialign_output");
+
+  const std::string& org_output_name = node_unit.Outputs()[0].name;
+  std::vector<uint32_t> output_shape = output_info.shape;
+  output_shape[1] = output_info.shape[2];
+  output_shape[2] = output_info.shape[3];
+  output_shape[3] = output_info.shape[1];
+  QnnTensorWrapper roialign_output(roialign_output_name,
+                                   QNN_TENSOR_TYPE_NATIVE,
+                                   input_info.qnn_data_type,
```
qti-yuduo: Shouldn't this be `output_info.qnn_data_type`?

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,151 @@
...
+                                   QnnQuantParamsWrapper(),
```
qti-yuduo: Why dropping original quant params?

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,151 @@
...
+                                   std::vector<uint32_t>(output_shape));
```
qti-yuduo: `std::move()`

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -414,6 +414,27 @@ class QnnParamWrapper {
     SetQnnTensorClientBuf(qnn_param_.tensorParam, param_data_);
   }
 
+  QnnParamWrapper(size_t node_index,
```
qti-yuduo: Maybe missed something, can't `createQnnParamWrapper` do the job?

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,244 @@
...
+  /*
+    Using fixed graph with golden output to test roi align. Because onnxruntime cpu reference only support
+    aligned = False, but HTP only support aligned = True.
+    So the ort-cpu reference is not applicable to roi aligned node.
+
+    Fixed graph
+    x: [[[1, 2], [3, 4]]]
+    rois: [0, 0, 1, 1]
+    batch_indices: [0]
+    expect output: [2.5]
```
qti-yuduo: Could you comment on the source of these ground truth values?

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,244 @@
...
+  const auto num_bytes = narrow<size_t>(stream.tellg());
+  byte_buffer.resize(num_bytes);
+  stream.seekg(0);
+  ORT_ENFORCE(stream.read(reinterpret_cast<char*>(byte_buffer.data()), num_bytes), "Failed to read file.");
+
+  byte_buffer_out = std::move(byte_buffer);
+  return gsl::span<const std::byte>(byte_buffer_out);
+}
+
+static void RunRoiAlignTest(
```
qti-yuduo: Missing quantized op tests.

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,244 @@
...
+    const std::string& backend_type = "htp",
+    const std::string& soc_model = "87",
```
qti-yuduo: Try if this is needed?

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,244 @@
...
+    x: [[[1, 2], [3, 4]]]
+    rois: [0, 0, 1, 1]
+    batch_indices: [0]
+    expect output: [2.5]
+  */
+  // HTP option
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = backend_type;
+  provider_options["offload_graph_io_quantization"] = "0";
+  provider_options["dump_json_qnn_graph"] = "1";
```
qti-yuduo: For debug? remove.

onnxruntime/test/onnx/main.cc
```diff
@@ -915,7 +915,11 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
         ORT_TSTR("mod_mixed_sign_int16"),
         ORT_TSTR("mod_mixed_sign_int8"),
         ORT_TSTR("mod_uint16"),
-        ORT_TSTR("mod_uint64")};
+        ORT_TSTR("mod_uint64"),
+        // QNN only support aligned = True.
+        ORT_TSTR("roialign_aligned_false"),
```
qti-yuduo, line 921: Why don't we properly reject in EP?

qti-yuduo: And why 3 times?

quic-muchhsu: 1.Why don't we properly reject in EP?
Because "aligned" attribute is not allowed in ONNX, and ONNX graph assumes every roialign is always aligned=False
2. Why 3 times. 
Typo my mad.

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
quic-muchhsu: Good call, adding the safeguard.

onnxruntime/test/providers/qnn/roialign_op_test.cc
quic-muchhsu: Yes, roialign is a newly added op in HTP, the default 30 won't work.

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
quic-muchhsu: Yes, I am adding sanity checks to make sure the onnx attr are legal in qnn.

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,173 @@
...
+                                                     roialign_output_name,
+                                                     org_output_name,
+                                                     /*input_shape=*/std::move(qnn_roialign_output_shape),
+                                                     /*transpose_perm=*/std::move(transpose_roialign_perm),
+                                                     /*output_shape=*/std::move(NCHW_roialign_output),
+                                                     output_info.qnn_data_type,
+                                                     output_info.quant_param,
+                                                     do_op_validation,
+                                                     false,
+                                                     qnn_model_wrapper.IsGraphOutput(org_output_name)));  // is_for_output
```
qti-yuduo: nit: `/*is_for_output=*/` (following same coding style as reset of the arguments). Same for other places.

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,322 @@
...
+    NodeArg* output = builder.MakeOutput();
+    Node& roialign_node = builder.AddNode("RoiAlign", inputs, {output});
+
+    for (const auto& attr : attrs) {
+      roialign_node.AddAttributeProto(attr);
+    }
+  };
+}
+
+// Returns a function that creates a graph with a QDQ Pad operator.
```
qti-yuduo: Pad? (remove entire line preferred)

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,322 @@
...
+template <typename QuantType>
+GetTestQDQModelFn<QuantType> BuildRoialignQDQTestCase(const TestInputDef<float>& input_def,
```
qti-yuduo, line 46:
```suggestion
GetTestQDQModelFn<QuantType> BuildRoiAlignQDQTestCase(const TestInputDef<float>& input_def,
```

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,322 @@
...
+      roialign_node.AddAttributeProto(attr);
+    }
+
+    // op_output -> Q -> DQ -> output
+    AddQDQNodePairWithOutputAsGraphOutput<QuantType>(builder, output, output_qparams[0].scale,
+                                                     output_qparams[0].zero_point);
+  };
+}
+
+// Runs an Pad model on the QNN CPU backend. Checks the graph node assignment, and that inference
```
qti-yuduo: Pad? Many other tests also has "Pad" maybe do a global search.

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,322 @@
...
+#include "core/graph/onnx_protobuf.h"
+#include "gtest/gtest.h"
+
+#include "test/util/include/test_utils.h"
+
+namespace onnxruntime {
+namespace test {
+
+// Returns a function that creates a graph with a single Pad operator.
+static GetTestModelFn BuildRoialignTestCase(const TestInputDef<float>& input_def,
```
qti-yuduo, line 24:
```suggestion
static GetTestModelFn BuildRoiAlignTestCase(const TestInputDef<float>& input_def,
```

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,322 @@
...
+// QNN rejects wrong output_width.
+TEST_F(QnnCPUBackendTests, TestRoialign_Fail_output_width) {
+  RunRoiAlignOpTest(TestInputDef<float>({1, 1, 2, 2}, false, {1.0f, 2.0f, 3.0f, 4.0f}),
+                    TestInputDef<float>({1, 4}, true, {0.0f, 0.0f, 1.0f, 1.0f}),
+                    TestInputDef<int64_t>({1}, true, {0}),
+                    {utils::MakeAttribute("coordinate_transformation_mode", "output_half_pixel"),
+                     utils::MakeAttribute("mode", "avg"),
+                     utils::MakeAttribute("sampling_ratio", static_cast<int64_t>(1)),
+                     utils::MakeAttribute("output_height", static_cast<int64_t>(1)),
+                     utils::MakeAttribute("output_width", static_cast<int64_t>(2)),  // expects 1
```
qti-yuduo: not following the comment: what is expected to be 1?

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,173 @@
...
+  // HTP doesn't support sampling_ratio = 0(adaptive mode)
+  int sampling_ratio = node_helper.Get("sampling_ratio", 0);
+  RETURN_IF_NOT(sampling_ratio != 0, "HTP doesn't support sampling ratio = 0.");
+
+  // Sanity check on output_width output_height
+  // Output tensor size [N, C, H, W]
+  TensorInfo output_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Outputs()[0], output_info));
+  int output_height = node_helper.Get("output_height", 1);
+  RETURN_IF_NOT(output_height == static_cast<int>(output_info.shape[3]), "Expect output_height == output_tensor.shape[3]");
```
qti-yuduo: Comment (line 78) says NCHW, but we're check NHWC here -- which one is wrong?

qti-yuduo: Also, better to include a size check before accessing `shape[2/3]`?

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,173 @@
...
+
+  // Transpose feature: [N C H W] to [N H W C]
+  const std::string& input_name = input_names[0];
+  const std::string& transposed_feature_name = utils::GetUniqueName(node_unit, "_transposed_feature");
+  std::vector<uint32_t> transpose_perm{0, 2, 3, 1};
+  std::vector<uint32_t> transpose_output_shape = input_info.shape;
+  transpose_output_shape[1] = input_info.shape[2];
+  transpose_output_shape[2] = input_info.shape[3];
+  transpose_output_shape[3] = input_info.shape[1];
+  RETURN_IF_ERROR(qnn_model_wrapper.AddTransposeNode(node_unit.Index(),
```
qti-yuduo: Forgot to ask last time I saw this, we kind of manually handling layout requirements as this op is missing from `GetORTLayoutSensitiveOps` -- could you try if we can avoid manual insertions?

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,322 @@
...
+                     utils::MakeAttribute("mode", "avg"),
+                     utils::MakeAttribute("sampling_ratio", static_cast<int64_t>(1)),
+                     utils::MakeAttribute("output_height", static_cast<int64_t>(1)),
+                     utils::MakeAttribute("output_width", static_cast<int64_t>(1)),
+                     utils::MakeAttribute("spatial_scale", 1.0f)},
+                    ExpectedEPNodeAssignment::All);
+}
+
+// QNN doesn't support coordinate_transformation_mode = output_half
+TEST_F(QnnCPUBackendTests, TestRoialign_Fail_output_half) {
```
qti-yuduo: Good idea to add failure test cases!

nit: I would avoid `Fail` in test name as that's people use for searching through logs. It would just give noise in those cases. Perhaps something else: e.g., `Reject/Unsupported/NotSupported/Negative`.

onnxruntime/test/providers/qnn/roialign_op_test.cc
quic-muchhsu: Edited: I am removing these 2 tests because the node's output tensor is generated using output_height and output_width.

output_height and output_width expected to have same value as the output tensor shapr in [N, C, H, W] formar.
Although onnx should have already validated these 2 values, but there's no harm to check it again.

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
quic-muchhsu: Yes, let me check GetORTLayoutSensitiveOps to see if I can avoid manual transpose.

onnxruntime/test/providers/qnn/roialign_op_test.cc
quic-muchhsu: Got it, changing to Unsupported

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,130 @@
...
+  RETURN_IF_NOT(4 == static_cast<int>(output_info.shape.size()), "Roialign requires 4d output (num_rois, C, output_height, output_width).");
+  int output_height = node_helper.Get("output_height", 1);
+  RETURN_IF_NOT(output_height == static_cast<int>(output_info.shape[2]), "Expect output_height == output_tensor.shape[2]");
+  int output_width = node_helper.Get("output_width", 1);
+  RETURN_IF_NOT(output_width == static_cast<int>(output_info.shape[3]), "Expect output_width == output_tensor.shape[3]");
+
+  // RoiAlignOp are sensitive with data layout, requires NHWC data layout.
+  // Continue RoiAlign if NHWC format.
+  if (node_unit.Domain() == kMSInternalNHWCDomain) {
+    return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
```
qti-yuduo: Nit: could we reverse the logic (everything else in this function is early return on false).

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1838,6 +1827,11 @@ OrtStatus* ORT_API_CALL QnnEp::ShouldConvertDataLayoutForOpImpl(_In_ OrtEp* this
     *should_convert = 1;
   }
 
+  if (std::string(domain) == kOnnxDomain && std::string(op_type) == "RoiAlign") {
+    // RoiAlign is translated to QNN's RoiAlign, which requires the NHWC layout for processing.
+    *should_convert = 1;
```
qti-yuduo, line 1843: Nice.

onnxruntime/test/providers/qnn/roialign_op_test.cc
qti-yuduo: reopened -- I think it got lost due to rebase or something. You can do a global search `Roialign`

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,130 @@
...
+}
+
+Ort::Status RoiAlignOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                                           const OrtNodeUnit& node_unit,
+                                                           std::vector<std::string>&& input_names,
+                                                           const Ort::Logger& logger,
+                                                           bool do_op_validation) const {
+  ORT_UNUSED_PARAMETER(logger);
+
+  TensorInfo input_info = {};
```
qti-yuduo: Looks like `input_info` is not used?

onnxruntime/test/providers/qnn/roialign_op_test.cc
```diff
@@ -0,0 +1,265 @@
...
+                    ExpectedEPNodeAssignment::None);
+}
+
+// QNN only supports pooling mode = average
+TEST_F(QnnCPUBackendTests, TestRoialign_Unsupported_mode_max) {
+  RunRoiAlignOpTest(TestInputDef<float>({1, 1, 2, 2}, false, {1.0f, 2.0f, 3.0f, 4.0f}),
+                    TestInputDef<float>({1, 4}, true, {0.0f, 0.0f, 1.0f, 1.0f}),
+                    TestInputDef<int64_t>({1}, true, {0}),
+                    {utils::MakeAttribute("coordinate_transformation_mode", "output_half_pixel"),
+                     utils::MakeAttribute("mode", "max"),  //  mode = max will Unsupported
```
qti-yuduo: `will unsupported` -> `is unsupported`

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,130 @@
...
+  RETURN_IF(spatial_scale == 0, "Roialign got invalid spatial_scale=0");
+  std::vector<float> img_size_ratio{1 / spatial_scale, 1 / spatial_scale};
+  QnnParamWrapper roi_align_param_img_size_ratio = createQnnParamWrapper<float>(
+      node_unit.Index(),
+      node_unit.Name(),
+      QNN_OP_ROI_ALIGN_PARAM_IMG_SIZE_RATIO,
+      std::move(img_size_ratio_shape),
+      std::move(img_size_ratio));
+  param_tensor_names.push_back(roi_align_param_img_size_ratio.GetParamTensorName());
+  qnn_model_wrapper.AddParamWrapper(std::move(roi_align_param_img_size_ratio));
```
qti-yuduo, line 113: https://docs.qualcomm.com/doc/80-63442-10/topic/MasterOpDef.html#roialign -- we have many unset params -- are they not needed? We're silently ignore them (drop values) which might causing unexpected behavior? Make sure do another pass so we have proper test cases, and properly reject unsupported ones.

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
quic-muchhsu: But if `node_unit.Domain() == kMSInternalNHWCDomain` pass and return node , won't it skip the rest of sanity checks?

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
quic-muchhsu: Do you mean these QNN params: `num_samples_y` `num_samples_x` `aligned` `allow_invalid_roi`?
Yes, they are not needed. Because:
- `num_samples_y` `num_samples_x` are for asymmetry box, which is not supported by onnx (source)
- `aligned==True` is `coordinate_transformation_mode=output_half_pixel.` in ONNX, and HTP doesn't support False cases. So this is not required.
- `allow_invalid_roi` is default to false, and ONNX doesn't have flag it. It is not settable from onnx.

onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
```diff
@@ -252,6 +252,10 @@ OpBuilderRegistrations::OpBuilderRegistrations() {
   {
     CreateTanOpBuilder("Tan", *this);
   }
+
+  {}
```
qti-yuduo: typo?

quic-muchhsu: Rebase fail

onnxruntime/core/providers/qnn/builder/opbuilder/roialign_op_builder.cc
```diff
@@ -0,0 +1,127 @@
...
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+};
+
+Ort::Status RoiAlignOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                             const OrtNodeUnit& node_unit,
+                                             const Ort::Logger& logger) const {
+  // RoiAlignOp are sensitive with data layout, requires NHWC data layout.
+  // Continue RoiAlign if NHWC format.
+  if (node_unit.Domain() == kMSInternalNHWCDomain) {
+    return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
```
qti-yuduo, line 44: I meant if `node_unit.Domain() !=` nhwc, we should early return (can be implemented as `RETURN_IF_NOT`), and finally we can `return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);`

qti-yuduo: Same logic as your previous implementation-- just clear on style.

quic-muchhsu: Got it. Will make the change.

quic-muchhsu: @qti-yuduo Looks like we can't do in that way. Saying:
```
RETURN_IF_NOT(node_unit.Domain() ==NHWC, ...);
sanity checks ...;
return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
```
Because IsOpSupported will be called twice, first time at `qnn_backend_manager.cc:1433 SetupBackend] QNN SetupBackend succeed` -> This stage allows NCHW because it will add a layout transformation wrapper around it to make it NHWC, and mark the node group to NHWC format.
In `CreateQnnNode` stage it will get called the second time, and this time the transform wrapper is already added to QNN graph, and it required to have NHWC format.

So we can only write in current way:
```
if (node_unit.Domain() == kMSInternalNHWCDomain) {
    return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
Sanity check...
```
That the first call it will not add node to model and do sanity check, and then on the second call it will add the node to QNN graph.

qti-yuduo (approved):
LGTM.

PR #94, opened 2026-02-27
Use SKIP_IF() instead of raw #ifdef for skipping tests

qti-yuduo:
## Motivation

Currently many tests are enabled or DISABLED based on complex `#ifdef`s. It's very hard to read, missing reasons on skipping (slightly skip) and easy-to-misuse (for example, additional skipping on more than necessary platforms)

## Changes

Defined macros that help to enforce putting a skipping reasons and narrow skipping platforms. 

Follow up will be needed to cleanup additional `DISABLE_*` tests.. Test should only be DISBALED_ when it expect to fail on all platforms.

### DIFF
--- a/onnxruntime/test/providers/qnn/gemm_test.cc
+++ b/onnxruntime/test/providers/qnn/gemm_test.cc
@@ -384,13 +384,8 @@ TEST_F(QnnHTPBackendTests, Gemm_Broadcast_Bias_DynamicA_StaticB_StaticC) {
 // Expected val: 120.73912048339844
 // QNN QDQ val: 0 (err 120.73912048339844)
 // CPU QDQ val: 120.73889923095703 (err 0.00022125244140625)
-// Issue fixed in 2.30
-#ifdef __linux__
-// Failed on Linux with 2.31
-TEST_F(QnnHTPBackendTests, DISABLED_Gemm_Dynamic_A_Static_B_Dynamic_Bias_U16) {
-#else
 TEST_F(QnnHTPBackendTests, Gemm_Dynamic_A_Static_B_Dynamic_Bias_U16) {
-#endif
+  QNN_SKIP_TEST_ON_LINUX_X86_64("Output value mismatch with QNN SDK 2.31");
   std::vector<float> input_a_data = GetFloatDataInRange(-10.0f, 10.0f, 6);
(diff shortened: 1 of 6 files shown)


onnxruntime/test/providers/qnn/gemm_test.cc
```diff
@@ -384,13 +384,8 @@ TEST_F(QnnHTPBackendTests, Gemm_Broadcast_Bias_DynamicA_StaticB_StaticC) {
...
 // QNN QDQ val: 0 (err 120.73912048339844)
 // CPU QDQ val: 120.73889923095703 (err 0.00022125244140625)
-// Issue fixed in 2.30
-#ifdef __linux__
-// Failed on Linux with 2.31
-TEST_F(QnnHTPBackendTests, DISABLED_Gemm_Dynamic_A_Static_B_Dynamic_Bias_U16) {
-#else
 TEST_F(QnnHTPBackendTests, Gemm_Dynamic_A_Static_B_Dynamic_Bias_U16) {
-#endif
+  QNN_SKIP_TEST_ON_LINUX_X86_64("Output value mismatch with QNN SDK 2.31");
```
qti-jkilpatrick: Did this condition change to now allow running on ARM64 Linux?

qti-jkilpatrick (commented):
This is much clearer. Thank you! Just one question.

onnxruntime/test/providers/qnn/gemm_test.cc
qti-yuduo: You're right.. Do we have ARM64 Linux on CI? Want to see if this can be relaxed. I can revert to be safe if it's not covered. 

As follow up, I think we should try relax for DISABLE_* to increase coverage for all non-failing cases.

qti-jkilpatrick: Your PR reminded me that I need to [re-enable ARM64 Linux](https://github.com/onnxruntime/onnxruntime-qnn/pull/95).

I totally agree that we should make a pass over things to increase coverage.

qti-yuduo: I'll revert for this PR. Keep original coverage

qti-jkilpatrick (approved):

qti-kromero (approved):

PR #104, opened 2026-03-03
Fix nightly workflow permissions

qti-kromero:


qti-jkilpatrick (approved):

qti-kromero:
tested on nightly run here: https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/22646623962

can't really test the schedule job manually so we will see how it goes tonight :)

PR #109, opened 2026-03-04
add scatter element op

quic-muchhsu:
### Description

Add scatter element op build from base op builder.
This is required because scatter element's input 0 and input 2 doesn't support int 32.
Casting to fp32 if it's int32.
Moving out scatter element from base op builder for there's a big change only for scatter element, not making sense to modify shared function for single element.


onnxruntime/core/providers/qnn/builder/opbuilder/scatter_elements_op_builder.cc
```diff
@@ -0,0 +1,324 @@
...
+
+ private:
+  static constexpr std::array<std::string_view, 4> scatterelements_supported_reduction = {"none", "add", "mul", "max"};
+};
+
+Ort::Status ScatterElementsOpBuilder::ProcessInput(QnnModelWrapper& qnn_model_wrapper,
+                                                   const OrtNodeUnitIODef& input,
+                                                   const Ort::Logger& logger,
+                                                   std::vector<std::string>& input_names,
+                                                   bool allow_int_32) const {
```
quic-muchhsu, line 46: allow_int_32 flag should be true for input[1], because in an input[0] [1] [2] are int64 case, all of them will be casted to int 32, but only input[1] supports int 32. Other will be casted to f32.

quic-muchhsu:
Close due to qnn 2.45 supports Scatter element with int32

PR #110, opened 2026-03-04
Fix graph i/o name mapping for partition added input

qti-yuduo:
Fix graph i/o name mapping to correctly preserve the partition-added input.

Corner case missed now also added as tests. 

Thanks @minfhong-qti for providing the repro test!!


minfhong-qti (approved):
LGTM

PR #111, opened 2026-03-05
Add monolithic lstm for HTP

chenweng-quic:
### Description
- Change unrolled LSTM to one single LSTM op in LSTM op builder.
- Add option monolithic lstm in HTP, see https://docs.qualcomm.com/doc/80-63442-10/topic/htp_backend.html#qnn-htp-monolithic-lstm.

### Motivation and Context
With this option it may reduce finalization time of models with LSTM.

### DIFF
--- a/cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
+++ b/cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
@@ -1,0 +1,67 @@
+From c60a332e9a65dbae2920914adb8537de60f566e7 Mon Sep 17 00:00:00 2001
+From: "Cheng-Hsin Weng (Brian)" <chenweng@qti.qualcomm.com>
+Date: Fri, 13 Mar 2026 15:57:10 +0800
+Subject: [PATCH 4/4] Update argument for monolithic lstm
+
+---
+ onnxruntime/test/onnx/main.cc                    | 3 ++-
+ onnxruntime/test/perftest/command_args_parser.cc | 1 +
+ onnxruntime/test/perftest/ort_test_session.cc    | 3 ++-
+ 3 files changed, 5 insertions(+), 2 deletions(-)
+
+diff --git a/onnxruntime/test/onnx/main.cc b/onnxruntime/test/onnx/main.cc
+index f08b974d60..a50d771234 100644
+--- a/onnxruntime/test/onnx/main.cc
++++ b/onnxruntime/test/onnx/main.cc
+@@ -87,6 +87,7 @@ void usage() {
+       "\t    Options are '0', '68', '69', '73', '75', '81'. Defaults to '0' (none). \n"
+       "\t    [QNN only] [device_id]: The ID of the device to use when setting 'htp_arch'. Defaults to '0' (for single device). \n"
+       "\t    [QNN only] [enable_htp_fp16_precision]: Enable the HTP_FP16 precision so that the float32 model will be inferenced with fp16 precision. \n"
++      "\t    [QNN only] [disable_htp_monolithic_lstm]: Disable the HTP monolithic lstm. Default to 0\n"
+       "\t    Otherwise, it will be fp32 precision. Works for float32 model for HTP backend. Defaults to '1' (with FP16 precision.). \n"
+       "\t    [QNN only] [offload_graph_io_quantization]: Offload graph input quantization and graph output dequantization to another EP (typically CPU EP). \n"
+       "\t    Defaults to '0' (QNN EP handles the graph I/O quantization and dequantization). \n"
+@@ -612,7 +613,7 @@ int real_main(int argc, char* argv[], Ort::Env& env) {
+             std::string str = str_stream.str();
+             ORT_THROW("Wrong value for htp_arch. select from: " + str);
+           }
+-        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization") {
++        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization" || key == "disable_htp_monolithic_lstm") {
+           std::unordered_set<std::string> supported_options = {"0", "1"};
+           if (supported_options.find(value) == supported_options.end()) {
+             std::ostringstream str_stream;
+diff --git a/onnxruntime/test/perftest/command_args_parser.cc b/onnxruntime/test/perftest/command_args_parser.cc
+index cb583812e1..5aa0d556a1 100644
+--- a/onnxruntime/test/perftest/command_args_parser.cc
++++ b/onnxruntime/test/perftest/command_args_parser.cc
+@@ -110,6 +110,7 @@ ABSL_FLAG(std::string, i, "",
+           "  Options are '0', '68', '69', '73', '75'. Defaults to '0' (none). \n"
+           "  [QNN only] [device_id]: The ID of the device to use when setting 'htp_arch'. Defaults to '0' (for single device). \n"
+           "  [QNN only] [enable_htp_fp16_precision]: Enable the HTP_FP16 precision so that the float32 model will be inferenced with fp16 precision. \n"
++          "  [QNN only] [disable_htp_monolithic_lstm]: Disable the HTP monolithic lstm. Default to 0\n"
+           "  Otherwise, it will be fp32 precision. Works for float32 model for HTP backend. Defaults to '1' (with FP16 precision.). \n"
+           "  [QNN only] [offload_graph_io_quantization]: Offload graph input quantization and graph output dequantization to another EP (typically CPU EP). \n"
+           "  Defaults to '0' (QNN EP handles the graph I/O quantization and dequantization). \n"
+diff --git a/onnxruntime/test/perftest/ort_test_session.cc b/onnxruntime/test/perftest/ort_test_session.cc
+index 1e879bf5df..795ee2fadd 100644
+--- a/onnxruntime/test/perftest/ort_test_session.cc
++++ b/onnxruntime/test/perftest/ort_test_session.cc
+@@ -256,7 +256,7 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
+                         {"backend_type", "backend_path", "profiling_file_path", "profiling_level",
+                               "rpc_control_latency", "vtcm_mb", "soc_model", "device_id", "htp_performance_mode", "op_packages",
+                              "qnn_saver_path", "htp_graph_finalization_optimization_mode", "qnn_context_priority",
+-                             "htp_arch", "enable_htp_fp16_precision", "offload_graph_io_quantization",
++                             "htp_arch", "enable_htp_fp16_precision", "disable_htp_monolithic_lstm", "offload_graph_io_quantization",
+                              "enable_htp_spill_fill_buffer", "enable_htp_shared_memory_allocator", "dump_json_qnn_graph",
+                              "json_qnn_graph_dir", "disable_file_mapped_weights", "htp_bf16_enable", "enable_vtcm_backup_buffer_sharing"});
+     for (const auto& provider_option : provider_options) {
+@@ -319,6 +319,7 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
+           ORT_THROW("Wrong value for htp_arch. select from: " + str);
+         }
+       } else if (key == "enable_htp_fp16_precision" ||
++                 key == "disable_htp_monolithic_lstm" ||
+                  key == "offload_graph_io_quantization" ||
+                  key == "enable_htp_spill_fill_buffer" ||
+                  key == "enable_htp_shared_memory_allocator" ||
+--
+2.53.0.windows.1
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
@@ -249,8 +249,6 @@ Ort::Status LSTMOpBuilder::AddStridedSliceOrReshape(QnnModelWrapper& qnn_model_w
 Ort::Status LSTMOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
                                          const OrtNodeUnit& node_unit,
                                          const Ort::Logger& logger) const {
-  ORT_UNUSED_PARAMETER(qnn_model_wrapper);
-  ORT_UNUSED_PARAMETER(node_unit);
   ORT_UNUSED_PARAMETER(logger);
   if (node_unit.Inputs().size() > 4 && node_unit.Inputs()[4].Exists()) {
     TensorInfo tensor_info = {};
@@ -309,7 +307,6 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
                                                const bool& is_bidirection,
                                                std::vector<std::string>& uni_lstm_output_names) const {
   ORT_UNUSED_PARAMETER(logger);
-
   const auto& onnx_inputs = node_unit.Inputs();
   const auto& onnx_outputs = node_unit.Outputs();
   const std::string& node_name = node_unit.Name();
@@ -344,7 +341,7 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
   std::vector<std::string> param_names;
 
   // direction
-  RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(),
+  RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name() + "_" + direction,
                                          direction == "forward" ? QNN_OP_LSTM_DIRECTION_FORWARD : QNN_OP_LSTM_DIRECTION_REVERSE,
                                          QNN_OP_LSTM_PARAM_DIRECTION, param_names));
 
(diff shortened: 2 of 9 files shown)


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -712,6 +712,29 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
 
+  // HTP monolithic lstm
+  std::string enable_htp_monolithic_lstm_str;
+  GetSessionConfigEntryOrDefault(ort_api,
+                                 session_options_,
+                                 FormatEPConfigKey("enable_htp_monolithic_lstm"),
+                                 "0",
+                                 enable_htp_monolithic_lstm_str);
+  if (enable_htp_monolithic_lstm_str == "1") {
+    enable_HTP_monolithic_lstm_ = true;
```
tirupath-qti: May i know the reason for Client to pass this information? Can't the QNN backend make a guided decision by looking at the graph?

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -789,8 +885,39 @@ Ort::Status LSTMOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
 
   if (direction == "bidirectional") {
     std::vector<std::string> uni_lstm_output_names_forward, uni_lstm_output_names_reverse;
+
+    // const uint32_t hidden_size = node_helper.Get("hidden_size", 0);
```
tirupath-qti: Please remove the commented code.

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -826,6 +953,7 @@ Ort::Status LSTMOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
       }
     }
   } else {
+    // not used, just a placeholder
```
tirupath-qti, line 760: why it is not used?

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -344,7 +409,7 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
   std::vector<std::string> param_names;
 
   // direction
-  RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(),
+  RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name() + "_" + direction,
```
tirupath-qti, line 344: any reason for appending "_direction"?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chenweng-quic: Thanks for the review.
There is guidance for this option; please refer to:
https://docs.qualcomm.com/doc/80-63442-10/topic/htp_backend.html#recommended-use-cases
From the QNN side, I think it is difficult to make a one-size-fits-all decision. For example, in a model with two LSTM layers, one might benefit from enabling this option while the other might not.
Given this, it seems better to leave the option exposed so that users can make the decision based on their specific model and use case.

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chenweng-quic: Will remove

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
tirupath-qti: If User can inspect the model and determine the possible use of this option why can't the backend does the same kind of inspection and take appropriate action?

Also another important point, what could be the guideline if an onnx model contains more than one of the LSTM variants? Instead of user explicitly selecting this option, can't the Backend apply different logic for different LSTm layers based on their configuration?

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chenweng-quic: to avoid situation where param name are the same while param value are different.
Before the PR: the direction is actually not used since it is unrolled.
With the PR: the direction is now important and they need to have different names

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chenweng-quic: Some information from HTP team:
The monolithic_lstm_flag is a trade-off setting.
Enabling it can significantly improve prepare time and reduce the context binary size, but it may lead to slower execution time.
Users should decide whether to use this option based on their specific scenario.

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chenweng-quic: The logic of bidirection LSTM:
1. ONNX LSTM (bidirection) = concat of QNN LSTM(forward) + QNN LSTM(reverse).
2. ONNX LSTM (forward) = QNN LSTM(forward)
3. ONNX LSTM (reverse) = QNN LSTM()
In case 2,3, we don't need to get the name from the AddUnidirectionLSTM, so just a placeholder to pass to function.

qcom/packages.yml
```diff
@@ -161,23 +177,7 @@ python_314_windows_arm64:
   version: 3.14.3
   md5: 1183951d055ef6c0ae1fb686b3342a6f
 qairt:
-  version: 2.42.0.251225
+  version: 2.44.0.260225
```
kuanyul-qti, line 180: Please rebase on the latest mainline.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -175,6 +175,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
   qnn::HtpGraphFinalizationOptimizationMode htp_graph_finalization_opt_mode_ = qnn::HtpGraphFinalizationOptimizationMode::kDefault;
   int32_t vtcm_size_in_mb_ = 0;
   bool enable_HTP_FP16_precision_ = true;
+  bool enable_HTP_monolithic_lstm_ = true;
```
kuanyul-qti: Why set the default to true? In my imagination, the default value is false, and it will only be set to true when the user provides this option.

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
```diff
@@ -64,6 +64,12 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
               "QNN EP does not support ScatterND op on CPU backend. Falling back to ORT CPU.");
   }
 
+  // TODO: Remove once QNN HTP PRelu bug is fixed
```
kuanyul-qti, line 67: Please rebase on the latest mainline.

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -357,41 +354,42 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
                                       QNN_OP_LSTM_PARAM_OUTPUT_CLIP_THRESHOLD, param_names));
 
   // time_major
-  RETURN_IF_ERROR(AddQnnScalar<bool>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), false,
+  RETURN_IF_ERROR(AddQnnScalar<bool>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), true,
```
kuanyul-qti, line 358: Could you explain why this is set to true?

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -357,41 +354,42 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
...
 
-  // // hidden_state_qscale
+  // hidden_state_qscale
   RETURN_IF_ERROR(AddQnnScalar<float>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), 0.0,
                                       QNN_OP_LSTM_PARAM_HIDDEN_STATE_QSCALE, param_names));
 
   // Common LSTM cell inputs
   const std::string null_tensor_name = "null_tensor";
-  QnnTensorWrapper null_tensor_wrapper(null_tensor_name, QNN_TENSOR_TYPE_NULL, QNN_DATATYPE_UNDEFINED,
+  QnnTensorWrapper null_tensor_wrapper(null_tensor_name, QNN_TENSOR_TYPE_NULL, QNN_DATATYPE_FLOAT_32,
```
kuanyul-qti, line 387: Is there any reason to change from undefined to float32?

chenweng-quic: I recall that this is for IR backend to dump DLC since they don't support QNN_DATATYPE_UNDEFINED.
Let me do some experiments and make a confirmation.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
chenweng-quic: Thanks for point this out!
Will update!

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chenweng-quic: In unrolled LSTM this parameter is not used actually.
With this PR we must align the value to layout in onnx.
Since the builder only support layout 0 (time_major), I set it to true.

I will check if I can add support for layout 1, if yes I can update it to `layout==0`

chenweng-quic: I saw CPU EP not support this as well...
https://github.com/microsoft/onnxruntime/blob/0411d41d234a7b1930cb8720bb6d7dd29ac2fa6a/onnxruntime/core/providers/cpu/rnn/lstm_base.h#L53
I believe HTP can run, but considering that there is no use case and it is hard to refactor current test to discard cpu ep for comparision, keep this intact and just update comment.

qcom/packages.yml
chenweng-quic: done

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
chenweng-quic: done

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chenweng-quic: Yes, I confirm this change it for IR backend to dump DLC.

onnxruntime/test/perftest/ort_test_session.cc
qti-hungjuiw, line 1: The change for `perftest` will need to use PATCH as currently it belongs to ORT Core. You may refer to https://github.com/onnxruntime/onnxruntime-qnn/pull/118/changes#diff-038272abbf046fd1e03986de25a4c9b9a8c91a8dba8935e84c4e23f338e453f9

onnxruntime/test/providers/qnn/lstm_test.cc
```diff
@@ -279,10 +281,12 @@ static void RunHtpFp16LSTMOpTest(const TestInputDef<float>& X_def,
                                  const int64_t hidden_size,
                                  const int64_t layout,
                                  ExpectedEPNodeAssignment expected_ep_assignment,
-                                 int opset = 22,
-                                 float tolerance = 0.004f) {
+                                 bool enable_htp_monolithic_lstm = false,
+                                 float tolerance = 0.04f,
```
qti-hungjuiw: Just to confirm— is this a typo, or do we actually need to change the `tolerance` from 0.004 to 0.04?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -712,6 +712,29 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+  } else if (enable_htp_monolithic_lstm_str == "0") {
+    enable_HTP_monolithic_lstm_ = false;
+  } else {
+    ORT_CXX_LOG(logger_,
+                ORT_LOGGING_LEVEL_VERBOSE,
```
qti-hungjuiw: Would it be better to turn this into `ORT_LOGGING_LEVEL_WARNING` or `ORT_LOGGING_LEVEL_ERROR`?

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -654,111 +653,47 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
...
+      utils::GetUniqueName(node_unit, "_QNN_LSTM_output_cell_state_" + direction),
+      utils::GetUniqueName(node_unit, "_QNN_LSTM_output_hidden_state_" + direction)};
+
+  // QNN output order: all_hidden, last_cell, last_hidden
+  // ONNX output order: all_hidden, last_hidden, last_cell
+  std::map<size_t, int> output_position_map = {
+      {0, 0},
+      {1, 2},
+      {2, 1}};
+  for (size_t j = 0; j < 3; j++) {
```
qti-hungjuiw:
```suggestion
  for (size_t j = 0; j < qnn_lstm_output_names .size(); j++) {
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chenweng-quic: Fixed, thanks.

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chenweng-quic: Fixed, thanks.

onnxruntime/test/providers/qnn/lstm_test.cc
chenweng-quic: Fixed, thanks.

onnxruntime/test/perftest/ort_test_session.cc
chenweng-quic: Fixed, thanks.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
tirupath-qti: Can you please respond to below comment ?
"
Also another important point, what could be the guideline if an onnx model contains more than one of the LSTM variants? Instead of user explicitly selecting this option, can't the Backend apply different logic for different LSTm layers based on their configuration?
"

tirupath-qti: > Some information from HTP team: The monolithic_lstm_flag is a trade-off setting. Enabling it can significantly improve prepare time and reduce the context binary size, but it may lead to slower execution time. Users should decide whether to use this option based on their specific scenario.

Could you please share the trade-off data for the review? is the slower execution time appears for all models or only for few models? what are the characteristics of this behavior when could this happen and when it could not?

cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
```diff
@@ -0,0 +1,68 @@
+From c66acc1ea78522cf91c8fdd796474340e17299de Mon Sep 17 00:00:00 2001
```
qti-yuduo: How are these patches related to the LSTM change? Patch 1 and 2 both seem completely unrelated. 

Higher level question: is this the step to follow when requires ort-core change? This does not sounds scalable.. I thought we already decoupled the test harness?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -712,6 +712,29 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    ORT_CXX_LOG(logger_,
+                ORT_LOGGING_LEVEL_ERROR,
+                ("Invalid enable_HTP_monolithic_lstm: " +
+                 enable_htp_monolithic_lstm_str +
+                 " only 0 or 1 allowed. Set to 0.")
+                    .c_str());
+  }
+  ORT_CXX_LOG(logger_,
+              ORT_LOGGING_LEVEL_VERBOSE,
+              ("User specified enable_htp_monolithic_lstm: " + enable_htp_fp16_precision_str).c_str());
```
qti-yuduo: ("User specified enable_htp_monolithic_lstm: " + enable_htp_fp16_precision_str).c_str()
//                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// Should be: enable_htp_monolithic_lstm_str

onnxruntime/test/providers/qnn/lstm_test.cc
```diff
@@ -884,7 +867,686 @@ TEST_F(QnnHTPBackendTests, DISABLED_LSTM_Fp16_sanity_bidirectional_Y_c_only) {
...
+      std::ref(P_def),                                                                         // P
+      true,                                                                                    // has_Y
+      true,                                                                                    // has_Y_h
+      true,                                                                                    // has_Y_c
+      direction,                                                                               // direction
+      hidden_size,                                                                             // hidden_size
+      0,                                                                                       // layout
+      ExpectedEPNodeAssignment::All,
+      true,
+      0.7);
```
qti-yuduo, line 1336: 0.7 is surprisingly high threshold..

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -654,111 +653,47 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
...
+
+  std::vector<std::string> qnn_lstm_output_names = {
+      utils::GetUniqueName(node_unit, "_QNN_LSTM_output_all_hidden_state_" + direction),
+      utils::GetUniqueName(node_unit, "_QNN_LSTM_output_cell_state_" + direction),
+      utils::GetUniqueName(node_unit, "_QNN_LSTM_output_hidden_state_" + direction)};
+
+  // QNN output order: all_hidden, last_cell, last_hidden
+  // ONNX output order: all_hidden, last_hidden, last_cell
+  std::map<size_t, int> output_position_map = {
+      {0, 0},
```
qti-yuduo: Key here is simply index (wasted), `int map[] = {0, 2, 1}` / `std::array` is much simpler.

onnxruntime/test/providers/qnn/lstm_test.cc
```diff
@@ -230,7 +230,7 @@ static GetTestQDQModelFn<InputQType> BuildQDQLSTMTestCase(const TestInputDef<flo
   };
 }
 
-#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
```
qti-yuduo, line 233: Could you explain why removing?

onnxruntime/test/providers/qnn/lstm_test.cc
chenweng-quic: This threshold is for android test in CI, for homoa device I recall the value is not so high.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chenweng-quic: Thanks, will fix

cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
chenweng-quic: In current build, the onnxruntime_provider_test is not built from onnxruntime-qnn, instead, it is built from origin onnxruntime.
We have to patch ort so these executable can accept new argument.
I think once we discard onnxruntime in the future, we don't need these patches.

onnxruntime/test/providers/qnn/lstm_test.cc
chenweng-quic: I recall that some finalization error happen in linux because the soc_model is not able to set.
So I disable it.

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chenweng-quic: Fixed, thanks

qti-chuteng (approved):

onnxruntime/test/providers/qnn/lstm_test.cc
```diff
@@ -513,12 +500,12 @@ TEST_F(QnnHTPBackendTests, DISABLED_LSTM_QDQ_sanity_bidirectional_all_initialize
...
-                               22,
+                               false,
                                QDQTolerance(0.008f));
 }
 
-// Fails with QNN SDK 2.35.0:
-// Failed to finalize QNN graph. Error code: 1002
+// disable since there is a bug in optional output
+// see https://github.com/microsoft/onnxruntime/pull/27301
 TEST_F(QnnHTPBackendTests, DISABLED_LSTM_QDQ_sanity_bidirectional_Y_only) {
```
kuanyul-qti, line 510: The https://github.com/microsoft/onnxruntime/pull/27301 is closed, and it seems to be fixed by https://github.com/microsoft/onnxruntime/pull/27644 which is merged. Could you check the ORT latest mainline to see if the issue exists? If not, we can re-enable the tests.

chenweng-quic: We need to wait for it be released, and our build flow uplevel the ort version so we can enable the tests.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -187,6 +187,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
   qnn::HtpGraphFinalizationOptimizationMode htp_graph_finalization_opt_mode_ = qnn::HtpGraphFinalizationOptimizationMode::kDefault;
   int32_t vtcm_size_in_mb_ = 0;
   bool enable_HTP_FP16_precision_ = true;
+  bool disable_htp_monolithic_lstm_ = false;
```
kuanyul-qti, line 190: Should it be true as you change the flag to disable_htp_monolithic_lstm_ ?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -732,6 +732,29 @@ QnnEp::QnnEp(QnnEpFactory& factory,
               ORT_LOGGING_LEVEL_VERBOSE,
               ("User specified enable_htp_fp16_precision: " + enable_htp_fp16_precision_str).c_str());
 
+  // HTP monolithic lstm
+  std::string disable_htp_monolithic_lstm_str;
+  GetSessionConfigEntryOrDefault(ort_api,
+                                 session_options_,
+                                 FormatEPConfigKey("disable_htp_monolithic_lstm"),
+                                 "0",
```
kuanyul-qti, line 740: Should the default value be 1?

onnxruntime/test/providers/qnn/lstm_test.cc
kuanyul-qti: Could you leave a TODO in the code?

onnxruntime/core/providers/qnn/qnn_execution_provider.h
chenweng-quic: per Tiru's comment, he wants the flag default on (enable_htp_monolithic_lstm=true in previous commint), I am updating it to disable_htp_monolithic_lstm_ = false

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chenweng-quic: per Tiru's comment, he wants the flag default on (enable_htp_monolithic_lstm=true in previous commint), I am updating it to disable_htp_monolithic_lstm_ = false

onnxruntime/test/providers/qnn/lstm_test.cc
chenweng-quic: done

kuanyul-qti (approved):

PR #117, opened 2026-03-07
[QNN-EP] Implement file mapped weights feature (#26952)

quic-calvnguy:
**** This is a missing PR/commit from the ORT repo. Changes are based on this commit: 
https://github.com/microsoft/onnxruntime/commit/f83d4d06e4283d53a10c54ce84da3455cfb4e21d

**** Additionally, the following fix for an edge case has been cherry-picked over too:
https://github.com/microsoft/onnxruntime/commit/b0b5de39eaab2ae24e2dc489ddff4cf0aaa8ccc4

Description
Enables the file mapping of weights as well as the overall context bin. This feature is currently only enabled for ARM64 WIN devices

Motivation and Context
Currently, when reading the context bin, ORT allocates a large buffer on the heap. Assuming the same model is used, each ORT session will allocate a buffer for the context bin. This is incredibly wasteful when large models are used. Instead, WIN file mapping can be leveraged to map the context bin, then every time a context needs to be created with the context bin, the pointer to the context bin can be retrieved and used instead of some pre-allocated buffer, thus making QNN EP more memory-efficient. In the case of multiple ORT sessions, the context bin will only be loaded once for all sessions, increasing memory efficiency and overall initialization performance. This is very useful regarding the use of LLMs going forward.


minfhong-qti:
Thanks for migrating this feature back to our repo. I've tried it myself in #114 with all necessary ABI replacements.
During my migration, I've discovered a bug in a corner case, specifically in `QnnBackendManager::LoadCachedQnnContextFromBuffer` not considering the case that `context_bin_filepath` could be empty. I've tried to solve this bug in my PR but still failed to pass all CIs.
Maybe starting from my branch could save you a lot of efforts.

quic-calvnguy:
@minfhong-qti The bug was found internally, and its fix was what notified me of our repo missing the file mapping feature. I just cherry-picked the fix from the main ORT repo in case you want to verify the change

quic-calvnguy:
@minfhong-qti I can fix the logging issues on Monday

onnxruntime/core/providers/qnn/builder/qnn_windows_file_mapper.h
```diff
@@ -0,0 +1,42 @@
...
+  // Creates a file mapping of the context binary and returns the
+  // mapview pointer of the file mapping
+  Status GetContextBinMappedMemoryPtr(const std::string& bin_filepath,
+                                      void** mapped_data_ptr) override;
+
+ private:
+  // A container of smart pointers of mapview memory pointers to mapped context bins
+  // key: filepath to context bin, value: smart pointer of mapview memory pointers
+  std::mutex map_mutex_;
+  std::unordered_map<std::string, onnxruntime::Env::MappedMemoryPtr> mapped_memory_ptrs_;
```
quic-calvnguy: @minfhong-qti  do you know what i need to do to use this MappedMemoryPtr class? it seems onnxruntime::Env is no longer accessible as written

minfhong-qti: Suggest to directly define it here.
`using MappedMemoryPtr = std::unique_ptr<char[], std::function<void(void*)>>;`

quic-calvnguy:
This PR includes commits from this PR that resolves unit test failures due to ordering of destruction between the EP and ORT session:
https://github.com/onnxruntime/onnxruntime-qnn/pull/175

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -819,22 +823,145 @@ Ort::Status SetQnnContextConfig(ContextPriority context_priority, QnnContext_Con
...
+    return QNN_CONTEXT_ERROR_ABORTED;
+  }
+
+  if (mapped_base_ptr == nullptr) {
+    ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_ERROR, "Attempting to map DMA data for null memory mapped base pointer");
+    return QNN_CONTEXT_ERROR_INVALID_ARGUMENT;
+  }
+  ORT_CXX_LOG_PTR(logger_ptr_,
+                  ORT_LOGGING_LEVEL_INFO,
+                  ("Mapping DMA data for request: memory mapped base pointer(" + utils::PtrToString(mapped_base_ptr) + "), offset(" + std::to_string(request.offset) + "), size(" + std::to_string(request.size) + "), total file size(" + std::to_string(file_size) + ") isBackendMappingNeeded(" + std::to_string(request.isBackendMappingNeeded) + ")").c_str());
```
quic-calvnguy, line 862: I tried splitting these log lines up into multiple lines, but lintrunner always reverts it back to one line

qti-jkilpatrick (approved):

PR #119, opened 2026-03-09
Update wheels for deployment

kuanyul-qti:
### Description
Build x86_64 wheels and add a step to merge wheels related to amd.
Minor fix to the uplevel script to update the version in build_and_package_info.py.

### Motivation and Context
We want to run QNN-EP on windows x64 device, but we only have arm64ec wheel currently. It doesn't work if we install the arm64ec wheel on windows x64 device, so we need to add x64 wheel for release.
However, another problem is that the name of x64 wheel is the same as the name of arm64ec wheel, which could cause confusion and result in one wheel overwriting the other.

- x64 wheel name: onnxruntime_qnn-2.0.0-cp311-cp311-win_amd64.whl
- arm64ec wheel name: onnxruntime_qnn-2.0.0-cp311-cp311-win_amd64.whl

Therefore, this PR aims to merge the x64 and arm64ec wheels into a single wheel so that the wheel can run on both Windows-x64 and Windows-arm64 devices.

### DIFF
--- a/qcom/scripts/upleveling/merge_wheels.py
+++ b/qcom/scripts/upleveling/merge_wheels.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+# ==============================================================================
+#
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# All Rights Reserved.
+# Confidential and Proprietary - Qualcomm Technologies, Inc.
+#
+# ==============================================================================
+
+"""
+Merge two architecture-specific wheels into a unified wheel package.
+
+This script takes AMD64 and ARM64EC wheels and combines them into a single
+wheel with both library sets in separate subdirectories.
+"""
+
+import argparse
+import logging
+import shutil
+import subprocess
+import tempfile
+import zipfile
+from pathlib import Path
+
+
+def extract_wheel(wheel_path, extract_dir):
+    """Extract a wheel file to a directory."""
+    logging.info(f"  Extracting: {wheel_path}")
+    with zipfile.ZipFile(wheel_path, "r") as zip_ref:
+        zip_ref.extractall(extract_dir)
+    return extract_dir
+
+
+def get_package_dir(extract_dir):
+    """Find the main package directory in extracted wheel."""
+    extract_path = Path(extract_dir)
+    # Look for onnxruntime_qnn directory
+    for item in extract_path.iterdir():
+        if item.is_dir() and item.name.startswith("onnxruntime"):
+            if not item.name.endswith(".dist-info"):
+                return item
+    raise ValueError(f"Could not find package directory in {extract_dir}")
+
+
+def is_library_file(filename):
+    """Check if a file is a library file that should be moved."""
+    suffixes = {".dll", ".so", ".cat", ".pyd", ".dylib"}
+    return Path(filename).suffix.lower() in suffixes
+
+
+def merge_wheels(amd64_wheel, arm64ec_wheel, output_folder):
+    """
+    Merge two architecture-specific wheels into a unified wheel.
+
+    Args:
+        amd64_wheel: Path to AMD64 wheel file
+        arm64ec_wheel: Path to ARM64EC wheel file
+        output_folder: Output folder path for output unified wheel
+    """
+    logging.info("=" * 80)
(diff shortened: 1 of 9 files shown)


qti-jkilpatrick:
Before I dig into the implementation, could you share a little more on the plan? I'm not familiar with this strategy for making a single Python wheel that works on both Windows architectures. Can you share a link to a PEP or discussion where this has been done previously? Or perhaps point to some packages where this is done?

kuanyul-qti:
> Before I dig into the implementation, could you share a little more on the plan? I'm not familiar with this strategy for making a single Python wheel that works on both Windows architectures. Can you share a link to a PEP or discussion where this has been done previously? Or perhaps point to some packages where this is done?

Hi Jeff, I've added you in a teams channel for this issue.

qti-jkilpatrick:
> > Before I dig into the implementation, could you share a little more on the plan? I'm not familiar with this strategy for making a single Python wheel that works on both Windows architectures. Can you share a link to a PEP or discussion where this has been done previously? Or perhaps point to some packages where this is done?
> 
> Hi Jeff, I've added you in a teams channel for this issue.

I'm sorry, but I seem to have missed the notification somewhere. Could you please ping me on Teams directly?

kuanyul-qti:
> > > Before I dig into the implementation, could you share a little more on the plan? I'm not familiar with this strategy for making a single Python wheel that works on both Windows architectures. Can you share a link to a PEP or discussion where this has been done previously? Or perhaps point to some packages where this is done?
> > 
> > 
> > Hi Jeff, I've added you in a teams channel for this issue.
> 
> I'm sorry, but I seem to have missed the notification somewhere. Could you please ping me on Teams directly?

Sure

qcom/scripts/upleveling/merge_wheels.py
qti-jkilpatrick, line 1: * I'm not sure I'm comfortable editing `__init__.py` in this way. Can we have one that simply does the right thing when it's on x86_64?
* Assuming we can live with just a single `__init__.py`, we might be able to simplify this by simply opening a copy of the wheel as a zip file and appending any new files.

kuanyul-qti: Hi Jeff, as the table I posted in the chat -

- If we live with just a single `__init__.py`, which means changing the architecture for all platforms from `onnxruntime_qnn` to `onnxruntime_qnn/libs/<arch>`, there are two issues:

  1. It will affect all the wheels including x64, arm64ec, Linux x64 and Windows ARM regardless of whether we plan to merge them or not.
  2. It is hard to determine the `arch` when calling `setup.py` (used to build wheel); the arch is affected by multiple factors, and the command line arguments are currently not available in `setup.py`.

      -  Example of the factors that affect the `arch`: the host machine arch, the target machine arch, the command line arguments (e.g., --arm64, --arm64ec), etc.
 
- If we only change the architecture when it's on x86_64, there are two issues:

  1. The architectures of x64 and arm64ec wheels are affected regardless of whether we plan to merge them or not.
  2. It is hard to determine the `arch` when calling `setup.py`.

  - Besides, I'm not sure if maintaining two `__init__.py` files in the code repository is allowed.

qti-jkilpatrick: > It will affect all the wheels including x64, arm64ec, Linux x64 and Windows ARM regardless of whether we plan to merge them or not.

It's totally reasonable to have branches that are irrelevant to the current architecture. On Windows, `__init__.py` needs to determine the interpreter and underlying architectures. Is there a reason this can't done in an `if`?

> It is hard to determine the arch when calling `setup.py`

No need. Just do it all at runtime.

kuanyul-qti: Update the `__init__.py` so that we maintain one init file.

qti-jkilpatrick: Great! What about this one?

> Assuming we can live with just a single `__init__.py`, we might be able to simplify this by simply opening a copy of the wheel as a zip file and appending any new files.

kuanyul-qti: Done

qti-chuteng (approved):

PR #121, opened 2026-03-09
Fix ScatterElements UT failures by aligning input encodings

yuhuchua-qti:
### Description
Fix ScatterElements unit test failures by aligning input encodings. Aggregated min/max values of both inputs (data and updates) when calculating encodings and re-enabled the previously disabled unit test.

### Motivation and Context
QNN HTP assumes that ScatterElements inputs share the same encoding. The previous UT implementation did not consider this implicit assumption, leading to failures. Combining min/max values ensures consistent encoding and resolves the issue.

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_test_utils.h
+++ b/onnxruntime/test/providers/qnn/qnn_test_utils.h
@@ -1452,19 +1452,26 @@ inline GetTestQDQModelFn<QuantType> BuildQDQOpTestCase(
     const std::vector<ONNX_NAMESPACE::AttributeProto>& attrs,
     const std::string& op_domain = kOnnxDomain,
     bool use_contrib_qdq = false,
-    AllocatorPtr input_allocator = nullptr) {
+    AllocatorPtr input_allocator = nullptr,
+    bool combine_quant_inputs_qparams = false) {
   return [node_name, op_type, quant_input_defs, non_quant_input_defs, quant_input_defs_2, attrs, op_domain,
-          use_contrib_qdq, input_allocator](
+          use_contrib_qdq, input_allocator, combine_quant_inputs_qparams](
              ModelTestBuilder& builder, std::vector<QuantParams<QuantType>>& output_qparams) {
     std::vector<std::string> op_input_names;
 
     op_input_names.reserve(quant_input_defs.size() + non_quant_input_defs.size() + quant_input_defs_2.size());
 
+    std::vector<TestInputDef<float>> combined_input_defs;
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -328,6 +328,22 @@ inline QuantParams<QType> GetTestInputQuantParams(const TestInputDef<float>& inp
   return QuantParams<QType>::Compute(frange.first, frange.second, symmetric);
 }
 
+template <typename QType>
+inline QuantParams<QType> GetTestInputsQuantParams(const std::vector<TestInputDef<float>>& input_defs,
+                                                   bool symmetric = false) {
+  std::pair<float, float> frange = {std::numeric_limits<float>::max(), std::numeric_limits<float>::lowest()};
```
qti-hungjuiw, line 343: Should this be `std::numeric_limits<float>::lowest()` or `std::numeric_limits::min()`?

onnxruntime/test/providers/qnn/simple_op_test.cc
```diff
@@ -1457,7 +1472,11 @@ TEST_F(QnnHTPBackendTests, ScatterElements_int8_reduction_mul) {
                                      test::MakeAttribute("reduction", "mul"),
                                  },
                                  17,
-                                 ExpectedEPNodeAssignment::All);
+                                 ExpectedEPNodeAssignment::All,
+                                 kOnnxDomain,
+                                 false,
+                                 QDQTolerance(),
+                                 true);
```
qti-hungjuiw: Please indicate this argument in comment. Here and other places.
```suggestion
                                 true); // combine_quant_inputs_qparams
```

qti-hungjuiw:
LGTM

onnxruntime/test/providers/qnn/qnn_test_utils.h
qti-hungjuiw: Perhaps there can be a comment to emphasize that we use `std::numeric_limits<float>::lowest()` to include negative values.

yuhuchua-qti: Sure. I've added a comment about this.

onnxruntime/test/providers/qnn/simple_op_test.cc
yuhuchua-qti: I've updated this in PR. Thanks.

qti-hungjuiw (approved):

qti-jkilpatrick (approved):

PR #118, opened 2026-03-09
Fix roialign_op_test.cc on the new qnn test infra

qti-hungjuiw:
### Description

- Updated roialign_op_test.cc to run correctly on the latest test infrastructure introduced in #23.
- Added a patch to disable related test cases in `onnxruntime_plugin_ep_onnx_test`.

### Motivation and Context

- The pull request for roialign_op_test.cc (#89) was merged before being validated against the latest main branch.
- This change addresses the resulting CI failures and ensures the tests run properly on the updated infrastructure.

PR #123, opened 2026-03-09
Revert change to fix build

qti-jkilpatrick:
Revert #89, which broke the build, and #118, which built on top of it.


qti-jkilpatrick:
This is wrong. The oldest reverted change is fixed by the other one, but the nightly wheel build was broken by something else.

PR #124, opened 2026-03-09
Don't enable ccache in the "prebuilt" build unless it's enabled in the host build

qti-jkilpatrick:
### Description

CCache is used on demand in the ORT build: don't enable it in nested builds unless it's requested in the main build.

### Motivation and Context

When building for aarch64 manylinux, ccache is disabled by default: a sensible location for a ccache root must be specified in the environment. Otherwise, ccache will waste temporary disk space on a root that will not persist between runs. Note that CI runners are configured to use ccache via the environment variable `ORT_BUILD_DOCKER_CCACHE_ROOT`.

### DIFF
--- a/cmake/external/onnxruntime_prebuilt.cmake
+++ b/cmake/external/onnxruntime_prebuilt.cmake
@@ -67,8 +67,10 @@ else()
         --use_qnn "${QNN_LIBRARY_KIND}"
         --qnn_home "${onnxruntime_QNN_HOME}"
         --no_kleidiai
-        --use_cache
     )
+    if(onnxruntime_BUILD_CACHE)
+        list(APPEND ORT_BUILD_COMMAND "--use_cache")
+    endif()
     if (${CMAKE_SYSTEM_NAME} STREQUAL "Android")
         list(APPEND ORT_BUILD_COMMAND --android)
         list(APPEND ORT_BUILD_COMMAND --android_sdk_path)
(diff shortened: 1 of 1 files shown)


qti-hungjuiw (approved):
Thanks @qti-jkilpatrick! I am also considering the modification.

cmake/external/onnxruntime_prebuilt.cmake
```diff
@@ -180,6 +182,9 @@ ExternalProject_Add(
     LOG_INSTALL ON
     LOG_OUTPUT_ON_FAILURE ON
     LOG_MERGED_STDOUTERR ON
+
+    # Don't run more than one Ninja build at a time
+    USES_TERMINAL_BUILD ON
```
qti-hungjuiw, line 187: Just curious, do we need the `USES_TERMINAL_BUILD` given that we can access the build log in `Release\ort_core_target-prefix\src\ort_core_target-stamp\ort_core_target-build.log`?

qti-jkilpatrick: The issue isn't log accessibility, but rather that we don't want to run `ninja` more than once at a time since every instance will attempt to use all CPU cores. This [is a side-effect](https://cmake.org/cmake/help/latest/module/ExternalProject.html#terminal-access-options) of `USE_TERMINAL_*`.

PR #138, opened 2026-03-10
[QNN EP] Enable offline x64 compilation with memhandle IO type (#27479)

quic-calvnguy:
*** This is cherry-pick from this PR merge: https://github.com/microsoft/onnxruntime/commit/d626b568e01e394e5f17c92e95f0e21e36a73a7e#diff-ee58cc5d11eb6509cb8fc2f796a8fa847761fdd3d9fe15b2364a8594475f659a

This change is required for the ABI and non-ABI repos.

*** Original Description:

Enable offline compilation for QNN EP with MEMHADNLE IO type. It was previously enabled only on ARM because the QNN EP was loading rpcmem library (only available on ARM drivers), which is not actually used for the compilation (it is required only for inference to allocate the shared memory).

Ensured that the MEMHANDLE IO type is correctly set regardless of the way how `QnnTensorWrapper` is created (either through factory function or by creating it directly). This ensures that mem type will be correctly configured regardless of the op builder implementation.


quic-calvnguy:
This commit resolves two issues:

1.  Currently if the HTP shared memory allocator is enabled during context bin generation, libcdsprpc will be loaded and will result in failure for hosts that do not have libcdsprpc installed (e.g. Linux & WIN x86 systems). An if statement is added to resolve this issue
2. The IO binding test using the HTP shared allocator is failing due to how memtype is handled, resulting in input/output tesnors allocating to RAM rather than the NPU memory space. This is also resolved.

qti-jkilpatrick (approved):

qti-chuteng (approved):

PR #140, opened 2026-03-11
Add support for BN with activation in u8, scale in s8

quic-ykota:
### Description
Updates BN op builder to support the following combination: input/output -> uint8, scale -> int8

### Motivation and Context
Several models fail compilation because of unsupported optype for batchnorm op. This PR addresses the issue with sfixed8 input for scale by converting it into ufixed8 instead.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
@@ -466,6 +466,11 @@ void OverrideParamTypeForRequantize(Qnn_DataType_t x_dtype,
     scale_dtype = is_scale_has_negative_values ? QNN_DATATYPE_SFIXED_POINT_16 : QNN_DATATYPE_UFIXED_POINT_8;
   }
 
+  // QNN HTP with UFIXED_POINT_8 input doesn't support SFIXED_POINT_8 scale
+  if (x_dtype == QNN_DATATYPE_UFIXED_POINT_8 && scale_dtype == QNN_DATATYPE_SFIXED_POINT_8) {
+    scale_dtype = QNN_DATATYPE_UFIXED_POINT_8;
+  }
+
   // QNN HTP requires quantized bias for quantized ops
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -466,6 +466,11 @@ void OverrideParamTypeForRequantize(Qnn_DataType_t x_dtype,
     scale_dtype = is_scale_has_negative_values ? QNN_DATATYPE_SFIXED_POINT_16 : QNN_DATATYPE_UFIXED_POINT_8;
   }
 
+  // QNN HTP with UFIXED_POINT_8 input doesn't support SFIXED_POINT_8 scale
```
qti-yuduo: I would add a short explanation on why mathematically this safe.

onnxruntime/test/providers/qnn/batchnormalization_test.cc
```diff
@@ -550,6 +550,26 @@ TEST_F(QnnHTPBackendTests, BatchNorm2dQdqParams) {
                        21,
                        ExpectedEPNodeAssignment::All);
 }
+// Test BatchNorm with U8 input, S8 scale (converted to U8), float bias (converted to S32)
```
qti-yuduo, line 554: nit: new space between tests.

onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
qti-yuduo, line 463: We're not setting `is_scale_has_negative_values` from caller side. With this function extended, it feels more and more risky. 

For a model with U16 input and all-positive S8 scale (`BatchNorm2dQdqParams` test setup), validation checks `{U16, S16, S32, U16}` against the dtype table, while the built QNN graph actually contains `{U16, U8, S32, U16}`, see `ProcessInputs()`. Should we set this `is_scale_has_negative_values` in validation call?

quic-ykota: I am not sure why we would need is_scale_has_negative_values here, and this decision could be taken without checking the actual scale min values. How about I refactor this separately and put up a different PR for this?

qti-yuduo (approved):

PR #144, opened 2026-03-11
[QNN EP] Add support for all params for Slice Op in QNN EP

qti-ashwshan:
### Description
- Slice Op in QNN EP has some missing params such as begin_mask, end_mask, shrink_axis and new_axes_mask
- This PR adds support for the missing params to run fully on the NPU
- Added unit tests and additional helper function to fetch the params added to the Op after the run on NPU.

### Motivation and Context
- A few params of QNN HTP Op for Slice was not supported as part of the original QNN EP op builder

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/slice_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/slice_op_builder.cc
@@ -282,14 +282,34 @@ Ort::Status SliceOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mod
   RETURN_IF_ERROR(PrepareForComputeHelper(raw_starts, raw_ends, raw_axes, raw_steps, compute_metadata));
 
   const size_t input_rank = input_dimensions.size();
+
+  // Determine which axes were explicitly specified by the ONNX Slice op.
+  // Unspecified axes use default ranges (start=0, end=dim, step=1) and will have
+  // their begin_mask/end_mask bits set so QNN uses the full range for those dimensions.
+  InlinedHashSet<size_t> specified_axes;
+  if (raw_axes.empty()) {
+    // When axes are omitted, ONNX Slice implicitly specifies axes [0, ..., len(starts)-1].
+    for (size_t i = 0; i < raw_starts.size(); i++) {
+      specified_axes.insert(i);
+    }
+  } else {
+    for (const auto& axis : raw_axes) {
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/slice_op_builder.cc
```diff
@@ -282,28 +282,73 @@ Ort::Status SliceOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mod
...
+  } else {
+    for (const auto& axis : raw_axes) {
+      auto resolved = axis < 0 ? axis + static_cast<int64_t>(input_rank) : axis;
+      specified_axes.insert(static_cast<size_t>(resolved));
+    }
+  }
+
+  // Build the ranges parameter [rank, 3] with (begin, end, stride) per axis.
+  // Cast through int32_t first to correctly preserve the sign bit for negative values
+  // (e.g., end=-1 for backward slicing), since QNN interprets this data as INT_32.
```
minfhong-qti, line 304: I think there should be no negative value here as `PrepareForComputeHelper` already handles it.

onnxruntime/core/providers/qnn/builder/opbuilder/slice_op_builder.cc
```diff
@@ -282,28 +282,73 @@ Ort::Status SliceOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mod
...
                                       std::move(ranges_dims),
                                       std::move(ranges_data),
-                                      true);
-  std::string param_tensor_name(ranges_paramwrapper.GetParamTensorName());
+                                      /*is_signed=*/true);
+  std::vector<std::string> param_tensor_names;
+  param_tensor_names.push_back(ranges_paramwrapper.GetParamTensorName());
   qnn_model_wrapper.AddParamWrapper(std::move(ranges_paramwrapper));
+
+  // Compute begin_mask and end_mask bitmasks. For axes not explicitly specified by the
```
minfhong-qti, line 325: I think `PrepareForComputeHelper` as well handles such cases.

qti-ashwshan: Looks like PrepareForComputeHelper  is handling the attributes begin_mask and end_mask and ensuring that the math is changed so the attributes are not seen in the Final operator in QNN EP. So this change is not required.

PR #145, opened 2026-03-11
[QNN EP] Support for Beta Parameter in Softmax and Log Softmax operation

qti-mbadnara:
### Description

- Add support for the QNN-specific beta attribute for the `Softmax` and `LogSoftmax` operator in the QNN EP `SoftmaxOpBuilder`. 
- The beta value is set to default `1.0f` forwarded to QNN.

### Motivation and Context

The QNN backend supports a beta parameter for the `Softmax` and `LogSoftmax` operator that scales the input before applying the operation. `ORT QNN EP` was not forwarding this attribute to QNN as it is not an ORT attribute. This resulted in operator disparity between `QAIRT Native` and `ORT Graph`. This change resolved the disparity.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/softmax_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/softmax_op_builder.cc
@@ -176,6 +176,20 @@ Ort::Status SoftmaxOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_m
   RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(outputs[0], output_info));
   size_t output_rank = output_info.shape.size();
 
+  // Builds the axis param and, for LogSoftmax, the beta param, into param_tensor_names.
+  auto AddAxisAndBetaParams = [&](std::vector<std::string>& param_tensor_names,
+                                  Qnn_Scalar_t effective_axis_scalar) {
+    QnnParamWrapper axis_param(node_unit.Index(), node_unit.Name(), QNN_OP_SOFTMAX_PARAM_AXIS, effective_axis_scalar);
+    param_tensor_names.push_back(axis_param.GetParamTensorName());
+    qnn_model_wrapper.AddParamWrapper(std::move(axis_param));
+
+    if (is_log_softmax) {
+      QnnParamWrapper beta_param(node_unit.Index(), node_unit.Name(), QNN_OP_LOG_SOFTMAX_PARAM_BETA, beta_qnn_scalar);
+      param_tensor_names.push_back(beta_param.GetParamTensorName());
+      qnn_model_wrapper.AddParamWrapper(std::move(beta_param));
+    }
+  };
+
   if (opset_version < 13) {
     std::string reshape_input_name = utils::GetUniqueName(orig_output_name, "_reshape");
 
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/simple_op_test.cc
```diff
@@ -751,6 +751,85 @@ TEST_F(QnnHTPBackendTests, UnaryOp_LogSoftmax11_SetAxis) {
...
+                        ExpectedEPNodeAssignment::All);
+}
+
+// Test QNN EP correctly handles the QNN-specific beta attribute on LogSoftmax (opset 13).
+// With beta=1.0 (the default), the output must match the standard LogSoftmax.
+TEST_F(QnnHTPBackendTests, UnaryOp_LogSoftmax13_WithBeta) {
+  std::vector<float> input_data = GetFloatDataInRange(-5.0f, 5.0f, 6);
+  RunQDQOpTest<uint8_t>("LogSoftmax",
+                        {TestInputDef<float>({1, 2, 3}, false, input_data)},
+                        {test::MakeAttribute("beta", 1.0f)},  // Explicit default beta; QNN passes it through
```
qti-ashwshan: beta is a QNN param not an ONNX attribute.

onnxruntime/core/providers/qnn/builder/opbuilder/softmax_op_builder.cc
```diff
@@ -177,10 +199,8 @@ Ort::Status SoftmaxOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_m
       axis_qnn_scalar.uint32Value = 1;
     }
 
-    QnnParamWrapper axis_param(node_unit.Index(), node_unit.Name(), QNN_OP_SOFTMAX_PARAM_AXIS, axis_qnn_scalar);
     std::vector<std::string> param_tensor_names;
-    param_tensor_names.push_back(axis_param.GetParamTensorName());
-    qnn_model_wrapper.AddParamWrapper(std::move(axis_param));
+    AddAxisAndBetaParams(param_tensor_names, axis_qnn_scalar);
```
qti-ashwshan: We don't need an additional function for adding the params for Axis and Beta.

qti-mbadnara: Updated as part of the latest change

onnxruntime/test/providers/qnn/simple_op_test.cc
qti-mbadnara: Updated as part of the latest change

qti-yuduo (commented):
Not following the motivation of adding this attribute. How can we use it as it's not in standard ONNX?

qti-mbadnara:
The QAIRT SDK supports it but QNNX doesnt support it natively. This PR resolves the disparity.

qti-ashwshan:
> Not following the motivation of adding this attribute. How can we use it as it's not in standard ONNX?

According to the QNN HTP Op spec we have an attribute beta which defaults at 1. and When beta is 1 the operation is same as LogSoftMax Onnx OP. We can either propagate the same beta = 1 value or decide to avoid it but have graph differences in the native and delegate workflow.

minfhong-qti:
> > Not following the motivation of adding this attribute. How can we use it as it's not in standard ONNX?
> 
> According to the QNN HTP Op spec we have an attribute beta which defaults at 1. and When beta is 1 the operation is same as LogSoftMax Onnx OP. We can either propagate the same beta = 1 value or decide to avoid it but have graph differences in the native and delegate workflow.

May I know whether there is a real usecase for this or any motivation behind this change?

qti-mbadnara:
Closing this PR as the beta flag is not present in the EP path today. Without it being applied in either ORT or the EP, configuring a beta flag does not introduce any functional difference.

PR #151, opened 2026-03-13
Install Qualcomm Device Cloud SDK from Software Center instead of internal PyPI

qti-mbadnara:
### Description

Switch the QDC Python client install from the internal self-hosted PyPI serve to the publicly released Qualcomm Device Cloud SDK on Software Center. The SDK zip is registered in `packages.yml` and downloaded via `package_manager.py`. The extracted wheel is passed directly to `uv pip install` alongside `requirements-qdc.txt`, letting the resolver satisfy the existing `qdc-public-api-client` requirement from the local wheel. The wheel is resolved lazily at execution time so the download only occurs when the QDC venv task actually runs.

### Motivation and Context

Even after `Qualcomm Device Cloud SDK v0.3.0` was released on the `Qualcomm Software Center`, the build was still pulling it from an internal PyPI service, creating an unnecessary dependency on internal infrastructure. This change removes that dependency

### DIFF
--- a/qcom/ep_build/tasks/python.py
+++ b/qcom/ep_build/tasks/python.py
@@ -19,12 +19,21 @@ from ..task import (
     RunInTempDirectoryTask,
     Task,
 )
-from ..tools import PythonExecutableArchT, get_model_zoo_root, get_onnx_models_root, get_python_executable
+from ..tools import PythonExecutableArchT, get_model_zoo_root, get_onnx_models_root, get_package_content_dir, get_python_executable
 from ..util import (
     MSFT_CI_REQUIREMENTS_RELPATH,
     REPO_ROOT,
 )
 from .build import BuildConfigT, TargetPyVersionT, get_ort_version
+
+
+def _get_qdc_sdk_wheel() -> Path:
+    """Return the path to the QDC SDK wheel, downloading and extracting it if necessary."""
+    sdk_dir = get_package_content_dir(None, "qualcomm_device_cloud_sdk")
+    wheels = list(sdk_dir.glob("qualcomm_device_cloud_sdk-*.whl"))
+    if not wheels:
+        raise FileNotFoundError(f"Could not find Qualcomm Device Cloud SDK wheel in {sdk_dir}")
+    return wheels[0]
 
 
 def uv_pip_install_cmd(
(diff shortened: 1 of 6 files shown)


qcom/ep_build/tasks/python.py
```diff
@@ -19,14 +19,29 @@
...
 from ..util import (
     MSFT_CI_REQUIREMENTS_RELPATH,
     REPO_ROOT,
 )
 from .build import BuildConfigT, TargetPyVersionT, get_ort_version
 
 
+def _get_qdc_sdk_wheel(venv: Path | None = None) -> Path:
+    """Return the path to the QDC SDK wheel, downloading and extracting it if necessary."""
+    sdk_dir = get_package_content_dir(venv, "qualcomm_device_cloud_sdk")
```
qti-jkilpatrick: Create a helper function (perhaps `get_qualcomm_device_cloud_sdk_root`) in `qcom/ep_build/tools.py` so that knowledge of the name of the package remains confined to the package/tools system.

qcom/ep_build/tasks/python.py
```diff
@@ -19,14 +19,29 @@
...
+    wheels = list(sdk_dir.glob("qualcomm_device_cloud_sdk-*.whl"))
+    if not wheels:
+        raise FileNotFoundError(f"Could not find Qualcomm Device Cloud SDK wheel in {sdk_dir}")
+    return wheels[0]
```
qti-jkilpatrick: What if there's more than one?

qcom/ep_build/tasks/python.py
```diff
@@ -122,12 +150,19 @@ def __init__(self, python_executable: Path, venv_path: Path) -> None:
             tasks=[
                 CreateVenvTask(python_executable=python_executable, venv_path=venv_path),
                 PipInstallTask(
-                    f"Installing QDC build requirements into {venv_path}",
+                    f"Installing package manager requirements into {venv_path}",
                     venv_path,
-                    requirements=[
-                        REPO_ROOT / "qcom" / "requirements-qdc.txt",
+                    requirements=[REPO_ROOT / "qcom" / "requirements.txt"],
```
qti-jkilpatrick, line 152: Rather than having a single venv with build and QDC dependencies, can we keep them separate (even if the QDC one decides to install build dependencies under the covers -- we should not know/care if that's the case at this layer)?

qcom/requirements-qdc.txt
```diff
@@ -2,4 +2,3 @@
 # SPDX-License-Identifier: MIT
 
 -r requirements.txt
-qdc-public-api-client == 0.0.14
```
qti-jkilpatrick, line 5: Can we keep this in here (with the version number adjusted)? I would think that pip is smart enough to figure things out if you leave this and also give a wheel path in the same `pip install` invocation. The idea is that this file is the canonical description of requirements, while passing the wheel path is just an implementation detail.

qti-jkilpatrick (commented):
This is great. I just have a couple tweaks to maintain some not so obvious abstraction boundaries.

qcom/ep_build/tasks/python.py
qti-mbadnara: Resolved as part of the latest changes

qcom/ep_build/tasks/python.py
qti-mbadnara: Resolved as part of the latest changes

qcom/requirements-qdc.txt
qti-mbadnara: Resolved as part of the latest changes

qcom/ep_build/tasks/python.py
qti-mbadnara: Resolved as part of the latest changes

qti-jkilpatrick (approved):
Nice work!

PR #157, opened 2026-03-18
[QNN EP] Add mode attribute and set it to DCR for SpaceToDepth

qti-mbadnara:
### Description

Add support for the QNN-specific mode attribute on the `SpaceToDepth` operator in the `QNN EP`. 
The mode is set to `DCR` and forwarded to QNN. Added test case for `FP32`.

### Motivation and Context

The QNN backend supports a mode parameter for the `SpaceToDepth` operator that controls how spatial elements are packed into the channel dimension. `ORT QNN EP` was not forwarding this attribute to `QNN` as it is not an ORT attribute. This resulted in operator disparity between `QAIRT Native` and `ORT Graph`. This change resolved the disparity.


qti-ashwshan:
Please fix lint issues

qti-mbadnara:
> Please fix lint issues

Resolved!

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
```diff
@@ -426,6 +426,14 @@ Ort::Status SimpleOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mo
 
   if (op_type == "SpaceToDepth") {
     RETURN_IF_ERROR(ProcessBlockSizeAttribute(qnn_model_wrapper, node_unit, param_tensor_names));
+
+    Qnn_Scalar_t mode_qnn_scalar = QNN_SCALAR_INIT;
+    mode_qnn_scalar.dataType = QNN_DATATYPE_UINT_32;
+    mode_qnn_scalar.uint32Value = QNN_OP_SPACE_TO_DEPTH_MODE_DCR;
```
qti-shubham, line 432: Please add test cases to check if there is accuracy issue with HTP FP16 SpaceToDepth op with DCR mode. 

For UINT8 and UINT16 activations it works fine but for floating point precision I see accuracy mismatch.

qti-shubham: The CRD mode works fine for HTP but we can't use it as S2D ONNX spec don't support (there is no mode attribute and default behaves in DCR mode)

qti-shubham:
The accuracy issue with SpaceToDepth FP16 kernel in DCR mode seems to be limited with x86 libQnnHtp.so only.

On-device the delta is insignificant (tested on SM8650, SM8450 and SA8295P ADP):
[htp_dcr] max_abs_diff=0.000243 mean_abs_diff=0.000081 max_rel_diff=0.000366

So we will just keep the test disabled and map the ONNX SpaceToDepth op to QNN SpaceToDepth with DCR mode.

qti-shubham:
Changes in this PR looks good to me.

qti-shubham (approved):

qti-yuduo (approved):
LGTM

onnxruntime/test/providers/qnn/simple_op_test.cc
```diff
@@ -875,7 +875,23 @@ TEST_F(QnnHTPBackendTests, DepthToSpaceOp_DCR) {
                         ExpectedEPNodeAssignment::All);
 }
 
-// Test QDQ SpaceToDepth.
+// Test float32 SpaceToDepth on HTP.
+// TODO: Disabling it due to known issues
```
qti-yuduo, line 879: would be good if you can comment on details of the issue

PR #159, opened 2026-03-18
Bump ORT core to 1.24.4

qti-mattsinc:
### Description
* ORT 1.24.4 includes a fix for an access violation when the EP queries producer/consumer info on a missing optional input/output: https://github.com/microsoft/onnxruntime/pull/27644.
* Bumping to 1.24.4 in anticipation of unit tests that will require that fix.


cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
qti-mattsinc, line 1: I had to manually modify this patch file. The existing patch on tip *should* be able to apply cleanly on top of v1.24.1, after which I would have been able to rebase on v1.24.4, resolve the merge conflicts, and have git generate a new patch file, but I could not get `git am`, `git apply`, or `patch` to cooperate.

qti-mattsinc:
note to self: might want to update this table to include 1.24.4

https://github.com/onnxruntime/onnxruntime-qnn/blob/5789ddc5154a6cfefed9257bf37412fef4380d89/docs/execution_providers/QNN-ExecutionProvider.md?plain=1#L37-L40

qti-mattsinc:
Closing. Will be handled by #194.

PR #160, opened 2026-03-19
Restore dynamic-shape fallback using TensorTypeAndShape_HasShape

yuhuchua-qti:
### Description
This PR uses ort_api.TensorTypeAndShape_HasShape to first verify whether a valid input shape actually exists before attempting to retrieve it in ort_api.cc. This ensures that operators receiving dynamic shapes correctly fall back to the CPU EP when needed, preventing the issue where downstream tensor shapes are incorrectly inferred as [1]. The behavior now aligns with the intended and previously existing handling before the ABI cleanup.

### Motivation and Context
In the ABI flow, a previous cleanup removed the node_arg parameter from QnnModelWrapper::GetOnnxShape(), since node_arg is an internal ORT variable. This cleanup also eliminated the associated nullptr check on node_arg.Shape().
However, removing this check inadvertently caused certain dynamic‑shape ops to be assigned to the QNN EP when they should have fallen back to CPU. This resulted in invalid shape propagation —often filling shapes as [1]—and ultimately triggered execution errors.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
@@ -679,14 +679,19 @@ bool QnnModelWrapper::ComposeQnnGraph(bool build_json_qnn_graph) {
   return true;
 }
 
-bool QnnModelWrapper::GetOnnxShape(const std::vector<int64_t>& onnx_shape, std::vector<uint32_t>& shape) {
+bool QnnModelWrapper::GetOnnxShape(const std::optional<std::vector<int64_t>>& onnx_shape, std::vector<uint32_t>& shape) {
+  // Don't support dynamic shape.
+  if (onnx_shape == std::nullopt) {
+    return false;
+  }
+
   // Set shape to 1 for scalar.
-  if (onnx_shape.size() < 1) {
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -679,14 +679,19 @@ bool QnnModelWrapper::ComposeQnnGraph(bool build_json_qnn_graph) {
   return true;
 }
 
-bool QnnModelWrapper::GetOnnxShape(const std::vector<int64_t>& onnx_shape, std::vector<uint32_t>& shape) {
+bool QnnModelWrapper::GetOnnxShape(const std::optional<std::vector<int64_t>>& onnx_shape, std::vector<uint32_t>& shape) {
+  // Don't support dynamic shape.
+  if (onnx_shape == std::nullopt) {
```
```suggestion
  if (!onnx_shape.has_value()) {
```

yuhuchua-qti: I've applied this suggestion. Thanks.

minfhong-qti (approved):

qti-hungjuiw (approved):

qti-chuteng (approved):

PR #161, opened 2026-03-19
Fix content in nuget package

kuanyul-qti:
### Description
Fix content in nuget package.

### Motivation and Context
Fix content in nuget package.

### DIFF
--- a/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
+++ b/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
@@ -123,6 +123,10 @@ env:
   ORT_VERSION_SUFFIX: "${{ inputs.version_suffix != '' && inputs.version_suffix || '' }}"
   BUILD_ARTIFACTORY_REPO: "${{ secrets.BUILD_ARTIFACTORY_REPO }}"
 
+  BUILD_SOURCEBRANCH: "${{ github.ref }}"
+  BUILD_SOURCEVERSION: "${{ github.sha }}"
+  BUILD_BUILDID: "${{ github.run_id }}"
+
 jobs:
   build:
     name: "build-${{inputs.target_os}}-${{inputs.target_arch}}-py${{ inputs.target_py_vsn != 'None' && inputs.target_py_vsn || 'none'}}"
--- a/csharp/src/Qualcomm.ML.OnnxRuntime.QNN/Qualcomm.ML.OnnxRuntime.QNN.csproj
+++ b/csharp/src/Qualcomm.ML.OnnxRuntime.QNN/Qualcomm.ML.OnnxRuntime.QNN.csproj
@@ -4,6 +4,7 @@
     <LangVersion>latest</LangVersion>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageId>Qualcomm.ML.OnnxRuntime.QNN</PackageId>
+    <Version>$([System.IO.File]::ReadAllText('$(MSBuildThisFileDirectory)..\..\..\VERSION_NUMBER').Trim())</Version>
     <Description>Helper utilities for QNN Execution Provider - provides GetLibraryPath(), GetEpNames() and GetEpName() functions</Description>
     <Authors>Qualcomm</Authors>
     <Company>Qualcomm</Company>
--- a/qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
+++ b/qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
@@ -34,6 +34,38 @@ function Set-NuGetCredentials {
         $source_name = "testnuget.org"
     } elseif ($server -eq "nuget") {
         $source_name = "nuget.org"
+
+        # qnn_ep_uplevel.py (called later) runs "nuget sources Add" with credentials.
+        # If nuget.org already exists under any name, the Add command fails with a duplicate URL error.
+        # Remove it here so it can be re-added with credentials.
+        # Get a list of all NuGet sources
+        $nugetSources = nuget sources list
+        if ($LASTEXITCODE -ne 0) {
+            throw "Failed to get the list of NuGet sources"
+        }
+
+        # Parse sources
+        for ($i = 0; $i -lt $nugetSources.Count; $i++) {
+            $line = $nugetSources[$i].Trim()
+            # Match lines that start with a number (source name lines)
+            if ($line -match '^\d+\.\s+(.+?)\s+\[') {
+                $existedSourceName = $matches[1]
+                # The next line should contain the URL/path
+                if ($i + 1 -lt $nugetSources.Count) {
+                    $url = $nugetSources[$i + 1].Trim()
+                    if ($url -eq "https://api.nuget.org/v3/index.json") {
+                        # Remove the nuget.org source; it will be added back later in the script.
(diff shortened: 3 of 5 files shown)


qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
```diff
@@ -34,6 +34,29 @@ function Set-NuGetCredentials {
         $source_name = "testnuget.org"
     } elseif ($server -eq "nuget") {
         $source_name = "nuget.org"
+
+        # Get a list of all NuGet sources
+        $nugetSources = nuget sources list
```
qti-kromero, line 42: what happens if this or something below fails?

csharp/src/Qualcomm.ML.OnnxRuntime.QNN/Qualcomm.ML.OnnxRuntime.QNN.csproj
```diff
@@ -1,9 +1,12 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <LangVersion>latest</LangVersion>
```
qti-kromero, line 4: are we intentionally removing this LangVersion being set to latest or was this a mistake in setting the version below? this will potentially change which C# compiler the package is compiled with? unsure what history is here

qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
```diff
@@ -34,6 +34,29 @@ function Set-NuGetCredentials {
...
+        # Parse sources
+        for ($i = 0; $i -lt $nugetSources.Count; $i++) {
+            $line = $nugetSources[$i].Trim()
+            # Match lines that start with a number (source name lines)
+            if ($line -match '^\d+\.\s+(.+?)\s+\[') {
+                $existedSourceName = $matches[1]
+                # The next line should contain the URL/path
+                if ($i + 1 -lt $nugetSources.Count) {
+                    $url = $nugetSources[$i + 1].Trim()
+                    if ($url -eq "https://api.nuget.org/v3/index.json") {
```
qti-kromero, line 56: not really following why this specific index is being removed - we need this specific URL to be removed and then are adding back a different URL for nuget.org source later?

qti-kromero:
we should look to add a page for release notes to the nuget listing for this release as well if possible, currently 2.0.0 is empty: 
https://www.nuget.org/packages/Qualcomm.ML.OnnxRuntime.QNN#releasenotes-body-tab

qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
kuanyul-qti: The task of uploading to Artifactory/external server will be terminated.

qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
kuanyul-qti: We add the url to the source in the qnn_ep_uplevel.py script (called by this file). If the url has been existed in the source, the command of "nuget souces Add" will fail. Since we handle the source name and add credentials below, would like to remove the existing source in the first stage.

tools/nuget/generate_nuspec_for_native_nuget.py
kuanyul-qti, line 273: Need to update release notes

kuanyul-qti:
> we should look to add a page for release notes to the nuget listing for this release as well if possible, currently 2.0.0 is empty: https://www.nuget.org/packages/Qualcomm.ML.OnnxRuntime.QNN#releasenotes-body-tab

Leave a comment in https://github.com/onnxruntime/onnxruntime-qnn/pull/161#discussion_r2963490594
That is the source of the release notes in the NuGet package.

qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
qti-jkilpatrick: Kyle can correct me if this is wrong, but I suspect his concern is that PowerShell doesn't typically abort on failure. There is no equivalent for `set -euo pipefail`, which is why we defined `Assert-Success` in `qcom/scripts/windows/utils.ps1`.

csharp/src/Qualcomm.ML.OnnxRuntime.QNN/Qualcomm.ML.OnnxRuntime.QNN.csproj
kuanyul-qti: Revert the modification.

qcom/scripts/upleveling/qnn_ep_uplevel_nuget.ps1
kuanyul-qti: Add `if ($LASTEXITCODE -ne 0)` condition below to handle the case that the command is failed.

tools/nuget/generate_nuspec_for_native_nuget.py
kuanyul-qti: Done. Also set environment variables in .github/workflows/qualcomm-internal-build-and-test-single-os.yml

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -120,6 +120,10 @@ env:
 
   ORT_VERSION_SUFFIX: "${{ inputs.version_suffix != '' && inputs.version_suffix || '' }}"
 
+  BUILD_SOURCEBRANCH: "${{ github.ref }}"
+  BUILD_SOURCEVERSION: "${{ github.sha }}"
+  BUILD_BUILDID: "${{ github.run_id }}"
```
qti-jkilpatrick, line 128: How are these used? I see possibly relevant references in `tools/nuget/generate_nuspec_for_native_nuget.py`. Any reason we can't use the [GitHub equivalents](https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables) rather than re-creating an Azure-like environment?

kuanyul-qti: Yes. It's used in tools/nuget/generate_nuspec_for_native_nuget.py.
If we use `GITHUB_REF`, `GITHUB_SHA`, and `GITHUB_RUN_ID`, it has error parsing called workflow with: `(Line: 123, Col: 23): Unrecognized named-value: 'GITHUB_REF'. Located at position 1 within expression: GITHUB_REF, (Line: 124, Col: 24): Unrecognized named-value: 'GITHUB_SHA'. Located at position 1 within expression: GITHUB_SHA, (Line: 125, Col: 18): Unrecognized named-value: 'GITHUB_RUN_ID'. Located at position 1 within expression: GITHUB_RUN_ID`.
The issue is that GITHUB_REF, GITHUB_SHA, and GITHUB_RUN_ID are being used as bare variable names, but in GitHub Actions expressions, these need to be accessed through the github context object.

1. One solution is to use `env.GITHUB_REF`, but since we are at `env` part, we can't reference environment variables that are defined in the same env block.
2. Another solution is to define GITHUB_REF, GITHUB_SHA, and GITHUB_RUN_ID first, and then reference them. But this still needs using `github.xxx` to define them.

So, I will keep it as is to achieve the simplest usage.

qti-chuteng (approved):

PR #163, opened 2026-03-23
[QnnBackendManager] Fix logger usage in shared-context scenario.

minfhong-qti:
### Description

Update QnnBackendManager to cache logger as a pointer instead of a constant reference to allow updating. Revise all logging usage accordingly. Note that any class below QnnBackendManager should not cache logger anymore but pass the logger in each function call.
Affected classes
- `QnnContextMemHandleManager`
- `HtpPowerConfigManager`

### Motivation and Context

In shared-context scenario, QnnBackendManager is shared accross several ORT sessions. Since the cached logger in QnnBackendManager is a session logger, it is destroyed along with the session desctruction. That is, the logger cannot be used in the following sessions and must be updated to a new session logger.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -161,12 +161,12 @@ Ort::Status ReadBinaryFromFile(const std::string& file_path, uint8_t* buffer, si
 }
 
 Ort::Status QnnBackendManager::ParseLoraConfig(std::string lora_config_path) {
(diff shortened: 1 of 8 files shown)


qti-hungjuiw (approved):

qti-hungjuiw:
Per offline discussion 
- Another choice can be also adding `ResetLogger` into `QnnContextMemHandleManager` and `HtpPowerConfigManager` so that we can still cache the logger and reset by needed. However, this might not scalable if more XXXManager are added in the future.

huaychou (approved):
LGTM :shipit:

qti-chuteng (approved):

PR #23, opened 2026-02-02
Download ORT Core as ExternalProject and Reconstruct Qnn Test Framework

qti-hungjuiw:
### Description

- This PR introduces the **ORT Core source code** from Microsoft as a downloaded dependency via `deps.txt`.
- The PR adds an `ExternalProject` to the build system to compile the Microsoft **ORT Core source code**.
    - As part of this decoupling, the end-to-end test binary `onnxruntime_plugin_ep_onnx_test.exe` now depends on the Microsoft codebase.
- This change reconstructs the QNN Test Framework by eliminating the use of ORT internal/private libraries such as `onnxruntime_graph` and `onnxruntime_common`.
    - To create test vectors without using the ORT internal library `onnxruntime_graph`, the PR directly leverages third-party **ONNX** dependency.

### Build System Dependency Graph (Before)
```mermaid
graph LR
    onnxruntime_unittest_utils --> onnxruntime_graph
    onnxruntime_unittest_utils --> onnxruntime_session
    onnxruntime_unittest_utils --> onnxruntime_test_utils 
    onnxruntime_unittest_utils --> onnxruntime_framework 
    onnxruntime_unittest_utils --> onnxruntime_providers 
    onnxruntime_unittest_utils --> onnxruntime_optimizer 
    onnxruntime_unittest_utils --> onnxruntime_common

    onnxruntime_session --> onnxruntime_lora
    onnxruntime_session --> onnxruntime_flatbuffers
    onnxruntime_session --> onnxruntime_graph 
    onnxruntime_session --> onnxruntime_test_utils 
    onnxruntime_session --> onnxruntime_framework 
    onnxruntime_session --> onnxruntime_providers 
    onnxruntime_session --> onnxruntime_optimizer 
    onnxruntime_session --> onnxruntime_util 
    onnxruntime_session --> onnxruntime_common 
    onnxruntime_session --> onnxruntime_mlas

    onnxruntime_graph --> onnxruntime_flatbuffers
    onnxruntime_graph --> onnxruntime_framework 
    onnxruntime_graph --> onnxruntime_common 
    onnxruntime_graph --> onnxruntime_mlas

    onnxruntime_optimizer --> onnxruntime_graph
    onnxruntime_optimizer --> onnxruntime_framework 
    onnxruntime_optimizer --> onnxruntime_providers 
    onnxruntime_optimizer --> onnxruntime_util 
    onnxruntime_optimizer --> onnxruntime_common 
    onnxruntime_optimizer --> onnxruntime_mlas

[...]

PR #164, opened 2026-03-23
Re-enable matmulnbits_test.cc

qti-hungjuiw:
### Description

- The change modifies the `matmulnbits_test.cc` with public API and re-enables it.

### Motivation and Context

- To avoid relying on private ORT Core headers/libraries, #23 reconstructed the QNN Test Framework using only ORT’s public headers and libraries.
- The newly added op test cases required several adjustments to work with the updated QNN test infrastructure. To keep the scope of #23 manageable, several tests were temporarily disabled and are now being re‑enabled with the necessary updates.


onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -1719,16 +1715,244 @@ bool ReduceOpHasAxesInput(const std::string& op_type, int opset_version);
...
-  } else                                      \
-    static_assert(true, "")
+  } while (0)
 #else
 #define QNN_SKIP_TEST_ON_LINUX_X86_64(reason) \
   do {                                        \
   } while (0)
 #endif
 
+/**
```
minfhong-qti: Better to be placed in `unittest_util/qdq_test_utils.h`?

qti-hungjuiw: Sure, thanks!

minfhong-qti (approved):

qti-chuteng (approved):

PR #167, opened 2026-03-24
[QNN EP] Add LayerNorm Fusion node group pattern

qti-ashwshan:
### Description

- Add a QNN EP node group fusion that detects decomposed LayerNorm patterns in ONNX graphs and fuses them into a single QNN_OP_LAYER_NORM operator.


qti-yuduo (commented):
Adding unit test?

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,600 @@
...
+    OrtNodeAttrHelper rm2_helper(*reduce_mean2_node_unit);
+    if (rm2_helper.Get("keepdims", static_cast<int64_t>(1)) != 1) {
+      return nullptr;
+    }
+  }
+
+  // Both ReduceMeans must have the same axes, pointing to the last dimension.
+  auto axes1_opt = GetReduceMeanAxes(qnn_model_wrapper, reduce_mean_node_unit);
+  auto axes2_opt = GetReduceMeanAxes(qnn_model_wrapper, *reduce_mean2_node_unit);
+  if (!axes1_opt.has_value() || !axes2_opt.has_value() || axes1_opt.value() != axes2_opt.value()) {
```
qti-yuduo: Can `axes` be `{1, 2}` vs. `{2, 1}`? They are still valid pattern so maybe normalize before comparing. E.g., handling negative, sort, dedupe, etc.

qti-yuduo: same concern for other places accessing axes, e.g., ValidateAndSqueezeScaleShape

onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc
```diff
@@ -132,6 +134,7 @@ static std::unique_ptr<IQnnNodeGroup> TryQnnFusions(
       starting_node_unit.OpType() != "Gather" &&
       starting_node_unit.OpType() != "MatMul" &&
       starting_node_unit.OpType() != "Erf" &&
+      starting_node_unit.OpType() != "ReduceMean" &&
```
minfhong-qti: I don't think this case needs adding here.

onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc
```diff
@@ -16,6 +16,7 @@
 #include "core/providers/qnn/builder/qnn_node_group/dq_q_fusion.h"
 #include "core/providers/qnn/builder/qnn_node_group/gather_transpose_reshape_fusion.h"
 #include "core/providers/qnn/builder/qnn_node_group/gelu_fusion.h"
+#include "core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.h"
```
minfhong-qti: Sort alpabetacally.

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,600 @@
...
+  if (!qmw.IsConstantInput(input_name)) {
+    return std::nullopt;
+  }
+
+  const OrtValueInfo* value_info = qmw.GetConstantTensor(input_name);
+  if (!value_info) {
+    return std::nullopt;
+  }
+
+  const OrtValue* value = nullptr;
```
minfhong-qti: Use CXX API which is cleaner. https://github.com/onnxruntime/onnxruntime-qnn/blob/main/include/onnxruntime/core/session/onnxruntime_cxx_api.h

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,600 @@
...
+    return std::nullopt;
+  }
+
+  const void* raw_data = nullptr;
+  RETURN_DEFAULT_IF_API_FAIL(ort_api.GetTensorData(value, &raw_data), ort_api, std::nullopt);
+
+  return *static_cast<const float*>(raw_data);
+}
+
+static std::optional<std::vector<uint32_t>> GetReduceMeanAxes(const QnnModelWrapper& qmw,
```
minfhong-qti: Is it possible to somehow share the logic with op builder `ReduceOpBuilder::GetAxesSet`?

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,600 @@
...
+
+// Validates that a gamma/beta shape is compatible with LayerNorm over the given axes,
+// then squeezes it to 1D. Rules:
+//   - All non-normalized dims must be 1 (no scaling along non-normalized axes).
+//   - All normalized dims must match the input shape at those axes.
+// Returns std::nullopt if the shape is invalid.
+static std::optional<std::vector<uint32_t>> ValidateAndSqueezeScaleShape(
+    const std::vector<uint32_t>& scale_shape,
+    const std::vector<uint32_t>& input_shape,
+    gsl::span<const uint32_t> axes) {
```
minfhong-qti:
```suggestion
static Ort::Status ValidateAndSqueezeScaleShape(
    const std::vector<uint32_t>& scale_shape,
    const std::vector<uint32_t>& input_shape,
    gsl::span<const uint32_t> axes,
    /*out*/std::vector<uint32_t> squeezed_shape) {
```
Prefer to define like this.

minfhong-qti (commented):
A general question. If there are more and more LayerNorm pattern, any plan for the fusion to be scalable?

qti-ashwshan:
> A general question. If there are more and more LayerNorm pattern, any plan for the fusion to be scalable?

Yes... Since these patterns have the root node as ReduceMean for 90% of the patterns. We can easily scale it using this single fusion node group and add more patterns to it.

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,570 @@
...
+  const std::array<std::string_view, 1> mul_types{"Mul"};
+  const OrtNodeUnit* mul_gamma_node_unit = GetOnlyChildOfType(qnn_model_wrapper, *div_node_unit,
+                                                              mul_types, node_to_node_unit,
+                                                              node_unit_to_qnn_node_group);
+  if (mul_gamma_node_unit == nullptr) {
+    return nullptr;
+  }
+
+  // Find gamma: the constant input to Mul.
+  const auto& mul_inputs = mul_gamma_node_unit->Inputs();
```
quic-muchhsu, line 274: Scale is somehow optional for layernorm, we can just fill scale=1 when mul is not found.
bias is optional for layernorm.
You might want to include these 2 patterns.

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,570 @@
...
+  if (axes1 != axes2) {
+    return nullptr;
+  }
+  std::vector<uint32_t> input_shape;
+  if (!qnn_model_wrapper.GetOnnxShape(rm1_inputs[0].shape, input_shape)) {
+    return nullptr;
+  }
+  const size_t input_rank = input_shape.size();
+
+  // Axes must point to the last dimension
```
quic-muchhsu: I think we should do normalizing the axes -> compare the axes.
Because `ReduceMean(n:K-1)` on a K dim tensor can be fused to `Layernorm(n)`.
The check should focus on if the axes in reduce mean is (1) continuous and (2) includes the last dim.

quic-muchhsu (approved):

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
quic-muchhsu: It should be a comment instead of an approve. Please wait for others before merge.

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,570 @@
...
+  if (!data) {
+    return std::nullopt;
+  }
+
+  return *data;
+}
+
+static std::optional<std::vector<uint32_t>> GetReduceMeanAxes(const QnnModelWrapper& qmw,
+                                                              const OrtNodeUnit& reduce_mean_node_unit) {
+  return GetReduceAxes(qmw, reduce_mean_node_unit);
```
qti-yuduo: Now this is shared, this `GetReduceMeanAxes` provides no abstraction. We can just remove this indirect?

onnxruntime/core/providers/qnn/builder/qnn_node_group/utils.cc
```diff
@@ -14,6 +16,58 @@
...
+    const auto* axes_tensor = qmw.GetConstantTensor(axes_input_name);
+    if (!axes_tensor) {
+      return std::nullopt;
+    }
+    std::vector<uint8_t> axes_bytes;
+    if (!qmw.UnpackInitializerData(axes_tensor, axes_bytes).IsOK()) {
+      return std::nullopt;
+    }
+    raw_axes.resize(axes_bytes.size() / sizeof(int64_t));
+    std::memcpy(raw_axes.data(), axes_bytes.data(), axes_bytes.size());
```
qti-yuduo, line 52: Is this safe? Can axes be non 64bit? 

Better to Check alignment before proceeding
if (axes_bytes.size() % sizeof(int64_t) != 0) {
  return std::nullopt;
}

qti-yuduo: Maybe not needed, needs to double check all reduce ops https://onnx.ai/onnx/operators/

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,570 @@
...
+static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qmw,
+                                         gsl::span<const OrtNodeUnit* const> node_units,
+                                         const OrtNodeUnitIODef& root_input,
+                                         const OrtNodeUnitIODef& gamma_input,
+                                         const OrtNodeUnitIODef& beta_input,
+                                         const OrtNodeUnitIODef& final_output,
+                                         float epsilon,
+                                         gsl::span<const uint32_t> axes,
+                                         bool validate) {
+  assert(!node_units.empty());
```
qti-yuduo: why we need to assert? (also this won't happen on release build -- so if really need to check use ORT_ENFORCE)

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,570 @@
...
+    return nullptr;
+  }
+
+  // Normalize axes for comparison: sort and deduplicate.
+  auto axes1 = axes1_opt.value();
+  auto axes2 = axes2_opt.value();
+  std::sort(axes1.begin(), axes1.end());
+  std::sort(axes2.begin(), axes2.end());
+  axes1.erase(std::unique(axes1.begin(), axes1.end()), axes1.end());
+  axes2.erase(std::unique(axes2.begin(), axes2.end()), axes2.end());
```
qti-yuduo: Redundant? `GetReduceMeanAxes` already did normalizations?

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
qti-ashwshan: Yes!
I am removing that ..

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -65,7 +65,6 @@ static std::optional<float> GetConstantFloatScalar(const QnnModelWrapper& qmw,
   return *data;
 }
 
-
```
qti-yuduo: FYI, do `pre-commit install`, then this would auto apply. so this won't happen

qti-yuduo: @qti-ashwshan

qti-ashwshan: Sure

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,568 @@
...
+  if (!ort_value_info.GetInitializer(ort_value).IsOK()) {
+    return std::nullopt;
+  }
+
+  auto type_info = ort_value_info.TypeInfo();
+  auto tensor_info = type_info.GetTensorTypeAndShapeInfo();
+  if (tensor_info.GetElementType() != ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT) {
+    return std::nullopt;
+  }
+
```
qti-chuteng, line 59: Check if the tensor is the scalar.
----
if (tensor_info.GetElementCount() != 1) {
  return std::nullopt;
}

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -0,0 +1,572 @@
...
+static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qmw,
+                                         gsl::span<const OrtNodeUnit* const> node_units,
+                                         const OrtNodeUnitIODef& root_input,
+                                         const OrtNodeUnitIODef& gamma_input,
+                                         const OrtNodeUnitIODef& beta_input,
+                                         const OrtNodeUnitIODef& final_output,
+                                         float epsilon,
+                                         gsl::span<const uint32_t> axes,
+                                         bool validate) {
+  const std::string node_name = utils::GetUniqueName(*node_units[0]);
```
qti-yuduo: Sorry you'll need a follow up change: see https://github.com/onnxruntime/onnxruntime-qnn/pull/184

qti-yuduo (approved):

qti-chuteng (approved):

PR #168, opened 2026-03-24
Use Artifactory for CI artifacts

qti-jkilpatrick:
### Description

Re-enable building on x64 Windows runners whenever possible, making use of Artifactory to host temporary artifacts.

* Introduce `upload-ci-artifact` and `download-ci-artifact` actions.
* Use new actions in place of GitHub's artifact actions in most places.
* Add a scheduled job to prune after a day.

### Motivation and Context

When this repo moved from internally-hosted GitHub to public GitHub, our self-hosted runners started randomly suffering very long artifact download times -- fetching a Windows test archive could take half an hour instead of a few seconds. Our analysis concludes this is the result of network security measures outside of our control. To mitigate, we avoided temporary artifacts and started building and testing on the same machine whenever possible. This caused a huge increase in pressure on our limited ARM64 Windows runners, motivating this change.

### DIFF
--- a/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
+++ b/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
@@ -118,7 +118,10 @@ env:
   # Cross compilation on Windows adds a folder to the build directory.
   BUILD_CONFIG_DIR: "${{ (inputs.target_os == 'windows' && ((inputs.build_runner_arch == 'X64' && inputs.target_arch != 'x86_64') || (inputs.build_runner_arch == 'ARM64' && inputs.target_arch != 'arm64'))) && 'Release/Release' || 'Release' }}"
 
+  TEST_ARCHIVE_EXT: "${{ inputs.target_os == 'linux' && 'tar.bz2' || 'zip' }}"
+
   ORT_VERSION_SUFFIX: "${{ inputs.version_suffix != '' && inputs.version_suffix || '' }}"
+  BUILD_ARTIFACTORY_REPO: "${{ secrets.BUILD_ARTIFACTORY_REPO }}"
 
 jobs:
   build:
@@ -157,21 +160,28 @@ jobs:
           report_paths: |
             ${{ github.workspace }}/build/${{inputs.target_os}}-${{inputs.target_arch}}/${{ env.BUILD_CONFIG_DIR }}/*.results.xml
 
-      - uses: actions/upload-artifact@v4
-        if: ${{ inputs.upload_test_archive }}
+      - name: Setup JFrog CLI
+        uses: jfrog/setup-jfrog-cli@v4
         with:
-          name: ORT ${{inputs.target_os}} ${{inputs.target_arch}} py${{ inputs.target_py_vsn != 'None' && inputs.target_py_vsn || 'none'}} Tests
-          retention-days: 7
-          path: |
-            ${{ github.workspace }}/build/onnxruntime-tests-${{inputs.target_os}}-${{ inputs.target_arch }}.*
+          disable-auto-build-publish: true
+          disable-job-summary: true
+        env:
+          JF_URL: ${{ secrets.BUILD_ARTIFACTORY_URL }}
+          JF_ACCESS_TOKEN: ${{ secrets.BUILD_ARTIFACTORY_TOKEN }}
 
-      - uses: actions/upload-artifact@v4
-        if: ${{ inputs.upload_wheel }}
+      - name: Upload Test Archive to Artifactory
+        if: ${{ inputs.upload_test_archive }}
+        uses: ./.github/actions/upload-ci-artifact
         with:
-          name: ORT ${{inputs.target_os}} ${{ inputs.target_arch }} Python ${{ inputs.target_py_vsn }} Wheel
-          retention-days: 7
-          path: |
-            ${{ github.workspace }}/build/${{inputs.target_os}}-${{ inputs.target_arch }}/${{ env.BUILD_CONFIG_DIR }}/dist/onnxruntime_qnn*.whl
+          name: "${{ inputs.target_os }}-${{ inputs.target_arch }}-py${{inputs.target_py_vsn}}-tests"
+          src_pattern: "onnxruntime-tests-${{inputs.target_os}}-${{ inputs.target_arch }}.${{ env.TEST_ARCHIVE_EXT }}"
+
+      - name: Upload Wheel to Artifactory
+        if: ${{ inputs.upload_wheel }}
+        uses: ./.github/actions/upload-ci-artifact
+        with:
+          name: "${{ inputs.target_os }}-${{ inputs.target_arch }}-py${{inputs.target_py_vsn}}-wheels"
+          src_pattern: "${{ inputs.target_os }}-${{ inputs.target_arch }}/${{ env.BUILD_CONFIG_DIR }}/dist/onnxruntime_qnn*.whl"
 
       - uses: actions/upload-artifact@v4
         if: ${{ inputs.upload_nuget }}
@@ -202,11 +212,19 @@ jobs:
         with:
           submodules: recursive
 
-      - uses: actions/download-artifact@v4
+      - name: Setup JFrog CLI
+        uses: jfrog/setup-jfrog-cli@v4
         with:
-          name: ORT ${{inputs.target_os}} ${{inputs.target_arch}} py${{ inputs.target_py_vsn != 'None' && inputs.target_py_vsn || 'none'}} Tests
-          path: |
-            ${{ github.workspace }}/build
+          disable-auto-build-publish: true
+          disable-job-summary: true
+        env:
+          JF_URL: ${{ secrets.BUILD_ARTIFACTORY_URL }}
+          JF_ACCESS_TOKEN: ${{ secrets.BUILD_ARTIFACTORY_TOKEN }}
+
+      - name: Download Tests from Artifactory
+        uses: ./.github/actions/download-ci-artifact
+        with:
+          name: "${{ inputs.target_os }}-${{ inputs.target_arch }}-py${{inputs.target_py_vsn}}-tests"
 
       - name: Test ${{inputs.target_os}} ${{inputs.target_arch}} py${{ inputs.target_py_vsn != 'None' && inputs.target_py_vsn || 'none'}}
         run: >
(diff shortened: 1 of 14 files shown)


.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -118,7 +118,10 @@ env:
   # Cross compilation on Windows adds a folder to the build directory.
   BUILD_CONFIG_DIR: "${{ (inputs.target_os == 'windows' && ((inputs.build_runner_arch == 'X64' && inputs.target_arch != 'x86_64') || (inputs.build_runner_arch == 'ARM64' && inputs.target_arch != 'arm64'))) && 'Release/Release' || 'Release' }}"
 
+  TEST_ARCHIVE_EXT: "${{ inputs.target_os == 'linux' && 'tar.bz2' || 'zip' }}"
+
   ORT_VERSION_SUFFIX: "${{ inputs.version_suffix != '' && inputs.version_suffix || '' }}"
+  BUILD_ARTIFACTORY_REPO: "${{ secrets.BUILD_ARTIFACTORY_REPO }}"
```
qti-jkilpatrick, line 124: This is the simplest pattern I could get working:
* Set `BUILD_ARTIFACTORY_REPO` in the workflow's environment so that the artifactory scripts know where to find things. I failed to get the actions to read this from secrets, despite the internet suggesting this is possible. Future work.
* Install the JFrog CLI within each job that needs it. I tried to accomplish this with an action, but I couldn't get the environment to propagate to subsequent steps.
* Use `upload-ci-artifact` / `download-ci-artifact`

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -157,21 +160,28 @@ jobs:
...
+        uses: ./.github/actions/upload-ci-artifact
         with:
-          name: ORT ${{inputs.target_os}} ${{ inputs.target_arch }} Python ${{ inputs.target_py_vsn }} Wheel
-          retention-days: 7
-          path: |
-            ${{ github.workspace }}/build/${{inputs.target_os}}-${{ inputs.target_arch }}/${{ env.BUILD_CONFIG_DIR }}/dist/onnxruntime_qnn*.whl
+          name: "${{ inputs.target_os }}-${{ inputs.target_arch }}-py${{inputs.target_py_vsn}}-wheels"
+          src_pattern: "${{ inputs.target_os }}-${{ inputs.target_arch }}/${{ env.BUILD_CONFIG_DIR }}/dist/onnxruntime_qnn*.whl"
 
       - uses: actions/upload-artifact@v4
```
qti-jkilpatrick, line 186: I haven't tackled the newer packages yet. I'm a little less worried about these because they don't block CI, but I'll get to them soon.

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -202,11 +212,19 @@ jobs:
         with:
           submodules: recursive
 
-      - uses: actions/download-artifact@v4
+      - name: Setup JFrog CLI
```
qti-jkilpatrick, line 215: Unfortunately, this has to happen once per *job*, not once per workflow.

.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -26,13 +26,14 @@ on:
       # Cross-compiling from X64 to ARM64 is faster with our runners, but cannot produce a wheel.
       windows_arm64_runner_arch:
         type: string
-        default: "ARM64"
+        default: X64
```
qti-jkilpatrick, line 29: Yay, we can go back to building ARM64 test archives on X64.

.github/workflows/qualcomm-internal-clean-artifacts.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Delete old CI artifacts
+        run: >
+          jf rt delete
+          --spec "${{ github.workspace }}/qcom/scripts/artifactory/aql/expired_ci_artifacts.json"
+          --spec-vars "repo=${{ env.BUILD_ARTIFACTORY_REPO }}"
+
+      # This is a bit of a hack, but there does not appear to be a way to do this nicely without
+      # a server-side plugin. Instead, we just use a CLI-side plugin several times. Eventually
+      # it will converge and do nothing.
```
qti-jkilpatrick, line 40: Even Gemini hallucinated some AQL file properties that might have enabled this, but alas, they do not exist.

.github/workflows/qualcomm-internal-release.yml
```diff
@@ -44,12 +47,25 @@ jobs:
...
+        uses: ./.github/actions/download-ci-artifact
         with:
-          name: ORT linux x86_64 Python 3.10 Wheel
-          path: ${{ github.workspace }}/build/dist
+          name: linux-x86_64-py3.10-wheels
+
+      - name: Consolidate Wheels
+        run: |
+          mkdir "${{ github.workspace }}/build/dist"
+          cp "${{ github.workspace }}/build/linux-x86_64/Release/dist/"*.whl "${{ github.workspace }}/build/dist"
```
qti-jkilpatrick, line 68: We have a new invariant that usually simplifies things, but in this case added a manual step: any file uploaded to artifactory includes full path of that file, relative to the build folder.

qcom/docs/setup_linux_build_vm.sh
```diff
@@ -61,6 +61,8 @@ env_tmpfile=$(mktemp --suffix=-runner-env)
 cat > "${env_tmpfile}" <<EOF
 LANG=en_US.UTF-8
 HOME=${runner_home}
+TMPDIR=${runner_root}/tmp
+JFROG_CLI_HOME_DIR=${runner_root}/jfrog-home
```
qti-jkilpatrick, line 65: * `TMPDIR` should have been here already. I added it to all the runners sometime back and forgot to add it here.
* Our Artifactory server's certificate chain goes back to an internally-issued root so we have to add it manually on Linux. On Windows, the certificate store is managed by the OS, where security tools keep things up to date.
* I also updated the VM/device setup docs on the wiki.

qcom/scripts/all/package_manager.py
```diff
@@ -230,7 +235,8 @@ def install(self) -> None:
             return
         package_path = self.__fetch()
 
-        if package_path.suffix == ".exe":
+        on_win = is_host_windows()
+        if (on_win and package_path.suffix == ".exe") or (not on_win and package_path.suffix == ""):
```
qti-jkilpatrick, line 239: Early on, I thought I was going to have to install the JFrog CLI manually so I added support for downloading bare binaries on non-Windows platforms. It ended up being unnecessary, but this adds parity so I'd like to keep it.

.github/workflows/qualcomm-internal-clean-artifacts.yml
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+name: Qualcomm Internal Clean Artifacts
+description: Periodically impose a retention policy on Artifactory Artifacts
```
qti-jkilpatrick, line 5: As it turns out, retention policies are not baked into Artifactory. There is a server-side plugin we can request that kinda does this, but it's not sufficiently configurable for our purposes.

qti-mbadnara:
@qti-yuduo / @qti-kromero

qti-jkilpatrick:
> qti-yuduo / qti-kromero

These guys, along with Tiru, Jerry, and me are automatically included as reviewers on every PR.

image

@qti-mbadnara

.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -76,12 +77,14 @@ jobs:
...
     secrets: inherit
     with:
-      nightly_build: ${{ inputs.nightly_build }}
-      build_runner_arch: ${{ inputs.windows_arm64_runner_arch }}
       target_arch: arm64
-      upload_wheel: ${{ inputs.windows_arm64_runner_arch == 'ARM64' }}
-      target_py_vsn: ${{ inputs.windows_arm64_runner_arch == 'ARM64' && inputs.target_py_vsn || 'None' }}
+      build_runner_arch: ${{ inputs.windows_arm64_runner_arch }}
       test_runner_arch: arm64
+      upload_wheel: ${{ inputs.windows_arm64_runner_arch == 'ARM64' && inputs.target_py_vsn }}
```
qti-kromero, line 83: is there a reason for this reordering? I can't seem to find the pattern

.github/workflows/qualcomm-internal-clean-artifacts.yml
```diff
@@ -0,0 +1,52 @@
...
+          --spec "${{ github.workspace }}/qcom/scripts/artifactory/aql/expired_ci_artifacts.json"
+          --spec-vars "repo=${{ env.BUILD_ARTIFACTORY_REPO }}"
+
+      # This is a bit of a hack, but there does not appear to be a way to do this nicely without
+      # a server-side plugin. Instead, we just use a CLI-side plugin several times. Eventually
+      # it will converge and do nothing.
+      - name: Prune empty folders
+        run: |
+          jf plugin install rm-empty
+          jf rm-empty folders ${{ env.BUILD_ARTIFACTORY_REPO }}/ci/ --quiet
```
qti-kromero, line 44: is calling this like 10 times for redundancy or?

.github/workflows/qualcomm-internal-build-and-test.yml
qti-jkilpatrick: It's hard to find natural groupings with this set of parameters, but it's essentially

1. architecture options
2. which artifacts to save
3. build options

TBH, I was mostly reacting to relatively unimportant options like `nightly_build` being so high in the list and the architectures being separated. I'm open to other ideas.

.github/workflows/qualcomm-internal-clean-artifacts.yml
qti-jkilpatrick: Check out the comment on lines 38-40.

qti-jkilpatrick: Oh, but what happens is that it deletes the leaf empty folders, creating more empty folders that will then be the next set of leaves to delete, etc.

qti-kromero (approved):

PR #173, opened 2026-03-25
Reset unique name generator between compile

qti-yuduo:
## Changes:

- Static counter in GetUniqueName leaked across ComposeGraph calls in
the same process;
- Refactor into UniqueNameGeneratorImpl class and reset at ComposeGraph entry.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_utils.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_utils.h
@@ -127,50 +127,16 @@ Ort::Status GetQnnDataType(const bool is_quantized_tensor,
 // (e.g., "_2") when the same base + suffix combination is requested more than once.
 class UniqueNameGeneratorImpl {
  public:
-  void Reset() {
-    std::lock_guard<std::mutex> lock(mutex_);
-    counter_.clear();
-  }
-
-  std::string New(std::string_view base, std::string_view suffix = {}) {
-    std::string name(base);
-    if (!suffix.empty()) {
-      name.append(suffix);
-    }
-    {
-      std::lock_guard<std::mutex> lock(mutex_);
-      int& count = counter_[name];
-      if (count++ > 0) {
-        name.append("_").append(std::to_string(count));
-      }
-    }
-    return name;
-  }
-
-  std::string New(const OrtNodeUnit& node_unit, std::string_view suffix = {}) {
-    const std::string base = node_unit.Name();
-    if (base.empty()) {
-      return New(node_unit.OpType() + std::to_string(node_unit.Index()), suffix);
-    }
-    return New(base, suffix);
-  }
+  void Reset();
+  std::string New(std::string_view base, std::string_view suffix = {});
+  std::string New(const OrtNodeUnit& node_unit, std::string_view suffix = {});
 
  private:
   std::unordered_map<std::string, int> counter_;
   std::mutex mutex_;
 };
 
-inline UniqueNameGeneratorImpl& UniqueNameGenerator() {
-  static UniqueNameGeneratorImpl instance;
-  return instance;
-}
-
-inline std::string GetUniqueName(std::string_view base, std::string_view suffix = {}) {
-  return UniqueNameGenerator().New(base, suffix);
-}
-inline std::string GetUniqueName(const OrtNodeUnit& node_unit, std::string_view suffix = {}) {
-  return UniqueNameGenerator().New(node_unit, suffix);
-}
+UniqueNameGeneratorImpl& UniqueNameGenerator();
 
 bool OnnxDataTypeToQnnDataType(const ONNXTensorElementDataType onnx_data_type,
                                Qnn_DataType_t& qnn_data_type,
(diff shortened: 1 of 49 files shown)


onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -2357,6 +2358,51 @@ TEST_F(QnnCPUBackendTests, GraphInputOutputOrderMatchesOnnxOffloadGraphIoQuantiz
...
+
+    ProviderOptions provider_options;
+    provider_options["backend_type"] = "cpu";
+    provider_options["offload_graph_io_quantization"] = "0";
+    provider_options["dump_json_qnn_graph"] = "1";
+    provider_options["json_qnn_graph_dir"] = json_dir.string();
+
+    RunQnnModelTest(model_fn, provider_options, 13, ExpectedEPNodeAssignment::All);
+
+    std::set<std::string> node_names;
```
qti-hungjuiw: Could we use **std::unordered_set**?
```suggestion
    std::unordered_set<std::string> node_names;
```

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -120,11 +122,54 @@ Ort::Status GetQnnDataType(const bool is_quantized_tensor,
...
+    std::string name(base);
+    if (!suffix.empty()) {
+      name.append(suffix);
+    }
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
+      int& count = counter_[name];
+      if (count++ > 0) {
+        name.append("_").append(std::to_string(count));
+      }
```
qti-hungjuiw: The flow appears to generate `name`, `name_2`, `name_3`, and so on.
Just confirmed, is skipping the "_1" expected behavior?

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -120,11 +122,54 @@ Ort::Status GetQnnDataType(const bool is_quantized_tensor,
...
+  static UniqueNameGeneratorImpl instance;
+  return instance;
+}
+
+inline std::string GetUniqueName(std::string_view base, std::string_view suffix = {}) {
+  return UniqueNameGenerator().New(base, suffix);
+}
+inline std::string GetUniqueName(const OrtNodeUnit& node_unit, std::string_view suffix = {}) {
+  return UniqueNameGenerator().New(node_unit, suffix);
+}
```
qti-hungjuiw: Could we make `GetUniqueName` a **public member function** of `UniqueNameGeneratorImpl`?

This way, `UniqueNameGeneratorImpl::GetUniqueName` would be public, while `UniqueNameGeneratorImpl::New` could remain private.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
qti-yuduo: Intentional to keep things original/shorter.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
qti-yuduo: This is the idiomatic I used to avoid changing caller side (so the change is smaller/cleaner). If you prefer, I can do a follow up to use `New` (already a public method) on all callers. Lemme know! Leaving as follow up.

qti-hungjuiw: Sure. We can leave this as follow up.

qti-hungjuiw (approved):

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -121,11 +123,54 @@ Ort::Status GetQnnDataType(const bool is_quantized_tensor,
                            const ONNXTensorElementDataType onnx_data_type,
                            Qnn_DataType_t& tensor_data_type);
 
-// Returns an unique name string based on a base string and an optional suffix.
-std::string GetUniqueName(const std::string& base, std::string_view suffix = {});
+// Name generator that produces unique QNN node names by appending a counter suffix,
+// (e.g., "_2") when the same base + suffix combination is requested more than once.
+class UniqueNameGeneratorImpl {
```
qti-jkilpatrick, line 128: I'm a little sad that this is in the header, but I won't complain too loudly.

qti-jkilpatrick (approved):

onnxruntime/core/providers/qnn/builder/qnn_utils.h
qti-yuduo: fair, moving to cc

onnxruntime/core/providers/qnn/builder/qnn_utils.h
qti-yuduo: Needed another change. So updated caller side directly here.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
qti-yuduo: moved.

PR #174, opened 2026-03-25
[QNN-EP] Parallelize Graph Prepare

quic-calvnguy:
### Description 
- QnnJobThreadPool created
 - Parallelize graph finalization to improve prepare time

### Motivation and Context
This PR is to shorten the time it takes to compile a model and, from the user's perspective, ORT session creation.
The goal is to parallelize graph compilation as much as possible. At the moment, only graph finalization can be done asychronously.

There are plans to add metrics to the profiling .csv files instead of verbose logging. Additionally, there are plans to check the possibility of parallelizing graph composition and adding a new HTP custom config option for offline prepare use cases (i.e. no execution needed).


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -731,6 +732,56 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    max_num_supported_threads = 4;
+  }
+  // 8 threads provided the best initialization performance from testing
+  // Default to max number of supported threads if less than 8. Otherwise default to 8 threads
+  uint8_t def_num_graph_prepare_threads = max_num_supported_threads > 8 ? 8 : max_num_supported_threads;
+#endif
+
+  if (!num_graph_prepare_threads_str.empty()) {
+#if defined(_WIN32) && (defined(__aarch64__) || defined(_M_ARM64))
+    uint8_t value = static_cast<uint8_t>(std::stoi(num_graph_prepare_threads_str));
```
qti-chuteng, line 782: if num_graph_prepare_threads_str is not a valid non-zero number, it should raise the error

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -128,6 +128,15 @@ class QnnEp : public OrtEp, public ApiPtrs {
     QnnEp& ep;
   };
 
+  typedef struct GraphFinalizationInfo {
+    std::string model_name;
+    std::unique_ptr<qnn::QnnModel> model;
+    std::unique_ptr<std::thread> thread;
```
qti-chuteng: it seems not be used.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -731,6 +732,56 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    ORT_CXX_LOG(logger_,
+                ORT_LOGGING_LEVEL_VERBOSE,
+                (" User specified num_graph_prepare_threads: " + std::to_string(value)).c_str());
+
+    if (value > max_num_supported_threads) {
+      ORT_CXX_LOG(logger_,
+                  ORT_LOGGING_LEVEL_WARNING,
+                  ("Specified number of graph prepare threads (" + std::to_string(value) + ") is outside of the allowable range [1," + std::to_string(max_num_supported_threads) + "]. Defaulting to " + std::to_string(def_num_graph_prepare_threads) + "threads.").c_str());
```
qti-chuteng: " threads."

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -640,6 +647,70 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+3. Input/Ouput tensor setup
+
+For a typical model, this is done sequentially. However, for larger models, such as LLMS, one model may contain multiple subgraphs of substantial size. If done sequentially, the total time to prepare a model may seem like an eternity, thus resulting in the need for parallelization.
+
+Parallel graph preparation was implemented to parallelize the prepare process as much as possible, saving a considerable amount of time. The QNN API currently only supports asynchronous graph finalization. As such, parallelized prepare is as follows:
+1. Create and start a thread pool
+2. For each subgraph, compose the QNN graph
+3. For each subgraph, submit a job to finalize the QNN graph into the thread pool
+4. Wait for all jobs to finish
+6. For each subgraph, setup the I/O tensors
```
qti-chuteng: 5.

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void QnnJobThreadPool::Start() {
+  if (IsRunning()) {
+    return;
+  }
+
+  ORT_CXX_LOG(OrtLoggingManager::GetDefaultLogger(), ORT_LOGGING_LEVEL_VERBOSE, "QnnJobThreadPool: Start");
+  std::unique_lock<std::mutex> s_lock(state_mutex_);
+  std::unique_lock<std::mutex> tp_lock(thread_pool_mutex_);
```
qti-chuteng: does there anywhere else use thread_pool_mutex_ ?

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+
+void QnnJobThreadPool::WaitForAllJobsToFinish() {
+  ORT_CXX_LOG(OrtLoggingManager::GetDefaultLogger(),
+              ORT_LOGGING_LEVEL_VERBOSE,
+              "QnnJobThreadPool: Waiting for all jobs to finish");
+
+  // Block all newly submitted jobs from entering the queue
+  std::unique_lock<std::mutex> lock(queue_mutex_);
+  // Only wait until queue is empty if thread pool has not been stopped
+  if (IsRunning() && !job_queue_.empty()) {
```
qti-chuteng, line 149: IsRunning requires  queue_mutex_, may cause the deadlock

I think that you can try to change statement from "IsRunning() && !job_queue_.empty()" to "!job_queue_.empty()"

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -640,6 +647,70 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+## Parallel Graph Preparation
+For general context, please read the [QNN context binary cache feature](#qnn-context-binary-cache-feature) section above.
+
+The process to prepare a model to run on the HTP backend is to do the following for each subgraph:
+
+1. Compilation/composition into a QNN Graph
+2. Finalization into a HTP-friendly binary
+3. Input/Ouput tensor setup
+
+For a typical model, this is done sequentially. However, for larger models, such as LLMS, one model may contain multiple subgraphs of substantial size. If done sequentially, the total time to prepare a model may seem like an eternity, thus resulting in the need for parallelization.
```
qti-jkilpatrick: LLMS --> LLMs

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -640,6 +647,70 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+Parallel graph preparation was implemented to parallelize the prepare process as much as possible, saving a considerable amount of time. The QNN API currently only supports asynchronous graph finalization. As such, parallelized prepare is as follows:
+1. Create and start a thread pool
+2. For each subgraph, compose the QNN graph
+3. For each subgraph, submit a job to finalize the QNN graph into the thread pool
+4. Wait for all jobs to finish
+6. For each subgraph, setup the I/O tensors
+
+### General Usage
+To utilize this feature, set the value of `num_graph_prepare_threads` to the desired number of threads as shown in the example code below:
+```
```
qti-jkilpatrick: \```python

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+QnnJobThreadPool::QnnJobThread::QnnJobThread(uint8_t thread_num, QnnJobThreadPool* thread_pool_ptr)
+    : thread_num_(thread_num), tp_(thread_pool_ptr) {
+  ORT_CXX_LOG(OrtLoggingManager::GetDefaultLogger(),
+              ORT_LOGGING_LEVEL_VERBOSE,
+              ("QnnJobThread: Thread " + std::to_string(thread_num_) + " created").c_str());
+
+  // Used to exit out of QnnJobThreadPool::WaitForJobQueueUpdate() regardless of job queue status
+  exit_predicate_ = [this]() {
+    return IsStopped();
+  };
```
qti-jkilpatrick, line 20: If we're capturing `this`, I think we need to ensure that this type isn't movable. To do that, perhaps add something like this to the header:

```c++
// This class captures `this` so it cannot be safely moved.
QnnJobThread::QnnJobThread(QnnJobThread&&) = delete;
QnnJobThread& operator=(QnnJobThread&&) = delete;
```

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void QnnJobThreadPool::QnnJobThread::Start() {
+  // Only created thread if no thread exists and the current state is stopped
+  if (!thread_ && IsStopped()) {
+    std::unique_lock<std::mutex> lock(thread_state_mutex_);
+    thread_stopped_ = false;
+  } else {
+    return;
+  }
```
qti-jkilpatrick: Is there a race condition here? If two threads make it into the `if` side, only one will actually set `thread_stopped_`, but both will initialize `thread_`.

qti-jkilpatrick: OTOH, `Start` isn't `const` so you're not making any promises about thread safety here so maybe you can ignore me.

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+  // Only stop if the thread exists and current state is running
+  if (thread_ && !IsStopped()) {
+    ORT_CXX_LOG(OrtLoggingManager::GetDefaultLogger(),
+                ORT_LOGGING_LEVEL_VERBOSE,
+                ("QnnJobThread: Thread " + std::to_string(thread_num_) + " stopping").c_str());
+    std::unique_lock<std::mutex> lock(thread_state_mutex_);
+    thread_stopped_ = true;
+  } else {
+    return;
+  }
```
qti-jkilpatrick: Same comments as `Start` that could probably be ignored.

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+}
+
+QnnJobThreadPool::QnnJobThreadPool(uint8_t max_num_threads)
+    : max_num_threads_(max_num_threads), running_(false) {
+  thread_pool_.reserve(max_num_threads);
+}
+
+QnnJobThreadPool::~QnnJobThreadPool() {
+  Stop();
+  WaitForAllJobsToFinish();
```
qti-jkilpatrick: These might want to be in a `try` since exceptions in destructors cause the process to be aborted (unless a sensible state cannot be recovered and crashing really is the right thing to do).

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  ORT_CXX_LOG(OrtLoggingManager::GetDefaultLogger(), ORT_LOGGING_LEVEL_VERBOSE, "QnnJobThreadPool: Start");
+  std::unique_lock<std::mutex> s_lock(state_mutex_);
+  std::unique_lock<std::mutex> tp_lock(thread_pool_mutex_);
+  running_ = true;
+
+  while (thread_pool_.size() < max_num_threads_) {
+    StartJobThread();
+  }
```
qti-jkilpatrick: Since `StartJobThread` doesn't necessarily create a thread, it might be safer to make this a `for` loop.

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void QnnJobThreadPool::StartJobThread() {
+  uint8_t thread_num = static_cast<uint8_t>(thread_pool_.size());
+  auto job_thread = std::make_unique<QnnJobThread>(thread_num, this);
+  job_thread->Start();
+  thread_pool_.push_back(std::move(job_thread));
+}
+
+void QnnJobThreadPool::WaitForJobQueueUpdate(const uint8_t thread_num, std::function<bool()>& exit_predicate) {
```
qti-jkilpatrick: Can `exit_predicate` be `const`?

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.h
```diff
@@ -0,0 +1,137 @@
...
+  // Notifies a waiting thread that a new job has started
+  // Implies job has been taken from the job queue
+  void NotifyJobStarted() {
+    job_started_cv_.notify_all();
+  }
+
+  bool IsRunning() {
+    std::unique_lock<std::mutex> lock(state_mutex_);
+    return running_;
+  }
```
qti-jkilpatrick: This feels like a method that should be `const`. Of course, `state_mutex_` would have to be `mutable`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -731,6 +732,56 @@ QnnEp::QnnEp(QnnEpFactory& factory,
               ORT_LOGGING_LEVEL_VERBOSE,
               ("User specified enable_htp_fp16_precision: " + enable_htp_fp16_precision_str).c_str());
 
+  std::string num_graph_prepare_threads_str;
+  GetSessionConfigEntryOrDefault(ort_api,
+                                 session_options_,
+                                 FormatEPConfigKey("num_graph_prepare_threads"),
+                                 "",
+                                 num_graph_prepare_threads_str);
+#if defined(_WIN32) && (defined(__aarch64__) || defined(_M_ARM64))
```
qti-jkilpatrick, line 741: * Why only ARM64 Windows?
* IIRC, MSVC doesn't define `__arch64__`.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -203,6 +203,13 @@ Alternatively to setting profiling_level at compile time, profiling can be enabl
 |'0'|Default. Version check enabled.|
 |'1'|Skip QNN API interface version check to use other QNN library versions.|
 
+|`"num_graph_prepare_threads"`|Description|
```
qti-jkilpatrick, line 206: It might be worth mentioning here and in the section describing what happens that this is Windows-only (but also, why no love for Linux and Android?).

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -2649,6 +2649,137 @@ TEST_F(QnnHTPBackendTests, CompileApi_OutputStream_ReturnStatus) {
...
+    output_names_c.push_back(output_names[i].c_str());
+  }
+
+  ProviderOptions qnn_options = {{"backend_type", "htp"}};
+  Ort::SessionOptions so;
+  RegisteredEpDeviceUniquePtr registered_ep_device;
+  RegisterQnnEpLibrary(registered_ep_device, so, onnxruntime::kQnnExecutionProvider, qnn_options);
+
+  std::string output_model_file_str(output_model_file.string());
+  std::wstring compiled_model_file(output_model_file_str.begin(), output_model_file_str.end());
```
qti-jkilpatrick: I think you can get the same result with `std::filesystem::path::to_wstring`.

qti-jkilpatrick: (or `native()` to avoid Windows-specific code)

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
quic-calvnguy: IsRunning() uses state_mutex_ so there is no deadlock here.

The queue mutex lock is there to block any job submissions from entering the queue while the wait is on-going so the calling thread does not hang while another thread submits new jobs.

Additionally, it is possible for the thread pool to not be running but allow job submissions to queue up. If the conditional does not include IsRunning() in this case, then WaitForAllJobsToFinish() will block until the thread pool starts & the queue empties out.

I can change this to WaitForQueuedJobsToFinish() instead so it is clearer. I realize "AllJobs" may be misleading.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
quic-calvnguy: The feature will roll out in multiple phases. For now, it's restricted to ARM64 Windows for a more controlled roll-out.

Not sure about MSVC defining __aarch64__ but the guard does work when building for Windows x86.

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.h
```diff
@@ -0,0 +1,134 @@
...
+
+class QnnJobThreadPool {
+ private:
+  // Class that contains main job thread
+  // - Checks the thread pool job queue for jobs
+  // - Run any available jobs. Wait for new submission to queue otherwise
+  class QnnJobThread {
+   public:
+    QnnJobThread(uint8_t thread_num, QnnJobThreadPool* thread_pool_ptr);
+    QnnJobThread::QnnJobThread(QnnJobThread&&) = delete;
```
qti-jkilpatrick: Some compilers do not allow the extra qualification.
```suggestion
   QnnJobThread(QnnJobThread&&) = delete;
```

onnxruntime/core/providers/qnn/builder/qnn_thread_pool.cc
```diff
@@ -0,0 +1,197 @@
...
+              ("QnnJobThread: Thread " + std::to_string(thread_num_) + " created").c_str());
+
+  // Used to exit out of QnnJobThreadPool::WaitForJobQueueUpdate() regardless of job queue status
+  exit_predicate_ = [this]() {
+    return IsStopped();
+  };
+}
+
+QnnJobThreadPool::QnnJobThread::~QnnJobThread() {
+  Stop();
```
qti-jkilpatrick: I think `std::unique_lock` and `std::thread::join` can throw under some circumstances. Consider adding a `try` here unless failure to `Stop` should cause the process to `std::abort`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-jkilpatrick: It makes sense that it blocks compilation on x64 Windows since `_M_ARM64` isn't defined and `__aarch64__` is never defined by MSVC -- it's a gcc/clang thing.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -2649,6 +2815,134 @@ TEST_F(QnnHTPBackendTests, CompileApi_OutputStream_ReturnStatus) {
...
```
qti-jkilpatrick: I think the comment about using either `.to_wstring()` or (better) `.native()` might have been lost in the churn.

qti-jkilpatrick (commented):
Good stuff. We're down to nits (sorry!).

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
quic-calvnguy: Oop this was code that should have been removed that I apparently missed

qti-jkilpatrick (approved):
I do wish you'd remove `__aarch64__` from the preprocessor guards since it is unnecessary and confusing, but I won't block approval because it's harmless.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1759,7 +1759,8 @@ Ort::Status QnnBackendManager::SetupBackend(
   bool enable_htp_weight_sharing = false;
   if (share_ep_contexts && !load_from_cached_context) {
 #if (defined(__aarch64__) || defined(_M_ARM64)) && \
-    !(QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 34))
+    (QNN_API_VERSION_MAJOR < 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR < 34))
+    ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_WARNING, ("FINDME: " + std::to_string(QNN_API_VERSION_MAJOR) + "." + std::to_string(QNN_API_VERSION_MINOR) + "." + std::to_string(QNN_API_VERSION_PATCH)).c_str());
```
minfhong-qti, line 1763: Sorry for the late comment. Should this log be removed?

PR #175, opened 2026-03-26
Fix UT failures

huaychou:
### Description

- Fix qnn_basic_test.cc

### Motivation and Context

Ensure that all the UTs are destructed session before Ep Factory, otherwise it may introduce segmentation fault.


huaychou:
All changes have been merged in https://github.com/onnxruntime/onnxruntime-qnn/pull/117, so close this PR. However, we need a better solution for this destruction order issue.

PR #183, opened 2026-03-27
[QNN EP] Skip WIN_AMD 64 Wheel upload

qti-mbadnara:
### Description
This change skips AMD64 wheels when they are uploaded to PyPI. 

### Motivation and Context

Change is needed as the AMD64 wheels are all greater than 100 MB in size.

### DIFF
--- a/qcom/scripts/upleveling/qnn_ep_uplevel.py
+++ b/qcom/scripts/upleveling/qnn_ep_uplevel.py
@@ -244,7 +244,7 @@ class WheelUpleveler(ArtifactUpleveler):
 
     def _should_skip_artifact(self, artifact_filename: str) -> bool:
         """Skip win_amd64 wheels"""
-        return "win_amd64.whl" in artifact_filename
+        return artifact_filename.endswith("win_amd64.whl")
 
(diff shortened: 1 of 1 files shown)


qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -229,6 +242,10 @@ class WheelUpleveler(ArtifactUpleveler):
     def artifact_format(self) -> str:
         return "wheel"
 
+    def _should_skip_artifact(self, artifact_filename: str) -> bool:
+        """Skip win_amd64 wheels"""
+        return "win_amd64.whl" in artifact_filename
```
qti-ashwshan: Instead of checking the string can we do ->

return artifact_filename.endswith("win_amd64.whl")

Since the whl format always ends with amd64.

qti-ashwshan (approved):

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Fixed as part of the latest change

tirupath-qti (approved):

PR #172, opened 2026-03-25
Publish nightly artifacts for QA

qti-jkilpatrick:
### Description

Update nightly release process to create artifacts for QA testing.

PR #185, opened 2026-03-31
Add Linux aarch64 OE GCC11.2 cross-compilation support for wheels and…

huaychou:
### Description

- Add CI workflow jobs to build Python 3.11–3.14 wheels for linux/aarch64_oe_gcc11_2
- Fix build_and_test.py to pass target_py_version and build_zip flag to aarch64 OE GCC11.2 tasks
- Add --build-zip flag to build.sh; invoke pkg_assets.py post-build to produce zip archives
- Add llvm_linux_x86_64 package entry to packages.yml and get_llvm_bindir() helper in tools.sh
- Replace hardcoded LLVM paths in toolchain.cmake with ORT_BUILD_LLVM_BINDIR env var
- Set ONNXRUNTIME_WHEEL_PLAT_NAME=linux_aarch64 during cross-compilation for correct wheel naming
- Support ONNXRUNTIME_WHEEL_PLAT_NAME override in setup.py bdist_wheel
- Add --target_arch flag to pkg_assets.py for cross-compilation zip naming

### Motivation and Context

- Add Linux aarch64 OE GCC11.2 cross-compilation support for wheels and zips

### DIFF
--- a/qcom/packages.yml
+++ b/qcom/packages.yml
@@ -72,6 +72,13 @@ linux_oe_gcc112_toolchain:
   version: 20250402d
   url: http://ort-ep-win-01.na.qualcomm.com:8000/linux-oe-aarch64-gcc11.2-toolchain/linux-oe-aarch64-gcc11.2-toolchain-1-universal-{version}.tar.gz
   sha256: 1031825247827ad60189323aa5530490deae61547f4d9794ef9e0645e9207266
+llvm_linux_x86_64:
+  version: 16.0.4
+  platform_version: ubuntu-22.04
+  url: https://github.com/llvm/llvm-project/releases/download/llvmorg-{version}/clang+llvm-{version}-x86_64-linux-gnu-{platform_version}.tar.xz
+  sha256: fd464333bd55b482eb7385f2f4e18248eb43129a3cda4c0920ad9ac3c12bdacf
+  content_root: clang+llvm-{version}-x86_64-linux-gnu-{platform_version}
+  bindir: bin
 java_linux_x86_64: &java_linux_common
   major_version: 21
   version: 21.0.5
--- a/qcom/scripts/linux/build.sh
+++ b/qcom/scripts/linux/build.sh
@@ -38,8 +38,13 @@ target_py_version=
 use_cache=1
 warnings_as_errors=1
 build_java=
+build_zip=
 for i in "$@"; do
   case $i in
+    --build-zip)
+      build_zip=1
+      shift
+      ;;
     --config=*)
       config="${i#*=}"
       shift
@@ -177,9 +182,15 @@ case "${target_platform}" in
           toolchain_root="$(get_linux_oe_gcc112_toolchain_root)"
           toolchain_cmake="${REPO_ROOT}/qcom/scripts/linux/linux-aarch64-gcc11.toolchain.cmake"
 
-          # We need $toolchain_root from the toolchain.cmake, but the toolchain.cmake is sometimes
-          # evaluated without the project's CMakeCache.txt entries. Pass it through the environment :-/
+          # We need $toolchain_root and $llvm_bindir from the toolchain.cmake, but the toolchain.cmake
+          # is sometimes evaluated without the project's CMakeCache.txt entries. Pass them through
+          # the environment :-/
           export ORT_BUILD_LINUX_TOOLCHAIN_ROOT="${toolchain_root}"
+          export ORT_BUILD_LLVM_BINDIR="$(get_llvm_bindir)"
+
+          # The wheel is built on an x86_64 host but targets aarch64; override the
+          # platform tag so the wheel is correctly named linux_aarch64.
+          export ONNXRUNTIME_WHEEL_PLAT_NAME="linux_aarch64"
 
           platform_args+=(--cmake_extra_defines
                           CMAKE_TOOLCHAIN_FILE:FILEPATH="${toolchain_cmake}"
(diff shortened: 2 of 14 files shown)


qti-jkilpatrick:
We built the `manylinux_2_34` build for this purpose. In what way does it not already solve this problem?

qcom/scripts/linux/build.sh
```diff
@@ -277,6 +288,24 @@ else
...
+        --use_ninja
+        --verbose
+      )
+      if [ -n "${ORT_VERSION_SUFFIX:-}" ]; then
+        zip_args+=(--version_suffix "${ORT_VERSION_SUFFIX}")
+      fi
+      if [ "${target_arch}" == "aarch64_oe_gcc11_2" ]; then
+        zip_args+=(--target_arch "aarch64")
+      fi
+      "${python_for_build}" "${REPO_ROOT}/tools/ci_build/pkg_assets.py" "${zip_args[@]}"
```
kuanyul-qti: Can we build zip when calling `${REPO_ROOT}/tools/ci_build/build.py` (L286) by passing the argument `build_zip_asset`?

qcom/packages.yml
```diff
@@ -72,6 +72,12 @@ linux_oe_gcc112_toolchain:
   version: 20250402d
   url: http://ort-ep-win-01.na.qualcomm.com:8000/linux-oe-aarch64-gcc11.2-toolchain/linux-oe-aarch64-gcc11.2-toolchain-1-universal-{version}.tar.gz
   sha256: 1031825247827ad60189323aa5530490deae61547f4d9794ef9e0645e9207266
+llvm_linux_x86_64:
+  version: 16.0.4
+  url: https://github.com/llvm/llvm-project/releases/download/llvmorg-{version}/clang+llvm-{version}-x86_64-linux-gnu-ubuntu-22.04.tar.xz
```
kuanyul-qti: Should we add the ubuntu version in the package name, e.g, `llvm_linux_x86_64_ubuntu_2204`?

kuanyul-qti (commented):
Please also update https://github.com/onnxruntime/onnxruntime-qnn/blob/dev/huaychou/add-linux-aarch64-release/.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml#L118 to download the linux wheel and zip for release.

huaychou:
> We built the `manylinux_2_34` build for this purpose. In what way does it not already solve this problem?

Since IoT platform requires ORT QNN EP artifacts of `linux-aarch64-oe-gcc11.2` and we don't release these artifacts now.
In addition, I think `manylinux_2_34` is not built in CI pipeline now. Another solution is to re-enable `manylinux_2_34` in CI pipeline and release pipeline.

huaychou:
I think we need to wait until https://github.com/onnxruntime/onnxruntime-qnn/pull/172 merged since the nightly and release artifacts are quite different now.

qcom/packages.yml
huaychou: Done

qcom/scripts/linux/build.sh
huaychou: Done

huaychou:
> Reference in

Done

huaychou:
> We built the `manylinux_2_34` build for this purpose. In what way does it not already solve this problem?

We found that `manylinux_2_34` uses QNN ABI `aarch64-ubuntu-gcc9.4`, while IoT platform requires `aarch64-oe-linux-gcc11.2`, so `linux-aarch64-oe-gcc11_2` is still required.

qti-jkilpatrick:
> > We built the `manylinux_2_34` build for this purpose. In what way does it not already solve this problem?
> 
> Since IoT platform requires ORT QNN EP artifacts of `linux-aarch64-oe-gcc11.2` and we don't release these artifacts now. In addition, I think `manylinux_2_34` is not built in CI pipeline now. Another solution is to re-enable `manylinux_2_34` in CI pipeline and release pipeline.

There is a lot of confusion about ABI compatibility in AISW. The `manylinux_2_34` wheels should work on IoT devices; I have tested them on Rubik PI running Ubuntu. [They are built every night](https://github.com/onnxruntime/onnxruntime-qnn/actions/workflows/qualcomm-internal-release.yml), but not with every PR because they are expensive (requiring docker with ARM64 emulation on an x86 runner) and we don't yet have any test devices available to CI (though we are working on solving this problem).

qti-jkilpatrick:
> > We built the `manylinux_2_34` build for this purpose. In what way does it not already solve this problem?
> 
> We found that `manylinux_2_34` uses QNN ABI `aarch64-ubuntu-gcc9.4`, while IoT platform requires `aarch64-oe-linux-gcc11.2`, so `linux-aarch64-oe-gcc11_2` is still required.

There is a lot of confusion about how ABI compatibility works. On Linux, it depends not on the compiler version, but rather the version of glibc against which it was built. This is why the manylinux project exists, to create a predictable environment in which to build artifacts that work on a great number of Linux distributions. I chose `manylinux_2_34` because it's the oldest one that I could get QAIRT artifacts to use.

If there are meaningful differences between the gcc9.4 and gcc11.2 artifacts that would prevent the former from running on IoT devices, please let me know what they are. The only thing I can think of is that we might be missing stub/skel files for various HTP architectures, but the binaries that actually run on Linux should be fine as far as I can tell.

qti-jkilpatrick:
Also, if the gcc11.2 SDK is actually required, why not make the easy change and do that instead of inventing a new build?

qti-jkilpatrick:
We continue to discuss wheels for aarch64 and it is not going in this direction. We hope for our support to be as close to ORT's as possible and that means manylinux, not an _ad hoc_ approach such as this.

huaychou:
> We continue to discuss wheels for aarch64 and it is not going in this direction. We hope for our support to be as close to ORT's as possible and that means manylinux, not an _ad hoc_ approach such as this.

Sure. we will create another PR to release manylinux wheel with QNN ABI `linux-aarch64-oe-gcc11_2` for IoT platforms.

PR #187, opened 2026-03-31
Fix dump dlc create dir with full .dlc path bug.

quic-muchhsu:
### Description

The issue: std::filesystem::create_directories(dlc_path) was being called with the full file path (e.g., /path/to/file.dlc), which tried to create a directory named file.dlc instead of just the parent directory.

The fix: Changed to std::filesystem::create_directories(dlc_path.parent_path()) which correctly creates only the directory portion of the path, not the filename.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -94,7 +94,7 @@ class QnnIrConfig : public QnnSerializerConfig {
     dlc_path_custom_config->option = QNN_GRAPH_CONFIG_OPTION_CUSTOM;
     dlc_path_custom_config->customConfig = dlc_path_config;
 
-    std::filesystem::create_directories(dlc_path);
+    std::filesystem::create_directories(dlc_path.parent_path());
 
     // Keep the pointer to dlc_path_str's null-terminated string alive.
     std::swap(dlc_path_str, dlc_path_str_);
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -94,8 +94,6 @@ class QnnIrConfig : public QnnSerializerConfig {
     dlc_path_custom_config->option = QNN_GRAPH_CONFIG_OPTION_CUSTOM;
     dlc_path_custom_config->customConfig = dlc_path_config;
 
-    std::filesystem::create_directories(dlc_path.parent_path());
-
     // Keep the pointer to dlc_path_str's null-terminated string alive.
     std::swap(dlc_path_str, dlc_path_str_);
 
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -94,6 +94,8 @@ class QnnIrConfig : public QnnSerializerConfig {
     dlc_path_custom_config->option = QNN_GRAPH_CONFIG_OPTION_CUSTOM;
     dlc_path_custom_config->customConfig = dlc_path_config;
 
+    std::filesystem::create_directories(dlc_path.parent_path());
+
     // Keep the pointer to dlc_path_str's null-terminated string alive.
     std::swap(dlc_path_str, dlc_path_str_);
 


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -94,7 +94,7 @@ class QnnIrConfig : public QnnSerializerConfig {
     dlc_path_custom_config->option = QNN_GRAPH_CONFIG_OPTION_CUSTOM;
     dlc_path_custom_config->customConfig = dlc_path_config;
 
-    std::filesystem::create_directories(dlc_path);
+    std::filesystem::create_directories(dlc_path.parent_path());
```
qti-jkilpatrick: I suspect something more is going on. IIRC, the `dlc_path` is mean to be the directory into which DLC files are written ([see docs](https://github.com/onnxruntime/onnxruntime-qnn/blob/main/docs/execution_providers/QNN-ExecutionProvider.md?plain=1#L195)). Perhaps there is an issue where a nested directory is specified and this fails?

quic-muchhsu: At L86: https://github.com/onnxruntime/onnxruntime-qnn/blob/bf5ed1a389733709d27e2e15e0a262b843e53ae7/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc#L86 it changes to the full dlc path.

Unless we actually want to create a dir with dlc suffix?

We can also change the implementation to `std::filesystem::create_directories(dlc_dir_ );` which will just do the same job.

qti-jkilpatrick: Good point. Do you have any idea how this was working before? What caused you to find this issue?

quic-muchhsu: I am trying to dump a dlc file to reproduce an error
I am setting my dlc_path here `provider_options["dump_qnn_ir_dlc_dir"] = "/local/mnt/workspace/onnxruntime_mlg_1/";`
and I hit this error:
```
❯ terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'                                                  
  what():  filesystem error: cannot create directories: Not a directory                                                                        
[/local/mnt/workspace/onnxruntime_mlg_1/QNNExecutionProvider_QNNExecutionProvider_9969863190718819140_0_0.dlc]                                 
Aborted (core dumped)           
```
I was just working, with some error before, not a segmentation fault until recent. I can still get the dlc.

Actually meanwhile let me double check the if the intent is to create a dir with .dlc suffix, and the real issue is dir created after dlc is written with the same path.

quic-muchhsu: Okay I have dig into it more. When dumping json, it just dumping to the full path with `.json`, and in dump dlc, it creates `.dlc` file and dumps dlc to the full path. It's fine on the first pass, because file can overwrite a dir.
But on the second pass, create dir will fail because create dir can't overwrite a file.
I think we should just remove the `std::filesystem::create_directories(dlc_path);` since the created dir will be overwrite anyway in current code base.

I will keep things works the same and solve dev bug on second run.

See JSON code snip:
```
  JSON graph (qnn_model.cc:267-278):
  if (build_json_graph) {
      const nlohmann::json& json_graph = qnn_model_wrapper.GetQnnJSONGraph();
      std::ofstream ofs(context.json_qnn_graph_path);  // Just opens file directly

      if (ofs.is_open()) {
        ofs << json_graph.dump();
        ofs.close();
      } else {
                    ("Could not open JSON graph file: " + context.json_qnn_graph_path).c_str());
      }
  }
```

qti-jkilpatrick: So the first time, it creates something like `some_dir/dlc_path.dlc/dlc_path.dlc` and then the second time it fails when trying to create `some_dir/dlc_path.dlc`?

quic-muchhsu: Not exactly,
Say we set dlc_dir_ = `some_dlc_path`
The 1st pass, it will create a directory named `some_dlc_path/dlc_graph.dlc` and writes dlc file at  `some_dlc_path/dlc_graph.dlc` that overwirte a dir to a file.
So the create dir part is like a noop.

The 2nd pass, which most people are not doing.
It again tries to create a dir name  `some_dlc_path/dlc_graph.dlc` -> this is where it error outs, because the dlc is already exist, and you can create a dir with the same name.

I am thinking to just remove the create dir part since it's a noop anyway. We could add more line to make sure there's nested structure (though my humble opinion thinks it not necessary), and it could break existing code in the Hub side.

qti-jkilpatrick: So ORT creates `some_dlc_path/dlc_graph.dlc` as a directory and then QAIRT deletes the directory and writes a file in that path?

If `some_dlc_path/` does not exist, will QAIRT create it?

quic-muchhsu: So ORT creates some_dlc_path/dlc_graph.dlc as a directory and then QAIRT deletes the directory and writes a file in that path? -> Yes.

If some_dlc_path/ does not exist, will QAIRT create it? -> No. If I removed the create_dir line, it won't create a dir and will throw error `[E:onnxruntime:, qnn_model.cc:311 FinalizeGraphs] Failed to finalize QNN graph. Error code: 6022`

qti-jkilpatrick: That is not the most intuitive behavior.

Since QAIRT won't create the destination directory and the existing behavior here is nominally to do that, I think your first iteration is the way to go.

Thanks for patiently explaining all this to me.

quic-muchhsu: Okay, I'll go back to create a parent dir implementation.

qti-jkilpatrick (approved):

qti-chuteng (approved):

PR #188, opened 2026-03-31
Add another case we need to split GEMM

qti-yuduo:
### Description
Split when bias is an intermediate (NATIVE) tensor produced by another op.

### Motivation and Context

1. ORT's `MatMulAddFusion`** (runs before QNN EP sees the graph) fuses `MatMul + Add → Gemm`. This is a general ORT graph optimization that fires whenever it sees the pattern — it doesn't know or care what the Add's second input is.

2. **The problematic pattern** — CLIP-style text projection averaging:
   ```
   mm1 = MatMul(A1, W)
   mm2 = MatMul(A2, W)
   add1 = mm1 + mm2          ← MatMulAddFusion fires: becomes Gemm(A1, W, C=mm2)
   add2 = add1 + mm3         ← MatMulAddFusion fires: becomes Gemm(A1(?), W, C=add1)
   ```
   After fusion, the `C` (bias) input to Gemm is `mm2` or `add1` — an **intermediate NATIVE tensor**, not a constant or graph input.

3. **QNN `FullyConnected` constraint**: QNN's FC op requires the bias to be `STATIC` (initializer/constant) or `APP_WRITE` (graph input). It **cannot accept a `NATIVE` tensor** (intermediate output of another op) as bias. This would cause a validation/compilation failure.

4. **The fix**: When the bias is `NATIVE`, decompose `Gemm → FC(A, W) + ElementWiseAdd(fc_output, C)`. The Add op has no such restriction on input types.

Fixes following issue: 
```
Node(s) placed on [CPUExecutionProvider]. Number of nodes: 17
DequantizeLinear (QcQuantizeOp_model.box_head.conv1_ctr.0.weight_dq)
DequantizeLinear (QcQuantizeOp_model.box_head.conv1_offset.0.weight_dq)
DequantizeLinear (QcQuantizeOp_model.box_head.conv1_size.0.weight_dq)
DequantizeLinear (QcQuantizeOp_model.box_head.conv5_offset.weight_dq)
DequantizeLinear (QcQuantizeOp_val_176_dq)
DequantizeLinear (QcQuantizeOp_val_22_dq)
DequantizeLinear (QcQuantizeOp_val_26_dq)
Gemm (node_MatMul_21/MatMulAddFusion)
Gemm (node_MatMul_23/MatMulAddFusion)
Gemm (node_MatMul_131/MatMulAddFusion)
DequantizeLinear (node_MatMul_21/MatMulAddFusion_bias_dq)
DequantizeLinear (node_MatMul_23/MatMulAddFusion_bias_dq)
DequantizeLinear (node_MatMul_131/MatMulAddFusion_bias_dq)
DequantizeLinear (node_Conv_145_bias_dq)
DequantizeLinear (node_Conv_137_bias_dq)
DequantizeLinear (node_Conv_141_bias_dq)
DequantizeLinear (node_conv2d_11_bias_dq)
```

### DIFF
--- a/onnxruntime/test/providers/qnn/gemm_test.cc
+++ b/onnxruntime/test/providers/qnn/gemm_test.cc
@@ -496,6 +496,68 @@ TEST_F(QnnHTPBackendTests, Gemm_TransAB_Dynamic_B_And_Bias) {
                                         ExpectedEPNodeAssignment::All);
 }
 
+// Reproduces the CLIP text projection averaging pattern where ORT's MatMulAddFusion
+// creates Gemm nodes with intermediate (NATIVE) bias:
+//   A1 -> MatMul(W) -> mm1  (stays as MatMul)
+//   A2 -> Gemm(W, C=mm1)   -> add1  (C is NATIVE)
+//   A3 -> Gemm(W, C=add1)  -> add2  (C is NATIVE)
+//   A4 -> Gemm(W, C=add2)  -> add3  (C is NATIVE)
+namespace {
+GetTestModelFn BuildGemmFromMatMulAddTestCase(int64_t K, int64_t N) {
+  return [K, N](ModelTestBuilder& builder) {
+    constexpr int64_t batch = 1;
+    const std::vector<int64_t> input_shape = {batch, K};
+    const std::vector<int64_t> weight_shape = {K, N};
+
+    // 4 dynamic inputs
+    builder.MakeInput<float>("A1", input_shape, -1.0f, 1.0f);
+    builder.MakeInput<float>("A2", input_shape, -1.0f, 1.0f);
+    builder.MakeInput<float>("A3", input_shape, -1.0f, 1.0f);
+    builder.MakeInput<float>("A4", input_shape, -1.0f, 1.0f);
+
+    // Shared static weight
+    builder.MakeInitializer<float>("W", weight_shape, -1.0f, 1.0f);
+
+    // 4 MatMul nodes
+    builder.AddNode("matmul_1", "MatMul", {"A1", "W"}, {"mm1"});
+    builder.AddNode("matmul_2", "MatMul", {"A2", "W"}, {"mm2"});
+    builder.AddNode("matmul_3", "MatMul", {"A3", "W"}, {"mm3"});
+    builder.AddNode("matmul_4", "MatMul", {"A4", "W"}, {"mm4"});
+
+    // Chain of Adds: add1 = mm1 + mm2, add2 = add1 + mm3, add3 = add2 + mm4
+    builder.AddNode("add_1", "Add", {"mm1", "mm2"}, {"add1"});
+    builder.AddNode("add_2", "Add", {"add1", "mm3"}, {"add2"});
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/gemm_test.cc
```diff
@@ -496,6 +496,68 @@ TEST_F(QnnHTPBackendTests, Gemm_TransAB_Dynamic_B_And_Bias) {
...
+TEST_F(QnnHTPBackendTests, GemmFromMatMulAddNonStaticBias) {
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = "htp";
+  provider_options["offload_graph_io_quantization"] = "0";
+
+  RunQnnModelTest(BuildGemmFromMatMulAddTestCase(/*K=*/4, /*N=*/3),
+                  provider_options,
+                  /*opset=*/18,
+                  ExpectedEPNodeAssignment::All,
+                  /*fp32_abs_err=*/1e-2f);
```
quic-ykota: Why is the accuracy threshold kept so high? I see couple of other tests using 1e-5f. Could we use that here?

quic-ykota (commented):
I wish there was a way to subscribe to ORT when to fuse and when not to fuse, which would avoid a lot of unnecessary logic in EP to undo what ORT has done. But, this change looks good to me.

onnxruntime/test/providers/qnn/gemm_test.cc
qti-yuduo: Linux HTP simulator gives slightly higher error in this case. I do see Windows/Android not complaining even with default tolerance which is a good signal.

qti-yuduo:
> I wish there was a way to subscribe to ORT when to fuse and when not to fuse, which would avoid a lot of unnecessary logic in EP to undo what ORT has done. But, this change looks good to me.

Yeah the flexibility would be better. However, note in this case, `MatMulAddFusion` generally helps, thus, we do not actually want to disable entire pass. Undo partial corner cases probably the best thing we could do here.

onnxruntime/test/providers/qnn/gemm_test.cc
quic-ykota: Is it possible to change the tolerance for Windows/Android to be 1e-5f and keep it as 1e-2 for linux? This way we can catch any regressions in the future. Keeping it at 1e-2 might mask subtle errors

qti-yuduo: We can do macro I guess, not elegant but doable.

qti-yuduo: Added. Let's see if it passes.

quic-ykota (approved):

qti-chuteng (approved):

onnxruntime/test/providers/qnn/gemm_test.cc
qti-yuduo: Default still not good enough.. one more try.

quic-muchhsu (approved):

PR #190, opened 2026-04-01
[QNN EP] Add python script for perf test tool

qti-ashwshan:
### Description
- Add Python equivalents of onnxruntime_perf_test.exe targeting the QNN plugin EP.

### Motivation and Context
- Gives the ability to users to easily hook up this script for running models on the NPU in a python env

### DIFF
--- a/tools/scripts/onnxruntime_perf_test.py
+++ b/tools/scripts/onnxruntime_perf_test.py
@@ -1,0 +1,446 @@
+#!/usr/bin/env python3
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+"""QNN EP performance benchmarking script.
+
+Replicates the functionality of onnxruntime_perf_test.exe for the QNN Execution Provider.
+Measures latency (min/max/mean/median/p99) and throughput for an ONNX model on QNN backends.
+
+Usage examples:
+    # Basic HTP benchmark with random inputs
+    python onnxruntime_perf_test.py model.onnx
+
+    # Run for 10 seconds on HTP with burst performance mode
+    python onnxruntime_perf_test.py model.onnx --backend htp --duration_secs 10 \\
+        --htp_performance_mode burst
+
+    # Use specific input data files
+    python onnxruntime_perf_test.py model.onnx --input_data input_0.pb input_1.pb
+
+    # Enable context caching and save outputs
+    python onnxruntime_perf_test.py model.onnx --context_cache --output_dir ./outputs
+"""
+
+from __future__ import annotations
+
+import argparse
+import logging
+import statistics
+import sys
+import time
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import numpy as np
+
+# Allow running this script directly from any working directory
+sys.path.insert(0, str(Path(__file__).parent))
+import qnn_ep_utils as utils
+
+# ---------------------------------------------------------------------------
+# Argument parser
+# ---------------------------------------------------------------------------
+
+
+def build_arg_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        prog="onnxruntime_perf_test.py",
+        description="Benchmark an ONNX model on the QNN Execution Provider.",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog=__doc__,
+    )
+
+    # Positional
+    parser.add_argument("model", type=Path, metavar="MODEL", help="Path to the ONNX model file.")
+
+    # Backend
+    parser.add_argument(
+        "--backend",
+        "-b",
+        default="htp",
+        choices=["cpu", "gpu", "htp"],
+        help="QNN backend to use. (default: htp)",
+    )
+
+    # Run counts / duration
+    run_group = parser.add_mutually_exclusive_group()
+    run_group.add_argument(
+        "--num_runs",
+        "-r",
+        type=int,
+        default=10,
+        metavar="N",
+        help="Number of timed inference runs. Ignored if --duration_secs is set. (default: 10)",
+    )
+    run_group.add_argument(
+        "--duration_secs",
+        "-d",
+        type=float,
+        default=None,
+        metavar="SECS",
+        help="Run inference for this many seconds instead of a fixed count.",
+    )
+    parser.add_argument(
+        "--warmup_runs",
+        "-w",
+        type=int,
+        default=3,
+        metavar="N",
+        help="Number of warmup runs before timing begins. (default: 3)",
+    )
+
+    # Threading
+    parser.add_argument(
+        "--intra_op_threads",
+        type=int,
+        default=None,
+        metavar="N",
+        help="Number of intra-op threads for ORT. (default: ORT default)",
+    )
+    parser.add_argument(
+        "--inter_op_threads",
+        type=int,
+        default=None,
+        metavar="N",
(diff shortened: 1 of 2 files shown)


tools/scripts/qnn_ep_utils.py
```diff
@@ -0,0 +1,374 @@
...
+# ---------------------------------------------------------------------------
+
+_ep_registered = False
+
+# ---------------------------------------------------------------------------
+# EP loading and registration
+# ---------------------------------------------------------------------------
+
+
+def load_qnn_ep() -> tuple[types.ModuleType, types.ModuleType]:
```
minfhong-qti, line 67: I did not get why importing these two modules and passing them around instead of importing at module level at the start. Could you explain a bit more?

qti-hungjuiw:
Do we need to ensure whether test tool in Python is executable on Android device?

tools/scripts/onnxruntime_perf_test.py
```diff
@@ -0,0 +1,446 @@
...
+    if all(p.suffix == ".pb" for p in paths):
+        logging.info(f"Loading {len(paths)} input .pb files")
+        return dict(utils.load_pb_tensor(f) for f in paths)
+
+    # Mixed or unknown: try .pb first, then .npy
+    logging.warning("Mixed or unrecognized input file types. Attempting .pb load.")
+    try:
+        return dict(utils.load_pb_tensor(f) for f in paths)
+    except Exception as e:
+        logging.error(f"Failed to load inputs: {e}. Falling back to random inputs.")
```
qti-chuteng, line 276: ## Concern
when no input_*.pb is found, the script only logs a warning and silently switches to random inputs.

When the user explicitly passes --input_data, they expect that data to drive the benchmark. Misnamed paths, wrong formats, missing onnx package (ImportError raised by load_pb_tensor), or a proto.name mismatch should surface as a hard error, not degrade silently to random data.

## Suggested Fix
1. Narrow the catch to specific exceptions (FileNotFoundError, ImportError), each with a clear message.

2. Remove the "fall back to random on failure" behavior. If the user passed --input_data, failures should raise / sys.exit(1). Random inputs should only be used when --input_data is not provided at all (already handled at the top of the function).

tools/scripts/onnxruntime_perf_test.py
```diff
@@ -0,0 +1,446 @@
...
+
+    # Single directory: look for input_*.pb inside it
+    if len(paths) == 1 and paths[0].is_dir():
+        pb_files = sorted(paths[0].glob("input_*.pb"))
+        if pb_files:
+            logging.info(f"Loading {len(pb_files)} input .pb files from {paths[0]}")
+            return dict(utils.load_pb_tensor(f) for f in pb_files)
+        # Fall through to try as a single file
+        logging.warning(f"No input_*.pb files found in {paths[0]}. Generating random inputs.")
+        return utils.generate_random_inputs(session)
```
qti-chuteng, line 251: ## Concern
when no input_*.pb is found, the script only logs a warning and silently switches to random inputs.

When the user explicitly passes --input_data, they expect that data to drive the benchmark. Misnamed paths, wrong formats, missing onnx package (ImportError raised by load_pb_tensor), or a proto.name mismatch should surface as a hard error, not degrade silently to random data.

## Suggested Fix
Remove the "fall back to random on failure" behavior. If the user passed --input_data, failures should raise / sys.exit(1). Random inputs should only be used when --input_data is not provided at all (already handled at the top of the function).

tools/scripts/onnxruntime_perf_test.py
```diff
@@ -0,0 +1,446 @@
...
+                logging.error(f"Invalid --provider_options entry (expected KEY|VALUE): {item!r}")
+                sys.exit(1)
+            k, v = item.split("|", 1)
+            extra_provider_options[k.strip()] = v.strip()
+
+    # Auto-create any output directories referenced in provider options
+    for dir_key in ("json_qnn_graph_dir", "profiling_file_path", "qnn_saver_path", "dump_qnn_ir_dlc_dir"):
+        if dir_key in extra_provider_options:
+            dir_path = Path(extra_provider_options[dir_key])
+            dir_path.mkdir(parents=True, exist_ok=True)
```
qti-chuteng, line 418: ## Concern
profiling_file_path is not a folder path

## Suggested Fix
# Directories to pre-create (user-provided via --provider_options)
_DIR_KEYS = ("json_qnn_graph_dir", "dump_qnn_ir_dlc_dir", "qnn_saver_path")
_FILE_KEYS = ("profiling_file_path",)

for k in _DIR_KEYS:
    if k in extra_provider_options:
        Path(extra_provider_options[k]).mkdir(parents=True, exist_ok=True)

for k in _FILE_KEYS:
    if k in extra_provider_options:
        Path(extra_provider_options[k]).parent.mkdir(parents=True, exist_ok=True)

tools/scripts/onnxruntime_perf_test.py
```diff
@@ -0,0 +1,446 @@
...
+    @property
+    def mean_ms(self) -> float:
+        return statistics.mean(self.latencies_ms)
+
+    @property
+    def median_ms(self) -> float:
+        return statistics.median(self.latencies_ms)
+
+    @property
+    def p99_ms(self) -> float:
```
qti-chuteng, line 312: ## Concern
No linear interpolation, inconsistent with industry convention (numpy.percentile default)

## Suggested Fix
@property
def p99_ms(self) -> float:
    if not self.latencies_ms:
        return 0.0
    return float(np.percentile(self.latencies_ms, 99))

tools/scripts/qnn_ep_utils.py
```diff
@@ -0,0 +1,374 @@
...
+
+    Dynamic dimensions (None or symbolic strings) are replaced with 1.
+    """
+    inputs: dict[str, np.ndarray] = {}
+    for inp in session.get_inputs():
+        name = inp.name
+        dtype = ort_type_str_to_numpy(inp.type)
+
+        # Resolve concrete shape: replace None / symbolic strings with 1
+        shape = [dim if isinstance(dim, int) and dim > 0 else 1 for dim in inp.shape]
```
qti-chuteng, line 292: ## Concern
The native onnxruntime_perf_test.exe supports user-specified symbolic dim values

## Suggested Fix
Add a CLI option (mirror the native ORT naming):
`
parser.add_argument("--free_dim_override", "-f", nargs="*", default=None,
                    metavar="NAME=VALUE",
                    help="Override free/symbolic dimensions, e.g. 'batch=4 seq_len=128'.")
`

tools/scripts/qnn_ep_utils.py
```diff
@@ -0,0 +1,374 @@
...
+
+    return sess_options
+
+
+# ---------------------------------------------------------------------------
+# ONNX dtype mapping
+# ---------------------------------------------------------------------------
+
+# Maps ONNX TensorProto.DataType integer values to numpy dtypes
+_ONNX_DTYPE_MAP: dict[int, np.dtype] = {
```
qti-chuteng, line 227: ## Concern
Missing BF16 (16), the FP8 family (17-20), INT4 (22), UINT4 (21) and other ONNX 1.17+ types. Models quantized to INT4 / FP16 and loaded via .pb test data will hit ValueError from onnx_dtype_to_numpy.

## Suggested Fix
Add the new type codes (BF16 can be stored as np.float32 with a note), or delegate to onnx.helper.tensor_dtype_to_np_dtype so the mapping stays in sync with onnx upstream automatically.

PR #191, opened 2026-04-01
[QNN EP] Uplevel QAIRT to 2.45.0

qti-mbadnara:
### Description

This change upgrade the mainline build to use `2.45.0 QAIRT SDK`

### Motivation and Context

Qualcomm released `QAIRT 2.45.0` as part of Qualcomm Software Center. This change ensures `onnxruntime-qnn` is built on the latest version of `QAIRT SDK`


qti-kromero (approved):

onnxruntime/test/providers/qnn/qnn_device_compatibility_test.cc
```diff
@@ -144,9 +144,16 @@ TEST_F(QnnDeviceCompatibilityTests, GPUDeviceWithQualcommVendorIsCompatible) {
       env_, onnxruntime::kQnnExecutionProvider, &gpu_device, &details));
   ASSERT_NE(details, nullptr);
 
-  // Verify compatible (no incompatibility reasons)
-  uint32_t reasons_bitmask = 0xFFFFFFFF;
+  // Check if GPU device is incompatible for any reason - if so, skip this test
```
qti-kromero, line 147: Explanation on this change from Claude code

The test GPUDeviceWithQualcommVendorIsCompatible was added in the same commit as the device compatibility feature (March 26), which was on top of QAIRT 2.44. The 2.44 libQnnGpu.so advertises QNN API version 2.33.0. The EP compiled against QAIRT 2.44 headers required req_version = 2.33.x. They matched, so GetQnnInterfaceProvider passed the version check.

Then QAIRT was upleveled to 2.45 (April 1). The 2.45 headers now define QNN_API_VERSION = 2.34.0. The EP is recompiled requiring req_version = 2.34.0. But libQnnGpu.so from 2.45 also advertises 2.34.0 — so that still matches.

Wait — both match. So the version check isn't the issue. The failure must be in InitializeBackend (backendCreate()) or CreateDevice (deviceCreate()). Something in the GPU backend initialization changed between 2.44 and 2.45 that causes it to fail on this x86_64 Linux machine without a physical GPU.

The most likely explanation: QAIRT 2.45's libQnnGpu.so dropped support for software/emulation initialization on x86_64 Linux, or tightened the backendCreate() requirements — whereas 2.44's version was more permissive and succeeded even without real GPU hardware. The test was written on a machine where 2.44 happened to work, and 2.45 broke it.

So to directly answer your question: the mock device is irrelevant to the failure. The mock is only used to pass the vendor ID gate. After that, GetHardwareDeviceIncompatibilityDetails ignores it entirely and does a real SetupBackend call against actual hardware. The tests were never truly mocked at the backend initialization level — they always required a working GPU backend on the test machine.

qti-ashwshan (approved):

qti-kromero:
updates to tolerances:

image

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1760,10 +1760,13 @@ Ort::Status QnnBackendManager::SetupBackend(
   if (share_ep_contexts && !load_from_cached_context) {
 #if (defined(__aarch64__) || defined(_M_ARM64)) && \
     (QNN_API_VERSION_MAJOR < 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR < 34))
-    ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_WARNING, ("FINDME: " + std::to_string(QNN_API_VERSION_MAJOR) + "." + std::to_string(QNN_API_VERSION_MINOR) + "." + std::to_string(QNN_API_VERSION_PATCH)).c_str());
     ORT_CXX_LOG_PTR(logger_ptr_,
                     ORT_LOGGING_LEVEL_WARNING,
                     "Weight sharing on Windows arm64 device requires QNN API version >=2.34. Since Current version is too old, disabling the Weight sharing.");
+#elif defined(__ANDROID__) && (defined(__aarch64__) || defined(_M_ARM64))
```
qti-jkilpatrick: `_M_ARM64` is never defined on `__ANDROID__`. It's only an MSVC thing. In any case, this should probably be gated on `__ANDROID__` since it isn't supported on any arch there.

qti-jkilpatrick: I'm surprised this works, since the previous branch has `__aarch64__`, which is defined on Android.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -2159,6 +2159,8 @@ TEST_F(QnnHTPBackendTests, QnnContextShareAcrossSessions) {
 #if (defined(__aarch64__) || defined(_M_ARM64)) && \
     !(QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 34))
   GTEST_SKIP() << "HTP weight sharing on ARM64 requires QNN API version >= 2.34.";
+#elif defined(__ANDROID__) && (defined(__aarch64__) || defined(_M_ARM64))
+  GTEST_SKIP() << "Weight sharing on Android devices is disabled";
```
qti-jkilpatrick: Same as above. I'm skeptical of this preprocessor expression.

qti-jkilpatrick: Here and below

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-kromero: Image

I think it's the second condition which was preventing this from working on Android before QAIRT 2.45RC4 - so it sounds like the correct condition should just be

#elif defined(__ANDROID__)

?

qti-jkilpatrick: Yes, though I don't understand why the proceeding `#if` doesn't include `!defined(__ANDROID__)` since `__aarch64__` should be defined on Android.

qti-kromero: I think we thought it was working on Android as per what @quic-calvnguy told me in offline discussion?

quic-calvnguy: We should also check with QAIRT team to understand why weight sharing was only recently disabled:

https://github.com/microsoft/onnxruntime/blame/main/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc#L1667

The feature has been around for about a year now at least.

tirupath-qti: The weight sharing is enabled in the past in AOT workflows. Recently in 2.45, the weight sharing was enabled in Windows JIT path. I doubt the weight sharing is enabled in Android JIT path.

quic-calvnguy: > I think we thought it was working on Android as per what @quic-calvnguy told me in offline discussion?

I think what Jeff is saying is that the warning log implies the guard was meant for Win Arm64 builds but it seemed to also apply to Android via \_\_aarch64__. 

Weight sharing was working on Android before the guards were added AFAIK since the shared context test has been around since before VTCM backup buffer sharing was implemented. The test has never changed from the AOT workflow (create 2 identical models, compile into two models sharing same context bin, execute one of the models at least once), so I'm not sure what changed that would cause failure to create EP.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
quic-calvnguy: Updated to use guard against \_\_ANDROID__ only

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
quic-calvnguy: Code has been updated to guard against \_\_ANDROID__ only

qti-jkilpatrick:
I need to sign off for the week. No need to wait for me to approve.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -37,7 +37,7 @@ download the Qualcomm AI Runtime SDK (QAIRT SDK) from [https://qpm.qualcomm.com/
 ONNX Runtime QNN EP has been built and tested with the following SDK version combinations on Windows:
 | QNN EP Version | QAIRT SDK Version | ONNX Runtime Version |
 |----------------|-------------------|----------------------|
-| v2.0.0         | v2.42.0           | v1.24.1              |
+| v2.0.0         | v2.45.0           | v1.24.4              |
```
kuanyul-qti, line 40: I think we should match QNN EP v2.0.0 with QAIRT SDK v2.42.0, and match QNN EP v2.1.0 with QAIRT SDK v2.45.0.

PR #198, opened 2026-04-02
Shorten CI artifact retention to 18 hours

qti-jkilpatrick:
### Description

Shorten the official retention of temporary CI artifacts to 18 hours.

### Motivation and Context

While the scheduled job continues to run only every 24 hours, this should alleviate problems with the daily release job attempting to publish the previous day's wheels.


qti-jkilpatrick:
Attempting to fix with a more direct solution #199.

PR #200, opened 2026-04-02
Add transpose reshape transpose fusion

quic-muchhsu:
### Description

We are seeing models with `Conv -> (ops) -> Reshape -> (ops) -> Conv` in onnx domain are adding transpose around conv for qnn due to NCHW -> NHWC and give this pattern:
`Conv -> Transpose(to NCHW) -> (ops) -> Reshape -> (ops) -> Transpose(NHWC) -> Conv`
Meanwhile ORT has it's own optimization that pushes Transpose up/down over elementwise nodes, but reshape is not an elementwise node. Which creates this inefficient pattern that those transpose should cancel out each other.
`Conv -> (ops) -> Transpose(to NCHW) -> Reshape -> Transpose(NHWC) -> (ops) -> Conv`

I am adding this change to merge a more generic pattern where `Transpose -> Reshape -> Transpose` are simply merging dims.


onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,413 @@
...
+///  Input [2,3,4]
+///      │
+///      ▼ Transpose perm1=[1,2,0]
+///  [3,4,2]  (B,C,A)
+///      │
+///      ▼ Reshape to [12,2]  (merge B*C=12, keep A=2)
+///  [12,2]
+///      │
+///      ▼ Transpose perm2=[1,0]
+///  [2,12]  (A, B*C)
```
quic-ykota, line 81: High level question: I agree converting from tensor of shape  [2, 3, 4] to [2, 12] should require only one reshape. I am not sure why someone would write the model this way. Irrespective of what the reason is, do you think we should instead address this in onnx-to-onnx transformations stage since this is not runtime specific?

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,413 @@
...
+        final_mapping[i].push_back(static_cast<size_t>(perm1[intermediate_dim]));
+      }
+    }
+  }
+
+  // Check if final_mapping represents a valid "reshape-only" transformation
+  // Valid if: original dimensions appear in strictly increasing order (0, 1, 2, ...)
+  // The order within each group matters for data layout, so we must NOT sort.
+  // E.g., [1, 2] means dim 1 has higher stride than dim 2 (correct for merge)
+  //       [2, 1] means dim 2 has higher stride than dim 1 (different data layout)
```
quic-ykota, line 144: Even under all the possible options for T->R->T, this optimization only kicks in for a small subset where the shape order hasn't been changed yet (I understand if the order has changed, we cant really combine them). Are there a lot of models where this situation might be present?

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: This is not written by someone. Unfortunately, it is an inevitable pattern from onnx2qnn. 
For example: 
Onnx model:
```
conv -> (element wise nodes ) -> reshape -> (element wise nodes ) -> conv
```
It will add transpose before and after conv for NCHW to NHWC.
```
transpose -> conv -> transpose -> (element wise nodes )-> reshape -> (element wise nodes ) -> transpose -> conv -> transpose
```
And ort will push transposes up and down through element wise nodes and we will see this pattern
```
transpose -> conv ->(element wise nodes ) ->  transpose -> reshape ->  transpose -> (element wise nodes ) -> conv -> transpose
```
You can notice there's a `transpose(to NCHW) -> reshape ->  transpose(to NHWC)` in the pattern.
These 2 transposes should cancel out each other, but we are lacking in such fusion.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: I haven't run a full statistic, but I've seen 2 models with this pattern in 2 weeks.
Another reason is qiart has this pattern fusion, and I want to align with qiart

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-ykota: Okay, thanks for explanation. Do you think there are other data movement ops where this fusion can be applied as well? Not saying we need to club all the fusion transforms into one PR, but I would design to keep it a bit generalized.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-ykota: Okay cool. Please share the models separately if you can.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+///      │
+///      ▼ Transpose perm2=[1,0]
+///  [2,12]  (A, B*C)
+/// @param input_shape Shape of the input tensor.
+/// @param perm1 Permutation of the first Transpose.
+/// @param reshape_shape Target shape of the Reshape (output of Reshape).
+/// @param perm2 Permutation of the second Transpose.
+/// @param[out] fused_shape The equivalent reshape shape if fusion is valid.
+/// @return true if fusion is valid, false otherwise.
+bool CanFuseToReshape(
```
quic-ykota, line 82: Does this support a case where [2, 3, 4] is eventually reshaped to [6, 8]? If so, please add a test.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
quic-ykota: Please use Qualcomm copyrights

onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
```diff
@@ -446,4 +446,4 @@ Ort::Status DataTypeCheckForCpuBackend(QnnModelWrapper& qnn_model_wrapper,
 }
 
 }  // namespace qnn
-}  // namespace onnxruntime
+}  // namespace onnxruntime
```
quic-ykota: Nit: Unwanted change

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.h
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
quic-ykota: Copyrights

onnxruntime/test/providers/qnn/qnn_node_group/transpose_reshape_transpose_fusion_test.cc
```diff
@@ -0,0 +1,292 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
quic-ykota: Copyrights

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.h
```diff
@@ -0,0 +1,62 @@
...
+  const OrtNodeUnit* GetTargetNodeUnit() const override { return node_units_[0]; }
+  std::string_view Type() const override { return "TransposeReshapeTransposeFusion"; }
+
+  /// </summary>
+  /// <param name="qnn_model_wrapper">The QNN model wrapper for graph access.</param>
+  /// <param name="transpose1_node_unit">The first Transpose node unit (starting point).</param>
+  /// <param name="node_to_node_unit">Maps Node* to NodeUnit*.</param>
+  /// <param name="node_unit_to_qnn_node_group">Maps NodeUnit* to existing IQnnNodeGroup*.</param>
+  /// <param name="logger">Logger for diagnostics.</param>
+  /// <returns>A TransposeReshapeTransposeFusion if pattern matches, nullptr otherwise.</returns>
```
quic-ykota, line 48: Is this expected to be rendered somewhere? If so it should start with `<summary>` and end with `</summary>` right? I suggest looking at the rendered artifact and fix the description here as needed.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-ykota, line 82: `CanFuseToReshape` has the main functionality. If possible, please add unit tests separately for this method instead of relying on model level testing.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+// Licensed under the MIT License.
+
+#include "core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.h"
```
quic-ykota: Nit: Club this with other similar imports

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+
+constexpr char kAttrTransposePerm[] = "perm";
+constexpr char kOpTranspose[] = "Transpose";
+constexpr char kOpReshape[] = "Reshape";
+
+using MapNodeToNodeUnit = std::unordered_map<const OrtNode*, const OrtNodeUnit*>;
+using MapNodeUnitToGroup = std::unordered_map<const OrtNodeUnit*, const IQnnNodeGroup*>;
+
+/// @brief Get the shape of a tensor from its OrtValueInfo
+std::optional<std::vector<int64_t>> GetTensorShape(const OrtApi& ort_api, const OrtValueInfo* value_info) {
```
quic-ykota: Possible to move this to common utils? This can be used by other classes as well.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+  if (value_info == nullptr) {
+    return std::nullopt;
+  }
+
+  const OrtTypeInfo* type_info = nullptr;
```
minfhong-qti: Use CXX API in `onnxruntime_cxx_api.h` instead, which looks much cleaner.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+    return std::nullopt;
+  }
+
+  size_t dims_count = 0;
+  if (ort_api.GetDimensionsCount(tensor_info, &dims_count) != nullptr) {
+    return std::nullopt;
+  }
+
+  std::vector<int64_t> dims(dims_count);
+  if (ort_api.GetDimensions(tensor_info, dims.data(), dims_count) != nullptr) {
```
minfhong-qti: Should add a check `HasShape` before acquiring.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+  // Only handle SingleNode type (not QDQ groups for now)
+  if (transpose1->UnitType() != OrtNodeUnit::Type::SingleNode) {
+    return std::nullopt;
+  }
+
+  // Get Reshape child
+  const std::array<std::string_view, 1> reshape_types{kOpReshape};
+  const OrtNodeUnit* reshape = GetOnlyChildOfType(qnn_model_wrapper, *transpose1, reshape_types,
+                                                  node_to_node_unit, node_unit_to_qnn_node_group);
+  if (reshape == nullptr || reshape->UnitType() != OrtNodeUnit::Type::SingleNode) {
```
minfhong-qti:
```suggestion
  if (reshape == nullptr) {
```
I believe `GetOnlyChildOfType` inside already checks it.
In addition, should `GetChildNodeUnitAllowQdq` be used to allow QDQ in between?

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+  const size_t input_rank = input_shape.size();
+  const size_t output_rank = output_shape.size();
+
+  // Basic validation
+  if (perm1.size() != input_rank) {
+    return false;
+  }
+  if (perm2.size() != reshape_shape.size()) {
+    return false;
+  }
```
minfhong-qti: Wouldn't these two checks never happen as they violate the ONNX spec?

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    // Handle case where dimensions match exactly
+    if (accumulated_size == target_size) {
+      // Good, this output dim is complete
+    } else if (intermediate_idx < intermediate_shape.size() &&
+               accumulated_size * intermediate_shape[intermediate_idx] == target_size) {
+      // Need one more dimension
+      reshape_mapping[out_idx].push_back(intermediate_idx);
+      intermediate_idx++;
```
minfhong-qti: Did not get the case for this check. Can you provide an example?

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+    } else {
+      // Reshape is not a simple merge - can't fuse
+      return false;
+    }
+
+    // If reshape_mapping is empty for this output, add current intermediate index
```
minfhong-qti: When will the mapping be empty? When `target_size=1`?
In that case, if `intermediate_shape[0] != 1`, shouldn't it be rejected above?

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+  // E.g. perm2=[1,0], reshape_mapping=[[0,1], [2]], perm1=[1,2,0]
+  // ->
+  //      final_mapping[0] = [perm1[d] for d in reshape_mapping[1]] = [perm1[2]] = [0]
+  //      final_mapping[1] = [perm1[d] for d in reshape_mapping[0]] = [perm1[0], perm1[1]] = [1, 2]
+  std::vector<std::vector<size_t>> final_mapping(output_rank);
+  for (size_t i = 0; i < output_rank; ++i) {
+    size_t src_idx = static_cast<size_t>(perm2[i]);
+    if (src_idx >= reshape_mapping.size()) {
+      return false;
+    }
```
minfhong-qti: Same question regarding violating ONNX spec. As well for LN194.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,414 @@
...
+
+  const OrtNodeUnit* transpose1 = pattern->at(0);
+  const OrtNodeUnit* reshape = pattern->at(1);
+  const OrtNodeUnit* transpose2 = pattern->at(2);
+
+  const OrtApi& ort_api = qnn_model_wrapper.GetOrtApi();
+
+  // Get input shape of first Transpose
+  size_t num_t1_inputs = 0;
+  RETURN_DEFAULT_IF_API_FAIL(ort_api.Node_GetNumInputs(&transpose1->GetNode(), &num_t1_inputs), ort_api, nullptr);
```
minfhong-qti: Use CXX API here as well.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: If you meant a case like [2, 3, 4] → [6, 4] (merging the first two dimensions), yes, this is supported. I can add a  test for it.

quic-muchhsu: CanFuseToReshape is intentionally defined in an anonymous namespace as an internal helper function - it's not intended to be reused by other files or components. The existing model-level tests already cover the core fusion logic through various scenarios (fusable patterns, non-fusable patterns, identity transposes, 4D tensors, etc.).
If we really feel like to have a unit test for it, we'll need to exposing this function (and currently there's no other usecase for it), which seems unnecessary for a private implementation detail.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: No problem

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: Yes, sounds good.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: Yes, it's moved to utils.h/utils.cc and refactored to use QnnModelWrapper& signature for consistency

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: Good point

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: Good call, removing dead code

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: You're right - the mapping is empty when target_size == 1, and if intermediate_shape[idx] != 1 in that case, it should be rejected as an invalid reshape pattern.
                                                 
I'll add a check:
```
  if (reshape_mapping[out_idx].empty()) {                                                                                                 
    if (intermediate_idx >= intermediate_shape.size() ||
        intermediate_shape[intermediate_idx] != 1) {
      return false;
    }
    reshape_mapping[out_idx].push_back(intermediate_idx);
    intermediate_idx++;
  }
```

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
quic-muchhsu: You're right - both checks guard against invalid ONNX models. I'll remove them.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
minfhong-qti: Reopen this thread. Please confirm again whether LN112-LN116 should be removed as I did not see change in the latest commit.

onnxruntime/core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.cc
```diff
@@ -0,0 +1,361 @@
...
+  std::vector<const OrtValueInfo*> t1_inputs(num_t1_inputs);
+  RETURN_DEFAULT_IF_API_FAIL(ort_api.Node_GetInputs(&transpose1->GetNode(), t1_inputs.data(), t1_inputs.size()),
+                             ort_api, nullptr);
+
+  auto input_shape = GetTensorShape(qnn_model_wrapper, t1_inputs[0]);
+  if (!input_shape.has_value()) {
+    return nullptr;
+  }
```
minfhong-qti:
```suggestion
  auto input_shape = transpose1->Inputs()[0].shape;
  if (!input_shape.has_value()) {
    return nullptr;
  }
```
As well for others. I believe no need to define `GetTensorShape` anymore.

quic-muchhsu: GetTensorShape is removed from utils as well.

minfhong-qti (approved):

quic-ykota (approved):

qti-chuteng (approved):

PR #205, opened 2026-04-04
[QNN EP] Enablement of 64-bit UDMA mode in QNN EP

qti-ashwshan:
### Description
- Ported enablement of extended UDMA mode with ABI EP adaptations including unit test gated to v81+ hardware.
- Adds extended_udma provider option that configures the mode on HTP context creation, enabling far-mapping of weights and spill/fill buffers for v81+ devices.


tirupath-qti (approved):
LKGM

PR #209, opened 2026-04-06
Relax BN node group selection

qti-yuduo:
### Description
- Relax `OrtBatchNormalizationNodeGroupSelector` to accept BN nodes with only 2 DQ inputs (x and scale), down from the previous minimum of 3
- This is common: quantized model contains BN with DQ on x (UINT16) and scale (INT8) only — bias/mean/var remain float initializers
- Without this fix, all BN nodes fall to CPU ("dynamic scale" error), causing quantized models to split into many subgraphs
- Add HTP accuracy tests for U16+S8 and U8+S8 with float params pattern


qti-yuduo:
=> Draft for now, will take a closer look at CI failures to see if it's related.

quic-muchhsu (approved):
This change make sense to me. Can we link a issue to AIMET as well? I remember there are some discussion around AIMET enabling quantizers on bias.

qti-jkilpatrick (approved):

PR #213, opened 2026-04-07
Add DX12 shared memory support for QNN-EP with GPU backend.

vjatoth-qti:
### Description
* Add DX12 shared memory allocator to the EP and associated provider option
  `"enable_dx12_shared_memory_allocator"`.
* This allocator leverages `QNN_MEM_TYPE_DX12` mem-handles added in QAIRT 2.46.

### Motivation and Context
* Improve LLM inferencing speed by eliminating unnecessary CPU<->GPU copies of
  KV cache tensors. See the corresponding change to onnxruntime-genai:
  [onnxruntime-genai/pull/2105](https://github.com/microsoft/onnxruntime-genai/pull/2105).


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2291,12 +2310,22 @@ OrtStatus* ORT_API_CALL QnnEp::CreateAllocatorImpl(_In_ OrtEp* this_ptr,
   *allocator = nullptr;
   QnnEp* ep = static_cast<QnnEp*>(this_ptr);
 
-  if (ep->IsHtpSharedMemoryAllocatorAvailable()) {
+  ep->qnn_backend_manager_->LoadBackend();
```
qti-mattsinc: Loading the backend here instead of in `LoadBackend` to get the device type for the allocator is one option. Alternatively, we could have one `QnnSharedMemoryAllocator` that gets returned in `CreateAllocatorImpl`, which later delegates to `HtpSharedMemoryAllocator` or `Dx12SharedMemoryAllocator` as appropriate. Edit: No longer applicable

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -2546,6 +2546,166 @@ TEST_F(QnnHTPBackendTests, ExtendedUdmaModeTest) {
...
+  }
+
+  // Verify the QnnGpuShared allocator is accessible and functional.
+  Ort::MemoryInfo info_qnn_gpu_shared("QnnGpuShared",
+                                      OrtMemoryInfoDeviceType_GPU,
+                                      OrtDevice::VendorIds::QUALCOMM,
+                                      0,
+                                      OrtDeviceMemoryType_HOST_ACCESSIBLE,
+                                      0,
+                                      OrtDeviceAllocator);
```
qti-mattsinc: Using the constructor that chooses `CreateMemoryInfo_V2` internally, so that "QnnGpuShared" is accepted. The other constructor uses the V1 `CreateMemoryInfo`, which allows only a few allocator names (incl. "QnnHtpShared", but not "QnnGpuShared").

[src/models/model.cpp:453-454](https://github.com/microsoft/onnxruntime-genai/blob/139001aade62ea90b3d17fe60a214b57a3fb71a1/src/models/model.cpp#L453-L454) in oga will probably  also need to be updated to use the V2 memory info API in order to not throw an exception for "QnnGpuShared".

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -2546,6 +2546,166 @@ TEST_F(QnnHTPBackendTests, ExtendedUdmaModeTest) {
...
+  void* p = qnn_gpu_shared_allocator.Alloc(1024);
+  ASSERT_NE(p, nullptr);
+  qnn_gpu_shared_allocator.Free(p);
+
+  auto mem_allocation = qnn_gpu_shared_allocator.GetAllocation(1024);
+  ASSERT_NE(mem_allocation.get(), nullptr);
+  ASSERT_EQ(mem_allocation.size(), size_t{1024});
+}
+
+TEST_F(QnnGPUBackendTests, io_binding_qnn_gpu_shared) {
```
qti-mattsinc, line 2990: Noticing that only the output tensor uses DX12 mem for this test, and the input resides on CPU and is bound to QNN as a client buf instead of a mem handle. This is the same behavior as in the equivalent HTP unit test, so it's not necessarily a problem, but we'll want to make sure inputs tensors use dx12 mem in oga.

qti-mattsinc: Investigated this further, and it seems like input residing on CPU is expected behavior?

`CopyOneInputAcrossDevices` is called in [IOBinding::BindInput](https://github.com/microsoft/onnxruntime/blob/a2742d0f945384eb1d83b38917f9f7a49569cad5/onnxruntime/core/session/IOBinding.cc#L33), which introduces a copy to put input tensors on the device determined in [SequentialExecutionPlan::SetLocation](https://github.com/microsoft/onnxruntime/blob/a2742d0f945384eb1d83b38917f9f7a49569cad5/onnxruntime/core/framework/allocation_planner.cc#L807), which is the CPU. An EP can override [IExecutionProvider::GetOrtDeviceByMemType](https://github.com/microsoft/onnxruntime/blob/a2742d0f945384eb1d83b38917f9f7a49569cad5/include/onnxruntime/core/framework/execution_provider.h#L424-L429) to change the default device for tensors, but AFAICT this was intented as an experimental feature for intermediate tensors (https://github.com/microsoft/onnxruntime/pull/24371), and none of the EPs seem to be using this function to keep input tensors on device memory. And there's also the issue that this API is inaccessible to plugin EPs anyway...

So the end result is that the model inputs end up on CPU memory instead of DX12 shared memory no matter how the `Ort::Value` is created client-side, which would be a problem as the inputs would not be zero-copy. I am probably missing something here. It would be a good idea to debug an equivalent WebGPU unit test to see if the inputs get copied to CPU and, if not, how that issue gets avoided.

qti-mattsinc: Forcing `PluginExecutionProvider` in ORT core to return the GPU device memory info in `GetOrtDeviceByMemType` actually does work, and I see mem handles for input tensors used both in the unit test and in a full model.

~~To do it properly, we would want to add a `GetOrtDeviceByMemType` function to the onnxruntime plugin EP API which `PluginExecutionProvider` would call, although that would require proposing the new API and getting it into ORT core.~~ EDIT: found a better way, although it will still require an ORT core change: [dev/mattsinc/plugin-ep-mem-type](https://github.com/CodeLinaro/onnxruntime/tree/dev/mattsinc/plugin-ep-mem-type).

onnxruntime/test/providers/qnn/qnn_basic_test.cc
qti-mattsinc: This branch has the oga changes: [CodeLinaro/onnxruntime-genai/tree/dev/mattsinc/gpu-zero-copy](https://github.com/CodeLinaro/onnxruntime-genai/tree/dev/mattsinc/gpu-zero-copy).

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -941,6 +941,25 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
   }
 
+#ifdef _WIN32
+  static const std::string QNN_DX12_SHARED_MEMORY_ALLOCATOR_ENABLED = "enable_dx12_shared_memory_allocator";
+  if (ParseBoolOption(ort_api,
+                      session_options_,
+                      FormatEPConfigKey(QNN_DX12_SHARED_MEMORY_ALLOCATOR_ENABLED),
+                      false,
+                      logger_)) {
+    if (!context_cache_enabled_) {
```
johnpaultaken: Does past-present buffer sharing work only if not running from context-cache model ?

onnxruntime/core/providers/qnn/qnn_allocator.cc
```diff
@@ -287,4 +287,249 @@ Ort::Status HtpSharedMemoryAllocator::AddAllocationCleanUpForThisAllocator(void*
...
+      D3D12_RESOURCE_STATE_COMMON,
+      nullptr,
+      IID_PPV_ARGS(&d3d12_resource));
+  if (FAILED(hr) || d3d12_resource == nullptr) {
+    ORT_CXX_API_THROW("Dx12SharedMemoryAllocator: CreateCommittedResource failed.", ORT_EP_FAIL);
+  }
+
+  // Map the resource to get a CPU-visible pointer (persistent mapping for UPLOAD heap).
+  void* mapped_ptr = nullptr;
+  hr = d3d12_resource->Map(0, nullptr, &mapped_ptr);
```
johnpaultaken: Do we need to map it to cpu ? KV cache don't need to be accessed by the CPU.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
qti-mattsinc: The ORT core change is actually not needed for ORT-genai use cases. After looking again, the KV cache inputs get setup correctly because of past-present buffer sharing. Other inputs like the input_ids do not, but that's OK as the KV cache is what actually needs to be zero copy.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -918,6 +918,25 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     model_settings_.htp_shared_memory = true;
   }
 
+#ifdef _WIN32
+  static const std::string QNN_DX12_SHARED_MEMORY_ALLOCATOR_ENABLED = "enable_dx12_shared_memory_allocator";
```
qti-mattsinc: I wonder if we should call this "enable_gpu_shared_memory_allocator" to better mimic the HTP option. Also, if we later enable zero-copy from ORT on non-windows platforms through DMABuf, then "enable_gpu_shared_memory_allocator" would still be an appropriate name. Thoughts?

Also, we'll want to update the MD file with the provider options documented when we have the name finalized.

onnxruntime/core/providers/qnn/qnn_allocator.cc
qti-mattsinc, line 1: A lot of the code in the file is duplicated between HTP and GPU and can be cleaned up a bit (e.g. combine the two AllocationTracker classes with one class templated over the `Allocator`). Edit: Done

onnxruntime/core/providers/qnn/builder/qnn_context_mem_handle_manager.cc
qti-mattsinc, line 131: We should add an informative error message in the event that the DX12 allocator is requested without recent enough gfx drivers to use `QNN_MEM_TYPE_DX12`. We could do that by checking if `memRegister` returns `QNN_MEM_ERROR_MAPPING` here. Edit: Done

onnxruntime/core/providers/qnn/qnn_allocator.cc
```diff
@@ -287,4 +287,273 @@ Ort::Status HtpSharedMemoryAllocator::AddAllocationCleanUpForThisAllocator(void*
...
+  buffer_desc.DepthOrArraySize = 1;
+  buffer_desc.MipLevels = 1;
+  buffer_desc.Format = DXGI_FORMAT_UNKNOWN;
+  buffer_desc.SampleDesc.Count = 1;
+  buffer_desc.Layout = D3D12_TEXTURE_LAYOUT_ROW_MAJOR;
+  buffer_desc.Flags = D3D12_RESOURCE_FLAG_NONE;
+
+  D3D12_HEAP_PROPERTIES heap_props = {};
+  heap_props.Type = D3D12_HEAP_TYPE_UPLOAD;
+  heap_props.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
```
johnpaultaken, line 325: Please check if using D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE  gives better performance.

johnpaultaken (approved):

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mattsinc: I've test with a Llama1B context model, and it works. The original check against `!context_cache_enabled_` that exists for the HTP shared memory allocator might be an HTP-specific restriction.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mattsinc: The provider option is removed now so this is resolved

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -118,7 +123,8 @@ class QnnEp : public OrtEp, public ApiPtrs {
   void ParseHtpGraphFinalizationOptimizationMode(const std::string& htp_graph_finalization_opt_mode_string,
                                                  const Ort::Logger& logger);
 
-  bool IsHtpSharedMemoryAllocatorAvailable() const { return rpcmem_library_ != nullptr; }
+  bool IsHtpSharedMemoryAllocatorAvailable() const { return model_settings_.htp_shared_memory; }
```
qti-mattsinc: The reason for the change to `IsHtpSharedMemoryAllocatorAvailable` is that `rpcmem_library_` also gets loaded if file-mapped weights are being used. That seems a bit suspicious to me, but perhaps enabling file mapped weights is meant to imply using the htp shared memory.

I'm not sure, so I'm making this comment to bring it to the attention of someone on the HTP side, who would know better and can say whether we should change it or leave it as-is.

qti-mattsinc:
Marking this change as ready for review. ~~However, it depends on 2.47 headers, so it can't be merged into main / won't pass CI until the EP upgrades to 2.47.~~ (nvm: 2.46 was what was needed, and the EP has already upgraded to 2.46).

onnxruntime/core/providers/qnn/builder/qnn_context_mem_handle_manager.cc
```diff
@@ -54,42 +55,79 @@ Ort::Status QnnContextMemHandleManager::GetOrRegister(void* shared_memory_addres
...
+      std::ostringstream oss1;
+      oss1 << "Registering QNN mem handle for context: " << context_
+           << ", shared memory (address: " << shared_memory_address
+           << ", offset: " << shared_memory_info.offset
+           << ", fd: " << shared_memory_info.fd
+           << ")";
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, oss1.str().c_str());
+    }
+#ifdef _WIN32
+    else if (IsGpuBackend(qnn_backend_type_)) {
```
johnpaultaken: why not make this allocator available for any device based on enable_d3d12_allocator provider option ?

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.h → IsHtpSharedMemoryAllocatorAvailable()`
Changing the predicate from `rpcmem_library_ != nullptr` to `model_settings_.htp_shared_memory` decouples it from actual availability. When `context_cache_enabled_=true` and `enable_htp_shared_memory_allocator=1`, the EP will not load `rpcmem_library_` (constructor only loads it when `!context_cache_enabled_`); if `enable_file_mapped_weights` is also explicitly disabled, the subsequent `CreateAllocatorImpl` will pass a null `rpcmem_library_` to `HtpSharedMemoryAllocator`'s constructor, which throws inside a `noexcept` entry and calls `std::terminate`. Please tie the predicate to actual state.
```suggestion
  bool IsHtpSharedMemoryAllocatorAvailable() const { return model_settings_.htp_shared_memory && rpcmem_library_ != nullptr; }
```

**[M-2]** `onnxruntime/core/providers/qnn/qnn_allocator.cc → Dx12SharedMemoryAllocator::AllocImpl`
On the global-tracker `RegisterAllocation` failure path, the function throws without unwinding: the D3D12 resource is still mapped with refcount=1, and `allocator->allocations_` already holds an entry for `mapped_ptr`. Please erase the local map entry and `Unmap+Release` the resource before throwing — or hold the resource as `Microsoft::WRL::ComPtr<ID3D12Resource>` so failure paths unwind automatically.

**[M-3]** `onnxruntime/core/providers/qnn/qnn_allocator.h → Dx12SharedMemoryAllocator`
Storing `const Ort::Logger& logger_` initialized from `ep->logger_` re-introduces the dangling-logger anti-pattern that PR #163 fixed for `HtpSharedMemoryAllocator`. The QNN allocator can outlive the session in cleanup callbacks. Mirror `HtpSharedMemoryAllocator`: drop the logger constructor argument and use `OrtLoggingManager::GetDefaultLogger()`.

---

### 🔵 Minor

**[N-1]** `docs/execution_providers/QNN-ExecutionProvider.md:L201`
The new table row for `enable_dx12_shared_memory_allocator` is missing the trailing `|`, breaking markdown rendering.
```suggestion
|'1'|Enable the QNN DX12 shared memory allocator. Requires a D3D12 capable system, and only available on Windows.|
```

**[N-2]** `onnxruntime/core/providers/qnn/qnn_execution_provider.h:L29-L32`
The new `#include <d3d12.h>` / `#include <wrl/client.h>` block is unused — `QnnEp` only adds a `bool enable_dx12_shared_memory_allocator_` member. The actual D3D12 / `ComPtr` usage lives in `qnn_allocator.h`. Remove these heavy includes from the EP header to keep compile-time and transitive-include surface small.

**[N-3]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc → io_binding_qnn_gpu_shared` (also `DISABLED_io_binding_qnn_gpu_shared_offset`)
`RegisteredEpDeviceUniquePtr registered_ep_device;` is declared in both the outer test scope and again inside the inner block, shadowing the outer. The outer is dead code. Remove one of the two declarations to match the existing `io_binding_qnn_htp_shared` style.

**[N-4]** `onnxruntime/core/providers/qnn/builder/qnn_model.cc → BindQnnTensorMemoryToOrtValueMemory`
Adding `OrtMemoryInfoDeviceType_GPU` to `uses_shared_memory` is currently dead code: the `QnnHtpShared` memory info is `OrtMemoryInfoDeviceType_CPU`, and the DX12 path reuses it. If this is preparation for a future GPU memory pathway, please add a comment; otherwise drop it to keep the diff aligned with the PR title.

onnxruntime/core/providers/qnn/builder/qnn_context_mem_handle_manager.cc
qti-mattsinc: Good catch, forgot to update this part

onnxruntime/core/providers/qnn/qnn_allocator.cc
qti-mattsinc: Using `D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE` causes `CreateCommittedResource` to fail:

```
hr               : 0x80070057 (The parameter is incorrect.) [Type: HRESULT]
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mattsinc: Ah, the HTP allocator loads another DLL, which is not required if in the context cache generation stage (`context_cache_enabled`), which is why that check is there. So this isn't required for the Dx12 allocator and anyway it's been removed

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -50,6 +50,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
                                                      OrtDeviceEpIncompatibilityDetails* details) noexcept;
 
   friend struct GenieNodeComputeInfo;
+  friend struct QnnEpFactory;
```
qti-mattsinc: The EP factory needs to know about the EP's allocator decision. I don't think it quite makes sense to expose that among the currently limited public members, so just adding the factory as a friend.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2694,5 +2715,123 @@ Ort::Status QnnBackendManager::GetGraphInfoAndBinVersion(QnnSystemContext_Handle
...
+        &buffer_desc,
+        D3D12_RESOURCE_STATE_COMMON,
+        nullptr,
+        IID_PPV_ARGS(&d3d12_resource));
+    if (FAILED(hr) || d3d12_resource == nullptr) {
+      supported = false;
+    }
+  }
+
+  if (supported) {
```
johnpaultaken, line 2964: Can we avoid this if (supported) check all over the function? Why not remove var supported and exit early ie return false  instead of 'supported = false'.  I know you need to cleanup context. Exceptions help write clean code. Is a try catch block allowed by coding standards here ?

johnpaultaken (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mattsinc: The function was more complex originally hence the reason for the supported check to avoid duplicate clean up code on early returns. But now the function is simpler and can avoid this. Done is latest commit

qti-mattsinc:
@qti-chuteng, your previous comments have been addressed. Can you take another look?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -500,7 +505,7 @@ void QnnLogging(const char* format,
   ORT_CXX_LOG(OrtLoggingManager::GetDefaultLogger(), ORT_LOGGING_LEVEL_VERBOSE, stream.str().c_str());
 }
 
-Ort::Status QnnBackendManager::InitializeQnnLogCommon(const QNN_INTERFACE_VER_TYPE& interface,
+Ort::Status QnnBackendManager::InitializeQnnLogCommon(const QNN_INTERFACE_VER_TYPE& qnn_interface,
```
qti-mattsinc, line 513: Renaming `interface` to `qnn_interface` to avoid collisions with the `interface` macro in combaseapi.h.

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:L2956`
`IsDx12SharedMemoryAllocatorSupported()` creates a temporary QNN context via `contextCreate`, but on the `memRegister` failure path it does `return false` without calling `contextFree(context, nullptr)` — the context is only freed on the success path. Because the negative result is also not cached (N-2), every session created on a DX12-unsupported machine leaks one QNN context.
```suggestion
  if (register_result != QNN_SUCCESS) {
    qnn_interface_.contextFree(context, nullptr);
    return false;
  }
```

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/qnn_allocator.cc:L364`
`ORT_CXX_LOGF` uses `%lu` for `dx12_info.offset` / `total_size`, which are `uint64_t`. This block compiles only under `_WIN32` (LLP64), where `unsigned long` is 32-bit — a printf format/variadic type mismatch. Use `%llu` or `PRIu64` from `<cinttypes>`.

**[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:L2956–L2964`
`dx12_shared_memory_allocator_supported_` (a `std::optional<bool>`) is only assigned `= true` on success; every failure path returns without writing it, so the cache is ineffective and the full probe (LoadBackend / InitializeBackend / CreateDevice / D3D12CreateDevice / contextCreate / memRegister) re-runs each call. Write `= false` before each failure return, or restructure to a single `bool result` written once at the end (also centralizes the M-1 cleanup).

**[N-3]** `onnxruntime/core/providers/qnn/qnn_execution_provider.h:L126`
`bool IsDx12SharedAllocatorSupported() const;` is declared but never defined or called (the real check lives in `QnnBackendManager::IsDx12SharedMemoryAllocatorSupported()`). Dead declaration — remove it along with its `#if defined(_WIN32)` guard.

qti-mattsinc:
@qti-chuteng, thanks for reviewing. Your previous comments were addressed. PTAL

Also, force-pushing now to rebase and squash

qti-chuteng:
**[C-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc` — DX12 allocator path

The DX12 allocator setup path never sets `model_settings_.htp_shared_memory = true`, but the HTP path does. `QnnModelWrapper::SetTensorMemTypeFromSettings()` depends on this flag to assign `QNN_TENSORMEMTYPE_MEMHANDLE` to graph I/O tensors at compile time. Without it, the DX12 path compiles tensors with `QNN_TENSORMEMTYPE_RAW`, while `BindQnnTensorMemoryToOrtValueMemory()` at execution time attempts to use `QNN_TENSORMEMTYPE_MEMHANDLE` based on `OrtDeviceMemoryType_HOST_ACCESSIBLE`. This compile-time/runtime mismatch may silently defeat the zero-copy feature.

```cpp
// Fix: add this line in the DX12 enable path (symmetric with HTP path)
if (qnn_backend_manager_->IsDx12SharedMemoryAllocatorSupported()) {
  qnn_allocator_type_ = qnn::QnnAllocatorType::DX12_SHARED;
  model_settings_.htp_shared_memory = true;  // ADD THIS
}
```

---

**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc` — `IsDx12SharedMemoryAllocatorSupported()`

This function calls `InitializeBackend()` and `CreateDevice()` as side effects (documented in a comment, good) — but it's unclear if these are fully idempotent when `SetupBackend()` later runs the same initialization. `LoadBackend()` now has a guard (`if (backend_lib_handle_)`), and `InitializeQnnLogCommon()` has `if (log_handle)` — does `InitializeBackend()` also have such a guard? Please confirm and add a comment on the function documenting the side effects and idempotency guarantee.

---

**[M-2]** `onnxruntime/core/providers/qnn/qnn_allocator.cc` — `FreeImpl` log format string

`Dx12AllocationInfo::offset` and `total_size` are `uint64_t`, but the `FreeImpl` log uses `%lu` (32-bit on Windows). Should be `%llu`, consistent with `AllocImpl` which correctly uses `%llu`.

```cpp
// Before (FreeImpl):
"\\n  offset             = %lu"
"\\n  total_size         = %lu"

// After:
"\\n  offset             = %llu"
"\\n  total_size         = %llu"
```

---

**[M-3]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc` — `CreateAllocatorImpl`

The DX12 allocator construction (`std::make_unique<Dx12SharedMemoryAllocator>(memory_info)`) has no try-catch. If `D3D12CreateDevice` fails here (race condition after the probe succeeded), `ORT_CXX_API_THROW` propagates uncaught, potentially causing a silent failure rather than a visible `OrtStatus` error. Wrap with try-catch or add a constructor that returns a status.

---

**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc` — `IsDx12SharedMemoryAllocatorSupported()`

`D3D12CreateDevice(nullptr, ...)` executes without an `if (supported)` guard — unlike every other step in the function. Even when prior QNN initialization has already failed, this D3D12 call still runs. Please wrap it consistently.

---

**[N-2]** `onnxruntime/core/providers/qnn/qnn_allocator.h` — `Dx12AllocationInfo::resource`

The field's comment says "not AddRef'd; lifetime tied to allocator record" but does not explicitly state the contract: this pointer becomes invalid after the allocation is freed. Current callers are safe (use immediately and don't retain), but please strengthen the comment to prevent future UAF bugs.

---

**[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc`

Typo in new guard: `// Log already intialized` → `// Log already initialized`

---

**[N-4]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc`

Per project convention, please document kill-test fence results in the PR description's Validation section for the new GPU shared memory tests (especially `io_binding_qnn_gpu_shared`). Did the test actually fail when the fix was temporarily disabled?

qti-mattsinc:
> **[C-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc` — DX12 allocator path
> 
> The DX12 allocator setup path never sets `model_settings_.htp_shared_memory = true`, but the HTP path does. `QnnModelWrapper::SetTensorMemTypeFromSettings()` depends on this flag to assign `QNN_TENSORMEMTYPE_MEMHANDLE` to graph I/O tensors at compile time. Without it, the DX12 path compiles tensors with `QNN_TENSORMEMTYPE_RAW`, while `BindQnnTensorMemoryToOrtValueMemory()` at execution time attempts to use `QNN_TENSORMEMTYPE_MEMHANDLE` based on `OrtDeviceMemoryType_HOST_ACCESSIBLE`. This compile-time/runtime mismatch may silently defeat the zero-copy feature.

This is intentional. The existing changes are sufficient for the GPU use-case, and AFAICT the `model_settings_.htp_shared_memory = true`  has some behavior useful for the HTP that we don't necessarily want to mimic here. Correct me if I'm wrong.


> **[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc` — `IsDx12SharedMemoryAllocatorSupported()`
> 
> This function calls `InitializeBackend()` and `CreateDevice()` as side effects (documented in a comment, good) — but it's unclear if these are fully idempotent when `SetupBackend()` later runs the same initialization. `LoadBackend()` now has a guard (`if (backend_lib_handle_)`), and `InitializeQnnLogCommon()` has `if (log_handle)` — does `InitializeBackend()` also have such a guard? Please confirm and add a comment on the function documenting the side effects and idempotency guarantee.

The side-affects are handled correctly. In a follow up, we could simplify this a bit. Note that `QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl` has a similar situation where it could run before the EP is created but handles it slightly differently (by creating a temp QNN EP instance).

> **[M-2]** `onnxruntime/core/providers/qnn/qnn_allocator.cc` — `FreeImpl` log format string
> ...

Done

> **[M-3]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc` — `CreateAllocatorImpl`
> 
> The DX12 allocator construction (`std::make_unique<Dx12SharedMemoryAllocator>(memory_info)`) has no try-catch. If `D3D12CreateDevice` fails here (race condition after the probe succeeded), `ORT_CXX_API_THROW` propagates uncaught, potentially causing a silent failure rather than a visible `OrtStatus` error. Wrap with try-catch or add a constructor that returns a status.

Done

> **[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc` — `IsDx12SharedMemoryAllocatorSupported()`
> 
> `D3D12CreateDevice(nullptr, ...)` executes without an `if (supported)` guard — unlike every other step in the function. Even when prior QNN initialization has already failed, this D3D12 call still runs. Please wrap it consistently.

Done

> **[N-2]** `onnxruntime/core/providers/qnn/qnn_allocator.h` — `Dx12AllocationInfo::resource`
> 
> The field's comment says "not AddRef'd; lifetime tied to allocator record" but does not explicitly state the contract: this pointer becomes invalid after the allocation is freed. Current callers are safe (use immediately and don't retain), but please strengthen the comment to prevent future UAF bugs.

I think the current comment is descriptive enough.

> **[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc`
> 
> Typo in new guard: `// Log already intialized` → `// Log already initialized`

Done

> **[N-4]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc`
> 
> Per project convention, please document kill-test fence results in the PR description's Validation section for the new GPU shared memory tests (especially `io_binding_qnn_gpu_shared`). Did the test actually fail when the fix was temporarily disabled?

I don't think this is a relevant nit. But yes, the added tests are passing only after adding the allocator.

qti-mattsinc:
@qti-chuteng, gentle reminder to review. Your latest comments have been addressed; is there anything blocking your approval?

tirupath-qti (approved):

tirupath-qti (approved):

onnxruntime/core/providers/qnn/qnn_execution_provider.h
tirupath-qti: @quic-calvnguy  Can you please review this ?

qti-mattsinc: FYI I ended up removing this part in a later patch. So the `IsHtpSharedMemoryAllocatorAvailable` check is unchanged in the PR and this comment is stale. (Though it might still be good for someone to check if `IsHtpSharedMemoryAllocatorAvailable` on main does what it's expected to do if also using file-mapped weights.)

qti-mattsinc:
Thanks Tiru for reviewing. I'll give some time for Jerry to take a final look if he wants to and plan to merge it on Monday.

PR #217, opened 2026-04-07
Add support for NonZero

quic-ykota:
### Description
Adds support for NonZero Op

### Motivation and Context
- NonZero op's output is of dynamic shape. This implementation sets the output shape to be [rank, num_inputs]. Current test framework also does not handle dynamic shapes well so the PR updates them (See: is_dynamic_output) or relaxes the checks in one place.
- HTP's output would be in transposed form of shape [num_inputs, rank]. So, this op builder explicitly adds a transpose to match with onnx format [rank, num_inputs]
- This op only supports a few combinations of inputs and outputs (See: https://docs.qualcomm.com/doc/80-63442-10/topic/HtpOpDefSupplement.html#nonzero). IsOpSupported, ProcessInputs account for the subset of combinations supported by HTP.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
@@ -96,23 +96,9 @@ Ort::Status NonZeroOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
     const std::string cast_output_name = utils::UniqueNameGenerator().New(original_input_name, "_cast_fp16");
 
     // Create fp16 intermediate tensor.
-    QnnTensorWrapper cast_output_tensorwrapper(cast_output_name,
-                                               QNN_TENSOR_TYPE_NATIVE,
-                                               QNN_DATATYPE_FLOAT_16,
-                                               QnnQuantParamsWrapper(),
-                                               std::vector<uint32_t>(input_info.shape));
-    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(cast_output_tensorwrapper)),
-                  "Failed to add fp16 cast tensor.");
-
-    // Add Cast node: fp32 -> fp16.
-    const std::string cast_node_name = cast_output_name + "_cast_node";
-    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(cast_node_name,
-                                                  QNN_OP_PACKAGE_NAME_QTI_AISW,
-                                                  QNN_OP_CAST,
-                                                  {original_input_name},
-                                                  {cast_output_name},
-                                                  {}),
-                  "Failed to add Cast fp32->fp16 node");
+    RETURN_IF_ERROR(qnn_model_wrapper.AddCastNode(cast_output_name + "_cast_node", original_input_name,
+                                                  cast_output_name, QNN_TENSOR_TYPE_NATIVE, QNN_DATATYPE_FLOAT_16,
+                                                  QnnQuantParamsWrapper(), std::vector<uint32_t>(input_info.shape), do_op_validation));
 
     // Replace the input name with the casted fp16 tensor.
     input_names.back() = cast_output_name;
@@ -200,7 +186,8 @@ Ort::Status NonZeroOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_m
                                                   QNN_OP_TRANSPOSE,
                                                   {nonzero_out_name},
                                                   {transpose_out_name},
-                                                  std::move(transpose_params)),
+                                                  std::move(transpose_params),
+                                                  do_op_validation),
                   "Failed to add Transpose node.");
 
     // Cast: int32 -> int64 for graph output.
@@ -217,7 +204,8 @@ Ort::Status NonZeroOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_m
                                                   QNN_OP_CAST,
                                                   {transpose_out_name},
                                                   {output_name},
-                                                  {}),
+                                                  {},
+                                                  do_op_validation),
                   "Failed to add Cast node.");
   } else {
     // Transpose output is the final native tensor.
@@ -234,7 +222,8 @@ Ort::Status NonZeroOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_m
                                                   QNN_OP_TRANSPOSE,
                                                   {nonzero_out_name},
                                                   {output_name},
-                                                  std::move(transpose_params)),
+                                                  std::move(transpose_params),
+                                                  do_op_validation),
                   "Failed to add Transpose node.");
   }
 
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
@@ -1,0 +1,249 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include <unordered_set>
+
+#include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/op_builder_factory.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+// NonZero op builder.
+// Output shape is [rank_of_input, num_elements], where num_elements is the total number of
+// elements in the input (the maximum possible number of non-zero elements).
+class NonZeroOpBuilder : public BaseOpBuilder {
+ public:
+  NonZeroOpBuilder() : BaseOpBuilder("NonZeroOpBuilder") {}
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(NonZeroOpBuilder);
+
+  // Override IsOpSupported to check backend type, consumer ops, and supported input types.
+  Ort::Status IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                            const OrtNodeUnit& node_unit,
+                            const Ort::Logger& logger) const override ORT_MUST_USE_RESULT;
+
+ protected:
+  Ort::Status ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
+                            const OrtNodeUnit& node_unit,
+                            const Ort::Logger& logger,
+                            std::vector<std::string>& input_names,
+                            bool do_op_validation) const override ORT_MUST_USE_RESULT;
+
+  Ort::Status ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                          const OrtNodeUnit& node_unit,
+                                          std::vector<std::string>&& input_names,
+                                          const Ort::Logger& logger,
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+};
+
+Ort::Status NonZeroOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                            const OrtNodeUnit& node_unit,
+                                            const Ort::Logger& logger) const {
+  // NonZero is only supported on HTP backend.
+  RETURN_IF_NOT(qnn_model_wrapper.GetQnnBackendType() == QnnBackendType::HTP,
+                "NonZero is only supported on HTP backend.");
+
+  // NonZero output must have a static shape (no dynamic dims).
+  // The ONNX model should set the output shape to [rank, num_elements] where num_elements
+  // is the total number of elements in the input.
+  const auto& output = node_unit.Outputs()[0];
+  std::vector<uint32_t> output_shape;
+  RETURN_IF_NOT(QnnModelWrapper::GetOnnxShape(output.shape, output_shape),
+                "NonZero output shape must be static. Set shape to [rank, num_elements].");
+
+  const std::string& output_name = output.name;
+  if (qnn_model_wrapper.IsGraphOutput(output_name)) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                "NonZero output is a graph output. QNN HTP pads unused elements with -1.");
+  }
+
+  // QNN HTP NonZero pads unused output elements with -1. Only these consumer ops handle -1 indices correctly.
+  static const std::unordered_set<std::string> allowed_consumers = {
+      "Gather", "GatherElements", "GatherND", "ScatterElements", "ScatterND", "Reshape", "Transpose"};
+
+  const OrtApi& ort_api = qnn_model_wrapper.GetOrtApi();
+  for (const OrtNode* consumer : node_unit.GetOutputNodes(ort_api)) {
+    std::string op = Ort::ConstNode(consumer).GetOperatorType();
+    RETURN_IF_NOT(allowed_consumers.count(op) == 1,
+                  ("NonZero consumer op '" + op + "' does not support -1 padded indices.").c_str());
+  }
+
+  // QNN HTP NonZero supported input types:
+  //   FLOAT_16, UFIXED_POINT_16, UFIXED_POINT_8, BOOL_8
+  // FLOAT_32 is also accepted here because ProcessInputs inserts a Cast(fp32 -> fp16).
+  return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
+}
+
+Ort::Status NonZeroOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
+                                            const OrtNodeUnit& node_unit,
+                                            const Ort::Logger& logger,
+                                            std::vector<std::string>& input_names,
+                                            bool do_op_validation) const {
+  ORT_UNUSED_PARAMETER(do_op_validation);
+
+  const auto& inputs = node_unit.Inputs();
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
+
+  // HTP NonZero only supports FLOAT_16 (not FLOAT_32) for float inputs.
+  // Insert a Cast(fp32 -> fp16) node if needed.
+  if (input_info.qnn_data_type == QNN_DATATYPE_FLOAT_32) {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[0], logger, input_names));
+
+    const std::string& original_input_name = input_names.back();
+    const std::string cast_output_name = utils::UniqueNameGenerator().New(original_input_name, "_cast_fp16");
+
+    // Create fp16 intermediate tensor.
+    QnnTensorWrapper cast_output_tensorwrapper(cast_output_name,
+                                               QNN_TENSOR_TYPE_NATIVE,
+                                               QNN_DATATYPE_FLOAT_16,
+                                               QnnQuantParamsWrapper(),
+                                               std::vector<uint32_t>(input_info.shape));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(cast_output_tensorwrapper)),
+                  "Failed to add fp16 cast tensor.");
+
+    // Add Cast node: fp32 -> fp16.
+    const std::string cast_node_name = cast_output_name + "_cast_node";
+    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(cast_node_name,
+                                                  QNN_OP_PACKAGE_NAME_QTI_AISW,
+                                                  QNN_OP_CAST,
+                                                  {original_input_name},
+                                                  {cast_output_name},
+                                                  {}),
+                  "Failed to add Cast fp32->fp16 node");
+
+    // Replace the input name with the casted fp16 tensor.
+    input_names.back() = cast_output_name;
+  } else {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[0], logger, input_names));
+  }
+
+  return Ort::Status();
+}
+
+Ort::Status NonZeroOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                                          const OrtNodeUnit& node_unit,
+                                                          std::vector<std::string>&& input_names,
+                                                          const Ort::Logger& logger,
+                                                          bool do_op_validation) const {
+  ORT_UNUSED_PARAMETER(logger);
+
+  // Get input shape to compute maximum output size.
+  const auto& inputs = node_unit.Inputs();
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
+
+  const auto& input_shape = input_info.shape;
+  uint32_t input_rank = static_cast<uint32_t>(input_shape.size());
+  SafeInt<uint32_t> num_elements = 1;
+  for (uint32_t dim : input_shape) {
+    num_elements *= dim;
+  }
+
+  // QNN NonZero output shape: [num_nonzero_elements, rank_of_input] (transposed vs ONNX).
+  // ONNX NonZero output shape: [rank_of_input, num_nonzero_elements].
+  // Set num_nonzero_elements to max possible (total elements in input).
+  std::vector<uint32_t> qnn_nonzero_shape = {num_elements, input_rank};
+  std::vector<uint32_t> onnx_output_shape = {input_rank, num_elements};
+
+  const auto& outputs = node_unit.Outputs();
+  const std::string& output_name = outputs[0].name;
+  bool is_graph_output = qnn_model_wrapper.IsGraphOutput(output_name);
+
+  Qnn_DataType_t output_data_type = QNN_DATATYPE_INT_32;
+
+  // Chain: NonZero(int32, [N, rank]) -> Transpose(int32, [rank, N]) -> Cast(int64) if graph output.
+  const std::string nonzero_out_name = output_name + "_nonzero_out";
+  const std::string transpose_out_name = output_name + "_transposed";
+
+  // 1. NonZero output tensor: [num_elements, input_rank], int32, native.
+  QnnTensorWrapper nonzero_out_tensor(nonzero_out_name,
+                                      QNN_TENSOR_TYPE_NATIVE,
+                                      output_data_type,
+                                      QnnQuantParamsWrapper(),
+                                      std::vector<uint32_t>(qnn_nonzero_shape));
+  RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(nonzero_out_tensor)),
+                "Failed to add NonZero output tensor.");
+
+  // Create the NonZero QNN node.
+  RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(utils::UniqueNameGenerator().New(node_unit),
+                                                QNN_OP_PACKAGE_NAME_QTI_AISW,
+                                                QNN_OP_NON_ZERO,
+                                                std::move(input_names),
+                                                {nonzero_out_name},
+                                                {},
+                                                do_op_validation),
+                "Failed to add NonZero node.");
+
+  // 2. Transpose: [num_elements, rank] -> [rank, num_elements] to match ONNX spec.
+  std::vector<uint32_t> perm = {1, 0};
+  QnnParamWrapper transpose_param(node_unit.Index(), node_unit.Name(), QNN_OP_TRANSPOSE_PARAM_PERM,
+                                  {static_cast<uint32_t>(perm.size())}, std::vector<uint32_t>(perm));
+  std::vector<std::string> transpose_params;
+  transpose_params.push_back(transpose_param.GetParamTensorName());
+  qnn_model_wrapper.AddParamWrapper(std::move(transpose_param));
+
+  if (is_graph_output) {
+    // Transpose output is native int32, then Cast to int64 for graph output.
+    QnnTensorWrapper transpose_out_tensor(transpose_out_name,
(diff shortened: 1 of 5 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,250 @@
...
+                "NonZero is only supported on HTP backend.");
+
+  // QNN HTP NonZero supported input types:
+  //   FLOAT_16, UFIXED_POINT_16, UFIXED_POINT_8, BOOL_8
+  // FLOAT_32 is also accepted here because ProcessInputs inserts a Cast(fp32 -> fp16).
+  const auto& inputs = node_unit.Inputs();
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
+
+  static const std::unordered_set<Qnn_DataType_t> supported_types = {
```
qti-yuduo: Do we have to check here rather than relying on QNN Validation APIs?

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,250 @@
...
+                                            Qnn_DataType_t qnn_data_type) const override ORT_MUST_USE_RESULT;
+
+  Ort::Status ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                          const OrtNodeUnit& node_unit,
+                                          std::vector<std::string>&& input_names,
+                                          const Ort::Logger& logger,
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+};
+
+Ort::Status NonZeroOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
```
qti-yuduo, line 41: QNN `NonZero` pads its output with `-1` in unused slots. Only 5 QNN ops (Gather, GatherElements, GatherNd, ScatterElements, ScatterNd) know to skip -1 per https://docs.qualcomm.com/doc/80-63442-10/topic/MasterOpDef.html. Every other op potentially risk producing unexpected behaviors / numerics. 

^ might be rare, but we don't want surprises, how about we more conservative here and gate on downstream consumer op types?

```
const OrtApi& ort_api = qnn_model_wrapper.GetOrtApi();
std::vector<const OrtNode*> consumers = node_unit.GetOutputNodes(ort_api);

// Empty consumers = graph output (no downstream nodes)
for (const OrtNode* consumer : consumers) {
    std::string op = Ort::ConstNode(consumer).GetOperatorType();
    // check against allowlist
}
```

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,250 @@
...
+                                            const OrtNodeUnit& node_unit,
```
qti-yuduo, line 42: Also should check `IsGraphOutput()` as well to avoid the semantic differences causing undefined behavior.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,250 @@
...
+  ORT_UNUSED_PARAMETER(logger);
+
+  // Get input shape to compute maximum output size.
+  const auto& inputs = node_unit.Inputs();
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
+
+  const auto& input_shape = input_info.shape;
+  uint32_t input_rank = static_cast<uint32_t>(input_shape.size());
+  uint32_t num_elements = 1;
```
qti-yuduo: `SafeInt` is better to handle potential overflow.

onnxruntime/test/providers/qnn/nonzero_test.cc
```diff
@@ -0,0 +1,91 @@
...
+  provider_options["offload_graph_io_quantization"] = "0";
+  RunQnnModelTest(BuildNonZeroTestCase<bool>({2, 2, 3},
+                                             {true, false, true, false, true, false,
+                                              true, true, false, false, false, true}),
+                  provider_options, 11, ExpectedEPNodeAssignment::All,
+                  1e-5f, ORT_LOGGING_LEVEL_ERROR, true, nullptr, /*is_dynamic_output=*/true);
+}
+
+// All non-zero: [[1, 2], [3, 4]] -> every element is non-zero
+TEST_F(QnnHTPBackendTests, NonZero_AllNonZero) {
```
qti-yuduo, line 170: Perhaps add all-zeros test case as well: NonZeroAllZero

quic-ykota: Addressed.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Nice class :) Using it now

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: My original intention was to provide a clear error message. But, I agree to keep it aligned with other op_builders and also to not take on the additional burden of maintaining the supported combination of inputs, its better to let the validation API handle it. Removing this check now.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Could you explain a bit more on what you are suggesting here?

qti-yuduo: User who use NonZero as output expect ONNX semantics, but would get -1s and different shape (than ONNX) that they don't know how to reason about.. I was just suggesting a safe guard to claim we do not support when NonZero is graph output so that they can continue getting existing ORT default behavior which aligns with ONNX spec.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Good suggestion. Added this check and also updated the list with "Reshape" which is being used in several models as a consumer before feeding into gather ops.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Ah got it. I thought about it early on, but supporting this would require a lot of changes to the way this is tested. More importantly, I feel user can easily update their handling code if at all this op is used at model output (just do a small post processing to remove all entries with -1s) which would allow us to still support nonZero op at the model outputs. To be more informative, I have added additional warning in the code to convey it. What do you think about it?

qti-yuduo: Would it be possible to avoid test infra change if we construct a NonZero test case model with additional node as output?

qti-yuduo: E.g., feed into => Gather, would that bring back ONNX semantics again?

qti-yuduo:
BTW if you pre-commit install, then it'll do lint automatically for you when commit -- so never forgot and wasting CI.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: trying it but failing right now. Let me dig a bit deeper and get back.

quic-ykota: So, here's the issue - 
Lets take this ONNX model 
Input -> <Some op> -> NonZero - (Dynamic shape) -> Gather -> (Dynamic shape) -> ... Op_with_fixed_output_shape -> (static shape)

Here the dynamic shape would continue to follow along in the onnx graph until we hit an op with fixed shape or output.
Op_with_fixed_output_shape above can be topK, mean, etc where the output shape is fixed.

This PR fixes the output shape for NonZero in DLC ie changes the above to  
Input -> <Some op> -> NonZero - (Fixed shape) -> Gather -> (Dynamic shape) -> ... <>

As you can see, fixing the NonZero's dynamic shape alone is not sufficient but the graph needs to be updated with a fixed shape for consumer branch of nonzero. So, along with this change, we need one more higher level change to update this. For example: onnx2onnx pass (?). Let me know what you think?

minfhong-qti: I previously had a similar PR(https://github.com/microsoft/onnxruntime/pull/26625) for NonZero, including adding a pass in preprocess stage, where the pass modified the output shape of NonZero to a fix shape. However, MS had concerns that this would introduce discrepancy between QNN and ORT CPU. I closed the PR since there is no actual requirement but if there is today, we probably should discuss to bring up a better solution.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,257 @@
...
+                  ("NonZero consumer op '" + op + "' does not support -1 padded indices.").c_str());
+  }
+
+  // QNN HTP NonZero supported input types:
+  //   FLOAT_16, UFIXED_POINT_16, UFIXED_POINT_8, BOOL_8
+  // FLOAT_32 is also accepted here because ProcessInputs inserts a Cast(fp32 -> fp16).
+  const auto& inputs = node_unit.Inputs();
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
+
```
qti-chuteng:
```suggestion
  static const std::unordered_set<Qnn_DataType_t> allowed_types = {
    QNN_DATATYPE_FLOAT_32, QNN_DATATYPE_FLOAT_16,
    QNN_DATATYPE_UFIXED_POINT_16, QNN_DATATYPE_UFIXED_POINT_8,
    QNN_DATATYPE_BOOL_8};
RETURN_IF_NOT(allowed_types.count(input_info.qnn_data_type),
              "NonZero: unsupported input data type.");
```

onnxruntime/test/util/test_utils.cc
```diff
@@ -137,6 +137,97 @@ static void VerifyOutputs(const std::vector<std::string>& output_names,
   }
 }
 
+// Verify outputs for ops with dynamic output shapes (e.g., NonZero).
+// The actual output may be padded to a max size. Uses the expected output's shape
+// to determine the valid region and compares only that portion row by row.
+static void VerifyDynamicOutputs(const std::vector<std::string>& output_names,
```
qti-chuteng: Change the naming to more straightforward one to indicate that it only support < 3d

onnxruntime/test/providers/qnn/nonzero_test.cc
```diff
@@ -0,0 +1,113 @@
...
+  RunQnnModelTest(BuildNonZeroTestCase<float>({2, 2}, {0, 0, 0, 0}),
+                  provider_options, 11, ExpectedEPNodeAssignment::All,
+                  1e-5f, ORT_LOGGING_LEVEL_ERROR, true, nullptr, /*is_dynamic_output=*/true);
+}
+
+#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
+
+}  // namespace test
+}  // namespace onnxruntime
+#endif  // !defined(ORT_MINIMAL_BUILD)
```
qti-chuteng, line 274: Would you consider to add UFIXED_POINT_16 and UFIXED_POINT_8 test?

onnxruntime/test/util/test_utils.cc
```diff
@@ -239,7 +330,8 @@ void RunAndVerifyOutputsWithEP(ModelPathOrBytes model_path_or_bytes,
                                std::string_view log_id,
                                const std::unordered_map<std::string, Ort::Value>& feeds,
                                const EPVerificationParams& params,
-                               bool verify_outputs) {
+                               bool verify_outputs,
+                               bool is_dynamic_output) {
```
qti-chuteng: could we put this in params?

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,257 @@
...
+  }
+  return qnn_data_type;
+}
+
+Ort::Status NonZeroOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                                          const OrtNodeUnit& node_unit,
+                                                          std::vector<std::string>&& input_names,
+                                                          const Ort::Logger& logger,
+                                                          bool do_op_validation) const {
+  ORT_UNUSED_PARAMETER(logger);
```
qti-chuteng, line 117: Could you refer to the conv_op_builder.cc and use the ORT_CXX_LOG for the logging?

qti-chuteng (dismissed):

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Are you able to explain what changes you are suggesting here? I am calling the macro ORT_CXX_LOG in this file as well.

onnxruntime/test/providers/qnn/nonzero_test.cc
quic-ykota: Done.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Could you explain why you are suggesting these changes?

onnxruntime/test/util/test_utils.cc
quic-ykota: Nice suggestion. Updated.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -162,8 +162,13 @@ Ort::Status QnnModel::SetGraphInputOutputInfo(const QnnModelContext& context) {
     ONNXTensorElementDataType elem_type = shape_info.GetElementType();
     std::vector<int64_t> shape = shape_info.GetShape();
 
-    for (const auto& s : shape) {
-      RETURN_IF(s < 0, ("Dynamic shape is not supported yet, for output: " + name).c_str());
+    bool has_dynamic_dim = std::any_of(shape.begin(), shape.end(), [](int64_t s) { return s < 0; });
```
minfhong-qti: May I know whether there is a real use case that NonZero output is the graph output?
Otherwise, I'm not a big fan of modifying the static shape check here.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,257 @@
...
+                                            std::vector<std::string>& input_names,
+                                            bool do_op_validation) const {
+  ORT_UNUSED_PARAMETER(do_op_validation);
+
+  const auto& inputs = node_unit.Inputs();
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
+
+  // HTP NonZero only supports FLOAT_16 (not FLOAT_32) for float inputs.
+  // Insert a Cast(fp32 -> fp16) node if needed.
```
minfhong-qti, line 91: Is the additional Cast here necessary? I thought HTP would automatically convert FP32 to FP16?

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,257 @@
...
+    RETURN_IF_NOT(allowed_consumers.count(op) == 1,
+                  ("NonZero consumer op '" + op + "' does not support -1 padded indices.").c_str());
+  }
+
+  // QNN HTP NonZero supported input types:
+  //   FLOAT_16, UFIXED_POINT_16, UFIXED_POINT_8, BOOL_8
+  // FLOAT_32 is also accepted here because ProcessInputs inserts a Cast(fp32 -> fp16).
+  const auto& inputs = node_unit.Inputs();
+  TensorInfo input_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
```
minfhong-qti: Can these lines be removed? I did not see their usage.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,257 @@
...
+    // Replace the input name with the casted fp16 tensor.
+    input_names.back() = cast_output_name;
+  } else {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[0], logger, input_names));
+  }
+
+  return Ort::Status();
+}
+
+Qnn_DataType_t NonZeroOpBuilder::GetSupportedOutputDataType(size_t index, Qnn_DataType_t qnn_data_type) const {
```
minfhong-qti: In `QnnTensorWrapper`, this is handled.

onnxruntime/test/providers/qnn/nonzero_test.cc
```diff
@@ -0,0 +1,158 @@
...
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = "htp";
+  provider_options["offload_graph_io_quantization"] = "0";
+  RunQnnModelTest(BuildNonZeroTestCase<float>({2, 3}, {1.0f, 0.0f, 3.0f, 0.0f, 5.0f, 0.0f}),
+                  provider_options, 11, ExpectedEPNodeAssignment::All,
+                  1e-5f, ORT_LOGGING_LEVEL_ERROR, true, nullptr, /*is_dynamic_output=*/true);
+}
+
+// Input (QDQ uint8): [[1, 0, 3], [0, 5, 0]] -> non-zero at (0,0), (0,2), (1,1)
+// Pattern: float input -> Q(u8) -> DQ(u8) -> NonZero -> int64 output
```
minfhong-qti, line 113: Since QDQ selector is not modified for NonZero, I think NonZero here is in fact added to QNN in float.

onnxruntime/test/util/test_utils.cc
```diff
@@ -278,7 +369,11 @@ void RunAndVerifyOutputsWithEP(ModelPathOrBytes model_path_or_bytes,
   RunWithEP(ort_session, ort_run_options, feeds, fetches);
 
   if (verify_outputs) {
-    VerifyOutputs(output_names, expected_fetches, fetches, params);
+    if (params.is_dynamic_output) {
+      VerifyDynamic1D2DOutputs(output_names, expected_fetches, fetches);
```
minfhong-qti: I think this function is only specifically for NonZero.
In an ***ideal*** dynamic shape support, QNN in fact will produce the same shape as ORT CPU does.
Is it possible that we make an input with known non-zero indices, followed by a Gather to gather them back? In this way, QNN and ORT CPU can be directly compared.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Updated the scope of the PR to only support nonZero op with static output shape (set to {rank, num_elements} shape). This fixes the issue discussed in this thread. We can revisit truly supporting dynamic shapes when the need arises.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Yes, it is necessary unfortunately. You'd see a validation error if its removed.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Resolving the comment for now. Please reopen if you think its still an issue.

onnxruntime/test/providers/qnn/nonzero_test.cc
quic-ykota: Nice catch.  I fixed this subtle bug.

onnxruntime/test/util/test_utils.cc
quic-ykota: The problem is that even with a gather op, the output shape would be different between cpu and htp (set to a shape of {rank, num_elements} always). So, this method is necessary unfortunately.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Resolving the comment. Please reopen if its still an issue.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
quic-ykota: Based on the updated nonZero op scope, this change is no longer necessary. I have reverted this change + another change in this file.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Done. Stale change from previous cleanup.

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
quic-ykota: Ah, looks like its a recent change. Updated the code.

onnxruntime/test/util/test_utils.cc
```diff
@@ -137,6 +137,97 @@ static void VerifyOutputs(const std::vector<std::string>& output_names,
   }
 }
 
+// Verify outputs for ops with padded outputs (e.g., NonZero pads unused elements with -1).
+// The actual output may be padded to a max size. Uses the expected output's shape
+// to determine the valid region and compares only that portion row by row.
+static void VerifyDynamic1D2DOutputs(const std::vector<std::string>& output_names,
```
qti-yuduo: Alternatively, I think it would be cleaner if we can avoid universal unit test infra change. Make a NonZero specific helper in-place. 
```
static void RunNonZeroTest(const GetTestModelFn& build_fn,
                           ProviderOptions& provider_options,
                           int opset) {
  // 1. Build model
  const std::unordered_map<std::string, int> domains = {{"", opset}};
  onnxruntime::Model model("nonzero_test", false, ModelMetaData(), PathString(),
                           IOnnxRuntimeOpSchemaRegistryList(), domains, {}, DefaultLoggingManager().DefaultLogger());
  build_fn(/* model builder */);

  // 2. Run on CPU EP → expected [rank, actual_count]
  // 3. Run on QNN EP → actual [rank, num_elements] (padded with -1)

  // 4. Pad expected to match actual shape
  auto expected_shape = expected.GetTensorTypeAndShapeInfo().GetShape();
  auto actual_shape = actual.GetTensorTypeAndShapeInfo().GetShape();
  int64_t rank = expected_shape[0];
  int64_t valid_cols = expected_shape[1];
  int64_t total_cols = actual_shape[1];

  std::vector<int64_t> padded(rank * total_cols, -1);
  const int64_t* src = expected.GetTensorData<int64_t>();
  for (int64_t r = 0; r < rank; ++r) {
    std::copy(src + r * valid_cols, src + r * valid_cols + valid_cols,
              padded.begin() + r * total_cols);
  }

  // 5. Element-wise compare padded vs actual
  const int64_t* act = actual.GetTensorData<int64_t>();
  for (int64_t i = 0; i < rank * total_cols; ++i) {
    EXPECT_EQ(padded[i], act[i]);
  }
}

Comment (quic-ykota) [reply onnxruntime/test/util/test_utils.cc]: Thanks for the recommendation. Updated to use this test logic.

**Review by qti-yuduo — APPROVED** [review]

File: onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc [hunk 18]
```diff
@@ -0,0 +1,249 @@
...
+                                         std::vector<uint32_t>(onnx_output_shape));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(graph_output_tensor)),
+                  "Failed to add graph output tensor.");
+
+    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(output_name + "_cast_node",
+                                                  QNN_OP_PACKAGE_NAME_QTI_AISW,
+                                                  QNN_OP_CAST,
+                                                  {transpose_out_name},
+                                                  {output_name},
+                                                  {}),
```
qti-chuteng: ditto

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,249 @@
...
+                                          std::vector<uint32_t>(onnx_output_shape));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(transpose_out_tensor)),
+                  "Failed to add Transpose output tensor.");
+
+    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(transpose_out_name + "_transpose_node",
+                                                  QNN_OP_PACKAGE_NAME_QTI_AISW,
+                                                  QNN_OP_TRANSPOSE,
+                                                  {nonzero_out_name},
+                                                  {transpose_out_name},
+                                                  std::move(transpose_params)),
```
qti-chuteng: Add do_op_validation in CreateQnnNode?

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,249 @@
...
+                                          std::vector<uint32_t>(onnx_output_shape));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(transpose_out_tensor)),
+                  "Failed to add Transpose output tensor.");
+
+    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(output_name + "_transpose_node",
+                                                  QNN_OP_PACKAGE_NAME_QTI_AISW,
+                                                  QNN_OP_TRANSPOSE,
+                                                  {nonzero_out_name},
+                                                  {output_name},
+                                                  std::move(transpose_params)),
```
qti-chuteng: ditto

onnxruntime/core/providers/qnn/builder/opbuilder/nonzero_op_builder.cc
```diff
@@ -0,0 +1,249 @@
...
+  // HTP NonZero only supports FLOAT_16 (not FLOAT_32) for float inputs.
+  // Insert a Cast(fp32 -> fp16) node if needed.
+  if (input_info.qnn_data_type == QNN_DATATYPE_FLOAT_32) {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[0], logger, input_names));
+
+    const std::string& original_input_name = input_names.back();
+    const std::string cast_output_name = utils::UniqueNameGenerator().New(original_input_name, "_cast_fp16");
+
+    // Create fp16 intermediate tensor.
+    QnnTensorWrapper cast_output_tensorwrapper(cast_output_name,
```
qti-chuteng: replace line#99-115 with below call
----
RETURN_IF_ERROR(qnn_model_wrapper.AddCastNode(cast_node_name, original_input_name, cast_output_name, QNN_TENSOR_TYPE_NATIVE, QNN_DATATYPE_FLOAT_16, QnnQuantParamsWrapper(), std::vector<uint32_t>(input_info.shape), do_op_validation));

onnxruntime/test/providers/qnn/nonzero_test.cc
```diff
@@ -0,0 +1,196 @@
...
+    MakeTestInput<float>(builder, "X", input_def);
+    std::string dq_out = AddQDQNodePair<uint8_t>(builder, "qdq_in", "X", 0.02f, uint8_t(0));
+    builder.AddNode("nonzero_node", "NonZero", {dq_out}, {"Y"}, kOnnxDomain);
+    builder.MakeOutput<int64_t>("Y", std::vector<int64_t>{input_rank, num_elements});
+  };
+
+  RunNonZeroTest(qdq_model_fn, 11);
+}
+
+// Input (QDQ uint8): [[1, 0, 3], [0, 4.1, 0]] -> non-zero at (0,0), (0,2), (1,1)
```
qti-chuteng, line 130: uint16
[0, 5, 0]

onnxruntime/test/providers/qnn/nonzero_test.cc
```diff
@@ -0,0 +1,196 @@
...
+  };
+
+  RunNonZeroTest(build_model, 11, ExpectedEPNodeAssignment::None);
+}
+
+#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
+
+}  // namespace test
+}  // namespace onnxruntime
+#endif  // !defined(ORT_MINIMAL_BUILD)
```
qti-chuteng, line 274: Could you add these UT?
1. (Positive) NonZero -> Gather UT (Accept)
2. (Negative) NonZero -> Add (Reject)
3. NonZero -> multiple consumers.

quic-ykota:
Hello @qti-chuteng! Your latest comments are addressed. Please stamp an approval (PR expects you to approve even though others have approved since you have requested changes here)

PR #218, opened 2026-04-08
Load target backend library for validation when dumping DLC via QnnIr

yuhuchua-qti:
### Description
This change refines the QNNEP backend loading and validation flow so that:
- QnnIr.dll is still used for graph serialization and DLC dumping
- The target backend library is loaded in parallel when "backend_path|QnnXXX.dll" and "ep.qnnexecutionprovider.dump_qnn_ir_dlc|1" are specified in command.
- Validation is performed using the target backend’s validator

### Motivation and Context
When the backend path is set to QnnIr, QNNEP currently only loads QnnIr.dll and uses its validator to serialize and dump DLC. This is insufficient to guarantee that the generated DLC is valid for the actual execution backend (e.g. QnnHTP).


qti-hungjuiw:
Would it be possible for us to have some unit tests for this feature/functionality?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1837,6 +1790,26 @@ Ort::Status QnnBackendManager::SetupBackend(
...
+    ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE, "SetValidatorLogger succeed.");
+  }
+
+  if (status.IsOK() && qnn_serializer_config_) {
+    status = InitializeValidatorBackend();
+  }
+  if (status.IsOK() && qnn_serializer_config_) {
+    ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE, "InitializeValidatorBackend succeed.");
+  }
+
```
minfhong-qti, line 1895:
```suggestion
  if (status.IsOK() && qnn_serializer_config_) {
    status = InitializeQnnValidatorLog();
    if (status.IsOK()) {
      ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE, "SetValidatorLogger succeed.");
    }
    if (status.IsOK()) {
      status = InitializeValidatorBackend();
    }
    if (status.IsOK()) {
      ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE, "InitializeValidatorBackend succeed.");
    }
  }

```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -531,18 +531,44 @@ Ort::Status QnnBackendManager::InitializeQnnLog() {
...
+            ("Failed to initialize backend (" + backend_label + "). Error: " +
+             QnnErrorHandleToString(result))
+                .c_str());
+
+  initialized_flag = true;
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::InitializeQnnLog() {
+  return CreateQnnLog(qnn_interface_, log_handle_, "primary");
```
minfhong-qti: I prefer "backend" here as in normal use case (i.e., w/o dumping DLC), it is simply the backend.
As well for others.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -595,30 +621,54 @@ Ort::Status QnnBackendManager::ResetQnnLogLevel(std::optional<OrtLoggingLevel> o
...
+                                              const std::string& backend_label) {
+  if (interface.backendFree == nullptr) {
+    initialized_flag = false;
+    return Ort::Status();
+  }
 
+  Qnn_ErrorHandle_t result = interface.backendFree(backend_handle);
+  initialized_flag = false;
+  RETURN_IF(QNN_BACKEND_NO_ERROR != result,
+            ("Failed to shutdown " + backend_label + " backend!").c_str());
```
minfhong-qti: I think the logic here is different from the original one. Please check again.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -238,11 +238,22 @@ Ort::Status QnnModelWrapper::ValidateQnnNode(const std::string& node_name,
                                        std::move(params));
 
   std::string error_msg;
-  RETURN_IF_NOT(op_config_wrapper.QnnGraphOpValidation(qnn_interface_, backend_handle_, error_msg), error_msg.c_str());
+  std::cout << "## QnnModelWrapper::ValidateQnnNode start!" << std::endl;
```
minfhong-qti: Remove these couts.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -487,7 +498,8 @@ bool QnnModelWrapper::CreateQnnNode(const std::string& qnn_node_name,
     ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, oss.str().c_str());
 
     std::string error_msg;
-    bool rt = op_config_wrapper.QnnGraphOpValidation(qnn_interface_, backend_handle_, error_msg);
+    // bool rt = op_config_wrapper.QnnGraphOpValidation(qnn_interface_, backend_handle_, error_msg);
```
minfhong-qti: Remove.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -238,11 +238,22 @@ Ort::Status QnnModelWrapper::ValidateQnnNode(const std::string& node_name,
...
 }
 
+bool QnnModelWrapper::DoOpValidation(QnnOpConfigWrapper& op_config, std::string& error_msg) const {
+  if (backend_validator_handle_ != nullptr) {
+    std::cout << "## Validate with validator backend handle!" << std::endl;
+    return op_config.QnnGraphOpValidation(qnn_validator_interface_, backend_validator_handle_, error_msg);
+  }
+  std::cout << "## Validate with default backend handle!" << std::endl;
+  return op_config.QnnGraphOpValidation(qnn_interface_, backend_handle_, error_msg);
+}
```
```suggestion
Ort::Status QnnModelWrapper::ValidateQnnNode(QnnOpConfigWrapper op_config_wrapper.
                                             const QNN_INTERFACE_VER_TYPE& qnn_interface.
                                             const Qnn_BackendHandle_t& backend_handle,
                                             std::string& error_msg) const {
  RETURN_IF_NOT(op_config.QnnGraphOpValidation(qnn_interface_, backend_handle_, error_msg), error_msg.c_str());
  return Ort::Status();
}
```
Define as overload.

minfhong-qti (commented):
Previously user can give `QnnIr.dll` as backend path and dump DLC. What is the new usage after this PR?

qti-jkilpatrick:
I'm not planning to review the code, but please do make sure that loading `QnnHtp.dll` is not required. Please be sure to allow for `QnnCpu.dll`, `QnnGpu.dll`, or even `QnnIr.dll` to handle validation. Requiring HTP will break AI Hub's usage.

This is a great idea though. This has been a consistent frustration.

yuhuchua-qti:
Hi @qti-jkilpatrick, 

Loading QnnHtp.dll is not required for this functionality.
The target backend library used for validation is only loaded conditionally, and only when both of the following are specified:
- backend_path|QnnXXX.dll
- ep.qnnexecutionprovider.dump_qnn_ir_dlc|1

Under this setup, the backend specified via backend_path (e.g., QnnCpu.dll, QnnGpu.dll, or QnnIr.dll) is loaded in parallel for validation purposes.
Before merging, we will make sure that this flow works correctly with QnnCpu.dll, QnnGpu.dll, and QnnIr.dll, without requiring QnnHtp.dll, to ensure compatibility with AI Hub usage.

Thanks

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
yuhuchua-qti: I've updated in latest PR. Thanks.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
yuhuchua-qti: I've fixed it. Thanks.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
yuhuchua-qti: I've fixed it. Thanks.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -531,18 +525,44 @@ Ort::Status QnnBackendManager::InitializeQnnLog() {
...
 
   RETURN_IF(QNN_BACKEND_NO_ERROR != result,
-            ("Failed to initialize logging in the QNN backend. Error: " + QnnErrorHandleToString(result)).c_str());
+            ("Failed to initialize logging in the " + backend_label + ". Error: " +
+             QnnErrorHandleToString(result))
+                .c_str());
+  return Ort::Status();
+}
+
+Ort::Status QnnBackendManager::CreateQnnBackend(const QNN_INTERFACE_VER_TYPE& interface,
```
minfhong-qti: Nit.
1. Move right **ahead** `InitializeBackend` since they are relevant.
2. Prefer to rename to `InitializeBackendCommon` since I think current naming is confusing and not easy to follow the relationship among these `InitializerBackend`-related functions.

Please as well apply to all the other similar functions.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
yuhuchua-qti: I've updated this in latest PR. Thanks.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -594,30 +597,69 @@ Ort::Status QnnBackendManager::ResetQnnLogLevel(std::optional<OrtLoggingLevel> o
...
+Ort::Status QnnBackendManager::ShutdownValidatorBackend() {
+  if (!validator_backend_initialized_) {
+    return Ort::Status();
+  }
+
+  return ShutdownBackendCommon(qnn_validator_interface_, validator_backend_handle_,
+                               validator_backend_initialized_, "validator");
+}
+
+Ort::Status QnnBackendManager::ShutdownBackendCommon(const QNN_INTERFACE_VER_TYPE& interface,
```
minfhong-qti: Nit. Move ahead `ShutdownBackend` to align with others' order.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -335,18 +338,27 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
   Ort::Status CreateDevice();
 
   Ort::Status ReleaseDevice();
 
   Ort::Status ShutdownBackend();
 
+  Ort::Status ShutdownValidatorBackend();
+
+  // Shared implementation for ShutdownBackend / ShutdownValidatorBackend.
+  Ort::Status ShutdownBackendCommon(const QNN_INTERFACE_VER_TYPE& interface,
```
minfhong-qti, line 356: Please reorder these new functions like in .cc as well.

yuhuchua-qti: I've fixed this in latest PR. Thanks.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
yuhuchua-qti: I've fixed this in latest PR. Thanks.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
yuhuchua-qti: I've fixed this in latest PR. Thanks.

minfhong-qti (approved):

yuhuchua-qti:
> Would it be possible for us to have some unit tests for this feature/functionality?

Hi. I've added this in latest PR. Thanks.

yuhuchua-qti:
> Previously user can give `QnnIr.dll` as backend path and dump DLC. What is the new usage after this PR?

Hi,
There is no new usage introduced. 
The existing option can be used to dump the DLC: `-C "ep.qnnexecutionprovider.dump_qnn_ir_dlc|1 ep.qnnexecutionprovider.dump_qnn_ir_dlc_dir|.\dlc_output"`.
The target validation backend can be specified via `-i "backend_path|QnnXXX.dll"`.
Thanks.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -979,15 +981,19 @@ TEST_F(QnnHTPBackendTests, TestNHWCResizeShapeInference_qdq_sizes_opset18) {
...
+// Test that QNN Ir generates the expected DLC file for a model meant to run on the QNN HTP backend,
+// using the HTP backend's validator.
+TEST_F(QnnHTPBackendTests, QnnIr_HtpValidator_OutputFiles) {
   BackendSupport ir_backend_support = IsIRBackendSupported();
   if (ir_backend_support == BackendSupport::UNSUPPORTED) {
     GTEST_SKIP() << "QNN IR backend is not available! Skipping test.";
   }
   ASSERT_NE(ir_backend_support, BackendSupport::SUPPORT_ERROR) << "Failed to check if QNN IR backend is available.";
 
+  if (QnnHTPBackendTests::ShouldSkipIfHtpArchIsLessThanOrEqualTo(QNN_HTP_DEVICE_ARCH_V68)) {
```
kuanyul-qti: Can we use `SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V68)`?

qti-chuteng:
## Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:363-411 / 2098-2154`

Removing the `gsl::finally` and storing the validator library handle directly into the member `validator_backend_lib_handle_` introduces a leak on the failure path. `ReleaseResources()` early-returns when `!backend_setup_completed_`, and `backend_setup_completed_` is only set to `true` at the very end of a fully successful `SetupBackend()`. If any of the following fails (a) the second `GetQnnInterfaceProvider` call inside `LoadQnnSerializerBackend`, (b) `LoadQnnSystemLib`, `InitializeQnnLog`, `InitializeBackend`, `InitializeQnnValidatorLog`, `InitializeValidatorBackend`, `CreateDevice`, `InitializeProfiling`, `LoadOpPackage`, `CreateContext` — `validator_backend_lib_handle_` is set but never unloaded.

**Suggested fix**: Either (a) keep a stack-local handle + `gsl::finally` inside `LoadQnnSerializerBackend()` and only transfer ownership after every step succeeds; or (b) drop the early-return in `ReleaseResources()` (or scope-narrow it) and unconditionally unload both library handles at the bottom. Consider a small RAII `LibUnloader` deleter (see PR #215's `PlatformInfoDeleter`).

---

## Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:355-358` — Stale comment now contradicts the PR's core behavior. The comment says "calls to QnnBackend_validateOpConfig will be to the saver backend, not the 'intended' one" but the PR makes the opposite true. Rewrite to describe the new behavior.

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h:210, 227` — `GetQnnValidatorInterface()` / `GetQnnValidatorBackendHandle()` return zero-initialized values in the non-serializer flow; `QnnModelWrapper::ValidateQnnNode` branches on the handle being null as the only signal. The invariant (handle null ⇔ interface zero) is implicit and not enforced by the type system. Suggest bundling into a `std::optional<ValidatorContext>` or at minimum an assert + comment.

**[M-3]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:2089-2177` — The new `QnnIr_GpuValidator_OutputFiles` and `QnnIr_CpuValidator_OutputFiles` tests rely only on the fixture's `SetUp()` for availability gating, but the fixture probes are weak: `QnnGPUBackendTests` uses `cached_gpu_support_` (probably OK), but `QnnCPUBackendTests::IsIRBackendSupported()` ultimately calls `GetCPUSupport()` which always returns `SUPPORTED` — i.e. dead code on the CPU fixture. CI hosts without `QnnGpu.dll` will hard-fail in `SetupBackend` rather than skipping. Add an explicit lib-load probe; or make `GetIRSupport()` actually try loading QnnIr.

**[M-4]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:585-598` — `ResetQnnLogLevel()` only updates `qnn_interface_.logSetLogLevel(log_handle_, ...)`. With this PR a separate `validator_log_handle_` exists, but it is never updated when the user changes the runtime log level (e.g. via `SetSessionLogSeverityLevel`). The validator backend's logging stays at the level captured at setup time. Refactor `ResetQnnLogLevel` into a shared helper and update both paths in the serializer flow.

---

## Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:46-66` — `QnnModelWrapper` ctor grew to 11 positional parameters; bundle `qnn_interface / backend_handle / qnn_validator_interface / backend_validator_handle` into a `QnnBackendRefs` aggregate.

**[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:107-108` — New `ValidateQnnNode(QnnOpConfigWrapper&, std::string&)` overload is public and takes `op_config` by non-const reference, but the implementation only reads from it. Make it `const&` and move the helper to `private:` ([C31]: internal helpers stay private).

**[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:2074-2096` — `TerminateQnnLog`'s best-effort tail uses `std::optional + std::move` with a comment justifying it; a plain "store both, return whichever failed first" pattern is more readable.

**[N-4]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:351-393` — `TestBackend::Gpu` is added to the enum but `AddSerializerConfigs`'s switch only handles Ir/Saver and falls through to `assert(false)`. Document the assumption on the enum or sharpen the assert message.

**[N-5]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:1867-1869` — Log says `"SetValidatorLogger succeed."` but the call is `InitializeQnnValidatorLog()`. Inherits a pre-existing inconsistency on the regular path (`"SetLogger succeed."` vs `InitializeQnnLog`). Align both messages.

yuhuchua-qti:
> ## Critical
> **[C-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:363-411 / 2098-2154`
> 
> Removing the `gsl::finally` and storing the validator library handle directly into the member `validator_backend_lib_handle_` introduces a leak on the failure path. `ReleaseResources()` early-returns when `!backend_setup_completed_`, and `backend_setup_completed_` is only set to `true` at the very end of a fully successful `SetupBackend()`. If any of the following fails (a) the second `GetQnnInterfaceProvider` call inside `LoadQnnSerializerBackend`, (b) `LoadQnnSystemLib`, `InitializeQnnLog`, `InitializeBackend`, `InitializeQnnValidatorLog`, `InitializeValidatorBackend`, `CreateDevice`, `InitializeProfiling`, `LoadOpPackage`, `CreateContext` — `validator_backend_lib_handle_` is set but never unloaded.
> 
> **Suggested fix**: Either (a) keep a stack-local handle + `gsl::finally` inside `LoadQnnSerializerBackend()` and only transfer ownership after every step succeeds; or (b) drop the early-return in `ReleaseResources()` (or scope-narrow it) and unconditionally unload both library handles at the bottom. Consider a small RAII `LibUnloader` deleter (see PR #215's `PlatformInfoDeleter`).
> 
> ## Major
> **[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:355-358` — Stale comment now contradicts the PR's core behavior. The comment says "calls to QnnBackend_validateOpConfig will be to the saver backend, not the 'intended' one" but the PR makes the opposite true. Rewrite to describe the new behavior.
> 
> **[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h:210, 227` — `GetQnnValidatorInterface()` / `GetQnnValidatorBackendHandle()` return zero-initialized values in the non-serializer flow; `QnnModelWrapper::ValidateQnnNode` branches on the handle being null as the only signal. The invariant (handle null ⇔ interface zero) is implicit and not enforced by the type system. Suggest bundling into a `std::optional<ValidatorContext>` or at minimum an assert + comment.
> 
> **[M-3]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:2089-2177` — The new `QnnIr_GpuValidator_OutputFiles` and `QnnIr_CpuValidator_OutputFiles` tests rely only on the fixture's `SetUp()` for availability gating, but the fixture probes are weak: `QnnGPUBackendTests` uses `cached_gpu_support_` (probably OK), but `QnnCPUBackendTests::IsIRBackendSupported()` ultimately calls `GetCPUSupport()` which always returns `SUPPORTED` — i.e. dead code on the CPU fixture. CI hosts without `QnnGpu.dll` will hard-fail in `SetupBackend` rather than skipping. Add an explicit lib-load probe; or make `GetIRSupport()` actually try loading QnnIr.
> 
> **[M-4]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:585-598` — `ResetQnnLogLevel()` only updates `qnn_interface_.logSetLogLevel(log_handle_, ...)`. With this PR a separate `validator_log_handle_` exists, but it is never updated when the user changes the runtime log level (e.g. via `SetSessionLogSeverityLevel`). The validator backend's logging stays at the level captured at setup time. Refactor `ResetQnnLogLevel` into a shared helper and update both paths in the serializer flow.
> 
> ## Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:46-66` — `QnnModelWrapper` ctor grew to 11 positional parameters; bundle `qnn_interface / backend_handle / qnn_validator_interface / backend_validator_handle` into a `QnnBackendRefs` aggregate.
> 
> **[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:107-108` — New `ValidateQnnNode(QnnOpConfigWrapper&, std::string&)` overload is public and takes `op_config` by non-const reference, but the implementation only reads from it. Make it `const&` and move the helper to `private:` ([C31]: internal helpers stay private).
> 
> **[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:2074-2096` — `TerminateQnnLog`'s best-effort tail uses `std::optional + std::move` with a comment justifying it; a plain "store both, return whichever failed first" pattern is more readable.
> 
> **[N-4]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:351-393` — `TestBackend::Gpu` is added to the enum but `AddSerializerConfigs`'s switch only handles Ir/Saver and falls through to `assert(false)`. Document the assumption on the enum or sharpen the assert message.
> 
> **[N-5]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc:1867-1869` — Log says `"SetValidatorLogger succeed."` but the call is `InitializeQnnValidatorLog()`. Inherits a pre-existing inconsistency on the regular path (`"SetLogger succeed."` vs `InitializeQnnLog`). Align both messages.

Hi @qti-chuteng ,
I've fixed above comments.

  - [C-1] ✅ Fixed — Refactored ReleaseResources(): moved UnloadLib for backend_lib_handle_ and
  validator_backend_lib_handle_ outside the backend_setup_completed_ guard to prevent resource leak on partial setup
  failure.
  - [M-1] ✅ Fixed — Rewrote stale comment to clarify that in the QnnIr flow, backendValidateOpConfig routes to the
  target backend; added TODO for QnnSaver.
  - [M-2] ✅ Fixed — Added invariant comment above qnn_validator_interface_ in qnn_backend_manager.h; added assert in
  QnnModelWrapper constructor.
  - [M-3] ✅ Fixed — GetIRSupport() now actually probes QnnIr.dll / libQnnIr.so via LoadDynamicLibraryImpl instead of
  always returning SUPPORTED.
  - [M-4] ✅ Fixed — Extracted SetQnnLogLevelCommon() as a private member function; ResetQnnLogLevel() now updates both
  main and validator backend log levels.
  - [N-1] ⏭ Deferred — The invariant is already enforced via [M-2] assert + comment; QnnBackendRefs refactor is out of
  scope for this PR — will do in a follow-up.
  - [N-2] ✅ Fixed — Moved ValidateQnnNode(QnnOpConfigWrapper&, std::string&) to private:. const& was attempted but
  dropped — C API (backendValidateOpConfig) does not guarantee const correctness, causing C2662.
  - [N-3] ✅ Fixed — Simplified TerminateQnnLog(): replaced std::optional pattern with two named Ort::Status variables
  and explicit std::move.
  - [N-4] ✅ Fixed — Sharpened assert message in AddSerializerConfigs to "only Ir and Saver are valid serializer
  backends.".
  - [N-5] ✅ Fixed — "SetLogger succeed." → "InitializeQnnLog succeed."; "SetValidatorLogger succeed." →
  "InitializeQnnValidatorLog succeed.".

Thanks

onnxruntime/test/providers/qnn/qnn_basic_test.cc
yuhuchua-qti: I've fixed this in latest PR. Thanks.

PR #222, opened 2026-04-08
Avoid private ORT Core header in Test Framework

qti-hungjuiw:
### Description

- Remove private/internal ORT headers from QNN EP unit tests, replacing them with public API         equivalents or self-contained test utilities.
- Remove unused files from `unittest_util` that were only needed for private header access (e.g.,     `base_tester`, `op_tester`, `function_test_util`, `dummy_allocator`, etc.).
- Refactor `graph_transform_test_builder.h/.cc` and `framework_test_utils.h` to avoid depending on       private ORT internals.
- Enable `ValidateCompiledModelCompatibilityInfoImpl` to work without a prior `CreateEp` call by creating a temporary QNN EP on-demand (mirroring the pattern used in `GetHardwareDeviceIncompatibilityDetailsImpl`) and update `ModelCompatibility_ApiValidate_NoEp` test to reflect the new behavior (empty compatibility info returns `EP_NOT_APPLICABLE` without error).

### Motivation and Context

- QNN EP is moving toward EP ABI distribution, where the EP is shipped as a separate plugin library rather than being statically linked into OnnxRuntime.
- In this model, the EP's unit tests cannot  include private ORT headers (e.g., `graph.h`, `node_attr_utils.h`) since they are not part of the public API surface.                                                                                                                                                   
- This PR eliminates all such dependencies so that QNN EP tests compile and run using only the public C/C++ API.
- As part of this, `ValidateCompiledModelCompatibilityInfoImpl` was updated to no longer require a full EP instance to be created first — it now lazily creates a temporary EP for validation, enabling the compatibility check to work in scenarios where the caller only has an `OrtEpDevice` (e.g., the Auto EP selection flow queries compatibility before session creation).


onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -3448,48 +3449,52 @@ TEST_F(QnnHTPBackendTests, ModelCompatibility_GetCompatibility) {
...
+  const OrtEpDevice* qcom_npu_device = nullptr;
+  for (size_t i = 0; i < num_ep_devices; i++) {
+    const char* name = Ort::GetApi().EpDevice_EpName(ep_devices[i]);
+    const char* vendor_name = Ort::GetApi().EpDevice_EpVendor(ep_devices[i]);
+    const OrtHardwareDevice* ep_hw_device = Ort::GetApi().EpDevice_Device(ep_devices[i]);
+    if (name && std::string(name) == "QNNExecutionProvider" &&
+        Ort::GetApi().HardwareDevice_Type(ep_hw_device) == OrtHardwareDeviceType_NPU) {
+      qcom_npu_device = ep_devices[i];
+    }
+  }
```
kuanyul-qti, line 3651: `qcom_npu_device` is initialized to `nullptr` and only set in the loop if a QNN NPU EP device is found. After the loop, it is immediately dereferenced via `EpDevice_EpName(qcom_npu_device)` etc. without a null check. This will crash on x86 CI or any environment without an NPU device.

```cpp
// After the loop, add:
if (qcom_npu_device == nullptr) {
  GTEST_SKIP() << "No QNN NPU EP device found";
}
```

cmake/onnxruntime_unittests.cmake
```diff
@@ -329,6 +329,9 @@ block()
   # For onnxruntime_cxx_api.h
   target_include_directories(onnxruntime_provider_test PRIVATE ${ONNXRUNTIME_APPLICATION_INCLUDE_ROOT}/core/session)
 
+  # For cpu_provider_factory.h
+  target_include_directories(onnxruntime_provider_test PRIVATE ${ONNXRUNTIME_APPLICATION_INCLUDE_ROOT}/core/providers/cpu)
```
kuanyul-qti, line 352: `cpu_provider_factory.h` at `core/providers/cpu/cpu_provider_factory.h` is a private ORT Core header. Could we use `OrtSessionOptionsAppendExecutionProvider_CPU` directly?

qti-hungjuiw: No worry, the `cpu_provider_factory.h` is included in the prebuilt released by Microsoft.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
qti-hungjuiw: Fixed, thanks!

qti-chuteng:
---

**[M-3]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:397-406` — exception handling in `ValidateCompiledModelCompatibilityInfoImpl`

The `try/catch` block around `std::make_unique<QnnEp>(...)` only catches `std::runtime_error`. `QnnEp` construction can throw `std::bad_alloc`, `std::system_error`, `OnnxRuntimeException` (from `ORT_THROW`), and other `std::exception` derivatives — these all fall through to `catch(...)`, dropping the actual `e.what()` message. Please catch `std::exception` instead. Also, `GetHardwareDeviceIncompatibilityDetailsImpl` has the same duplicated pattern and only catches `...`; consider extracting a shared helper.

```cpp
} catch (const std::exception& e) {
  *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
  return factory->ort_api.CreateStatus(ORT_FAIL, e.what());
} catch (...) {
  *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
  return factory->ort_api.CreateStatus(ORT_FAIL, "Unknown exception ...");
}
```

---

**[M-4] (was v1 M-2)** `onnxruntime/test/providers/qnn/qnn_test_utils.cc` — `SizeHelper` overflow guard still missing

Still unaddressed from v1. Replacing `SafeInt<int64_t, SafeIntExceptionHandler>` with plain `int64_t` lost the overflow protection that the original `// validate no overflow` comment explicitly relied on. Please add a manual check:

```cpp
int64_t size = 1;
for (size_t i = start; i < end; i++) {
  if (shape[i] < 0) return static_cast<size_t>(-1);
  if (shape[i] != 0 && size > INT64_MAX / shape[i]) return static_cast<size_t>(-1);
  size *= shape[i];
}
return static_cast<size_t>(size);
```

---

**[M-5]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:347-351` — contract change for null/empty `compatibility_info`

The new early-return path silently changes the API contract: previously, calling Validate without an EP returned `ORT_EP_FAIL`; now, null/empty `compatibility_info` returns `nullptr` (success) with `EP_NOT_APPLICABLE`. There is no test for null/empty input — every `TestModelCompatibilityApiValidate` call uses a valid string. Please confirm with mainline reviewers whether "success + EP_NOT_APPLICABLE" and "failure + EP_NOT_APPLICABLE" are treated equivalently by upper-layer callers, and add a unit test for this path.

---

**[N-1] (was v1 N-1)** `onnxruntime/test/providers/qnn/qnn_test_utils.h` — graph_checker removal still undocumented

The TODO comment is good, but the PR description still doesn't list which tests lost coverage (`AveragePool1DFusedQnnNodePresent`, `MaxPool1D_ReshapeNodesPresent`, `EPRejectsDynamicShapesF32`, `EPOffloadsGraphIOQuantDequant`) or what public API will replace them. Please open a follow-up issue and reference it in the PR description.

---

**[N-2] (was v1 N-3)** `onnxruntime/test/providers/qnn/qnn_test_utils.h` — `void*` allocator parameter

Still open. The `allocator` parameter on `MakeTestInput`, `BuildOpTestCase`, `BuildQDQOpTestCase` etc. is not used in any call path. If unused, please remove it entirely rather than keeping a typeless `void*`.

---

**[N-3]** `onnxruntime/test/providers/qnn/qnn_ep_context_test.cc:3475-3483` — unused variable + hardcoded EP name

`vendor_name` is fetched but never used. Also, the comparison uses `std::string(name) == "QNNExecutionProvider"` — should be `kQnnExecutionProvider` to match the rest of this file's migration.

```cpp
const char* name = Ort::GetApi().EpDevice_EpName(ep_devices[i]);
// remove vendor_name (unused)
const OrtHardwareDevice* ep_hw_device = Ort::GetApi().EpDevice_Device(ep_devices[i]);
if (name && std::string(name) == kQnnExecutionProvider &&
    Ort::GetApi().HardwareDevice_Type(ep_hw_device) == OrtHardwareDeviceType_NPU) {
  qcom_npu_device = ep_devices[i];
}
```

---

**[N-4]** `onnxruntime/test/util/include/test_utils.h` — `CreateInputOrtValueOnCPU` ownership change

The new implementation switches from copy semantics (`memcpy` into a self-owned Tensor) to non-owning semantics (`Ort::Value::CreateTensor` with raw data pointer). The returned `Ort::Value` only references `value.data()`, so callers must keep the input span alive. The comment only addresses `const_cast` safety, not the ownership change. Please either (a) restore copy semantics, or (b) update the comment + function name and audit call sites for use-after-free.

---

**[N-5]** `onnxruntime/test/unittest_util/graph_transform_test_builder.h` — duplicated ORT internal code

To avoid `core/framework/tensorprotoutils.h`, `ConvertRawDataInTensorProto` (200+ lines) plus the full `tensorproto_data_size` mapping is now inlined into the header. This is a maintenance hazard — when ORT mainline adds new tensor element types, this duplicate will silently lag (especially on big-endian). Please add a `// MIRRORED FROM core/framework/tensorprotoutils.h — keep in sync` banner and consider whether big-endian support is actually needed by the test framework.

---

**[N-6] (was v1 N-2)** `onnxruntime/test/providers/qnn/qnn_model_wrapper_test.cc` — private QNN EP includes

The top-of-file comment now explains the static-lib pattern (good), but doesn't explicitly call out that `qnn_model_wrapper.h` and `qnn_def.h` are intentionally retained. One additional line clarifying this would be helpful for future reviewers.

---

**[T-1] (was v1 T-1)** `onnxruntime/test/providers/qnn/qnn_node_group/qdq_node_group_selector_test.cpp`

Still `.cpp`. Since this PR modifies the file, it would be a good time to rename to `.cc` and update the CMake reference.

---

**[T-2] (was v1 T-2)** `onnxruntime/test/providers/qnn/qnn_test_utils.h`

`kOnnxDomain = ""` is defined but the default parameters in `BuildOpTestCase` / `BuildQDQOpTestCase` still use raw `""`. Replace with `kOnnxDomain` for consistency.

---

**[T-4]** `onnxruntime/test/providers/qnn/qnn_test_utils.h:37-44` — macro naming

```cpp
#ifndef ORT_UNUSED_PARAMETER
#define ORT_UNUSED_PARAMETER(x) (void)(x)
#endif

#ifndef ORT_HANDLE_EXCEPTION
#define ORT_HANDLE_EXCEPTION(func) func()
#endif
```

The `ORT_` prefix may mislead readers into thinking these are public APIs. Consider renaming to `QNN_TEST_UNUSED_PARAMETER` / `QNN_TEST_HANDLE_EXCEPTION`, or just inline `(void)(x)` / `try { f(); } catch (...) { ... }` without the macro indirection.

kuanyul-qti (commented):
`onnxruntime/test/providers/qnn/optimizer/transpose_optimizer_test.cc` was not migrated. It still `#include`s `core/graph/constants.h` and `core/optimizer/transpose_optimizer.h` (both private), and calls `DefaultLoggingManager()` and `CountOpsInGraph()` which were removed in this PR. This will cause a guaranteed build failure under `!defined(ORT_MINIMAL_BUILD)`.

Will the test pass in current changes?

qti-hungjuiw:
> **[M-3]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:397-406` — exception handling in `ValidateCompiledModelCompatibilityInfoImpl`
> 
> The `try/catch` block around `std::make_unique<QnnEp>(...)` only catches `std::runtime_error`. `QnnEp` construction can throw `std::bad_alloc`, `std::system_error`, `OnnxRuntimeException` (from `ORT_THROW`), and other `std::exception` derivatives — these all fall through to `catch(...)`, dropping the actual `e.what()` message. Please catch `std::exception` instead. Also, `GetHardwareDeviceIncompatibilityDetailsImpl` has the same duplicated pattern and only catches `...`; consider extracting a shared helper.
> 
> ```c++
> } catch (const std::exception& e) {
>   *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
>   return factory->ort_api.CreateStatus(ORT_FAIL, e.what());
> } catch (...) {
>   *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
>   return factory->ort_api.CreateStatus(ORT_FAIL, "Unknown exception ...");
> }
> ```
> 
> **[M-4] (was v1 M-2)** `onnxruntime/test/providers/qnn/qnn_test_utils.cc` — `SizeHelper` overflow guard still missing
> 
> Still unaddressed from v1. Replacing `SafeInt<int64_t, SafeIntExceptionHandler>` with plain `int64_t` lost the overflow protection that the original `// validate no overflow` comment explicitly relied on. Please add a manual check:
> 
> ```c++
> int64_t size = 1;
> for (size_t i = start; i < end; i++) {
>   if (shape[i] < 0) return static_cast<size_t>(-1);
>   if (shape[i] != 0 && size > INT64_MAX / shape[i]) return static_cast<size_t>(-1);
>   size *= shape[i];
> }
> return static_cast<size_t>(size);
> ```
> 
> **[M-5]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:347-351` — contract change for null/empty `compatibility_info`
> 
> The new early-return path silently changes the API contract: previously, calling Validate without an EP returned `ORT_EP_FAIL`; now, null/empty `compatibility_info` returns `nullptr` (success) with `EP_NOT_APPLICABLE`. There is no test for null/empty input — every `TestModelCompatibilityApiValidate` call uses a valid string. Please confirm with mainline reviewers whether "success + EP_NOT_APPLICABLE" and "failure + EP_NOT_APPLICABLE" are treated equivalently by upper-layer callers, and add a unit test for this path.
> 
> **[N-1] (was v1 N-1)** `onnxruntime/test/providers/qnn/qnn_test_utils.h` — graph_checker removal still undocumented
> 
> The TODO comment is good, but the PR description still doesn't list which tests lost coverage (`AveragePool1DFusedQnnNodePresent`, `MaxPool1D_ReshapeNodesPresent`, `EPRejectsDynamicShapesF32`, `EPOffloadsGraphIOQuantDequant`) or what public API will replace them. Please open a follow-up issue and reference it in the PR description.
> 
> **[N-2] (was v1 N-3)** `onnxruntime/test/providers/qnn/qnn_test_utils.h` — `void*` allocator parameter
> 
> Still open. The `allocator` parameter on `MakeTestInput`, `BuildOpTestCase`, `BuildQDQOpTestCase` etc. is not used in any call path. If unused, please remove it entirely rather than keeping a typeless `void*`.
> 
> **[N-3]** `onnxruntime/test/providers/qnn/qnn_ep_context_test.cc:3475-3483` — unused variable + hardcoded EP name
> 
> `vendor_name` is fetched but never used. Also, the comparison uses `std::string(name) == "QNNExecutionProvider"` — should be `kQnnExecutionProvider` to match the rest of this file's migration.
> 
> ```c++
> const char* name = Ort::GetApi().EpDevice_EpName(ep_devices[i]);
> // remove vendor_name (unused)
> const OrtHardwareDevice* ep_hw_device = Ort::GetApi().EpDevice_Device(ep_devices[i]);
> if (name && std::string(name) == kQnnExecutionProvider &&
>     Ort::GetApi().HardwareDevice_Type(ep_hw_device) == OrtHardwareDeviceType_NPU) {
>   qcom_npu_device = ep_devices[i];
> }
> ```
> 
> **[N-4]** `onnxruntime/test/util/include/test_utils.h` — `CreateInputOrtValueOnCPU` ownership change
> 
> The new implementation switches from copy semantics (`memcpy` into a self-owned Tensor) to non-owning semantics (`Ort::Value::CreateTensor` with raw data pointer). The returned `Ort::Value` only references `value.data()`, so callers must keep the input span alive. The comment only addresses `const_cast` safety, not the ownership change. Please either (a) restore copy semantics, or (b) update the comment + function name and audit call sites for use-after-free.
> 
> **[N-5]** `onnxruntime/test/unittest_util/graph_transform_test_builder.h` — duplicated ORT internal code
> 
> To avoid `core/framework/tensorprotoutils.h`, `ConvertRawDataInTensorProto` (200+ lines) plus the full `tensorproto_data_size` mapping is now inlined into the header. This is a maintenance hazard — when ORT mainline adds new tensor element types, this duplicate will silently lag (especially on big-endian). Please add a `// MIRRORED FROM core/framework/tensorprotoutils.h — keep in sync` banner and consider whether big-endian support is actually needed by the test framework.
> 
> **[N-6] (was v1 N-2)** `onnxruntime/test/providers/qnn/qnn_model_wrapper_test.cc` — private QNN EP includes
> 
> The top-of-file comment now explains the static-lib pattern (good), but doesn't explicitly call out that `qnn_model_wrapper.h` and `qnn_def.h` are intentionally retained. One additional line clarifying this would be helpful for future reviewers.
> 
> **[T-1] (was v1 T-1)** `onnxruntime/test/providers/qnn/qnn_node_group/qdq_node_group_selector_test.cpp`
> 
> Still `.cpp`. Since this PR modifies the file, it would be a good time to rename to `.cc` and update the CMake reference.
> 
> **[T-2] (was v1 T-2)** `onnxruntime/test/providers/qnn/qnn_test_utils.h`
> 
> `kOnnxDomain = ""` is defined but the default parameters in `BuildOpTestCase` / `BuildQDQOpTestCase` still use raw `""`. Replace with `kOnnxDomain` for consistency.
> 
> **[T-4]** `onnxruntime/test/providers/qnn/qnn_test_utils.h:37-44` — macro naming
> 
> ```c++
> #ifndef ORT_UNUSED_PARAMETER
> #define ORT_UNUSED_PARAMETER(x) (void)(x)
> #endif
> 
> #ifndef ORT_HANDLE_EXCEPTION
> #define ORT_HANDLE_EXCEPTION(func) func()
> #endif
> ```
> 
> The `ORT_` prefix may mislead readers into thinking these are public APIs. Consider renaming to `QNN_TEST_UNUSED_PARAMETER` / `QNN_TEST_HANDLE_EXCEPTION`, or just inline `(void)(x)` / `try { f(); } catch (...) { ... }` without the macro indirection.

Most comments have been resolved

qti-hungjuiw:
> transpose_optimizer_test.cc

The `transpose_optimizer_test.cc` is excluded in the `onnxruntime_unittests.cmake`.
https://github.com/onnxruntime/onnxruntime-qnn/blob/main/cmake/onnxruntime_unittests.cmake#L214

As discussed previously, I supposed the plan is to remove this testcase testing ORT Core transpose optimizer from QNN-EP repo and possibly submit it to MS's ORT Core repo.

cc. @minfhong-qti

qti-hungjuiw:
Close since the separated PRs #420, #425, #430 are merged.

PR #223, opened 2026-04-08
[WIP] docker linux prototype

qti-jkilpatrick:
Prototype of a containerized Linux build.

### DIFF
--- a/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
+++ b/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
@@ -122,7 +122,7 @@ jobs:
           "${{ inputs.target_os == 'windows' && 'python.exe' || 'python3' }}"
           qcom/build_and_test.py
           --target-py-version ${{inputs.target_py_vsn}}
-          test_ort_${{inputs.target_os}}_${{inputs.target_arch}} --only
+          create_venv test_ort_${{inputs.target_os}}_${{inputs.target_arch}} --only
 
       - name: Publish Test Report
         uses: mikepenz/action-junit-report@v6
--- a/.github/workflows/qualcomm-internal-build-and-test.yml
+++ b/.github/workflows/qualcomm-internal-build-and-test.yml
@@ -45,10 +45,7 @@ jobs:
       nightly_build: ${{ inputs.nightly_build }}
       target_os: linux
       target_arch: x86_64
-      # Specifying a python version here doesn't actually do anything at this point since we don't know what the Linux
-      # wheel distribution story is yet. For now, this is just so that all of the job and artifact names reflect the
-      # version of cpython that's installed on our Linux build machines.
-      target_py_vsn: "3.10"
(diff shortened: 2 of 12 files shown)


.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           "${{ inputs.target_os == 'windows' && 'python.exe' || 'python3' }}"
           qcom/build_and_test.py
           --target-py-version ${{inputs.target_py_vsn}}
-          test_ort_${{inputs.target_os}}_${{inputs.target_arch}} --only
+          create_venv test_ort_${{inputs.target_os}}_${{inputs.target_arch}} --only
```
qti-jkilpatrick, line 125: We're in the "build and test on the same machine" path here. `create_venv` was previously being handled by the build step, but since that's now inside docker, we need to do it explicitly here.

.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -45,9 +45,7 @@ jobs:
       nightly_build: ${{ inputs.nightly_build }}
       target_os: linux
       target_arch: x86_64
-      # Specifying a python version here doesn't actually do anything at this point since we don't know what the Linux
-      # wheel distribution story is yet. For now, this is just so that all of the job and artifact names reflect the
-      # version of cpython that's installed on our Linux build machines.
+      # CI machines only have Python 3.10 so we cannot test using other versions.
```
qti-jkilpatrick, line 48: If we want to test with Python versions other than 3.10, someone needs to go install them on the CI machines since tests still run outside of docker. Perhaps they shouldn't?

qcom/scripts/linux/build.sh
```diff
@@ -126,7 +126,7 @@ fi
 if [ -n "${target_py_version}" ]; then
   common_args+=(--build_wheel)
 
-  build_venv="${build_dir}/venv-${target_py_version}"
+  build_venv="${build_dir}/venv-${target_py_version}.${mode}"
```
qti-jkilpatrick, line 129: `build` now creates a venv inside docker that cannot be used for `test` or `archive` so just keep them separate. This is actually pretty cheap since we use `uv`.

qcom/scripts/linux/build.sh
```diff
@@ -163,7 +163,7 @@ case "${target_platform}" in
       qnn_args+=("--ort_home")
       qnn_args+=("${ort_prebuilt_root}")
     fi
-    platform_args=(--build_shared_lib)
+    platform_args=(--build_shared_lib --cmake_extra_defines CMAKE_BUILD_RPATH_USE_ORIGIN:BOOL=TRUE)
```
qti-jkilpatrick, line 166: CMake sets `RPATH` to the build directory by default, which is not helpful if you want to build in a container and then run from the host. This just swaps the hard coded bin directory path for `$ORIGIN`.

PR #228, opened 2026-04-09
Fix GCC 13 build errors in QNN provider and tests

yuhuchua-qti:
### Description
This PR fixes two GCC 13 related build failures caused by code patterns that were previously tolerated by older compilers.

1. Fix invalid memory_order for std::atomic::store() in QNN EP
- std::atomic::store() does not permit memory_order_acq_rel per the C++ standard. The previous usage was undefined behavior. Older GCC versions only emitted a warning, while GCC 13 (as used in Ubuntu 24.04) treats this as a hard error (-Werror=invalid-memory-model).
- Replace memory_order_acq_rel with memory_order_release, which is the correct memory ordering for a pure store and preserves the intended synchronization semantics.

2. Fix GCC 13 -Wmaybe-uninitialized false positive in TestInputDef and parsed_value

- GCC 13 may emit a false positive -Wmaybe-uninitialized after deep inlining involving std::variant copy construction. All TestInputDef members are correctly initialized; the warning is due to a GCC 13 data flow analysis limitation.
- Suppress the warning locally with a guarded GCC diagnostic pragma and add a comment explaining the root cause and rationale.
- Changed T parsed_value; to T parsed_value{};. Value-initialization ensures the variable has a defined state before use, silencing the GCC 13 -Werror=maybe-uninitialized warning.

### Reference
- https://en.cppreference.com/w/cpp/atomic/atomic/store

### DIFF
--- a/onnxruntime/test/providers/qnn/gemm_test.cc
+++ b/onnxruntime/test/providers/qnn/gemm_test.cc
@@ -210,9 +210,21 @@ TEST_F(QnnCPUBackendTests, ReshapeGemmFusion) {
   std::vector<int64_t> shape_data = {4, 2};
   std::vector<float> weight_data(6, 1.0f);
   std::vector<float> bias_data = {1.0f, 2.0f, 3.0f};
+// GCC 13 with -O2 inlines this call chain deeply enough that its data flow analyzer loses track of
+// std::variant's initialization state inside the copy constructor (variant:224), triggering a false
+// positive -Wmaybe-uninitialized. The warning is suppressed here because TestInputDef members are
+// properly initialized in all constructors; this is a known GCC 13 analysis limitation with
+// std::variant + lambda capture + deep inlining.
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
   RunReshapeGemmTest(TestInputDef<float>({2, 2, 2}, false, input_data), TestInputDef<int64_t>({2}, true, shape_data),
                      TestInputDef<float>({2, 3}, true, weight_data), TestInputDef<float>({3}, true, bias_data),
                      ExpectedEPNodeAssignment::All);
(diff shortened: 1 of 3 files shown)


qti-yuduo (approved):

qti-chuteng (approved):

onnxruntime/test/providers/qnn/gemm_test.cc
```diff
@@ -210,9 +210,21 @@ TEST_F(QnnCPUBackendTests, ReshapeGemmFusion) {
...
   std::vector<float> weight_data(6, 1.0f);
   std::vector<float> bias_data = {1.0f, 2.0f, 3.0f};
+// GCC 13 with -O2 inlines this call chain deeply enough that its data flow analyzer loses track of
+// std::variant's initialization state inside the copy constructor (variant:224), triggering a false
+// positive -Wmaybe-uninitialized. The warning is suppressed here because TestInputDef members are
+// properly initialized in all constructors; this is a known GCC 13 analysis limitation with
+// std::variant + lambda capture + deep inlining.
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
huaychou, line 220: I got this warning as error in `onnxruntime-qnn/onnxruntime/test/util/env_var_utils.h:136:10:`
`/usr/include/c++/13/optional:214:15: error: ‘parsed_value’ may be used uninitialized [-Werror=maybe-uninitialized]
`
Did you encounter this issue?

yuhuchua-qti: My original change only fixes the GCC13 issue in RelWithDebInfo and Debug build.
The error you mention occurs in Release build.
I've included the fix for that issue in latest PR.
Thanks.

huaychou: LGTM

huaychou (approved):

PR #231, opened 2026-04-09
Run tests on ARM64 Linux in CI

qti-jkilpatrick:


.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -186,7 +186,7 @@ jobs:
     name: "test-${{inputs.target_os}}-${{inputs.target_arch}}-py${{ inputs.target_py_vsn != 'None' && inputs.target_py_vsn || 'none'}}"
     if: ${{ inputs.run_tests && !inputs.build_test_same_runner }}
     needs: [build]
-    runs-on: ["self-hosted", "${{inputs.target_os}}", "${{ inputs.test_runner_arch }}", "Build"]
+    runs-on: ["self-hosted", "${{inputs.target_os}}", "${{ inputs.test_runner_arch }}", "Test"]
```
qti-jkilpatrick, line 189: Previously, we had two non-exclusive tags for a machine's role: `Build` and `QDC`. The former has now been split into `Build` and `Test`. All machines that used to have `Build` also have `Test`, but the new QCS6490-based Ubuntu runners only have `Test`.

https://github.com/onnxruntime/onnxruntime-qnn/settings/actions/runners

.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -72,6 +72,22 @@ jobs:
...
+    secrets: inherit
+    with:
+      nightly_build: ${{ inputs.nightly_build }}
+      target_os: linux
+      target_arch: aarch64_manylinux_2_34
+      target_py_vsn: ${{ inputs.target_py_vsn }}
+      upload_test_archive: true
+      upload_wheel: true
+      test_wheel: false  # Fails because the HTP is not discovered
+      run_tests: true
```
qti-jkilpatrick, line 87: Essentially all tests are disabled in `qcom/scripts/linux/run_tests.sh`.

cmake/CMakeLists.txt
```diff
@@ -473,6 +473,12 @@ if (MSVC OR ${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
       list(APPEND QNN_LIB_FILES ${EXTRA_HTP_LIB})
     elseif(${QNN_ARCH_ABI} STREQUAL "aarch64-ubuntu-gcc9.4")
       list(APPEND QNN_LIB_FILES "${onnxruntime_QNN_HOME}/lib/hexagon-v68/unsigned/libQnnHtpV68Skel.so")
+    elseif(${QNN_ARCH_ABI} STREQUAL "aarch64-oe-linux-gcc11.2")
+      list(APPEND QNN_LIB_FILES "${onnxruntime_QNN_HOME}/lib/hexagon-v68/unsigned/libQnnHtpV68Skel.so"
+                                "${onnxruntime_QNN_HOME}/lib/hexagon-v69/unsigned/libQnnHtpV69Skel.so"
+                                "${onnxruntime_QNN_HOME}/lib/hexagon-v73/unsigned/libQnnHtpV73Skel.so"
+                                "${onnxruntime_QNN_HOME}/lib/hexagon-v75/unsigned/libQnnHtpV75Skel.so"
+                                "${onnxruntime_QNN_HOME}/lib/hexagon-v81/unsigned/libQnnHtpV81Skel.so")
```
qti-jkilpatrick, line 482: Stay tuned: we switch from `aarch64-ubuntu-gcc9.4` to `aarch64-oe-linux-gcc11.2` in `build_and_test.py`.

qcom/scripts/linux/build.sh
```diff
@@ -163,7 +163,7 @@ case "${target_platform}" in
       qnn_args+=("--ort_home")
       qnn_args+=("${ort_prebuilt_root}")
     fi
-    platform_args=(--build_shared_lib)
+    platform_args=(--build_shared_lib --cmake_extra_defines CMAKE_BUILD_RPATH_USE_ORIGIN:BOOL=TRUE)
```
qti-jkilpatrick, line 166: By default, cmake uses absolute paths for `RPATH`/`RUNPATH` entries for binaries in the build tree. This is normally fine for running them on the machine on which they're built. However, since we're mapping our build directory into docker where it has a different path, we wind up with, for example, a test runner that can't find `libonnxruntime.so` because the absolute path in which it was written does not exist outside of docker. To work around this, we tell cmake to add `$ORIGIN` to the rpath at build time.

qcom/scripts/linux/run_tests.sh
```diff
@@ -116,11 +116,10 @@ cd "${onnx_models_root}"
...
 for runner in "${model_test_runners[@]}"; do
-    "${runner}" cpu node "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node"
-
-    "${runner}" cpu float32
 
     if [ "$(uname -m)" != "aarch64" ]; then  # TODO: [AISW-164203] ORT test failures on Rubik Pi
+        "${runner}" cpu node "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node"
+        "${runner}" cpu float32
         "${runner}" htp qdq
     fi
```
qti-jkilpatrick, line 124: When I last looked at this a few months ago, CPU tests were working. Now they're not.

setup.py
```diff
@@ -100,16 +100,19 @@ class bdist_wheel(_bdist_wheel):  # noqa: N801
 
         def finalize_options(self):
             _bdist_wheel.finalize_options(self)
-            if not is_manylinux:
-                self.root_is_pure = False
+            # Insist that this wheel contains more than Python source
+            self.root_is_pure = False
```
qti-jkilpatrick, line 103: One positive side effect of reexamining how to create manylinux wheels is that we now have a much simpler implementation. Instead of manually curating the wheel's compatibility tags, we just tell setuptools to figure it out from the system.

setup.py
```diff
@@ -100,16 +100,19 @@ class bdist_wheel(_bdist_wheel):  # noqa: N801
...
 
         def run(self):
+            qnn_dependencies = ["libcdsprpc.so"]
```
qti-jkilpatrick, line 106: Restored from the previous change: this file isn't distributed with QNN so don't try to inspect it when looking at the wheel's glibc imports.

qcom/build_and_test.py
```diff
@@ -820,6 +820,38 @@ def test_ort_linux_x86_64(self, plan: Plan) -> str:
...
 
+    if is_host_linux() and is_host_arm64():
+
+        @task
+        @depends(["build_ort_linux_aarch64_manylinux_2_34"])
+        def test_ort_linux_aarch64_manylinux_2_34_pygpu(self, plan: Plan) -> str:
+            assert self.__target_py_version is not None
+            return plan.add_step(
+                OrtWheelGpuModelTestTask(
+                    "Running GPU model tests on ARM64ec",
```
huaychou:
```suggestion
                    "Running GPU model tests on ARM64",
```

qcom/ep_build/tasks/python.py
huaychou, line 219: Should we also `WheelPeArchT` to `typing.py` or just remove it?

qcom/build_and_test.py
```diff
@@ -820,6 +820,38 @@ def test_ort_linux_x86_64(self, plan: Plan) -> str:
...
+                    self.__venv_path,
+                    "arm64",
```
huaychou:
```suggestion
                    "aarch64_manylinux_2_34",
```

qcom/ep_build/tasks/python.py
```diff
@@ -306,11 +314,23 @@ def __find_wheel(self) -> Path:
...
+                "arm64ec": "x86_64",
+                "x86_64": "x86_64",
+            }
+            wheel_pe_arch = pe_arches[self.__target_arch]
+            build_root = REPO_ROOT / "build" / f"windows-{wheel_pe_arch}"
+            wheel_arch = "amd64" if wheel_pe_arch in ["arm64ec", "arm64x"] else wheel_pe_arch
+            filename_glob = f"{package_name}-{get_ort_version()}*-{py_vsn}-{py_vsn}-win_{wheel_arch}.whl"
+        elif is_host_linux():
+            build_root = REPO_ROOT / "build" / f"linux-{self.__target_arch}"
+            filename_glob = f"{package_name}-{get_ort_version()}*-{py_vsn}-{py_vsn}-manylinux*_aarch64.whl"
```
huaychou, line 331: Could we also add a pe_arches map like Windows host for linux host?

setup.py
huaychou: can be removed

qcom/ep_build/tasks/python.py
```diff
@@ -306,11 +314,23 @@ def __find_wheel(self) -> Path:
...
-        build_root = REPO_ROOT / "build" / f"windows-{self.__wheel_pe_arch}"
         package_name = "onnxruntime_qnn_qcom_internal"
         py_vsn = f"cp{self.__py_version.replace('.', '')}"
-        wheel_arch = "amd64" if self.__wheel_pe_arch in ["arm64ec", "arm64x"] else self.__wheel_pe_arch
-        filename_glob = f"{package_name}-{get_ort_version()}*-{py_vsn}-{py_vsn}-win_{wheel_arch}.whl"
+        if is_host_windows():
+            pe_arches = {
+                "arm64": "arm64",
+                "arm64ec": "x86_64",
+                "x86_64": "x86_64",
```
huaychou:
```suggestion
                "arm64": "arm64",
                "arm64ec": "amd64",
                "arm64x": "amd64",
                "x86_64": "amd64",
```

qcom/ep_build/tasks/python.py
```diff
@@ -306,11 +314,23 @@ def __find_wheel(self) -> Path:
...
+            }
+            wheel_pe_arch = pe_arches[self.__target_arch]
+            build_root = REPO_ROOT / "build" / f"windows-{wheel_pe_arch}"
+            wheel_arch = "amd64" if wheel_pe_arch in ["arm64ec", "arm64x"] else wheel_pe_arch
+            filename_glob = f"{package_name}-{get_ort_version()}*-{py_vsn}-{py_vsn}-win_{wheel_arch}.whl"
```
huaychou:
```suggestion
            pe_arches_to_wheel_arches = {
                "arm64": "arm64",
                "arm64ec": "amd64",
                "arm64x": "amd64", # or "arm64", not sure
                "x86_64": "amd64",
            }
            build_root = REPO_ROOT / "build" / f"windows-{self.__target_arch}"
            wheel_arch= pe_arches_to_wheel_arches[self.__target_arch]
            filename_glob = f"{package_name}-{get_ort_version()}*-{py_vsn}-{py_vsn}-win_{wheel_arch}.whl"
```

huaychou: this blocks CI checks due to arm64ec wheel not found

qcom/ep_build/tasks/python.py
qti-jkilpatrick: `typing.py` is not intended to host every type definition, just those that are broadly applicable. Since `WheelPeArchT` is Python-specific, it is intentionally defined in `python.py`.

qcom/ep_build/tasks/python.py
qti-jkilpatrick: There is only one valid arch so no.

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
qti-kromero: so before the QDC tag was only on one machine right? has it always been the case that both 01 and 05 have the QDC tag?

.github/workflows/qualcomm-internal-build-and-test.yml
qti-kromero: nit: can we add that as inline comment here

qcom/ep_build/tasks/python.py
qti-kromero: does that reasoning contradict with having the Python version in typing.py?

https://github.com/onnxruntime/onnxruntime-qnn/pull/231/changes#diff-33762fca4d0ecfb1cb1e621eec052b1eb8526e3c4b4a49ec4b0c44c3787b1dc6R10

qcom/ep_build/tasks/python.py
```diff
@@ -306,11 +314,24 @@ def __find_wheel(self) -> Path:
...
-        filename_glob = f"{package_name}-{get_ort_version()}*-{py_vsn}-{py_vsn}-win_{wheel_arch}.whl"
+        if is_host_windows():
+            pe_arches = {
+                "arm64": "arm64",
+                "arm64ec": "amd64",
+                "x86_64": "amd64",
+            }
+            wheel_pe_arch = pe_arches[self.__target_arch]
+            build_root = REPO_ROOT / "build" / f"windows-{self.__target_arch}"
+            wheel_arch = "amd64" if wheel_pe_arch in ["arm64ec", "arm64x"] else wheel_pe_arch
```
huaychou: `wheel_pe_arch` can only be `arm64` or `amd64` here, so the if else is redundant.
In definition, `wheel_arch` can only be `arm64` or `amd64`, while `wheel_pe_arch` can be `arm64`, `arm64ec`, `arm64x`, `x86_64` (or `amd64`).

cmake/CMakeLists.txt
```diff
@@ -473,6 +473,12 @@ if (MSVC OR ${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
       list(APPEND QNN_LIB_FILES ${EXTRA_HTP_LIB})
     elseif(${QNN_ARCH_ABI} STREQUAL "aarch64-ubuntu-gcc9.4")
       list(APPEND QNN_LIB_FILES "${onnxruntime_QNN_HOME}/lib/hexagon-v68/unsigned/libQnnHtpV68Skel.so")
+    elseif(${QNN_ARCH_ABI} STREQUAL "aarch64-oe-linux-gcc11.2")
+      list(APPEND QNN_LIB_FILES "${onnxruntime_QNN_HOME}/lib/hexagon-v68/unsigned/libQnnHtpV68Skel.so"
+                                "${onnxruntime_QNN_HOME}/lib/hexagon-v69/unsigned/libQnnHtpV69Skel.so"
+                                "${onnxruntime_QNN_HOME}/lib/hexagon-v73/unsigned/libQnnHtpV73Skel.so"
+                                "${onnxruntime_QNN_HOME}/lib/hexagon-v75/unsigned/libQnnHtpV75Skel.so"
```
huaychou, line 480: missing `"${onnxruntime_QNN_HOME}/lib/hexagon-v79/unsigned/libQnnHtpV79Skel.so"`

tirupath-qti: Do we have any Linux IoT SoCs require V79 and V81 today?
We can add the missing one on demand.

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
qti-jkilpatrick: I added 05 recently when I got our QDC concurrency increased.

qcom/ep_build/tasks/python.py
qti-jkilpatrick: Sure, maybe, but the idea is that we have a central place for broadly applicable types. Since we use python versions everywhere, it seems to me like that is a reasonable case for putting it in `typing.py` and not forcing imports of `python.py` everywhere.

Here's what happens in practical terms. It's not uncommon for type definitions in our build system to start life in a single file where they're used. Typically, they're defined inline close to where they're used. When they become part of the file's interface (e.g., used to define function parameters), they usually get moved to the top of the file. When they become generic enough to be used in multiple files, they tend to get moved to `typing.py`.

There is no overarching high-level design for this detail. It's driven at implementation time by a desire to expose what's needed in a tasteful way without overexposure, which can lead to confusion, misuse, and leaky abstractions. `WheelPeArchT` is actually a good example of this: other comments on this PR suggest to me that its use is not well understood, which tells me that it is a poor candidate for broader exposure in `typing.py`.

.github/workflows/qualcomm-internal-build-and-test.yml
qti-jkilpatrick: I'd actually rather not. Experience tells me that whoever re-enables the tests in `run_tests.sh` will forget to update this file. Yes, I could put a note in that file telling them to update this, but really what problem are we solving? This file is not likely to be the place someone would go to determine if or which tests are running on a given platform. For that, they'd probably just look CI's output where we have tables summarizing the situation.

qti-kromero (approved):

PR #234, opened 2026-04-10
[QNN EP] Fix crash when offload_graph_io_quantization is enabled

qti-ashwshan:
### Description
- Remove tensor name overrides that caused I/O name mismatch during SetupTensors

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_basic_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_basic_test.cc
@@ -1524,7 +1524,7 @@ static GetTestModelFn QDQBuildSigmoidForTensorNameTest(const TestInputDef<float>
 }
 
 // Test that DLC I/O tensor names match original ONNX names when offload_graph_io_quantization=1.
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1524,7 +1524,7 @@ static GetTestModelFn QDQBuildSigmoidForTensorNameTest(const TestInputDef<float>
 }
 
 // Test that DLC I/O tensor names match original ONNX names when offload_graph_io_quantization=1.
-TEST_F(QnnHTPBackendTests, OffloadGraphIoQuantizationTensorNameOverrides) {
+TEST_F(QnnHTPBackendTests, DISABLED_OffloadGraphIoQuantizationTensorNameOverrides) {
```
qti-chuteng, line 1529: Could you add the reason for the disablement in comment?

tirupath-qti (approved):

PR #241, opened 2026-04-14
[QNN EP] Enable static MSVC runtime for providers_qnn DLL

qti-ashwshan:
### Description
- Add --enable_msvc_static_runtime to build args, removing MSVCP140.dll and VCRUNTIME140.dll dependencies


qti-kromero (approved):

qti-chuteng (approved):

qcom/scripts/windows/build.ps1
```diff
@@ -136,6 +136,12 @@ $CommonArgs = `
     "--config", $Config, `
     "--parallel"
 
+# Use static MSVC runtime for ARM64 builds to eliminate MSVCP140.dll and
+# VCRUNTIME140.dll dependencies from the shipping QNN EP DLL.
+if ($Arch -in @("aarch64", "arm64", "arm64ec")) {
```
kuanyul-qti, line 141: Should we also handle `x86_64` (`amd64`)?

qti-ashwshan: This case is only when EP is loaded for JIT and we were seeing some failures when enabling for x86. Could be a follow up.

PR #249, opened 2026-04-15
Disable FP16 UT for Linux Arm64 V68 platform

huaychou:
### Description

- Move platform check to ShouldSkipIfHtpArchIsLessThanOrEqualTo

### Motivation and Context

- Disable FP16 UT for Linux Arm64 V68 platform

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,40 @@
+# Repository Guidelines
+
+## Project Structure & Module Organization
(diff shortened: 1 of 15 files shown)


AGENTS.md
qti-mbadnara, line 1: Would suggest removal of this file

huaychou:
Use https://github.com/onnxruntime/onnxruntime-qnn/pull/280 to trace.

PR #250, opened 2026-04-16
[QNN EP] Genie and HtpNetRunExtensions as part of artifacts

qti-mbadnara:
Include Genie and HtpNetRunExtensions as part of artifacts


tools/ci_build/pkg_assets.py
```diff
@@ -64,6 +64,7 @@ def get_qnn_asset_file_list():
             "QnnSaver.dll",
             "QnnSystem.dll",
             "QnnHtpPrepare.dll",
+            "QnnHtpNetRunExtensions.dll",
```
qti-kromero, line 67: nit: missed this on initial reviews but these are not in alpha order now

qti-kromero (approved):

qti-chuteng (approved):

PR #251, opened 2026-04-16
[QNN EP] Generate PDB for Windows Release builds and wrap in zip

kuanyul-qti:
### Description
Enable debug symbol generation for Release config on Windows, wrap the PDB in a zip archive, and upload the pdb zip to internal artifactory.

### Motivation and Context
There is a request to generate the .pdb with Release config.

### DIFF
--- a/tools/ci_build/pkg_assets.py
+++ b/tools/ci_build/pkg_assets.py
@@ -46,7 +46,7 @@ def parse_version_number(source_dir):
 
 def get_qnn_asset_file_list():
     """
-    Returns the list of QNN asset files to include in the zip package.
+    Returns the list of QNN asset files to include in the archive package.
 
     Args:
         is_windows_platform: If None, auto-detect from platform.system()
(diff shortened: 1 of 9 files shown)


tools/ci_build/pkg_assets.py
```diff
@@ -56,6 +56,7 @@ def get_qnn_asset_file_list():
     qnn_assets = {
         "windows": [
             "onnxruntime_providers_qnn.dll",
+            "onnxruntime_providers_qnn.pdb",
```
huaychou: Do we need this only for zip or also for wheel?

huaychou (commented):
Do we need PDB files for all Release builds, including CI, nightly, and RC? If so, how much additional CI workload would this introduce?

tools/ci_build/pkg_assets.py
kuanyul-qti: Just for zip

kuanyul-qti:
> Do we need PDB files for all Release builds, including CI, nightly, and RC? If so, how much additional CI workload would this introduce?

Yes. For each build, there isn't much additional CI workload since it only adds the build of onnxruntime_providers_qnn.pdb.

huaychou (approved):
LGTM

qti-chuteng (approved):

PR #259, opened 2026-04-16
Dev/qti kromero/genie unit tests

qti-kromero:


onnxruntime/test/providers/qnn/genie/mock_genie_version_script.lds
```diff
@@ -0,0 +1,23 @@
...
+    GenieNode_create;
+    GenieNode_setData;
+    GenieNode_getData;
+    GenieNode_execute;
+    GenieNode_free;
+    GenieNode_reset;
+    GenieNodeConfig_free;
+    GenieLog_create;
+    GenieNodeConfig_bindLogger;
+    GenieLog_free;
```
qti-kromero, line 19: **Bug: symbol name mismatch**

`GetMockGenieCalls` here does not match the actual exported function name `GetMockGenieCallCount` in `genie_mock_dll.cc`. This will cause a linker failure when Linux support is enabled.

Fix: `GetMockGenieCalls;` → `GetMockGenieCallCount;`

onnxruntime/test/providers/qnn/genie/genie_mock_dll.cc
```diff
@@ -0,0 +1,190 @@
...
+  if (it == g_calls.end()) return 0;
+  return it->second.load();
+}
+
+MOCK_EXPORT
+void ResetMockGenieCalls() {
+  std::lock_guard<std::mutex> lock(g_call_mutex);
+  g_calls.clear();
+}
+
```
qti-kromero, line 57: **Potential data race in stub counters**

The stub functions do `++g_calls["NodeConfig_createFromJson"]` (and similar) without holding `g_call_mutex`. `std::unordered_map` is not thread-safe for concurrent inserts even when the value type is `std::atomic<int>`. If the Genie EP calls multiple stubs from different threads, this is a data race.

Options:
- (a) Protect every `++g_calls[...]` with `g_call_mutex`, or  
- (b) Pre-populate all expected keys in a static initialiser so that subsequent operations are purely atomic reads/increments without any map structural change.

The current mutex only guards `GetMockGenieCallCount` and `ResetMockGenieCalls`, leaving the write path unprotected.

onnxruntime/test/providers/qnn/genie/genie_mock_dll.cc
```diff
@@ -0,0 +1,190 @@
...
+Genie_Status_t GenieNode_getData(const GenieNode_Handle_t /*nodeHandle*/,
+                                 const GenieNode_IOName_t /*nodeIOName*/,
+                                 const char* /*ioConfig*/,
+                                 GenieNode_IOCallback_t ioCallback,
+                                 const void* userData) {
+  ++g_calls["Node_getData"];
+  // Invoke the callback with a minimal fake float payload so that ComputeImpl
+  // can populate the ORT output tensor without error.
```
qti-kromero, line 133: **`GenieNode_getData` callback hardcodes output shape `[1,1]`**

The shape `{"dimensions": [1,1]}` in the IOCallback is tightly coupled to `CreateDlcContextGraph()`'s declared output `{1, 1, 1}`. Any future integration test that declares a different graph output shape will silently receive wrong tensor dimensions.

Consider either making the output config configurable via a settable global, or at minimum adding a comment here that explicitly links this to `CreateDlcContextGraph`'s specific shape.

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,265 @@
...
+#include <gtest/gtest.h>
+#include <memory>
+#include <string>
+
+#if defined(_WIN32)
+#include <windows.h>
+static void* LoadMockLib() { return LoadLibraryA("MockGenie.dll"); }
+static void* GetSym(void* h, const char* s) { return reinterpret_cast<void*>(GetProcAddress(static_cast<HMODULE>(h), s)); }
+static constexpr const char* kMockGeniePath = "MockGenie.dll";
+#else
```
qti-kromero, line 13: **`RTLD_NOLOAD` will return NULL when called before session creation**

`LoadMockLib()` on Linux uses `dlopen("libMockGenie.so", RTLD_NOW | RTLD_NOLOAD)`. `RTLD_NOLOAD` only returns a handle if the library is *already mapped* into the process. Tests like `MockGenie_CallTrackingFunctions_Resolvable` and `CreateState_InvokesExpectedApiSequence` call `LoadMockLib()` *before* `MakeGenieSession()`, so at that point `libMockGenie.so` is not yet loaded and `dlopen` returns NULL. The subsequent `GetSym(nullptr, ...)` and `ASSERT_NE(handle, nullptr)` will fail.

This is currently masked by `GTEST_SKIP()` in `GenieSessionTest::SetUp()` on non-Windows, but it will break when Linux support is enabled.

Fix: use `RTLD_NOW` (without `RTLD_NOLOAD`) on Linux, or reorder the test to call `MakeGenieSession()` first so the library is already mapped when `LoadMockLib()` is called.

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,265 @@
...
+    attrs.push_back(builder.MakeStringAttribute("source", "QAIRTExport"));
+    attrs.push_back(builder.MakeStringAttribute("ep_dlc_context", "model.dlc"));
+    builder.AddNode("GenieDlcContextNode", "EPContext",
+                    {"genie_input"}, {"genie_output"}, kMSDomain, attrs);
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Shared helper: builds the serialised model bytes and creates a session
+// backed by MockGenie. Returns the session; registered_ep_device lifetime
```
qti-kromero, line 79: **`kMSDomain` shadows the constant already provided by includes**

`static constexpr const char* kMSDomain = "com.microsoft"` re-declares a constant that is already available through `core/graph/constants.h` (included transitively). This local declaration shadows the global one. Remove it and use the already-available `kMSDomain`.

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,265 @@
...
+  RegisteredEpDeviceUniquePtr ep1;
+  Ort::Session session1 = MakeGenieSession(*env_, ep1);
+  reset();
+
+  // Second session with the same process reuses the loaded library.
+  // SetupBackend() returns early, so no second library-init path is triggered.
+  // DlcConfig_create is called once per node (session-level), not per library load.
+  RegisteredEpDeviceUniquePtr ep2;
+  Ort::Session session2 = MakeGenieSession(*env_, ep2);
+
```
qti-kromero, line 257: **Loose assertion does not verify idempotency**

`EXPECT_GE(get_count("DlcConfig_create"), 1)` only asserts the count is non-zero — it would still pass if `SetupBackend()` ran twice and the count was 2. The inline NOTE acknowledges this, but an unfixed weak assertion is just a passing test that proves nothing about the described behaviour.

Before promoting to non-draft, either:
- Determine whether `GenieBackendManager` is EP-scoped or session-scoped and assert the exact expected count (1 or 2), or  
- Replace the test body with `SUCCEED() << "TODO: tighten assertion"` to make the gap explicit without misleading a reader into thinking this is verified.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -601,6 +601,29 @@ void QnnCPUBackendTests::SetUp() {
   }
 }
 
+static BackendSupport GetGenieSupport() {
```
qti-kromero, line 604: **`GetGenieSupport()` always returns `SUPPORTED` — fixture never skips**

Unlike `QnnHTPBackendTests` or `QnnGPUBackendTests` which probe actual hardware, `GetGenieSupport()` unconditionally returns `SUPPORTED`. This means `GenieBackendTests`-derived tests can never be skipped on the basis of backend availability.

If this is intentional (the unit tests use mocks and require no hardware), add a comment explaining why. Otherwise implement a real availability probe to match the pattern used by the other backends.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -528,6 +533,15 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     }
 
     ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, ("Using backend path: " + backend_path).c_str());
+
+    // If the resolved path is the Genie library (either via backend_type="genie" or a direct
+    // backend_path pointing to the Genie library), store it for use in GetGenieCapability().
+    // Otherwise, keep the default Genie path so the Genie pathway always has a valid path.
+    if (backend_path == kDefaultGenieBackendPath ||
+        (!backend_path_option.empty() && backend_path == backend_path_option &&
```
qti-kromero, line 541: **Genie path detection is case-sensitive and fragile**

`backend_path.find("Genie") != std::string::npos` is a case-sensitive substring match. A library named `genie_backend.dll`, `libgenie.so`, or an all-lowercase path from a user-supplied `backend_path` option would not match and `genie_backend_path_` would silently stay as the default instead of the user's path.

At this point in the code `backend_type_name` is already available. Consider using `backend_type_name == kGenieBackendTypeName` as the primary condition — it's already resolved and unambiguous — rather than doing a substring search on the resolved path string.

onnxruntime/core/providers/qnn/ort_api.h
```diff
@@ -89,10 +125,25 @@ namespace onnxruntime {
...
-// Convenient macro for logging with an Ort::Logger pointer, espeically in QnnBackendManager.
+// Returns true if an Ort::Logger has a null internal OrtLogger pointer (i.e., was default-constructed
+// and never initialized). Ort::Logger is standard-layout and trivially copyable; its first member
+// (const OrtLogger* logger_) is at offset 0, so memcpy is well-defined here.
+inline bool IsNullLogger(const Ort::Logger& logger) {
+  const OrtLogger* ptr = nullptr;
+  std::memcpy(&ptr, &logger, sizeof(ptr));
+  return ptr == nullptr;
+}
+
```
qti-kromero, line 136: **`IsNullLogger` depends on `Ort::Logger`'s undocumented internal layout**

The `memcpy` approach works because `logger_` is the first member of `Ort::Logger` today, but this is not guaranteed — `Ort::Logger` is not a standard-layout type and the C++ standard does not guarantee that the first data member is at offset 0 for non-trivial types.

Consider adding a `static_assert(offsetof(Ort::Logger, logger_) == 0, ...)` (requires making the member accessible or using `std::bit_cast` in C++20) or filing an upstream ORT issue requesting a public `IsNull()` API on `Ort::Logger`.

onnxruntime/test/util/test_environment.cc
```diff
@@ -21,6 +22,10 @@ const ::onnxruntime::Environment& GetEnvironment() {
   return ((OrtEnv*)*ort_env.get())->GetEnvironment();
 }
 
+::onnxruntime::logging::LoggingManager& DefaultLoggingManager() {
```
qti-kromero, line 25: **`DefaultLoggingManager()` appears unused in this PR**

This function is added here but has no declaration in a header and no visible callsite in this diff. If it was added for future use or removed from a prior version, either add the header declaration and usage or remove it to keep dead-code from accumulating.

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -212,3 +212,44 @@
...
+
+    # Copy MockGenie next to the test executable so GenieBackendManager
+    # finds it by name when backend_path="MockGenie.dll".
+    add_custom_command(
+      TARGET MockGenie POST_BUILD
+      COMMAND ${CMAKE_COMMAND} -E copy_if_different
+        $<TARGET_FILE:MockGenie>
+        $<TARGET_FILE_DIR:onnxruntime_provider_test>
+      COMMENT "Copying MockGenie to test output directory"
+    )
```
qti-kromero, line 254: **Nit: `endif()` missing an inline comment**

Other `endif()` blocks in this file are annotated (e.g., `endif() # onnxruntime_USE_QNN`). The closing `endif()` for the `if(WIN32)` MockGenie block should follow the same convention for readability, e.g., `endif()  # WIN32 (MockGenie)`.

qti-kromero (commented):
Review of the Genie unit test infrastructure. Found 3 bugs and 8 design/nit issues. Key items to address before promoting from draft:

**🔴 Bugs:**
1. Symbol name mismatch in `mock_genie_version_script.lds` (`GetMockGenieCalls` vs `GetMockGenieCallCount`) — will break Linux linker.
2. Data race in mock stub counters — `++g_calls[...]` increments happen outside `g_call_mutex`.
3. `RTLD_NOLOAD` in Linux `LoadMockLib()` returns NULL when called before session creation.

**🟡 Design:**
4. Genie path detection (`find("Genie")`) is case-sensitive/fragile.
5. `GetGenieSupport()` unconditionally returns `SUPPORTED` — differs from other backend fixtures.
6. `SetupBackend_AlreadySetup_IsIdempotent` uses `EXPECT_GE(..., 1)` which doesn't verify idempotency.
7. `GenieNode_getData` hardcodes output shape tightly coupled to one test graph.

**🔵 Nits:**
8. `kMSDomain` locally redeclared, shadowing the already-included constant.
9. `DefaultLoggingManager()` added but unused.
10. `IsNullLogger` relies on undocumented `Ort::Logger` memory layout.
11. `endif()` in CMake missing an inline annotation comment.

onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
```diff
@@ -52,24 +52,32 @@ OrtStatus* GenieNodeComputeInfo::CreateStateImpl(OrtNodeComputeInfo* this_ptr,
...
         }
       }
+    } catch (const std::filesystem::filesystem_error& e) {
+      std::string error_msg = std::string("Error searching for extension file: ") + e.what();
+      return ep.ort_api.CreateStatus(ORT_EP_FAIL, error_msg.c_str());
     }
-  } catch (const std::filesystem::filesystem_error& e) {
-    std::string error_msg = std::string("Error searching for extension file: ") + e.what();
-    return ep.ort_api.CreateStatus(ORT_EP_FAIL, error_msg.c_str());
   }
```
qti-kromero, line 81: add warning here that extension path was not found and we are proceeding without it

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -56,13 +56,15 @@ static bool ParseBackendTypeName(std::string_view backend_type_name,
                                  std::string& backend_path,
                                  const Ort::Logger& logger) {
   constexpr std::string_view kCpuBackendTypeName{"cpu"};
+  constexpr std::string_view kGenieBackendTypeName{"genie"};
```
qti-kromero, line 59: need to look into why this pathway was re-added

qti-kromero:
this PR just got absolutely destroyed by my force push - abandoning

qti-kromero:
this PR just got absolutely destroyed by my force push - abandoning

PR #242, opened 2026-04-14
Enable Linux Arm64 to build wheel and tgz artifacts

huaychou:
### Description

- Add argument to build tgz artifacts for Linux script
- Add all skel and stub into zip and wheel
- Enable Linux arm64 release pipeline

### Motivation and Context

- Enable Linux Arm64 to build wheel and tgz artifacts

PR #260, opened 2026-04-17
[QNN EP] Update wheel count for release

qti-mbadnara:
Update wheel count from 8 to 16 for internal release accounting for the merge

### DIFF
--- a/.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
+++ b/.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
@@ -159,10 +159,10 @@ jobs:
 
       - name: Verify wheel artifacts
         run: |
-          # After merge: 4 arm64 + 4 merged AMD wheels in build/dist/
+          # After merge: 4 arm64 + 4 x86_64 + 4 arm64ec + 4 merged wheels in build/dist/
           count=$(find "${{ github.workspace }}/build/dist" -name "*.whl" | wc -l)
           echo "Found $count wheel(s) in build/dist"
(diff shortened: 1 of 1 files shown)


.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -159,10 +159,10 @@ jobs:
 
       - name: Verify wheel artifacts
         run: |
-          # After merge: 4 arm64 + 4 merged AMD wheels in build/dist/
+          # After merge: 4 arm64 + 4 x86_64 + 4 arm64ec + 4 merged wheels in build/dist/
```
qti-ashwshan, line 162: AMD name can be kept the same instead of x86_64.

qti-ashwshan (approved):

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
qti-mbadnara: Will update this as part of the mainline when Linux changes are also incorporated

qti-kromero:
would like to echo Yuduo's comment on previous PR that we should generally not merge things directly to the release branch without it being on mainline

qti-mbadnara:
> would like to echo Yuduo's comment on previous PR that we should generally not merge things directly to the release branch without it being on mainline

I agree. This change is specific to the release branch.  Changes as part of this PR: https://github.com/onnxruntime/onnxruntime-qnn/pull/242 are being made to make the corresponding changes to the mainline.

tirupath-qti (approved):

PR #212, opened 2026-04-07
Avoid private ORT Core header in QNN-EP

qti-hungjuiw:
### Description

Remove dependencies on ORT Core private headers from the QNN Execution Provider (EP). 
- Removed unnecessary includes of `error_code_helper.h`, `provider_factory_creators.h`, and `abi_devices.h`
- Copied required utilities (macros from `TraceLoggingConfig.h`, container types from              `inlined_containers.h`/`inlined_containers_fwd.h`, int4 types from `int4.h`) directly into the QNN EP   under `core/providers/qnn/common/`, eliminating the dependency on ORT Core internal paths.
- Decoupled the QNN EP include directory from ORT Core's internal `core/common/` and `core/framework/` paths.

### Motivation and Context

The QNN EP can be built as a standalone shared library distributed independently of ORT Core. Depending on ORT Core's private/internal headers creates problems:                                                                                                                                                                                         

- **Fragility**: Private headers are implementation details of ORT Core and can change or be         removed without notice, breaking the QNN EP build.
- **Tight coupling**: It forces the QNN EP to track ORT Core's internal refactors even when the      EP's public behavior is unchanged.

PR #261, opened 2026-04-17
Genie backend unit and integration tests

qti-kromero:
### Description
Add Genie backend unit and integration tests
Attempts to clean up some API files where needed

### Motivation and Context
Helps ensure code quality of Genie integration feature; regressions less likely to be introduced


onnxruntime/test/providers/qnn/genie/mock_genie_version_script.lds
```diff
@@ -0,0 +1,23 @@
...
+    GenieNode_create;
+    GenieNode_setData;
+    GenieNode_getData;
+    GenieNode_execute;
+    GenieNode_free;
+    GenieNode_reset;
+    GenieNodeConfig_free;
+    GenieLog_create;
+    GenieNodeConfig_bindLogger;
+    GenieLog_free;
```
qti-kromero: **Bug: Symbol name mismatch**

This exports `GetMockGenieCalls` but the actual exported function in `genie_mock_dll.cc` is named `GetMockGenieCallCount` (with "Count"). The `.def` file has the correct name. When the Linux path is enabled, the linker will export the wrong (non-existent) symbol and tests that call `GetSym(h, "GetMockGenieCallCount")` will get null back.

```
GetMockGenieCallCount;  // <- should match genie_mock_dll.cc
```

onnxruntime/test/providers/qnn/genie/genie_mock_dll.cc
```diff
@@ -0,0 +1,190 @@
...
+#if defined(_WIN32)
+#define MOCK_EXPORT __declspec(dllexport)
+#else
+#define MOCK_EXPORT __attribute__((visibility("default")))
+#endif
+
+// ---------------------------------------------------------------------------
+// Call tracking — one counter per Genie API function.
+// Exported so the test EXE can read them after session creation.
+// ---------------------------------------------------------------------------
```
qti-kromero, line 36: **Bug: Race condition in call tracking**

The API stubs (e.g. `++g_calls["NodeConfig_createFromJson"]`) access the shared `std::unordered_map` *without* holding `g_call_mutex`. `std::atomic<int>` makes the integer increment safe, but looking up / inserting a key in `unordered_map` is not thread-safe on its own. `GetMockGenieCallCount` and `ResetMockGenieCalls` do lock the mutex, so there is a genuine TOCTOU race if two stubs are invoked concurrently (e.g. from multi-threaded ORT sessions).

Simplest fix: acquire `g_call_mutex` in each stub before touching the map:
```cpp
static Genie_Status_t GenieNodeConfig_createFromJson(...) {
  {
    std::lock_guard<std::mutex> lock(g_call_mutex);
    ++g_calls["NodeConfig_createFromJson"];
  }
  ...
}
```
Or pre-allocate a fixed set of `std::atomic<int>` counters (one per function) to avoid the map altogether.

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,265 @@
...
+
+#if defined(_WIN32)
+#include <windows.h>
+static void* LoadMockLib() { return LoadLibraryA("MockGenie.dll"); }
+static void* GetSym(void* h, const char* s) { return reinterpret_cast<void*>(GetProcAddress(static_cast<HMODULE>(h), s)); }
+static constexpr const char* kMockGeniePath = "MockGenie.dll";
+#else
+#include <dlfcn.h>
+static void* LoadMockLib() { return dlopen("libMockGenie.so", RTLD_NOW | RTLD_NOLOAD); }
+static void* GetSym(void* h, const char* s) { return dlsym(h, s); }
```
qti-kromero, line 16: **Bug: `RTLD_NOLOAD` causes null handle before library is loaded**

`RTLD_NOW | RTLD_NOLOAD` returns null unless the library is *already mapped* in the process. Tests that call `LoadMockLib()` **before** `MakeGenieSession()` (e.g. `MockGenie_CallTrackingFunctions_Resolvable`, `CreateState_InvokesExpectedApiSequence`) will receive a null handle here, then hit `ASSERT_NE(dll_handle, nullptr)` with a confusing failure rather than a clean skip.

Remove `RTLD_NOLOAD` so the library is actually opened:
```cpp
static void* LoadMockLib() { return dlopen("libMockGenie.so", RTLD_NOW); }
```

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,265 @@
...
+  // Second session with the same process reuses the loaded library.
+  // SetupBackend() returns early, so no second library-init path is triggered.
+  // DlcConfig_create is called once per node (session-level), not per library load.
+  RegisteredEpDeviceUniquePtr ep2;
+  Ort::Session session2 = MakeGenieSession(*env_, ep2);
+
+  // NOTE: If GenieBackendManager is EP-level (shared across sessions), this
+  // count will be 1. If it is session-level, it will be 1 per session.
+  // Verify the exact scope in qnn_execution_provider.cc and tighten as needed.
+  EXPECT_GE(get_count("DlcConfig_create"), 1);
```
qti-kromero: **Maintenance: weak test assertion — pin the expected count**

`EXPECT_GE(..., 1)` with the inline comment "verify the exact scope...and tighten as needed" means this test will pass whether the count is 1, 2, or 100. That provides almost no regression protection. Before merging, determine whether `GenieBackendManager` is EP-level or session-level and pin the expected value to that. If the behavior is genuinely still undecided, mark the test `DISABLED_` until it can be properly specified.

onnxruntime/core/providers/qnn/ort_api.h
```diff
@@ -89,10 +125,25 @@ namespace onnxruntime {
...
 
-// Convenient macro for logging with an Ort::Logger pointer, espeically in QnnBackendManager.
+// Returns true if an Ort::Logger has a null internal OrtLogger pointer (i.e., was default-constructed
+// and never initialized). Ort::Logger is standard-layout and trivially copyable; its first member
+// (const OrtLogger* logger_) is at offset 0, so memcpy is well-defined here.
+inline bool IsNullLogger(const Ort::Logger& logger) {
+  const OrtLogger* ptr = nullptr;
+  std::memcpy(&ptr, &logger, sizeof(ptr));
+  return ptr == nullptr;
+}
```
qti-kromero, line 112: **Design concern: `IsNullLogger` reads `Ort::Logger` internals via `memcpy`**

This reads the private layout of `Ort::Logger` by copying `sizeof(ptr)` bytes from its address and treating them as `const OrtLogger*`. Two fragile assumptions:
1. `logger_` is at offset 0 in `Ort::Logger`
2. The ORT C++ API class layout won't change (vtable, padding, leading member changes)

Neither is part of the public ABI contract. If the implementation changes this silently reads the wrong bytes.

At minimum, add a static_assert to catch layout changes at compile time:
```cpp
static_assert(offsetof(Ort::Logger, logger_) == 0,
              "IsNullLogger assumes logger_ is at offset 0");
```
Better long-term: pass `Ort::Logger*` (nullable pointer) to callers instead of `const Ort::Logger&`, making a null check direct and obvious without ABI assumptions.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -601,6 +601,29 @@ void QnnCPUBackendTests::SetUp() {
   }
 }
```
qti-kromero, line 604: **Design concern: `GetGenieSupport()` always returns `SUPPORTED` — availability check is a no-op**

`GetGenieSupport()` unconditionally returns `BackendSupport::SUPPORTED`. As a result:
- `SetUp()` returns immediately on the first call (early-return at the top)
- The `SUPPORT_ERROR` / `UNSUPPORTED` branches are dead code
- The `Ort::Logger logger = Ort::Logger()` and subsequent conditionals are unreachable

If Genie is always available in the test environment that runs these tests, simplify `SetUp()` to a no-op and remove the caching scaffolding. If Genie availability *should* be probed (e.g. by attempting `dlopen` of the real library), implement that here rather than returning a hard-coded value.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -528,6 +533,15 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     }
 
     ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, ("Using backend path: " + backend_path).c_str());
+
+    // If the resolved path is the Genie library (either via backend_type="genie" or a direct
+    // backend_path pointing to the Genie library), store it for use in GetGenieCapability().
+    // Otherwise, keep the default Genie path so the Genie pathway always has a valid path.
+    if (backend_path == kDefaultGenieBackendPath ||
+        (!backend_path_option.empty() && backend_path == backend_path_option &&
+         backend_path.find("Genie") != std::string::npos)) {
```
qti-kromero: **Design concern: fragile Genie path detection via substring search**

`backend_path.find("Genie") != std::string::npos` is case-sensitive and matches any path containing the substring "Genie", including paths like `/opt/MyGenieWrapper/htp.so`.

A more robust approach: set `genie_backend_path_` at the point where the user *explicitly* selects the Genie backend type (i.e. inside the `backend_type_name == kGenieBackendTypeName` branch already added above), rather than retroactively inferring it from the resolved path string. This eliminates the ambiguity entirely.

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -212,3 +212,45 @@
...
+    set_target_properties(MockGenie PROPERTIES
+      CXX_STANDARD 17
+      CXX_STANDARD_REQUIRED ON
+      FOLDER "ONNXRuntimeTest"
+    )
+
+    set_property(TARGET MockGenie APPEND_STRING PROPERTY LINK_FLAGS
+      " -DEF:${ONNXRUNTIME_ROOT}/test/providers/qnn/genie/mock_genie_symbols.def")
+
+    # else()
```
qti-kromero: **Maintenance: remove or implement the commented-out `else()` branch**

A multi-line block of commented-out CMake is easy to forget about and goes stale as paths and variables evolve. Two better options:

1. **If Linux support is planned soon**: convert to a `TODO` referencing a tracking issue:
   ```cmake
   # TODO(#NNN): Enable for Linux once the Genie pathway is validated.
   # target_link_options(MockGenie PRIVATE "-Wl,--version-script=...")
   ```
2. **If timeline is uncertain**: delete the block — `mock_genie_version_script.lds` is checked in and git history preserves the intent.

qti-kromero (commented):
## Review: Genie unit test infrastructure

Good overall direction — the MockGenie approach is clean, the defensive null-function-pointer guards in `genie_node.cc` are a genuine correctness improvement, the filesystem path guard in `genie_node_compute_info.cc` is a real robustness fix, and removing the copy-pasted utilities from `ort_api.h` in favour of the real headers is good housekeeping. The SafeInt handler explanation is well-documented.

A few things need attention before this is ready to merge:

**Bugs (must fix)**
1. Symbol name mismatch in `mock_genie_version_script.lds` — exports `GetMockGenieCalls` but the function is `GetMockGenieCallCount` (see inline comment).
2. Race condition in `genie_mock_dll.cc` — map accessed without mutex in the stubs, while `GetMockGenieCallCount`/`ResetMockGenieCalls` do lock it.
3. `RTLD_NOLOAD` on Linux causes a null handle before the library is loaded, breaking `MockGenie_CallTrackingFunctions_Resolvable` and `CreateState_InvokesExpectedApiSequence`.

**Design (should address)**
4. `IsNullLogger` reads `Ort::Logger` internals via `memcpy` — relies on undocumented ABI; at minimum add a `static_assert` on the offset.
5. `GetGenieSupport()` unconditionally returns `SUPPORTED`, making the availability check a no-op and leaving dead branches in `SetUp()`.
6. Genie path detection via `backend_path.find("Genie")` substring search is fragile — prefer setting `genie_backend_path_` at the explicit `kGenieBackendTypeName` parse site.

**Maintenance**
7. Commented-out CMake `else()` block should be removed or converted to a `TODO` with a tracking issue.
8. Weak `EXPECT_GE(..., 1)` in `SetupBackend_AlreadySetup_IsIdempotent` — pin the expected count once the lifecycle is defined.

**Minor:** the PR description is currently empty. Given 14 changed files and 1149 additions a short summary of the motivation, the MockGenie design rationale, and known limitations (Windows-only, draft status) would help future reviewers significantly.

onnxruntime/test/providers/qnn/genie/genie_mock_dll.cc
qti-kromero: don't really like the resolution of this as currently proposed - need to look at it more to maybe have some loops,etc.

cmake/onnxruntime_unittests.cmake
```diff
@@ -303,6 +303,20 @@ block()
     ${onnxruntime_test_framework_src}
   )
 
+  if(onnxruntime_USE_QNN)
```
qti-kromero, line 306: don't really like this here - need to look at why this isn't needed for other test sources

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -212,3 +212,41 @@
           LIBRARY   DESTINATION ${CMAKE_INSTALL_LIBDIR}
           RUNTIME   DESTINATION ${CMAKE_INSTALL_BINDIR}
           FRAMEWORK DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+  # ---------------------------------------------------------------------------
```
qti-kromero: are there other mocks in this file? should it be moved to a different file?

onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
```diff
@@ -52,24 +52,32 @@ OrtStatus* GenieNodeComputeInfo::CreateStateImpl(OrtNodeComputeInfo* this_ptr,
...
         }
       }
+    } catch (const std::filesystem::filesystem_error& e) {
+      std::string error_msg = std::string("Error searching for extension file: ") + e.what();
+      return ep.ort_api.CreateStatus(ORT_EP_FAIL, error_msg.c_str());
     }
-  } catch (const std::filesystem::filesystem_error& e) {
-    std::string error_msg = std::string("Error searching for extension file: ") + e.what();
-    return ep.ort_api.CreateStatus(ORT_EP_FAIL, error_msg.c_str());
   }
```
qti-kromero, line 85: add warning that we are continuing without using HTP extensions here

onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
```diff
@@ -52,24 +52,32 @@ OrtStatus* GenieNodeComputeInfo::CreateStateImpl(OrtNodeComputeInfo* this_ptr,
...
+  const bool parent_is_searchable = !parent_folder_path.empty() &&
+                                    std::filesystem::is_directory(parent_folder_path, dir_ec) &&
+                                    !dir_ec;
+  if (parent_is_searchable) {
+    try {
+      for (const auto& entry : std::filesystem::directory_iterator(parent_folder_path)) {
+        if (entry.is_regular_file()) {
+          std::string filename = entry.path().filename().string();
+          // Check if filename matches pattern tmp**.json
+          if (filename.size() >= 8 &&  // "tmp" + some random number + ".json"
```
qti-kromero: can this be cleaned up via some std library functions? maybe a regex matcher of some kind?

onnxruntime/core/providers/qnn/ort_api.h
```diff
@@ -3,13 +3,47 @@
...
+// We intentionally do NOT use core/common/safeint.h (which uses ORT_THROW → ORT_WHERE_WITH_STACK →
+// GetStackTrace()) because onnxruntime::GetStackTrace is a LOCAL symbol in libonnxruntime.so and is
+// not available to plugin EPs at runtime. Using std::runtime_error here avoids the
+// undefined-symbol dependency while still throwing on integer overflow/divide-by-zero.
+//
+// In test builds, core/common/safeint.h may be included transitively before this header. In that
+// case SafeIntDefaultExceptionHandler is already defined (using ORT_THROW), so we skip our
+// definition to avoid a redefinition conflict. The ORT_THROW-based handler is safe in tests
+// because GetStackTrace is available when linking against ort_core.
+#ifndef SafeIntDefaultExceptionHandler
```
qti-kromero: it feels like there should be a way to clean this up...isn't getting it from core/common/safeint.h pretty bad in terms of maintaining ABI-compatibility?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -52,6 +52,8 @@ const std::string kDefaultHtpBackendPath = MakeSharedLibraryPath("QnnHtp");
 const std::string kDefaultSaverBackendPath = MakeSharedLibraryPath("QnnSaver");
 const std::string kDefaultIrBackendPath = MakeSharedLibraryPath("QnnIr");
 
+constexpr std::string_view kGenieBackendTypeName{"genie"};
```
qti-kromero, line 58: where are the rest of these stored and why can't genie go with them?

onnxruntime/test/providers/qnn/genie/genie_mock_dll.cc
qti-kromero: can this use loops?

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,279 @@
...
+#include <dlfcn.h>
+static void* LoadMockLib() { return dlopen("libMockGenie.so", RTLD_NOW); }
+static void* GetSym(void* h, const char* s) { return dlsym(h, s); }
+static constexpr const char* kMockGeniePath = "libMockGenie.so";
+#endif
+
+#include "test/providers/qnn/qnn_test_utils.h"
+#include "core/session/onnxruntime_cxx_api.h"
+#include "core/session/onnxruntime_session_options_config_keys.h"
+#include "core/session/inference_session.h"
```
qti-chuteng: is this header used?

onnxruntime/core/providers/qnn/qnn_safeint.h
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
```
qti-chuteng: this is new file, should it use below copyright?

 // Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. // SPDX-License-Identifier: MIT

onnxruntime/test/providers/qnn/genie_integration_test.cc
qti-kromero: removed, it compiles without it on Windows ARM, let's see about the rest via CI

qti-mbadnara:
LGTM

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -222,6 +222,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
   mutable std::optional<std::pair<std::vector<std::string>, std::vector<std::string>>> onnx_graph_io_names_;
 
   // Genie pathway-specific variables
+  std::string genie_backend_path_;
```
quic-calvnguy: If CPU, HTP, IR, GPU all use the same backend_type EP option, and the genie backend path is determined via backend_type, do we need this var?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -513,11 +521,15 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     if (!backend_type.empty()) {
       if (std::string parsed_backend_path; ParseBackendTypeName(backend_type, parsed_backend_path, logger_)) {
```
quic-calvnguy, line 533: Reference for:
https://github.com/onnxruntime/onnxruntime-qnn/pull/261/changes#r3132848101

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Test: session Run() succeeds end-to-end through the mock Genie call flow.
+// MockGenie callback returns outputConfig "[1,1]"; ComputeImpl inserts a 1
+// at index 1, so the final output shape is {1, 1, 1}.
+// ---------------------------------------------------------------------------
+TEST_F(GenieSessionTest, Run_ProducesExpectedOutputShape) {
+  RegisteredEpDeviceUniquePtr registered_ep_device;
+  Ort::Session session = MakeGenieSession(*env_, registered_ep_device);
```
quic-calvnguy: One issue we ran into recently with the non-genie path is that if for whatever reason the session is destroyed after the EP is unregistered, there will be an SEH exception 0xc0000005 that will prevent future registration of the QNN EP for all unit tests thereafter. This may not be an issue now, but it may come up in the future if the timing of implicit destruction is affected somehow.

quic-calvnguy: PR from @huaychou for reference:
https://github.com/onnxruntime/onnxruntime-qnn/pull/175

This change was included in the parallel graph prepare work as this is where the issue was first observed (first 2 March 26 commits):
https://github.com/onnxruntime/onnxruntime-qnn/pull/117/commits

onnxruntime/core/providers/qnn/qnn_execution_provider.h
qti-kromero: yeah...I think this was originally needed due to the test overrides but I think I found a way to do without changing the private header contents which is cleaner and I like it more. tests/users can still override the backend path if they really want to via "backend_path" session options

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-kromero: see updated handling of this after new patch

onnxruntime/test/providers/qnn/genie_integration_test.cc
qti-kromero: added necessary brackets to ensure session is properly scoped

created the following ticket to investigate this issue further as possible ORT Core issue: https://jira-dc.qualcomm.com/jira/browse/AISW-181399

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -47,6 +47,7 @@
 
   message(STATUS ONNXRUNTIME_APPLICATION_SOURCE_ROOT ${ONNXRUNTIME_APPLICATION_SOURCE_ROOT})
   target_include_directories(onnxruntime_providers_qnn PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
+                                                               ${ONNXRUNTIME_APPLICATION_INCLUDE_ROOT}
```
minfhong-qti: After #212, we should not use headers other than those in "core/session" as those are private ones and not released.
cc. @qti-hungjuiw

onnxruntime/core/providers/qnn/ort_api.h
```diff
@@ -90,10 +107,31 @@ namespace onnxruntime {
...
+  std::memcpy(&ptr, &logger, sizeof(ptr));
+  return ptr == nullptr;
+}
+
+// Convenient macro for logging with an Ort::Logger pointer, especially in QnnBackendManager.
 // This macro avoids the necessity of parentheses (i.e., (*logger_ptr)) in every ORT_CXX_LOG call.
+// Guards against a null-constructed Ort::Logger (no-op when logger is uninitialized).
 // This macro can be removed once ORT_CXX_LOG is fixed to properly wrap given logger with parentheses.
-#define ORT_CXX_LOG_PTR(logger_ptr, message_severity, message) ORT_CXX_LOG((*logger_ptr), message_severity, message)
+#define ORT_CXX_LOG_PTR(logger_ptr, message_severity, message) \
```
minfhong-qti, line 118: Curious the motivation behind the change here and whether still needed.

cmake/onnxruntime_providers_qnn.cmake
qti-kromero: fairly certain this is resolved in latest patch, let me know if something still looks off @minfhong-qti @qti-hungjuiw

onnxruntime/core/providers/qnn/ort_api.h
qti-kromero: this was something that Claude found when working on my PR, this is the response for why it is needed:
"
The macro was needed because ORT_CXX_LOG uses . on its logger arg and therefore can't accept a pointer (logger_ptr_ is const Ort::Logger*)
The two guards (nullptr check and IsNullLogger) are both necessary: the pointer can be null, and Ort::Logger can be default-constructed with a null internal pointer (which would crash)
The code comment already documents when it can be removed (once ORT_CXX_LOG is fixed upstream to parenthesize its logger arg), but that hasn't happened yet
"

the proper fix requires a fix in the upstream ORT Core repo for ORT_CXX_LOG

"
Now I have the complete picture. Here's the answer:

Partially — the two null checks in ORT_CXX_LOG_PTR have different origins, and they need different fixes:

Check 1: !IsNullLogger(*logger_ptr) (default-constructed logger)

Ort::Logger's private member is const OrtLogger* logger_{} — zero-initialized to nullptr on default construction. The IsNullLogger function reads that private field via memcpy because there's no public API to check validity.

The clean fix in ORT Core would be to add explicit operator bool() to Ort::Logger:


// In onnxruntime_cxx_api.h, inside struct Logger
explicit operator bool() const noexcept { return logger_ != nullptr; }
This is a natural addition — Logger() and Logger(nullptr) already exist as "empty" constructors, and the class documentation even says "Must be initialized from a valid Ort::Logger before use." An operator bool just exposes that validity contract publicly. With this, QNN EP's IsNullLogger, the memcpy, and the static_assert could all be removed.

Check 2: (logger_ptr) (null pointer check)

This can't be moved into ORT_CXX_LOG since that macro takes a reference, not a pointer. ORT Core could add a companion macro, but that's a QNN EP-specific usage pattern — it's probably not worth upstreaming. Instead, with operator bool available, QNN EP's local macro simplifies to:


#define ORT_CXX_LOG_PTR(logger_ptr, message_severity, message) \
  do {                                                         \
    if ((logger_ptr) && *(logger_ptr)) {                       \
      ORT_CXX_LOG((*logger_ptr), message_severity, message);   \
    }                                                          \
  } while (false)
Summary of what to propose for ORT Core:

Add parentheses around logger in the four ORT_CXX_LOG* macros (the precedence fix)
Add explicit operator bool() const noexcept to struct Logger
After those two upstream changes, QNN EP could remove IsNullLogger, the static_assert, and simplify ORT_CXX_LOG_PTR to a one-liner guard — but the macro itself should stay since the pointer-taking pattern is QNN-local.
"

onnxruntime/core/providers/qnn/common/qnn_safeint.h
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Should we move this file into `qnn/common`?

onnxruntime/core/providers/qnn/ort_api.h
```diff
@@ -3,30 +3,39 @@
...
 #include <memory>
 #include <optional>
+#include <stdexcept>
 #include <string>
 #include <vector>
 
-#include <SafeInt.hpp>
+// SafeInt wrapper that avoids the GetStackTrace dependency from core/common/safeint.h.
+// See qnn_safeint.h for details.
+#include "core/providers/qnn/qnn_safeint.h"
```
minfhong-qti: Please help move to near "core/providers/qnn/common/int4.h" below.

onnxruntime/core/providers/qnn/ort_api.h
minfhong-qti: The check seems reasonable to me. Unfortunately, probably a follow-up, it should be applied to all ORT_CXX_LOG usage as well. The original motivation introducing this ORT_CXX_LOG_PTR is exactly that the pointer stored in Ort::Logger is already deleted and QNN-EP isn't aware of it. Due to the usage of shared QnnBackendManager and we could not replace the pointer inside Ort::Logger, I thus revise the logger usage in QnnBackendManager to pointer, allowing a pointer update.

cmake/onnxruntime_providers_qnn.cmake
minfhong-qti: Thanks for revising. I'll leave @qti-hungjuiw for double-checking and resolving this thread.

onnxruntime/core/providers/qnn/ort_api.h
```diff
@@ -130,22 +160,6 @@ namespace onnxruntime {
 #define FILEPATH_TO_STRING(filepath) (filepath).string();
 #endif
 
-// QNN-EP COPY START
```
qti-hungjuiw, line 133: Why are we moving this to a new file `qnn_graph_utils.h`?
Can’t we keep it in `ort_api.h`?

onnxruntime/core/providers/qnn/common/qnn_safeint.h
qti-hungjuiw, line 1: We already have `SafeIntExceptionHandler` defined in `qnn_test_utils.cc`.
Would it make sense to declare it in `qnn_test_utils.h` so it can be reused, rather than adding a new file?

cmake/onnxruntime_providers_qnn.cmake
qti-hungjuiw: Yes, I believe it’s good to resolve this thread.

onnxruntime/core/providers/qnn/common/qnn_safeint.h
qti-kromero: moved

onnxruntime/core/providers/qnn/common/qnn_safeint.h
qti-kromero: added comment explaining why that cannot be reused for this use case

onnxruntime/core/providers/qnn/ort_api.h
qti-kromero: leaving it here caused some test files to bring in a bunch of additional headers that it didn't need which also conflicted with something my new headers were including. I think it's probably fine to isolate to a new file

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -208,26 +212,28 @@ TEST_F(GenieSessionTest, Compute_InvokesExpectedApiSequence) {
...
-  session.Run(Ort::RunOptions{nullptr},
-              input_names, &input_tensor, 1,
-              output_names, 1);
-
-  EXPECT_EQ(get_count("Node_setData"), 1) << "GenieNode_setData";
-  EXPECT_EQ(get_count("Node_execute"), 1) << "GenieNode_execute";
-  EXPECT_EQ(get_count("Node_getData"), 1) << "GenieNode_getData";
-  // KV-cache rewind is not triggered on a fresh session (rewind_ starts at 1):
-  EXPECT_EQ(get_count("Node_reset"), 0) << "GenieNode_reset should not be called on first Run";
+  {
```
quic-calvnguy, line 215: Out of the scope of this task:
I feel we should find a better way to address this issue--brackets can easily be missed by reviewers and forgotten by submitters, so I don't think this is sustainable long-term.

quic-calvnguy (approved):
LGTM, but I also think we should start a discussion on how to address the session scoping and reg/dereg of the EP better

onnxruntime/core/providers/qnn/common/qnn_safeint.h
```diff
@@ -0,0 +1,49 @@
...
+  [[noreturn]] static void SafeIntOnOverflow() { throw std::runtime_error("Integer overflow"); }
+  [[noreturn]] static void SafeIntOnDivZero() { throw std::runtime_error("Divide by zero"); }
+};
+
+#define SAFEINT_EXCEPTION_HANDLER_CPP 1
+#define SafeIntDefaultExceptionHandler SafeIntExceptionHandler
+#endif  // !defined(SafeIntDefaultExceptionHandler)
+
+#if defined(__GNUC__)
+#include "onnxruntime_config.h"
```
qti-ashwshan, line 42: Should we add these includes as well ?

`#include <exception>
#include <stdexcept>`

Since SafeIntExceptionHandler uses std::exception and std::runtime_error but I think it only works because all consumers have ort_api.h included ?

onnxruntime/test/providers/qnn/genie_integration_test.cc
```diff
@@ -0,0 +1,289 @@
...
+  void* dll_handle = LoadMockLib();
+  ASSERT_NE(dll_handle, nullptr) << "MockGenie handle must be non-null after session creation";
+  auto reset = reinterpret_cast<void (*)()>(GetSym(dll_handle, "ResetMockGenieCallCounts"));
+  auto get_count = reinterpret_cast<int (*)(const char*)>(GetSym(dll_handle, "GetMockGenieCallCount"));
+  ASSERT_NE(reset, nullptr);
+  ASSERT_NE(get_count, nullptr);
+  reset();
+
+  RegisteredEpDeviceUniquePtr registered_ep_device;
+  {
```
qti-kromero, line 215: Agreed this is fragile — explicit scope brackets rely on reviewers catching every instance. The proper fix is a test utility that enforces `session` lifetime structurally rather than editorially. I've noted this for follow-up. Resolving as out-of-scope for the current PR per your note.

onnxruntime/core/providers/qnn/common/qnn_safeint.h
```diff
@@ -0,0 +1,51 @@
...
+class SafeIntExceptionHandler : public std::exception {
+ public:
+  [[noreturn]] static void SafeIntOnOverflow() { throw std::runtime_error("Integer overflow"); }
+  [[noreturn]] static void SafeIntOnDivZero() { throw std::runtime_error("Divide by zero"); }
+};
+
+#define SAFEINT_EXCEPTION_HANDLER_CPP 1
+#define SafeIntDefaultExceptionHandler SafeIntExceptionHandler
+#endif  // !defined(SafeIntDefaultExceptionHandler)
+
```
qti-kromero, line 40: Good catch — added `<exception>` and `<stdexcept>` inside the `#ifndef` guard. Both are only needed when the class is compiled, so keeping them scoped to the guard avoids pulling them into builds where the guard fires false.

qti-ashwshan (approved):
LGTM.

tirupath-qti (approved):
LGTM

PR #263, opened 2026-04-17
fix: re-enable offload_graph_io_quantization tensor name overrides

qti-yuduo:
fix: re-enable offload_graph_io_quantization tensor name overrides

The tensor name override mechanism was disabled in PR #234 to work
around a crash when loading context binaries (DLCs) compiled with
offload_graph_io_quantization=1. The root cause was that DLC tensors
carry overridden ONNX names, but CompileContextModel populates
graph I/O index maps using fused node names, causing a lookup failure
in SetupTensors.

Fix by adding QNN tensor name aliases in SetGraphInputOutputInfo when
a deserialized graph_info_ is already available, so that SetupTensors
can resolve either name. Re-enable the three disabled tests and add a
context binary round-trip regression test.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
@@ -94,6 +94,8 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
 
     if (qnn_model_wrapper.GetModelSettings().offload_graph_io_quantization &&
         qnn_model_wrapper.IsGraphOutput(node_unit.Outputs()[0].name)) {
+      qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Inputs()[0].name,
+                                              /*external=*/node_unit.Outputs()[0].name);
       return MAKE_EP_FAIL("QNN EP is configured to not take DQ nodes that generate a graph output.");
     }
   }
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
```diff
@@ -94,6 +94,8 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
 
     if (qnn_model_wrapper.GetModelSettings().offload_graph_io_quantization &&
         qnn_model_wrapper.IsGraphOutput(node_unit.Outputs()[0].name)) {
+      qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Inputs()[0].name,
```
huaychou, line 97: When one ONNX input is consumed by two QDQ nodes, we map two input tensors of QNN graph to the same name, which results in compose graph error.
Could you help resolve this issue and add the corresponding UT for this scenario?

qti-yuduo: I can take a look, could you share reproduce with me?

qti-yuduo: Nevertheless, it should be a follow up fix. Feel free to reach out directly offline.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -171,6 +171,35 @@ Ort::Status QnnModel::SetGraphInputOutputInfo(const QnnModelContext& context) {
                                    std::forward_as_tuple(i, static_cast<int32_t>(elem_type), std::move(shape)));
   }
 
+  // DLC tensors may carry overridden names that differ from the fused node I/O names.
+  if (graph_info_) {
```
qti-ashwshan, line 175: LGTM.
Follow-up: Worth exploring whether the lookup sites (SetupTensors at qnn_model , ExecuteGraph at qnn_model) should use GetResolvedTensorName() instead of GetName().

qti-yuduo:
Discussed with @qti-ashwshan offline, we should redesign to avoid multiple south of truth/discoveries. Leaving as follow up

qti-ashwshan (approved):
LGTM. Follow-up: Worth exploring whether the lookup sites (SetupTensors at qnn_model , ExecuteGraph at qnn_model) should use GetResolvedTensorName() instead of GetName().

PR #274, opened 2026-04-21
Fix dtype errors being ignored silently

quic-ykota:
### Description
Fixes silent errors being ignored while forming QDQ node groups. 

### Motivation and Context
GetNodeIODataType returned -1 on ORT API failure, but no caller was error checking it. When two API calls both failed, -1 == -1 evaluated as true, causing the selector to incorrectly form a QDQ node group with unknown data types :fearful:

This PR adds error checking and fixes the above issue. This PR also cleans up GetNodeIODataType by separating into GetNodeInputDataType and GetNodeOutputDataType. Unfortunately there is no test framework for selectors right now and adding tests would be a big task for this change.


onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -31,46 +31,51 @@ int NumActualValues(const OrtNode* node, const OrtApi& ort_api, bool input) {
...
   }
 
+  ONNXTensorElementDataType element_type;
+  RETURN_DEFAULT_IF_API_FAIL(ort_api.GetTensorElementType(tensor_info, &element_type), ort_api, std::nullopt);
+
+  return static_cast<int32_t>(element_type);
+}
+
+// Helper function to get the data type of a node's input at a given index
+std::optional<int32_t> GetNodeInputDataType(const OrtNode* node, const OrtApi& ort_api, int index) {
```
qti-yuduo: Use anonymous namespace or static

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -31,46 +31,51 @@ int NumActualValues(const OrtNode* node, const OrtApi& ort_api, bool input) {
...
-  // Get the OrtValueInfo at the specified index
-  const OrtValueInfo* value_info = io_array[index];
+  return GetDataTypeFromValueInfo(ort_api, inputs[index]);
+}
 
-  // Get the type info from the value info
-  const OrtTypeInfo* type_info = nullptr;
-  RETURN_DEFAULT_IF_API_FAIL(ort_api.GetValueInfoTypeInfo(value_info, &type_info), ort_api, -1);
+// Helper function to get the data type of a node's output at a given index
+std::optional<int32_t> GetNodeOutputDataType(const OrtNode* node, const OrtApi& ort_api, int index) {
```
qti-yuduo: Use anonymous namespace or static

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1389,12 +1483,16 @@ bool OrtScatterElementsNodeGroupSelector::Check(const OrtGraph* graph, const Ort
...
-  const int32_t dt_output = GetNodeIODataType(q_nodes[0], ort_api, false, 0);
+  const auto dt_input_1 = GetNodeInputDataType(dq_nodes[0], ort_api, 0);
+  const auto dt_input_2 = GetNodeInputDataType(dq_nodes[1], ort_api, 0);
+  const auto dt_output = GetNodeOutputDataType(q_nodes[0], ort_api, 0);
+
+  if (!dt_input_1.has_value() || !dt_input_2.has_value() || !dt_output.has_value()) {
+    return false;
+  }
 
   // All input and output types must match.
```
qti-yuduo, line 1496: Does it worth adding a simple test?

```
TEST_F(QnnHTPBackendTests, DtypeMismatch_DropQDQ_NotSelected) {
  // DQ(uint8) -> Reshape -> Q(uint16) : types deliberately differ
  auto model = BuildQDQReshapeWithTypeMismatch(/*dq_type=*/uint8,
                                               /*q_type=*/uint16);
  // Before fix: silently selected, backend compile fails / wrong output
  // After fix: selector rejects, node falls back to CPU
  RunQnnModelTest(model, provider_options, /*expect_fully_on_qnn=*/false);
}
```

yath1: Vibe coded this one. Its not testing exact functionality that I have updated but hopefully is a good enough test.

onnxruntime/test/providers/qnn/reshape_expand_test.cc
```diff
@@ -399,6 +399,72 @@ TEST_F(QnnHTPBackendTests, Expand_6D) {
                                         19);  // Opset
 }
 
+// Test that QDQ Reshape with mismatched DQ input type (uint4) and Q output type (uint8)
```
qti-yuduo: Rather than having test inside op builder test, perhaps find a better location?

qti-chuteng (approved):

onnxruntime/test/providers/qnn/reshape_expand_test.cc
quic-ykota: What place did you have in mind? In other tests, the model is programmatically built in a helper. Are you suggesting moving this model out of the test but keep in the same file? Or are you suggesting moving to a separate model directory?

qti-yuduo: Okay.. no existing fit.. perhaps worth it's own test file then.. Prefer to create a file just for QDQ node-group selector behavior tests

qti-yuduo (approved):

onnxruntime/test/providers/qnn/reshape_expand_test.cc
quic-ykota: Moved test to a different file.

PR #276, opened 2026-04-21
Support tf_half_pixel_for_nn coordinate transformation mode for Resize

pratha-qti:
### Description
Added support for tf_half_pixel_for_nn coordinate transformation mode for the Resize operator in QNN EP.
Since QNN has no dedicated tf_half_pixel_for_nn mode, it is mapped to HALF_PIXEL as the closest equivalent.

Changes;-

1. Added handling for tf_half_pixel_for_nn coordinate transformation mode in the QNN Resize op builder
2. Added 3 new QDQ unit tests on HTP backend covering all interpolation modes:
3. ResizeU8_2xLinearTfHalfPixelForNN — linear interpolation
4. ResizeU8_2xNearestTfHalfPixelForNN_RoundPreferFloor — nearest interpolation with round_prefer_floor
5. ResizeU8_2xCubicTfHalfPixelForNN — cubic interpolation

### Motivation and Context
The ONNX Resize operator supports tf_half_pixel_for_nn as a valid coordinate transformation mode. Previously, QNN EP did not handle this mode, causing models that use it to fail or fall back to CPU.
This change enables such models to run on the QNN HTP backend by mapping tf_half_pixel_for_nn to HALF_PIXEL, which is the closest supported equivalent in QNN. The reference float models in tests also use half_pixel to accurately validate what QNN computes.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
@@ -167,7 +167,6 @@ Ort::Status ResizeOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
   //                      asymmetric |    X     Resize    RBL     Resize       X
   //            tf_half_pixel_for_nn |    x     Resize    RBL     Resize       X
 
-
   // Resize w/ "nearest" mode.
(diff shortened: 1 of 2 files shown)


qti-mbadnara:
Please fix the linting errors

pratha-qti:
Lint error is solved

PR #284, opened 2026-04-22
[OpBuilder] GRU Op Enablement

qti-hungjuiw:
### Description

**Overview**
- This PR adds GRU support to the QNN Execution Provider, covering both QNN CPU and QNN HTP           backends

**Handle Qnn Backend Issue**
| QNN CPU | QNN HTP |
|------|----------|
| 1. Accuracy issue on `time_major=true` <br> 2. Accuracy issue on the second output `Y_H` |  Need to unroll time steps to avoid accuracy issue |

In this PR, we propose the workarounds/solutions
1. Always unroll time steps for GRU (aligned with qairt-converter)
2. If on QNN CPU
    - Use `time_major=false` with necessary Transpose
    - Use the first output `Y` to acquire `Y_H`

**Test Coverage**
- CPU FP32: forward, bidirectional
- HTP QDQ uint8: forward, reverse, bidirectional (with/without B, with/without H), all-initializer, linear_before_reset
- HTP FP16: same configurations as QDQ
- ONNX backend tests: test_gru_defaults, test_gru_seq_length, test_gru_with_initial_bias

### Motivation and Context

- Prevent CPU Fallback on GRU Op and leverage QNN backend accelerators

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
@@ -1,5 +1,5 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
@@ -132,6 +132,8 @@ Ort::Status GRUOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
   }
 
   OrtNodeAttrHelper node_helper(node_unit);
+  RETURN_IF(node_helper.Get("layout", static_cast<int64_t>(0)) != 0,
+            "QNN EP doesn't support layout=1 for GRU (ORT CPU EP cannot provide a reference for accuracy validation).");
   RETURN_IF(node_helper.HasAttr("clip"), "QNN EP doesn't support clip for GRU.");
   const std::vector<std::string> activations = node_helper.Get("activations", std::vector<std::string>{});
   RETURN_IF((activations.size() >= 2 && (activations[0] != "sigmoid" || activations[1] != "tanh")) ||
@@ -208,6 +210,9 @@ Ort::Status GRUOpBuilder::AddUnidirectionGRU(QnnModelWrapper& qnn_model_wrapper,
   RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(),
                                          static_cast<uint32_t>(linear_before_reset),
                                          QNN_OP_GRU_PARAM_LINEAR_BEFORE_RESET, param_names));
+  // TODO: Once the QNN CPU backend accuracy issue with time_major=true is resolved, remove the CPU
+  // workaround below and the associated Transpose nodes (X pre-transpose, per-step Y-to-h transpose,
+  // and post-concat transpose), then let CPU use time_major=true uniformly with layout=0.
   // time_major: on CPU, always use false to work around its batch dimension bug with time_major=true.
   // On other backends (HTP), follow the ONNX layout attribute: layout=0 -> true, layout=1 -> false.
   const bool is_cpu_backend = qnn_model_wrapper.GetQnnBackendType() == QnnBackendType::CPU;
@@ -330,7 +335,7 @@ Ort::Status GRUOpBuilder::AddUnidirectionGRU(QnnModelWrapper& qnn_model_wrapper,
   // so that per-step slicing produces [batch, 1, input] directly without per-step Reshapes.
   std::string x_source = input_names[0];                               // [seq, batch, input] for HTP, [batch, seq, input] for CPU
   std::vector<uint32_t> x_source_shape = input_tensor_infos[0].shape;  // [seq, batch, input]
-  if (!time_major) {
+  if (is_cpu_backend) {
     std::string x_transposed = utils::UniqueNameGenerator().New(input_names[0], "_transposed_" + direction);
     RETURN_IF_ERROR(qnn_model_wrapper.AddTransposeNode(
         node_unit.Index(), input_names[0], x_transposed,
@@ -410,7 +415,7 @@ Ort::Status GRUOpBuilder::AddUnidirectionGRU(QnnModelWrapper& qnn_model_wrapper,
     // Derive next step's initial_h.
     // QNN CPU's Y_h (out[1]) is unreliable, so on CPU we derive h from Y (out[0]) via Transpose.
     // On HTP (time_major=true), Y_h (out[1]) is already [1, batch, hidden] and can be used directly.
-    if (time_major) {
+    if (!is_cpu_backend) {
       // HTP: use Y_h directly as next step's initial_h
       prev_h_name = yh_name;
     } else {
@@ -454,7 +459,7 @@ Ort::Status GRUOpBuilder::AddUnidirectionGRU(QnnModelWrapper& qnn_model_wrapper,
 
   // For CPU (time_major=false), Transpose [batch, seq, hidden] -> [seq, batch, hidden].
   std::string y_all;
-  if (!time_major) {
+  if (is_cpu_backend) {
     y_all = utils::UniqueNameGenerator().New(node_unit, "_Y_all_" + direction);
     RETURN_IF_ERROR(qnn_model_wrapper.AddTransposeNode(
         node_unit.Index(), y_concat, y_all,
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
@@ -171,9 +171,9 @@ Ort::Status GRUOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
   return Ort::Status();
 }
 
-// Manually unrolls the GRU across time steps and batch elements.
-// Each QNN GRU node processes a single (time_step, batch_element) with batch_size=1, seq_length=1.
-// This works around QNN CPU/HTP backend bugs where batch_size > 1 produces incorrect results.
+// Manually unrolls the GRU across time steps only (time_major=true).
+// Each QNN GRU node processes a single time step with input shape [1, batch, input] (seq=1, time_major=true).
+// The full batch is processed together in each step, avoiding the batch-element loop entirely.
 Ort::Status GRUOpBuilder::AddUnidirectionGRU(QnnModelWrapper& qnn_model_wrapper,
                                              const OrtNodeUnit& node_unit,
                                              const std::string& direction,
@@ -211,15 +211,21 @@ Ort::Status GRUOpBuilder::AddUnidirectionGRU(QnnModelWrapper& qnn_model_wrapper,
   const uint32_t seq_length = input_tensor_infos[0].shape[0];
   const int32_t direction_idx = input_tensor_infos[1].shape[0] < 2 || direction == "forward" ? 0 : 1;
 
-  // GRU parameters - shared by all unrolled cells
+  // GRU parameters - shared by all unrolled time-step cells.
+  // Use the actual ONNX direction so QNN sees the correct gate ordering.
+  // Time step ordering for reverse is handled by the unrolling loop (iterating t from seq-1 down to 0).
   std::vector<std::string> param_names;
-  // Always use forward direction for individual cells; time step ordering handled by unrolling loop
+  const uint32_t qnn_direction = (direction == "reverse") ? QNN_OP_GRU_DIRECTION_REVERSE : QNN_OP_GRU_DIRECTION_FORWARD;
   RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name() + "_" + direction,
-                                         QNN_OP_GRU_DIRECTION_FORWARD, QNN_OP_GRU_PARAM_DIRECTION, param_names));
+                                         qnn_direction, QNN_OP_GRU_PARAM_DIRECTION, param_names));
   RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(),
                                          static_cast<uint32_t>(linear_before_reset),
                                          QNN_OP_GRU_PARAM_LINEAR_BEFORE_RESET, param_names));
-  RETURN_IF_ERROR(AddQnnScalar<bool>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), false,
+  // QNN CPU has a batch bug with time_major=true; QNN HTP has a batch bug with time_major=false.
+  // Pick time_major based on the backend to avoid the respective batch dimension bugs.
+  const bool is_npu_backend = qnn_model_wrapper.GetQnnBackendType() != QnnBackendType::CPU;
+  const bool time_major = is_npu_backend;  // true for HTP, false for CPU
+  RETURN_IF_ERROR(AddQnnScalar<bool>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), time_major,
                                      QNN_OP_GRU_PARAM_TIME_MAJOR, param_names));
 
   // Null tensor for optional inputs
@@ -301,6 +307,7 @@ Ort::Status GRUOpBuilder::AddUnidirectionGRU(QnnModelWrapper& qnn_model_wrapper,
     }
   }
   // initial_h: ONNX in[5] [num_directions, batch_size, hidden_size] -> [1, batch_size, hidden_size]
+  // Shape [1, batch, hidden] matches time_major=true initial_h for seq=1 GRU nodes.
   std::string initial_h_name;
   {
     std::vector<uint32_t> h_shape = {1, batch_size, hidden_size};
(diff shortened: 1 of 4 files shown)


onnxruntime/test/providers/qnn/gru_test.cc
```diff
@@ -0,0 +1,665 @@
...
+                      std::ref(H_def),                                                                         // initial_h
+                      true,                                                                                    // has_Y
+                      true,                                                                                    // has_Y_h
+                      direction,                                                                               // direction
+                      hidden_size,                                                                             // hidden_size
+                      0,                                                                                       // layout
+                      ExpectedEPNodeAssignment::All,
+                      1,  // linear_before_reset
+                      0.03f);
+}
```
kuanyul-qti, line 949: All 16 tests hard-code `layout=0`. The `layout=1` (batch-first) path in the op builder is a distinct code path with different shape computations (CPU slices along `seq_dim=1`; HTP inserts a pre-transpose before unrolling). This path is entirely untested.

Please add at least one `layout=1` test each for CPU FP32, HTP QDQ u8, and HTP FP16 (forward + bidirectional minimum).

kuanyul-qti, line 949: All tests request both `has_Y=true, has_Y_h=true`. The code branches for Y-only and Y_h-only (including the bidirectional Concat path when `fwd_out[1]` is empty) are untested.

Please add tests for `has_Y=true, has_Y_h=false` and `has_Y=false, has_Y_h=true` for at least one backend.

onnxruntime/test/providers/qnn/gru_test.cc
```diff
@@ -0,0 +1,665 @@
...
+TEST_F(QnnHTPBackendTests, GRU_QDQ_linear_before_reset) {
+  std::string direction = "forward";
+  uint32_t num_direction = 1;
+  uint32_t batch_size = 3;
+  uint32_t hidden_size = 4;
+  uint32_t input_size = 5;
+  uint32_t seq_len = 6;
+  auto B_def = TestInputDef<float>({num_direction, 6 * hidden_size}, false, -1.0f, 1.0f);
+  auto H_def = TestInputDef<float>({num_direction, batch_size, hidden_size}, false, -1.0f, 1.0f);
+  RunHtpQDQGRUOpTest<uint8_t>(TestInputDef<float>({seq_len, batch_size, input_size}, false, -1.0f, 1.0f),              // X
```
kuanyul-qti, line 659: Could you add a `RunHtpQDQGRUOpTest<uint16_t>` sanity test?

onnxruntime/test/providers/qnn/gru_test.cc
qti-hungjuiw: Done. Thanks!

onnxruntime/test/providers/qnn/gru_test.cc
qti-hungjuiw: Done.

onnxruntime/test/providers/qnn/gru_test.cc
qti-hungjuiw: This might not be feasible since the reference golden from ORT CPU-EP does not support `layout=1`.
https://github.com/microsoft/onnxruntime/blob/ce6aaeece1b538ef83aca26f141a947a68c6398c/onnxruntime/core/providers/cpu/rnn/deep_cpu_gru.h#L56

onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
```diff
@@ -0,0 +1,546 @@
...
+Ort::Status GRUOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                        const OrtNodeUnit& node_unit,
+                                        const Ort::Logger& logger) const {
+  ORT_UNUSED_PARAMETER(logger);
+  if (node_unit.Inputs().size() > 4 && node_unit.Inputs()[4].Exists()) {
+    TensorInfo tensor_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[4], tensor_info));
+    RETURN_IF_NOT(tensor_info.is_initializer, "QNN EP: dynamic sequence_length is not supported.");
+    std::vector<uint8_t> sequence_lens_bytes;
+    RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(tensor_info.initializer_tensor, sequence_lens_bytes));
```
huaychou, line 126: I think the unpack + reinterpret_cast can be a utility function in the future.

onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
```diff
@@ -0,0 +1,546 @@
...
+                                           concat_axis, QNN_OP_CONCAT_PARAM_AXIS, cp));
+    QnnTensorWrapper tw(y_concat, QNN_TENSOR_TYPE_NATIVE, output_tensor_infos[0].qnn_data_type,
+                        output_tensor_infos[0].quant_param.Copy(), std::move(concat_out_shape));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(tw)), "Failed to add Y Concat output.");
+    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(y_concat, QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_CONCAT,
+                                                  std::move(qnn_all_step_hidden_names), {y_concat}, std::move(cp), do_op_validation),
+                  "Failed to create Y Concat node.");
+  }
+
+  // For CPU (time_major=false), Transpose [batch, seq, hidden] -> [seq, batch, hidden].
```
huaychou, line 464: I think HTP also has time_major=false case?

onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
```diff
@@ -0,0 +1,546 @@
...
+    RETURN_IF_ERROR(qnn_model_wrapper.AddTransposeNode(
+        node_unit.Index(), y_concat, y_all,
+        {batch_size, seq_length, hidden_size}, {1, 0, 2}, {seq_length, batch_size, hidden_size},
+        output_tensor_infos[0].qnn_data_type, output_tensor_infos[0].quant_param,
+        do_op_validation, false, false));
+  } else {
+    y_all = y_concat;
+  }
+
+  // Map to ONNX output shapes:
```
huaychou, line 477: Do we need to map to ONNX output shape according to time_major?

qti-hungjuiw: Per offline discussion, we only support the case of `layout=0`, and thus we will always have `time_major=true` on HTP and GPU.
As for CPU, currently we forcibly use Transpose workaround with `time_major=false` due to accuracy issue, and it will yield correct ONNX output shape.

onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
qti-hungjuiw: Sure. HTP itself can support `time_major=false`. However, per offline discussion, we only support the case of `layout=0`, and thus we will always have `time_major=true` on HTP in QNN-EP. The comment and condition are also modified for clarification.

onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
qti-hungjuiw: Sure, per offline discussion, this can be separated as future work.

onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
```diff
@@ -0,0 +1,551 @@
...
+                      utils::UniqueNameGenerator().New(node_unit, "_cell" + sfx),
+                      QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_GRU,
+                      std::move(cell_inputs), {y_name, yh_name},
+                      std::vector<std::string>(param_names), do_op_validation),
+                  "Failed to create GRU node.");
+
+    // Derive next step's initial_h.
+    // QNN CPU's Y_h (out[1]) is unreliable, so on CPU we derive h from Y (out[0]) via Transpose.
+    // On HTP (time_major=true), Y_h (out[1]) is already [1, batch, hidden] and can be used directly.
+    if (!is_cpu_backend) {
```
huaychou, line 422: NIT, can we put is_cpu_backend first for better alignment?

onnxruntime/test/providers/qnn/gru_test.cc
huaychou: Could we add a test for `layout=1` that falls back to ORT CPU and is unable to run? This would help ensure we're notified with an upleveling PR once ORT CPU support becomes available.

onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc
qti-hungjuiw: Per offline discussion, will keep the current behavior to keep the short `if`.

onnxruntime/test/providers/qnn/gru_test.cc
qti-hungjuiw: Done, this is a good idea, thanks!

qti-chuteng:
**[M-1]** `onnxruntime/test/providers/qnn/gru_test.cc:323`

The HTP test guard omits `__linux__`, which differs from sibling `lstm_test.cc` and the team's review checklist. Currently all 30+ GRU HTP QDQ / FP16 tests are excluded from Linux x86_64 builds — a green Linux CI does not actually exercise the HTP path.

```diff
-#if defined(__aarch64__) || defined(_M_ARM64)
+#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
 // HTP QDQ + FP16 tests ...
-#endif  // defined(__aarch64__) || defined(_M_ARM64)
+#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
```

---

## Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:139`

`activations` length is only validated for sizes 2 and 4. Sizes 1 / 3 (spec-illegal but possible from a malformed model) silently pass the gate. Add an explicit length check.

```cpp
RETURN_IF(!activations.empty() && activations.size() != 2 && activations.size() != 4,
          "QNN EP: GRU activations attribute must have length 0, 2, or 4.");
```

---

**[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:224`

The literal name `"null_tensor"` collides with LSTM's null-tensor name and `AddTensorWrapper` silently returns true on collision (no overwrite). Worse, the dtypes diverge: LSTM uses `QNN_DATATYPE_UNDEFINED`, GRU uses `QNN_DATATYPE_FLOAT_32`. This is a latent footgun.

```cpp
const std::string null_tensor_name = utils::UniqueNameGenerator().New(node_unit, "_null_tensor_" + direction);
QnnTensorWrapper null_tensor_wrapper(null_tensor_name, QNN_TENSOR_TYPE_NULL, QNN_DATATYPE_UNDEFINED,
                                     QnnQuantParamsWrapper(), std::vector<uint32_t>{0});
```

---

**[N-3]** `docs/execution_providers/QNN-ExecutionProvider.md`

The supported-operator table lists `ai.onnx:LSTM` but is not updated to include GRU. New ops should be reflected in docs so users can discover support.

```diff
 |ai.onnx:LSTM||
+|ai.onnx:GRU||
```

---

**[N-4]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:192`

`hidden_size` is read with a literal `0` default (type `int`) and silently narrowed to `uint32_t`. Per checklist `[B4]`, use `SafeInt`:

```cpp
const int64_t hidden_size_i64 = node_helper.Get("hidden_size", static_cast<int64_t>(0));
RETURN_IF_NOT(hidden_size_i64 > 0, "hidden_size is not set for GRU");
const uint32_t hidden_size = SafeInt<uint32_t>(hidden_size_i64);
const int32_t hidden_size_sign = SafeInt<int32_t>(hidden_size_i64);
```

---

**[N-5]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:128`

`sequence_lens` is `reinterpret_cast<const int32_t*>` without checking `tensor_info.qnn_data_type`. A non-int32 input would be silently misinterpreted.

```cpp
RETURN_IF_NOT(tensor_info.qnn_data_type == QNN_DATATYPE_INT_32,
              "QNN EP: GRU sequence_lens must be int32 per ONNX spec.");
```

huaychou (approved):
LGTM

qti-hungjuiw:
> **[M-1]** `onnxruntime/test/providers/qnn/gru_test.cc:323`
> 
> The HTP test guard omits `__linux__`, which differs from sibling `lstm_test.cc` and the team's review checklist. Currently all 30+ GRU HTP QDQ / FP16 tests are excluded from Linux x86_64 builds — a green Linux CI does not actually exercise the HTP path.
> 
> ```diff
> -#if defined(__aarch64__) || defined(_M_ARM64)
> +#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
>  // HTP QDQ + FP16 tests ...
> -#endif  // defined(__aarch64__) || defined(_M_ARM64)
> +#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
> ```
> 
> ## Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:139`
> 
> `activations` length is only validated for sizes 2 and 4. Sizes 1 / 3 (spec-illegal but possible from a malformed model) silently pass the gate. Add an explicit length check.
> 
> ```c++
> RETURN_IF(!activations.empty() && activations.size() != 2 && activations.size() != 4,
>           "QNN EP: GRU activations attribute must have length 0, 2, or 4.");
> ```
> 
> **[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:224`
> 
> The literal name `"null_tensor"` collides with LSTM's null-tensor name and `AddTensorWrapper` silently returns true on collision (no overwrite). Worse, the dtypes diverge: LSTM uses `QNN_DATATYPE_UNDEFINED`, GRU uses `QNN_DATATYPE_FLOAT_32`. This is a latent footgun.
> 
> ```c++
> const std::string null_tensor_name = utils::UniqueNameGenerator().New(node_unit, "_null_tensor_" + direction);
> QnnTensorWrapper null_tensor_wrapper(null_tensor_name, QNN_TENSOR_TYPE_NULL, QNN_DATATYPE_UNDEFINED,
>                                      QnnQuantParamsWrapper(), std::vector<uint32_t>{0});
> ```
> 
> **[N-3]** `docs/execution_providers/QNN-ExecutionProvider.md`
> 
> The supported-operator table lists `ai.onnx:LSTM` but is not updated to include GRU. New ops should be reflected in docs so users can discover support.
> 
> ```diff
>  |ai.onnx:LSTM||
> +|ai.onnx:GRU||
> ```
> 
> **[N-4]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:192`
> 
> `hidden_size` is read with a literal `0` default (type `int`) and silently narrowed to `uint32_t`. Per checklist `[B4]`, use `SafeInt`:
> 
> ```c++
> const int64_t hidden_size_i64 = node_helper.Get("hidden_size", static_cast<int64_t>(0));
> RETURN_IF_NOT(hidden_size_i64 > 0, "hidden_size is not set for GRU");
> const uint32_t hidden_size = SafeInt<uint32_t>(hidden_size_i64);
> const int32_t hidden_size_sign = SafeInt<int32_t>(hidden_size_i64);
> ```
> 
> **[N-5]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:128`
> 
> `sequence_lens` is `reinterpret_cast<const int32_t*>` without checking `tensor_info.qnn_data_type`. A non-int32 input would be silently misinterpreted.
> 
> ```c++
> RETURN_IF_NOT(tensor_info.qnn_data_type == QNN_DATATYPE_INT_32,
>               "QNN EP: GRU sequence_lens must be int32 per ONNX spec.");
> ```

Fixed [M-1], [N-2], [N-4]
[N-1] is not correct as GRU cannot have three activations
[N-3] might need to be done in separated work
[N-5] might not be necessary as ONNX GRU cannot have sequence_lens other than int32

qti-chuteng:
**[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/gru_op_builder.cc:260` — `qnn_model_wrapper.AddTensorWrapper(std::move(null_tensor_wrapper));` still ignores the return value. N-3's unique-name change reduced the collision risk, but the call site is still inconsistent with the other 13 `RETURN_IF_NOT`-wrapped `AddTensorWrapper` calls in this file. One-line fix:

```cpp
RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(null_tensor_wrapper)),
              "Failed to add null tensor for GRU.");
```

**[M-3]** `gru_op_builder.cc:327-330` (zero-bias) and `gru_op_builder.cc:355-358` (zero initial_h) — Both branches build `QnnTensorWrapper` with `input_tensor_infos[0].qnn_data_type` and an empty `QnnQuantParamsWrapper()`. On HTP QDQ path the input dtype is `QNN_DATATYPE_UFIXED_POINT_{8,16}`; declaring a quantized dtype with no quant params makes the all-zero bytes represent a non-zero value, which biases sigmoid/tanh inputs and the recurrent path. The new `GRU_QDQ_sanity_bidirectional_wo_B` and `GRU_QDQ_sanity_bidirectional_wo_H` tests use the default `QDQTolerance()` and absorb the drift; they do not validate correctness. Recommended Option A (small fix): reject "quantized input + missing B/initial_h" in `IsOpSupported`.

---

## New Minor

**[N-9]** `gru_test.cc:1300-1307` and `gru_test.cc:1518-1525` — Linux x86_64 platform-aware tolerances of `0.15` (vs `0.006`, 25× wider) and `0.25` (vs `0.03`, 8.3× wider) are well outside reasonable FP16 rounding for a 6-step GRU unroll. The TODO marker says these are simulator-side bugs, but a wide tolerance does not auto-tighten when the simulator is fixed; a future real-device regression to 0.04 (still inside 0.15) would silently pass. Recommend `GTEST_SKIP()` on Linux x86_64 with a TODO ticket reference, or extracting the platform-aware block into a clearly-named macro (`LINUX_X86_HTP_SIMULATOR_TOLERANCE_OR(real_tolerance)`) so the workaround is grep-able.

qti-hungjuiw:
> [M-2]

[M-2] fixed.
[M-3] is invalid as an empty `QnnQuantParamsWrapper()` do not yield non-zero value.
[N-9] is unnecessary as TODO should be enough for grep.

qti-chuteng (approved):

PR #287, opened 2026-04-22
Support tf_half_pixel_for_nn coordinate transformation for the Resize

pratha-qti:
### Description
Added support for tf_half_pixel_for_nn coordinate transformation mode for the Resize operator in QNN EP.
Since QNN has no dedicated tf_half_pixel_for_nn mode, it is mapped to HALF_PIXEL as the closest equivalent.
Changes;-

Added handling for tf_half_pixel_for_nn coordinate transformation mode in the QNN Resize op builder
Added 3 new QDQ unit tests on HTP backend covering all interpolation modes:
ResizeU8_2xLinearTfHalfPixelForNN — linear interpolation
ResizeU8_2xNearestTfHalfPixelForNN_RoundPreferFloor — nearest interpolation with round_prefer_floor
ResizeU8_2xCubicTfHalfPixelForNN — cubic interpolation

### Motivation and Context
The ONNX Resize operator supports tf_half_pixel_for_nn as a valid coordinate transformation mode. Previously, QNN EP did not handle this mode, causing models that use it to fail or fall back to CPU.
This change enables such models to run on the QNN HTP backend by mapping tf_half_pixel_for_nn to HALF_PIXEL, which is the closest supported equivalent in QNN. The reference float models in tests also use half_pixel to accurately validate what QNN computes.


qti-yuduo:
@pratha-qti please click on the link and sign CLA.

onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
```diff
@@ -75,7 +75,9 @@ const std::unordered_map<std::string, uint32_t> ResizeOpBuilder::supported_coord
     {"half_pixel", QNN_OP_RESIZE_TRANSFORMATION_MODE_HALF_PIXEL},
     {"pytorch_half_pixel", QNN_OP_RESIZE_TRANSFORMATION_MODE_PYTORCH_HALF_PIXEL},
     {"align_corners", QNN_OP_RESIZE_TRANSFORMATION_MODE_ALIGN_CORNERS},
-    {"asymmetric", QNN_OP_RESIZE_TRANSFORMATION_MODE_ASYMMETRIC}};
+    {"asymmetric", QNN_OP_RESIZE_TRANSFORMATION_MODE_ASYMMETRIC},
+    // QNN has no dedicated tf_half_pixel_for_nn mode; map to HALF_PIXEL as the closest equivalent.
```
qti-yuduo: Am I understanding correctly that formulas differ by a constant −0.5 in source coordinates:
```
half_pixel:            x_src = (k + 0.5) / s - 0.5
tf_half_pixel_for_nn:  x_src = (k + 0.5) / s
```
?




Why can't we have full numerically equivalent support? 

QNN_OP_RESIZE:       output_shape[d]=2·L_out[d] on resized axes, mode=ASYMMETRIC
QNN_OP_STRIDED_SLICE: start=1, end=2·L_out[d], stride=2 on resized axes

pratha-qti: Yes, the formulas differ by a constant −0.5 shift in source coordinates. Full numerical equivalence is possible using a 2× ASYMMETRIC resize followed by a strided slice, which exactly reproduces tf_half_pixel_for_nn. However, this approach significantly increases compute cost (4× pixels for cubic resize), leading to ~3.7× higher resize cycles and ~2× end‑to‑end perf regression. Therefore, while two approaches exist (exact vs performant), we choose to map to half_pixel here because the ONNX→QNN converter already does this mapping, ensuring consistent behavior with existing conversion logic and better performance.

qti-chuteng:
### 🔴 Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc:L80-L81, L380, L412-L414`
`tf_half_pixel_for_nn` is mapped to QNN `HALF_PIXEL`, but the ONNX (opset 11) coordinate formulas differ by exactly 0.5: `half_pixel` = `(x+0.5)/scale - 0.5` vs `tf_half_pixel_for_nn` = `(x+0.5)/scale`. This is not a "closest equivalent" — it is a semantically wrong mapping that produces silent accuracy errors (different pixel selection for nearest, different interpolation weights for linear/cubic). If QNN has no native mode for this, reject it in `IsOpSupported()` and let the node fall back to CPU instead of silently approximating.

---

### 🟠 Major

**[M-1]** `onnxruntime/test/providers/qnn/resize_test.cc → ResizeU8_2x{Linear,Nearest,Cubic}TfHalfPixelForNN`
Tests build the Resize node at opset 19, but `tf_half_pixel_for_nn` exists only in the Resize **opset 11** spec (`cmake/external/onnx/onnx/defs/tensor/old.cc:L5447`); it was removed in opset 13 and 18. An opset-19 Resize using this mode is not a valid ONNX model. Clarify the target opset; if the goal is opset 13+ the mode is out-of-spec and should be rejected (see C-1).

**[M-2]** `onnxruntime/test/providers/qnn/resize_test.cc:L405, L431, L458`
The float reference models deliberately use `half_pixel` while only the QDQ models use `tf_half_pixel_for_nn`. Since the PR maps `tf_half_pixel_for_nn` → `half_pixel`, the comparison necessarily passes regardless of correctness — a fake fence that cannot detect C-1. Use the same `tf_half_pixel_for_nn` mode for the float reference (true CPU EP reference) so the test can actually validate the mapping, and run a kill-test (map to a wrong mode, confirm the test fails, then restore).

pratha-qti:
Yeah, I get the concern — agreed this isn’t a mathematically equivalent mapping. But from our side, we’re consciously choosing between two trade-offs:
* **Accuracy-first** → emulate `tf_half_pixel_for_nn` exactly (2× resize + slice)
  * Correct semantics
  * But heavy perf hit (especially for cubic)

* **Performance-first (current behavior)** → map to `HALF_PIXEL`
  * Slight coordinate shift (−0.5) → yes, not semantically exact
  * But much better runtime performance
  * Also aligns with what QNN converter/backend already does today

We picked the second option mainly for **consistency + performance** — since the existing QNN path already maps `tf_half_pixel_for_nn → half_pixel`, doing the same here avoids surprises and keeps behavior uniform across flows.

Totally fair point though — if strict correctness is required, we can consider:
* either adding the emulation path (opt-in), or
* explicitly rejecting in `IsOpSupported()` and falling back
Right now it’s an intentional approximation

qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🔴 Blocker` `[Arch]` — `resize_op_builder.cc` (IsOpSupported gate + ProcessTfHalfPixelForNN)
`IsOpSupported` only special-cases tf_half_pixel_for_nn when `interp_mode=="nearest"`. For `linear`/`cubic` + tf there is no gate, so the node is accepted onto HTP, and `ProcessAttributesAndOutputs` dispatches unconditionally on `transformation_mode == "tf_half_pixel_for_nn"` (ignoring interp_mode) into `ProcessTfHalfPixelForNN`, whose `local_modes` maps nearest/linear/cubic. But the intermediate QNN Resize node it builds never sets `QNN_OP_RESIZE_PARAM_CUBIC_COEFF` (the main path sets it from `onnx_cubic_coeff_a_attr`) nor `QNN_OP_RESIZE_PARAM_EXCLUDE_OUTSIDE`. So a cubic+tf Resize is lowered with QNN's default cubic coefficient — wrong output with no error. The coordinate math is actually exact for linear/cubic too; the root cause is dropped params + zero coverage, not a mapping error.
*Better design:* Either (a) restrict ProcessTfHalfPixelForNN and the gate to `interp_mode=="nearest"` and return unsupported for non-nearest tf so it falls back to CPU cleanly, or (b) fully parameterize the intermediate Resize — set cubic_coeff and exclude_outside for every admitted mode. IsOpSupported must not admit modes the decomposition cannot prove correct.

**[D-2]** `🟠 Concern` `[Arch]` — `resize_op_builder.cc` (ProcessTfHalfPixelForNN naming)
`resize_out_name`/`resize_node_name`/`slice_node_name` are all `node_unit.Name() + "_tfhp_*"`. ONNX nodes frequently have an empty Name(); two unnamed tf Resize nodes in one graph both produce the same node/tensor names → `CreateQnnNode`/`AddTensorWrapper` collision. The added comment only claims uniqueness between the two nodes of the same node_unit, not across node_units (the param wrappers are safe because QnnParamWrapper embeds node_unit.Index(), which makes the gap easy to miss).
*Better design:* Use the sibling idiom `utils::UniqueNameGenerator().New(node_unit, "_tfhp_resize")`, which falls back to OpType+Index when Name() is empty. See gru_op_builder.cc:86, lstm_op_builder.cc:200, cast_op_builder.cc:192, batchnormalization_op_builder.cc:770.

**[D-3]** `🟠 Concern` `[Arch]` — `resize_op_builder.cc` (final output tensor)
`ProcessTfHalfPixelForNN` builds the final graph-output tensor with `input_info.quant_param.Copy()` and returns, never calling `ProcessOutputs`. The normal Resize path derives output quant params from the node_unit's own output QDQ and then calls `OverrideOutputQuantParam` → `SetOutputQParamEqualToInputIfNearlyEqual`, forcing output=input only when nearly equal. The tf path forces output=input unconditionally, so a QDQ model whose output Q scale/zp legitimately differs gets the wrong (input) quant params at the graph boundary.
*Better design:* Route the StridedSlice output through the standard `ProcessOutputs` machinery, or at minimum reuse `SetOutputQParamEqualToInputIfNearlyEqual` instead of an unconditional copy, and document why output range is provably identical to input.

### 🟩 Maintainability

**[D-4]** `🟠 Concern` `[Maint]` — `resize_op_builder.cc` (ProcessTfHalfPixelForNN + pre-existing branches)
The diff replaces one-line `AddQnnScalar<T>(...)` calls with a hand-expanded 5-line ritual, both in the new helper and — gratuitously — in the `ResizeNearestNeighbor`/`ResizeBilinear`/fallback branches that the tf path never reaches (tf early-returns). AddQnnScalar is a free function in `namespace qnn`, fully callable here. Costs: shallow re-implementation of a deep helper; rewriting working pre-existing branches unrelated to the feature; and every manual site calls the bool-returning `AddParamWrapper` but discards the result, whereas AddQnnScalar wraps it in `RETURN_IF_NOT(...)` — a failed param insertion now goes silently.
*Better design:* Revert the pre-existing branches to `AddQnnScalar<T>`; in ProcessTfHalfPixelForNN use `AddQnnScalar<uint32_t>(...)` — uniqueness is already guaranteed by (node_name+index+param_name), the custom `_tfhp_` suffixes are unnecessary, and the error check is preserved.

**[D-7]** `🔵 Improvement` `[Maint]` — `resize_op_builder.cc` (ASCII table/comment) + `resize_test.cc` (test comments)
The added ASCII decision-table row reads `tf_half_pixel_for_nn | Resize+StridedSlice (all nearest_modes supported)` and the skip-site comment says all nearest_modes are supported — but the gate the PR adds does `RETURN_IF_NOT(tf_nearest_mode == "floor", …)`, floor only (the two _Unsupported tests assert exactly this). Separately, three test comments say "Maps to QNN ResizeNearestNeighbor(...)" but the code hardcodes op type "Resize".
*Better design:* Correct the ASCII row/comment to "floor only (round_prefer_floor/ceil fall back to CPU)"; fix the three test comments to "QNN Resize + StridedSlice" (or emit ResizeNearestNeighbor for the rank-4 nearest tf case, for perf).

### 🟩 Simplicity

**[D-6]** `🔵 Improvement` `[Simplicity]` — `resize_op_builder.cc` (local maps)
`ProcessTfHalfPixelForNN` declares function-local `static const unordered_map local_modes` and `local_nearest_modes` byte-identical to `ResizeOpBuilder::supported_modes` and `supported_nearest_modes`. `GetQnnModeValFromOnnxString` already takes the map as a parameter, and the else-branch already passes the file-scope tables in. A QNN SDK enum change must then be applied to both, or the two Resize paths silently diverge.
*Better design:* Delete both local maps; pass the file-scope tables into the helper, or make it a private static member.

**[D-8]** `🔵 Improvement` `[Simplicity]` — `resize_test.cc` (whole file)
The diff strips nearly every blank line between test functions, includes and helpers, and inserts a stray blank line before the license comment. Unrelated to the feature; bloats the diff, poisons git blame, raises merge-conflict risk, and buries the 4 real new tests in formatting noise.
*Better design:* Revert the whitespace-only changes; keep only the appended tf TEST_F blocks. Run a formatter in a separate PR if desired.

### 🔵 Completeness

**[D-5]** `🟠 Concern` `[Completeness]` — `resize_test.cc` (the 5 added TEST_F)
The PR description claims 3 tests covering linear/nearest/cubic, but the diff adds 5 tests all `interp_mode="nearest"` — zero linear, zero cubic tf_half_pixel_for_nn. This is exactly what lets D-1's dropped-param defect ship green (deleting the cubic_coeff logic turns no test red). Also the tf dispatch is not gated on `is_npu_backend`, so the QNN CPU backend also lowers via this decomposition, yet all 5 tests use the htp backend — the CPU path has zero coverage, breaking this file's CPU+HTP-per-mode convention.
*Better design:* If nearest-only, add a tripwire test asserting non-nearest tf falls back to CPU (ExpectedEPNodeAssignment::None); if linear/cubic ship, add linear+tf and cubic+tf (incl. non-default cubic_coeff_a) QDQ accuracy tests. Add at least one CPU tf test mirroring ResizeU8_2xLinearHalfPixel / ResizeU8_2xCubicHalfPixel.

qti-chuteng:
### 🟪 Architecture Fit

**[D-3]** `🔵 Improvement` `[Arch]` — `resize_op_builder.cc` (ProcessTfHalfPixelForNN) + PR description
The PR description says tf_half_pixel_for_nn "is mapped to HALF_PIXEL as the closest equivalent", but the code never maps to HALF_PIXEL — it builds a Resize(2x, ASYMMETRIC) intermediate then a StridedSlice picking odd indices. I independently verified the decomposition is **exact** (not approximate) under nearest+floor: intermediate 2M, ASYMMETRIC maps index k to `x_src=k·L/(2M)`; picking `k=2j+1` gives `(j+0.5)·L/M`, exactly tf_half_pixel_for_nn's `(j+0.5)/(M/L)`. With no LWD, the PR description is the only design record, and it currently misrepresents an exact decomposition as a lossy mapping.
*Better design:* Fix the description to state "Resize(2x, ASYMMETRIC) + StridedSlice (odd indices), exactly equivalent under nearest+floor, not an approximate mapping"; align the top-of-function comment's `s` notation with what the code does (double the output shape, take odd indices).

### 🟩 Maintainability

**[D-2]** `🟠 Concern` `[Maint]` — `resize_test.cc`
The first hunk inserts a UTF-8 BOM (verified: PR-branch file's first 3 bytes are `ef bb bf`; original starts directly with `// Copyright`), and every blank line between functions across the ~700-line file is deleted — including many tests unrelated to tf_half_pixel_for_nn. The BOM is an unintended encoding change inconsistent with the repo's other test files; the blank-line sweep makes most of the 94-add/104-del pure noise, buries the 6 new tests (and the RunQDQResizeAndAssertQnnOp skip-before-dump fix), and pollutes future blame/bisect.
*Better design:* Remove the BOM (save UTF-8 without BOM), restore the blank lines (start from main, re-add only the 6 new tests + the skip fix), and check the editor isn't auto-re-encoding or applying a whitespace rule on save.

### 🟩 Simplicity

**[D-1]** `🟠 Concern` `[Simplicity]` — `resize_op_builder.cc` (ResizeNearestNeighbor / ResizeBilinear / generic branches)
The PR only needs to add a tf_half_pixel_for_nn branch, but it also rewrites three existing, feature-unaffected branches from single-line `AddQnnScalar<bool>(...)` / `AddQnnScalar<uint32_t>(...)` into 5-line hand-rolled `Qnn_Scalar_t` + `QnnParamWrapper` + `push_back` + `AddParamWrapper` blocks. Checking `AddQnnScalar` (qnn_model_wrapper.h:620-652), its body is line-for-line equivalent to the hand-rolled code — and the new `ProcessTfHalfPixelForNN` itself still uses `AddQnnScalar`, proving the helper is sufficient. Zero functional gain, inflates the diff, adds regression surface to every existing branch, and fights the convention (~10 other op builders use `AddQnnScalar`).
*Better design:* Revert those three branches to their `AddQnnScalar<T>` calls (git checkout main for that region), so the diff carries only the new `ProcessTfHalfPixelForNN` + the IsOpSupported gate. If a param-construction style change is genuinely wanted, split it into a standalone refactor PR.

### 🔵 Completeness

**[D-4]** `🔵 Improvement` `[Completeness]` — `resize_op_builder.cc` / `resize_test.cc` (branch base vs main)
The PR head (`09117bc`) is 28 commits behind main. I confirmed main did **not** modify `resize_op_builder.cc` in that window, so the core file rebases cleanly (lower risk than PR #458's API drift). But `resize_test.cc` and shared test utils may have changed, and the passing tests were validated against the old base.
*Better design:* Rebase onto current main and re-run the Resize CPU/HTP tests; addressing D-2 first (restoring the test file's formatting) substantially reduces the rebase conflict surface.

qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🟠 Concern` `[Arch]` — `resize_op_builder.cc` (ProcessTfHalfPixelForNN vs base_op_builder.cc ProcessOutputs)
`ProcessTfHalfPixelForNN` ends by constructing the final graph-output `QnnTensorWrapper` itself (deciding `is_graph_output` / `QNN_TENSOR_TYPE_APP_READ` by hand) and calling `CreateQnnNode` directly for the StridedSlice node, entirely bypassing the `BaseOpBuilder::ProcessOutputs` (base_op_builder.cc:251) that the other three Resize dispatch branches all route through. That shared path does three things this tf branch now silently drops: (a) `OverrideOutputQuantParam` → the Resize builder's own `SetOutputQParamEqualToInputIfNearlyEqual` HTP quant optimization; (b) `GetSupportedOutputDataType` dtype downgrade; (c) the int64→int32 cast for int64/uint64 graph outputs. Numerically correct for the tested QDQ nearest+floor cases, but the tf path derives output quant params inconsistently with the other branches and leaves an untested int64/dtype hole. Flagged independently by all four sub-agents.
*Better design:* Create the trailing StridedSlice node through `ProcessOutputs(qnn_model_wrapper, node_unit, {resize_out_name}, std::move(slice_param_names), logger, do_op_validation, QNN_OP_STRIDED_SLICE)`, the way batchnormalization_op_builder.cc routes its terminal node. Also resolves D-6 and D-8.

**[D-8]** `🔵 Improvement` `[Arch]` — `resize_op_builder.cc`
The intermediate `resize_out_tensor` uses `input_info.quant_param` while the final output uses `output_info.quant_param`. StridedSlice is pure data movement, so ideally in and out qparams match (cf. rotary_embedding's slice reusing input qparams). Harmless for tested cases; auto-resolves once D-1 is adopted.
*Better design:* Resolved together with D-1; no separate action needed.

### 🟩 Maintainability


**[D-5]** `🔵 Improvement` `[Maint]` — `resize_op_builder.cc`
All intermediate names (`name_base + "_tfhp_resize_out"`, `"_tfhp_slice"`, param names) are hand-built string suffixes off the output tensor name. The repo idiom for inserted nodes/tensors is `utils::UniqueNameGenerator().New(node_unit, ...)` (gru_op_builder.cc:86, base_op_builder.cc:299). Collision-freeness here rests on a hand-proof that ONNX output names are unique — functionally fine, but inconsistent.
*Better design:* Use `utils::UniqueNameGenerator` for the intermediate names, consistent with the GRU StridedSlice insertion and other builders.

### 🟩 Simplicity

**[D-6]** `🔵 Improvement` `[Simplicity]` — `resize_op_builder.cc`
`ProcessTfHalfPixelForNN` takes 9 parameters, of which `output_shape` / `input_info` / `output_info` are fetched by the caller and threaded in, `nearest_modes_map` is just the static member `supported_nearest_modes`, and `nearest_mode` is gate-pinned to "floor". These are pass-through / self-derivable values inflating the signature.
*Better design:* Reduce the helper to ~4 params (`qnn_model_wrapper`, `node_unit`, `input_names`, `do_op_validation`) and derive the rest internally; hardcode nearest_mode to `QNN_OP_RESIZE_NEAREST_MODE_FLOOR`. Adopting D-1 naturally shrinks the helper.

### 🔵 Completeness

**[D-3]** `🟠 Concern` `[Completeness]` — `resize_test.cc` / `resize_op_builder.cc`
`ProcessTfHalfPixelForNN` and its comments claim rank-3/4/5 support, and the IsOpSupported rank check admits 3..5. But the six new tf test cells (one CPU + five HTP QDQ) are all rank-4. The rank-3 and rank-5 decomposition loop and StridedSlice ranges-building paths are exercised by no test, so there's no evidence the intermediate 2x Resize node passes the HTP validator at rank-5.
*Better design:* Add at least one rank-5 (or rank-3) HTP QDQ tf test cell; or, if non-rank-4 is out of scope, fail-closed the tf mode to rank==4 in IsOpSupported and drop the rank-3/5 claim from the comment.

**[D-4]** `🔵 Improvement` `[Completeness]` — `resize_test.cc` (RunQDQResizeAndAssertQnnOp)
`RunQDQResizeAndAssertQnnOp` was changed to first run a no-graph-dump probe run to detect whether to skip on this device (QCS6490 / HTP arch ≤ 68), then re-run with graph dump. The change is behavior-preserving and commented — not an unsafe Chesterton-fence removal — but it only serves the pre-existing pytorch_half_pixel tests and is unrelated to this PR's tf intent: bundled blast-radius creep that inflates the diff.
*Better design:* Split the probe-run helper change into a dedicated PR fixing the QCS6490/HTP-arch crash, leaving this PR scoped to tf_half_pixel_for_nn only.

**[D-7]** `🔵 Improvement` `[Completeness]` — `QNN-ExecutionProvider.md`
The doc's Resize row (QNN-ExecutionProvider.md:498) has an empty notes column and doesn't record the nearest+floor-only partial support. Note the table doesn't enumerate per-op coordinate-transformation modes generally (most rows have empty notes), so this is minor.
*Better design:* If recording partial support is the team convention, add "tf_half_pixel_for_nn: nearest+floor only (HTP)" to the Resize notes cell and a release note.

qti-chuteng (approved):

PR #288, opened 2026-04-22
[OpBuilder] Support HTP MatMulNBits.

minfhong-qti:
### Description

Extend MatMulNBits op builder from GPU only to HTP. Restrict the support in 2bits/4bits and block size 32/64 for HTP. Unlike GPU using BlockEncoding, HTP requires using BwFloatBlockEncoding with bitwidth set and quint8 for tensor. Thus, extend QnnQuantParamsWrapper to accept setting encodings for QNN_QUANTIZATION_ENCODING_BW_FLOAT_BLOCK type. Per HTP request, MatMulNBits is transformed into Conv2d with necessary Reshape around due to HTP timeline not able to complete the implementation for MatMul in time.
Test: UT while extending hardcoding utility functions for 2bit support.

### Motivation and Context

Enable for MSFT 2bit Phi3 model.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
@@ -1,6 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#include <cassert>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
 #include "core/providers/qnn/builder/qnn_model_wrapper.h"
@@ -10,39 +16,68 @@
 namespace onnxruntime {
 namespace qnn {
 /* Op Resolution
-    --> Incoming ONNX Node
+  --> Incoming ONNX Node
     1. MatMulNBits
-    Attributes : INT64
-      - accuracy_level  : 4
-      - bits            : 4
-      - block_size      : 32
-      - K               :
-      - N               :
-    Inputs
-      - A           :                 : (fp16/32) : [batch_size{1}, sequence_len, K]
-      - B           : Init            : (uint8)   : [N, K/block_size, (block_size * bits) / 8]
-      - scales      : Init            : (fp32)    : [N * K / block_size]
-      - zero_points : (optional)Init  : (uint8)   : [N * K / (block_size * 2)]
-      - bias        : (optional)Init  : [fp16/32] : [N]
-    Outputs
-      -  Y          :                 : (fp16/32) : [batch_size{1}, sequence_len, N]
-  <-- Outgoing QNN Node
-  1. FullyConnected
-  Attributes
-    -
-  Inputs
-    - Input           : (fp16/32)         : [batch_size{1}, sequence_len, K]
-    - Weight          : Static : (qint4)  : [N, K]
-      - Scales        : fp32              : [(N * K) / block_size{32}]
-      - Offsets       : int32_t           : [(N * K) / block_size{32}]
-    - Bias            : Static :(fp16/32) : [1, N]
-  Outputs
-    - Output          : (fp16/32)         : [batch_size{1} * sequence_len, N]
-  2. Reshape
-  Inputs
-    - Input           : (fp16/32)         : [batch_size{1} * sequence_len, N]
-  Outputs
-    - Output          : (fp16/32)         : [batch_size{1}, sequence_len, N]
+      Attributes : INT64
+        - bits            : 2(HTP), 4(GPU/HTP), 8(HTP)
+        - block_size      : 32(GPU), 16x(HTP if bits=2), 8x(HTP if bits=4), 4x(HTP if bits=8)
+        - K               :
+        - N               :
+      Inputs
+        - A           :                 : fp16/fp32 : [batch_size, sequence_len, K]
+        - B           : Init            : uint8     : [N, K / block_size, (block_size * bits) / 8]
+        - scales      : Init            : fp16/fp32 : [N, K / block_size]
+        - zero_points : (optional)Init  : uint8     : [N, (K / block_size) * bits / 8]
+      Outputs
+        -  Y          :                 : fp16/fp32 : [batch_size, sequence_len, N]
+  <-- Outgoing QNN Node (GPU)
+    1. FullyConnected
+      Attributes
+        -
+      Inputs
+        - Input           :        : fp16/fp32            : [batch_size, sequence_len, K]
+        - Weight          : Static : qint4(BlockEncoding) : [N, K]
+          - Scales        :        : fp32                 : [N * (K / block_size)]
+          - Offsets       :        : int32_t              : [N * (K / block_size)]
+      Outputs
+        - Output          :        : fp16/fp32            : [batch_size * sequence_len, N]
+    2. Reshape
+      Inputs
+        - Input           :        : fp16/fp32            : [batch_size * sequence_len, N]
+      Outputs
+        - Output          :        : fp16/fp32            : [batch_size, sequence_len, N]
+  <-- Outgoing QNN Node (HTP)
+    1. Reshape
+      Inputs
+        - Input           :        : fp16/fp32                    : [batch_size, sequence_len, K]
+      Outputs
+        - Output          :        : fp16/fp32                    : [batch_size, 1, sequence_len, K]
+    2. Cast (workaround for fp32)
+      Inputs
+        - Input           :        : fp32                         : [batch_size, 1, sequence_len, K]
+      Outputs
+        - Output          :        : fp16                         : [batch_size, 1, sequence_len, K]
+    3. Conv2d
+      Attributes : UINT32
+        - stride     : [1, 1]
+        - pad_amount : [[0, 0], [0, 0]]
+      Inputs
+        - Input           :        : fp16                         : [batch_size, 1, sequence_len, K]
+        - Weight          : Static : qint8(BwFloatBlockEncoding)  : [1, 1, K, N]
+          - Scales        :        : fp32                         : [N * (K / block_size)]
+          - Offsets       :        : fp32                         : [N * (K / block_size)]
+      Outputs
+        - Output          :        : fp16                         : [batch_size, 1, sequence_len, N]
+    4. Cast (workaround for fp32)
+      Inputs
+        - Input           :        : fp16                         : [batch_size, 1, sequence_len, N]
+      Outputs
+        - Output          :        : fp32                         : [batch_size, 1, sequence_len, N]
+    5. Reshape
+      Inputs
+        - Input           :        : fp16/fp32                    : [batch_size, 1, sequence_len, N]
+      Outputs
+        - Output          :        : fp16/fp32                    : [batch_size, sequence_len, N]
 */
 
 class MatMulNBitsOpBuilder : public BaseOpBuilder {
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -242,44 +381,101 @@ Ort::Status MatMulNBitsOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapp
...
       Qnn_TensorType_t weight_tensor_type = qnn_model_wrapper.GetTensorType(weight_tensor_name);
       const OrtValueInfo* weight_tensor_proto = qnn_model_wrapper.GetConstantTensor(weight_tensor_name);
       RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(weight_tensor_proto, quant_data, false));
 
-      // 2.2 Quantization Scales
+      // 2.2 Block-quantized scales.
       std::vector<uint8_t> per_block_uint8_scale;
       const OrtValueInfo* scale_tensor_proto = qnn_model_wrapper.GetConstantTensor(scales_tensor.name);
       RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(scale_tensor_proto, per_block_uint8_scale));
-      RETURN_IF_NOT(per_block_uint8_scale.size() == (num_blocks * sizeof(float)),
```
qti-hungjuiw, line 257: Do we need to keep this guard on `per_block_uint8_scale.size()` with `total_blocks`?

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -10,39 +13,69 @@
...
+      Inputs
+        - Input           :        : fp16                         : [batch_size, 1, sequence_len, K]
+        - Weight          : Static : qint8(BwFloatBlockEncoding)  : [1, 1, K, N]
+          - Scales        :        : fp32                         : [(K / block_size) * N]
+          - Offsets       :        : fp32                         : [(K / block_size) * N]
+      Outputs
+        - Output          :        : fp16                         : [batch_size, 1, sequence_len, N]
+    4. Cast (workaround for fp32)
+      Inputs
+        - Input           :        : fp16                         : [batch_size, sequence_len, K]
```
qti-hungjuiw: Should this `K` for Cast be `N`?
```suggestion
        - Input           :        : fp16                         : [batch_size, sequence_len, N]
```

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -242,44 +381,101 @@ Ort::Status MatMulNBitsOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapp
...
+        // 2.4 Block-quantized data.
+        std::vector<uint8_t> unpacked_quant_data;
+        // Unpack block-quantized data to uint8_t each.
+        UnpackDataToDatatype<uint8_t>(quant_data, bits, num_elements_per_uint8, unpacked_quant_data);
+        // Transform data from unsigned to signed fixed point.
+        TransformUnsignedToSignedFixedPoint(unpacked_quant_data, bits);
+
+        // Transpose block-quantized data to [K, N] which equals to [1, 1, K, N].
+        std::vector<uint8_t> transposed_unpacked_quant_data;
+        utils::TwoDimensionTranspose<uint8_t>(unpacked_quant_data,
```
qti-hungjuiw: Could we use `RETURN_IF_ERROR(utils::TwoDimensionTranspose<uint8_t>(...`?

onnxruntime/test/providers/qnn/matmulnbits_test.cc
```diff
@@ -11,63 +11,60 @@
...
   bool has_zero_point{false};
-  bool zp_is_4bit{true};
-  bool has_g_idx{false};
-  bool has_bias{false};
 };
 
-static void RunMatMul4BitsTest(const TestParams4Bits params,
-                               ExpectedEPNodeAssignment expected_ep_assignment = ExpectedEPNodeAssignment::All,
+template <int bits>
+static void RunMatMul4BitsTest(const TestParams params,
```
qti-hungjuiw: Should we rename this to `RunMatMulNBitsTest`?
```suggestion
static void RunMatMulNBitsTest(const TestParams params,
```

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -292,66 +488,142 @@ Ort::Status MatMulNBitsOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& q
...
+                                             {2, 2},
+                                             std::move(pad_amount));
+    param_tensor_names.push_back(pad_amount_param_wrapper.GetParamTensorName());
+    qnn_model_wrapper.AddParamWrapper(std::move(pad_amount_param_wrapper));
+
+    const std::string conv2d_output_name = utils::UniqueNameGenerator().New(output_tensor.name, "_conv2d");
+    std::vector<uint32_t> conv2d_output_shape = {output_info.shape[0], 1, output_info.shape[1], output_info.shape[2]};
+    QnnTensorWrapper conv2d_output_tensor_wrapper(conv2d_output_name,
+                                                  QNN_TENSOR_TYPE_NATIVE,
+                                                  QNN_DATATYPE_FLOAT_16 /*output_info.qnn_data_type*/,
```
qti-hungjuiw: Is this overriding the output datatype of conv2d?
If so, could we remove this `/*output_info.qnn_data_type*/`?

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -66,14 +99,19 @@ class MatMulNBitsOpBuilder : public BaseOpBuilder {
...
 };
 
-void MatMulNBitsOpBuilder::DQQToSignedFixedPoint4(std::vector<uint8_t>& quant_data,
-                                                  int64_t num_blocks,
-                                                  int64_t block_size) const {
+namespace {
+constexpr int64_t kByteBits = 8;
+
+// GPU-related constraints.
+static const std::unordered_set<int64_t> kGpuSupportedBits{4};
```
qti-hungjuiw: The `static` here can be removed as this is in anonymous namespace.
 
BTW,
Would it be better to use constexpr lookup to keep the things in the compile-time?
```suggestion
constexpr bool kGpuSupportedBits_lookup(int64_t bits) {
    return bits == 4;
}
```

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -10,39 +13,69 @@
...
+        - Output          :        : fp16/fp32            : [batch_size, sequence_len, N]
+  <-- Outgoing QNN Node (HTP)
+    1. Reshape
+      Inputs
+        - Input           :        : fp16/fp32                    : [batch_size, sequence_len, K]
+      Outputs
+        - Output          :        : fp16/fp32                    : [batch_size, 1, sequence_len, K]
+    2. Cast (workaround for fp32)
+      Inputs
+        - Input           :        : fp32                         : [batch_size, sequence_len, K]
```
qti-hungjuiw: Should this be `[batch_size, 1, sequence_len, K]`?
```suggestion
        - Input           :        : fp32                         : [batch_size, 1, sequence_len, K]
```

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -10,39 +13,69 @@
...
+      Attributes
+        -
+      Inputs
+        - Input           :        : fp16/fp32            : [batch_size, sequence_len, K]
+        - Weight          : Static : qint4(BlockEncoding) : [N, K]
+          - Scales        :        : fp32                 : [N * K / block_size)]
+          - Offsets       :        : int32_t              : [N * K / block_size)]
+      Outputs
+        - Output          :        : fp16/fp32            : [batch_size * sequence_len, N]
+    2. Reshape
```
qti-hungjuiw, line 44: Could we set the `keep_dims=true` in `FullyConnected` and avoid this additional `Reshape`?

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -94,99 +132,164 @@ void MatMulNBitsOpBuilder::DQQToSignedFixedPoint4(std::vector<uint8_t>& quant_da
...
+    }
+  }
+}
+
+void TransformUnsignedToSignedFixedPoint(std::vector<uint8_t>& quant_data, const int64_t bits) {
+  const int32_t zero_point = 1 << (bits - 1);
+  const uint32_t mask = (1u << bits) - 1u;
+
+  for (size_t idx = 0; idx < quant_data.size(); ++idx) {
+    uint32_t unsigned_value = static_cast<uint32_t>(quant_data[idx]) & mask;
```
huaychou: If the quant_data is not packed, why do we need mask?

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Added back.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Fixed.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Added. My mistake.

onnxruntime/test/providers/qnn/matmulnbits_test.cc
minfhong-qti: Yes. Renamed.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Originally I wanted to note it that it is overridden. I add a comment for this.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Thanks for pointing out. Remove the `static`.
However, I prefer not to define this `constexpr function` as it has limitations on the containers. And leaving the lookup below using `find` is more straightforward.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Revised.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: It seems GPU as well has constraint that `keep_dims` must be `false`. https://docs.qualcomm.com/doc/80-63442-10/topic/GpuOpDefSupplement.html#fullyconnected

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Just to make sure the upper bits are don't care.

onnxruntime/test/unittest_util/qdq_test_utils.h
```diff
@@ -58,6 +59,73 @@ GetQDQTestCaseFn BuildQDQReshapeTestCase(const std::vector<int64_t>& input_shape
...
+ * @brief Rectify min/max from a set of symmetric weights, and convert
+ *        to scale for quantization.
+ */
+template <typename ScaleT, int qbits, bool signed_quant>
+void range2scale(float min, float max, ScaleT& scale) {
+  max = std::fabs(max) > std::fabs(min) ? max : min;
+
+  // Original implementation allows negative scale to better fit larger half FP space. However, since HTP backend does
+  // not support negative scale, simplify to common formulation.
+  scale = ScaleT(max * 2 / BitsTraits<qbits, signed_quant>::fullRange);
```
qti-hungjuiw: Should this be absolute value to avoid negative scale?
```suggestion
  scale = ScaleT(std::fabs(max) * 2 / BitsTraits<qbits, signed_quant>::fullRange);
```

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
qti-hungjuiw: Got it. Thanks for the information!

onnxruntime/test/unittest_util/qdq_test_utils.h
minfhong-qti: Fixed.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -66,127 +101,189 @@ class MatMulNBitsOpBuilder : public BaseOpBuilder {
...
+
+// HTP-related constraints.
+const std::unordered_map<int64_t, int64_t> kHtpSupportedBitsAndBlockSizeMultipliers{{2, 16}, {4, 8}, {8, 4}};
+
+// This function exploits bit-manipulation trick to transform from unsigned to signed.
+// Formally, an unsigned value subtracts zero point (i.e.,  1 << (bits - 1)) to become a signed value. This subtraction
+// is essentially "flipping" the most-significant bit, which can be achieved through XOR with the zero point. While
+// taking the data being packed into consideration, simultaneously flipping all packed values can be faster than a
+// sequence of shifting, subtracting, and repacking operations for each packed value one by one.
+void TransformUnsignedToSignedFixedPoint(std::vector<uint8_t>& quant_data, const int64_t bits) {
```
minfhong-qti: Hi @huaychou and @qti-hungjuiw , I've modified the  unsigned to signed transform following Hua-Yu's suggestion. Now, quant data is first transformed to signed before unpack to avoid iterating over values multiple times.

qti-chuteng:
**[M-1]** `onnxruntime/test/providers/qnn/matmulnbits_test.cc:190-386`
All 18 new `QnnHTPBackendTests` testcases call `GTEST_SKIP() << "Skip this testcase before QAIRT 2.47."` — the entire HTP path has zero CI coverage. The kill-test fence is broken: even with the entire HTP branch commented out, this PR would still pass 100% of tests. The v2 force-push rewrote the quant-data transform with an XOR-mask trick — precisely the kind of high-risk refactor that needs runnable tests.

Suggested fix:
- Add at least 1–2 sanity tests that can actually run (e.g. `bits=4 + block_size=32`), or
- Replace `GTEST_SKIP` with the gtest `DISABLED_` prefix so they are visible to lint / dashboard tooling. Track "QAIRT 2.47 uplevel" as a TODO / ticket.

---

**[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc:305-314`
`ProcessInputs` / `ProcessAttributesAndOutputs` read `bits` / `block_size` with default `0` and immediately compute `kByteBits / bits` and `(N * K) / block_size`. Today QNN EP guarantees `IsOpSupported` runs first, so no actual SIGFPE; but this drops a previously-safe default (4 / 32) and any future "validate-only" path or refactor that exposes these methods will divide by zero.

Suggested fix:
```cpp
RETURN_IF_NOT(bits > 0 && block_size > 0,
              "Internal error: bits/block_size must be set before ProcessInputs.");
```
or restore defaults to legal values (`bits=4`, `block_size=32`).

---

**[M-3]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc:580`
HTP branch reads `output_info.shape[0..2]` without verifying rank=3 in `ProcessAttributesAndOutputs`. `IsOpSupported` only validates input A's rank, not the output's. Same pattern at line 327 for `reshape_output_shape`. ONNX spec implies output rank == input rank, but this cross-function inference is implicit.

Suggested fix:
```cpp
RETURN_IF(is_htp_backend && output_info.shape.size() != 3,
          "Unsupported output rank, expecting 3D for HTP backend.");
```
And add a static-shape guard in `IsOpSupported` so dynamic-shape input A doesn't reach the HTP path.

---

**[M-4]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc:17-81`
Doc-comment header is inconsistent with the implementation:
- `bits : 2(HTP), 4(GPU/HTP)` — missing **bits=8** which `kHtpSupportedBitsAndBlockSizeMultipliers` clearly supports.
- `block_size : 32(GPU.HTP), 64(HTP)` — period should be slash; per-bits multiplier rule (16 / 8 / 4) is not captured.
- `scales : ... fp/16fp32 ...` — token order mangled (should be `fp16/fp32`).
- GPU section (lines 40, 41): `[N * K / block_size)]` has an unbalanced bracket.
- HTP section omits the bias input even though `IsOpSupported` (line 286) still rejects bias.

---

## Minor

**[N-1]** `matmulnbits_op_builder.cc:162,301`
`bool is_htp_backend = IsNpuBackend(...)` — variable name conflates HTP with NPU. NPU == HTP today is a coincidence; if a future SDK adds another NPU backend, this code will silently mis-treat it as HTP. Either rename to `is_npu_backend` (matching `conv_op_builder.cc`) or add a `// QNN currently only has HTP as NPU backend` comment.

---

**[N-2]** `matmulnbits_op_builder.cc:583`
```cpp
QNN_DATATYPE_FLOAT_16,  // Explicitly override to float16.
```
Comment doesn't explain *why* — the rationale (HTP Conv2d FP kernel only supports fp16, hence the pre-Cast(fp16) and post-Cast(fp32) workaround) should be stated here so a future reader knows when to remove the override.

Suggested fix:
```cpp
QNN_DATATYPE_FLOAT_16, // HTP Conv2d FP kernel only supports fp16; cf. pre-Cast(fp16) above and post-Cast(fp32) below for fp32 graphs.
```

---

**[N-3]** `matmulnbits_op_builder.cc:114`
`kHtpSupportedBitsAndBlockSizeMultipliers{{2, 16}, {4, 8}, {8, 4}}` — value semantics ("multiplier") are unclear without reading `IsOpSupported`. Add a doc comment, or split into `kHtpSupportedBits` + `kHtpMinBlockBytes`.

---

**[N-4]** `matmulnbits_op_builder.cc:194`
`// TODO: Float16 DLC serialization failing.` — outdated. This PR does enable fp16 input support (line 205-206). Either delete the TODO or rephrase as "Validate fp16 DLC serialization end-to-end after QAIRT 2.47 uplevel".

---

**[N-5]** `qnn_utils.h:503-534`
New `TwoDimensionTranspose<T>(const std::vector<T>&, ...)` overload shares the name with the existing `TwoDimensionTranspose(const QnnModelWrapper&, ...)` despite very different APIs. Rename to `TwoDimensionTransposeBuffer` (or similar) and use `gsl::span<const T>` / `gsl::span<const uint32_t>` for view parameters per checklist [B2].

---

**[N-6]** `qnn_utils.h:528`
`std::memcpy(&transposed_data[dst_index], &data[src_index], sizeof(T));` is unnecessary for `T = uint8_t` and would be UB if the template were ever instantiated with a non-trivial `T`. Use `transposed_data[dst_index] = data[src_index];` or add `static_assert(std::is_trivially_copyable_v<T>)`.

---

**[N-7]** `matmulnbits_op_builder.cc:184-286`
Trailing-period style is inconsistent across `RETURN_IF` / `RETURN_IF_NOT` messages — e.g. line 184 has a period, line 187 / 224 don't, line 206 does. Standardize on trailing periods.

minfhong-qti:
> **[M-1]** `onnxruntime/test/providers/qnn/matmulnbits_test.cc:190-386` All 18 new `QnnHTPBackendTests` testcases call `GTEST_SKIP() << "Skip this testcase before QAIRT 2.47."` — the entire HTP path has zero CI coverage. The kill-test fence is broken: even with the entire HTP branch commented out, this PR would still pass 100% of tests. The v2 force-push rewrote the quant-data transform with an XOR-mask trick — precisely the kind of high-risk refactor that needs runnable tests.
> 
> Suggested fix:
> 
> * Add at least 1–2 sanity tests that can actually run (e.g. `bits=4 + block_size=32`), or
> * Replace `GTEST_SKIP` with the gtest `DISABLED_` prefix so they are visible to lint / dashboard tooling. Track "QAIRT 2.47 uplevel" as a TODO / ticket.
> 
> **[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc:305-314` `ProcessInputs` / `ProcessAttributesAndOutputs` read `bits` / `block_size` with default `0` and immediately compute `kByteBits / bits` and `(N * K) / block_size`. Today QNN EP guarantees `IsOpSupported` runs first, so no actual SIGFPE; but this drops a previously-safe default (4 / 32) and any future "validate-only" path or refactor that exposes these methods will divide by zero.
> 
> Suggested fix:
> 
> ```c++
> RETURN_IF_NOT(bits > 0 && block_size > 0,
>               "Internal error: bits/block_size must be set before ProcessInputs.");
> ```
> 
> or restore defaults to legal values (`bits=4`, `block_size=32`).
> 
> **[M-3]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc:580` HTP branch reads `output_info.shape[0..2]` without verifying rank=3 in `ProcessAttributesAndOutputs`. `IsOpSupported` only validates input A's rank, not the output's. Same pattern at line 327 for `reshape_output_shape`. ONNX spec implies output rank == input rank, but this cross-function inference is implicit.
> 
> Suggested fix:
> 
> ```c++
> RETURN_IF(is_htp_backend && output_info.shape.size() != 3,
>           "Unsupported output rank, expecting 3D for HTP backend.");
> ```
> 
> And add a static-shape guard in `IsOpSupported` so dynamic-shape input A doesn't reach the HTP path.
> 
> **[M-4]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc:17-81` Doc-comment header is inconsistent with the implementation:
> 
> * `bits : 2(HTP), 4(GPU/HTP)` — missing **bits=8** which `kHtpSupportedBitsAndBlockSizeMultipliers` clearly supports.
> * `block_size : 32(GPU.HTP), 64(HTP)` — period should be slash; per-bits multiplier rule (16 / 8 / 4) is not captured.
> * `scales : ... fp/16fp32 ...` — token order mangled (should be `fp16/fp32`).
> * GPU section (lines 40, 41): `[N * K / block_size)]` has an unbalanced bracket.
> * HTP section omits the bias input even though `IsOpSupported` (line 286) still rejects bias.
> 
> ## Minor
> **[N-1]** `matmulnbits_op_builder.cc:162,301` `bool is_htp_backend = IsNpuBackend(...)` — variable name conflates HTP with NPU. NPU == HTP today is a coincidence; if a future SDK adds another NPU backend, this code will silently mis-treat it as HTP. Either rename to `is_npu_backend` (matching `conv_op_builder.cc`) or add a `// QNN currently only has HTP as NPU backend` comment.
> 
> **[N-2]** `matmulnbits_op_builder.cc:583`
> 
> ```c++
> QNN_DATATYPE_FLOAT_16,  // Explicitly override to float16.
> ```
> 
> Comment doesn't explain _why_ — the rationale (HTP Conv2d FP kernel only supports fp16, hence the pre-Cast(fp16) and post-Cast(fp32) workaround) should be stated here so a future reader knows when to remove the override.
> 
> Suggested fix:
> 
> ```c++
> QNN_DATATYPE_FLOAT_16, // HTP Conv2d FP kernel only supports fp16; cf. pre-Cast(fp16) above and post-Cast(fp32) below for fp32 graphs.
> ```
> 
> **[N-3]** `matmulnbits_op_builder.cc:114` `kHtpSupportedBitsAndBlockSizeMultipliers{{2, 16}, {4, 8}, {8, 4}}` — value semantics ("multiplier") are unclear without reading `IsOpSupported`. Add a doc comment, or split into `kHtpSupportedBits` + `kHtpMinBlockBytes`.
> 
> **[N-4]** `matmulnbits_op_builder.cc:194` `// TODO: Float16 DLC serialization failing.` — outdated. This PR does enable fp16 input support (line 205-206). Either delete the TODO or rephrase as "Validate fp16 DLC serialization end-to-end after QAIRT 2.47 uplevel".
> 
> **[N-5]** `qnn_utils.h:503-534` New `TwoDimensionTranspose<T>(const std::vector<T>&, ...)` overload shares the name with the existing `TwoDimensionTranspose(const QnnModelWrapper&, ...)` despite very different APIs. Rename to `TwoDimensionTransposeBuffer` (or similar) and use `gsl::span<const T>` / `gsl::span<const uint32_t>` for view parameters per checklist [B2].
> 
> **[N-6]** `qnn_utils.h:528` `std::memcpy(&transposed_data[dst_index], &data[src_index], sizeof(T));` is unnecessary for `T = uint8_t` and would be UB if the template were ever instantiated with a non-trivial `T`. Use `transposed_data[dst_index] = data[src_index];` or add `static_assert(std::is_trivially_copyable_v<T>)`.
> 
> **[N-7]** `matmulnbits_op_builder.cc:184-286` Trailing-period style is inconsistent across `RETURN_IF` / `RETURN_IF_NOT` messages — e.g. line 184 has a period, line 187 / 224 don't, line 206 does. Standardize on trailing periods.

Most comments are applied.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -66,127 +101,167 @@ class MatMulNBitsOpBuilder : public BaseOpBuilder {
...
+              ("QNN HTP does not support MatMulNBits with bits=" + std::to_string(bits)).c_str());
+    RETURN_IF(block_size % it->second != 0,
+              ("QNN HTP does not support MatMulNBits with block_size=" + std::to_string(block_size)).c_str());
+  }
+  RETURN_IF_NOT((K % block_size) == 0, "K must be divisible by block_size.");
 
-  RETURN_IF_NOT(bits == 4, "Invalid bits. Qnn Gpu Only Supports MatMulNBits with bits == 4");
-  RETURN_IF_NOT(block_size == 32, "Invalid block_size. Qnn Gpu Only Supports MatMulNBits with block_size == 32");
-  RETURN_IF_NOT((K % block_size) == 0, "K must be divisible by block_size");
-  RETURN_IF_NOT(((N * K) % (2 * block_size)) == 0,
```
huaychou, line 118: Do we need this check?

huaychou (approved):
LGTM. Could you please check if https://github.com/onnxruntime/onnxruntime-qnn/pull/307 will impact MatmulNBits support?

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: I think it's redundant as `K % block_size` already covers it. The "2" originally indicates the two 4-bit values packed inside one 8-bit value which is naturally guaranteed.

minfhong-qti:
> LGTM. Could you please check if #307 will impact MatmulNBits support?

I think it will not impact as that PR is handling separate Q/DQ nodes while MatMulNBits inherently has scales/offsets as direct inputs.

qti-chuteng (approved):

PR #289, opened 2026-04-22
Add gcov-based code coverage support (Linux x86_64)

yuhuchua-qti:
## Summary      
                                                                                                                                                                                                                                                                                                                    
  Adds opt-in gcov-based code coverage measurement for the QNN EP on Linux x86_64.                                                                                                                                                                                                                                  
  Coverage is always RelWithDebInfo, never affects Release builds, and has zero impact on the                                                                                                                                                                                                                                
  existing CI matrix.                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                    
  ## Changes                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                    
  **Build system:**                                                                                                                                                                                                                                                                                                 
  - `version_script_coverage.lds` — new linker script that exports all symbols
    (used only when `ENABLE_COVERAGE=1`; production `version_script.lds` unchanged)                                                                                                                                                                                                                                 
  - `cmake/onnxruntime_providers_qnn.cmake` — adds `--coverage -g -O0 -fprofile-abs-path` for GCC                                                                                                                                                                                                                                          
  - `qcom/scripts/linux/build.sh` — new `--enable-coverage` flag that sets `ENABLE_COVERAGE:BOOL=ON` when invoking `build.py`                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                    
  **Coverage tooling:**                                                                                                                                                                                                                                                                                             
  - `qcom/packages.yml` — new `lcov_linux_x86_64` entry (v1.16; no extra Perl module dependencies unlike lcov 2.x)                                                                                                                                                                                                                                                                            
  - `qcom/scripts/linux/tools.sh` — new `get_lcov_bindir()` helper                                                                                                                                                                                                                                                  
  - `qcom/scripts/linux/generate_coverage.sh` — standalone script: clears stale counters, runs `onnxruntime_provider_test --gtest_filter="*Qnn*"`, collects and filters coverage data, generates HTML report via genhtml                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                    
  **Task orchestrator:**                                                                                                                                                                                                                                                                                            
  - `qcom/build_and_test.py` — new `coverage_linux_x86_64` public task (Linux x86_64 only): RelWithDebInfo build with coverage flags → generate_coverage.sh                                                                                                                                                                                                                                            
[...]

PR #294, opened 2026-04-23
Add function-level unit test infrastructure with coverage build support

yuhuchua-qti:
## Summary

Introduces a function-level / component-level unit test infrastructure for QNN EP internals, enabling fine-grained coverage of QnnModelWrapper and qnn_def logic that is unreachable through existing integration tests.

Problem: The QNN EP ships as a MODULE (dlopen-only) library. Its internal symbols are inaccessible to external test binaries, so tests have historically been limited to the public EP interface. This makes it impossible to write targeted unit tests for internal
components without running on-device hardware.

Solution: Under ENABLE_COVERAGE=ON only, rebuild the EP as a SHARED library that the test binary can link against directly.

## Infrastructure changes (coverage build only)

- EP rebuilt as SHARED so onnxruntime_provider_test can call internal symbols via target_link_libraries().
- version_script_coverage.lds exports all symbols (vs. production script which exports only the factory entry point); --gc-sections intentionally omitted to preserve gcov-instrumented sections.
- QNN_EP_FUNCTION_LEVEL_UT=1 defined on the test binary to activate test utilities guarded by the macro.

Production builds are unaffected — the EP is still a MODULE, the version script still hides internals, and all unit-test utility code compiles to empty translation units.

## QnnModelWrapper — zero test-infrastructure impact on production source

No changes to qnn_model_wrapper.h or .cc beyond the PR's feature changes. Tests call the production constructor directly through two adapters in qnn_unit_test_utils.h:

- fake_graph_sentinel_: a stable int member cast to const OrtGraph&. Stubs receive this pointer but never dereference it.
- MakeNullLogger(): constructs an Ort::Logger with cached_severity_level_=ORT_LOGGING_LEVEL_FATAL via memcpy. ORT_CXX_LOG checks the cached level before calling LogMessage, so all EP log calls (severity < FATAL) short-circuit without dereferencing the null
OrtLogger*. A static_assert(sizeof(Ort::Logger) == 2*sizeof(void*)) guards against future layout changes.

OrtApi function pointers that the path under test exercises are stubbed via the zero-initialized stub_ort_api member; all others remain null and are never reached.

Real-backend tests use QNN HTP

Tests covering ValidateQnnNode need a live Qnn_BackendHandle_t to exercise backendValidateOpConfig. They dlopen("libQnnHtp.so") on the x86-64 host — HTP supports the validation path without HTP hardware (no graph execution, no QNN context/session created). QNN
CPU is intentionally not used since CPU backend support is being removed. Tests are auto-skipped (GTEST_SKIP) when the SDK is unavailable.
  
## New test files

  | File | Tests | Source covered |
  |---|---|---|
  | `unit/qnn_def_test.cc` | 86 | `builder/qnn_def.cc` |
  | `unit/qnn_model_wrapper_test.cc` | 115 | `builder/qnn_model_wrapper.cc` |
  | `unit/qnn_unit_test_utils.h` | — | Shared fixtures, `OrtApi` stubs, real-HTP-backend context |
  | `unit/README.md` | — | Design doc + UT policy + agent prompt template |
  
The README codifies what reviewers had to repeatedly enforce on this PR — file layout, naming, mock-strategy table (lowest-cost layer per dependency), forbidden constructs (private ORT headers, noexcept lambda traps, dangling-reference parameter pitfalls), coverage targets, verification checklist — and ships a self-contained agent prompt block so future contributors (human or AI) can add tests without re-deriving the rules.
  
## Test plan
  
- 201/201 unit tests pass (./onnxruntime_provider_test --gtest_filter="QnnUnit_*")
- Unit-only coverage on directly tested files: qnn_def.cc 96.3% lines / 100% functions; qnn_model_wrapper.cc 91.8% lines / 96.0% functions
- Overall coverage (coverage_linux_x86_64, integration + unit): qnn_def.cc 96.3% / 100%; qnn_model_wrapper.cc 95.5% / 98.0%
- Release build (build_ort_linux_x86_64) succeeds — production source unchanged, no regression
- Lint passes — no private ORT headers, no unapproved cmake include paths


qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L120`
The `QnnModelWrapper` class now has an interleaved access-specifier layout: `public:` (production ctor) → `public:` (test-only ctor) → `private:` (implementation ctor) → `public:` (DISALLOW macro, destructor, methods). This public→private→public pattern violates QNN EP's convention of grouping access specifiers into contiguous sections, makes the public interface boundary harder to identify, and increases the risk that future contributors miss the constraints of the private implementation constructor.

**Suggested fix**: Reorganize into two contiguous blocks — all `public:` declarations first (production ctor, test-only ctor, DISALLOW, destructor, methods), then all `private:` members (implementation ctor, data members).

```suggestion
 public:
   // Production constructor (const OrtGraph& / const Ort::Logger& — non-null guaranteed)
   QnnModelWrapper(const OrtGraph& ort_graph, const ApiPtrs& api_ptrs, const Ort::Logger& logger, ...);
 #if QNN_EP_FUNCTION_LEVEL_UT
   // Test-only constructor (nullptr for ort_graph and logger)
   QnnModelWrapper(std::nullptr_t, const ApiPtrs& api_ptrs, std::nullptr_t, ...);
 #endif
   ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(QnnModelWrapper);
   ~QnnModelWrapper() = default;
   // ... public methods ...

 private:
   // Single implementation constructor; both public overloads delegate here.
   QnnModelWrapper(const OrtGraph* ort_graph, const ApiPtrs& api_ptrs, const Ort::Logger* logger, ...);
   // ... data members ...
```

---

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L197`
`GetInitializerTensors()` and `FindInitializer()` pass `ort_graph_ptr_` (which is null when constructed via the test-only ctor) directly into `api_ptrs_.ort_api.Graph_GetInitializers(ort_graph_ptr_, ...)`. Current tests are safe because `QnnModelWrapperTestContext` pre-stubs `Graph_GetNumInitializers` and `Graph_GetInitializers` by default. However, any future test that replaces or resets `stub_ort_api` without re-adding those two stubs will forward the null graph pointer to a zero-initialized function pointer → crash with no diagnostic message. Recommend adding a `QNN_EP_FUNCTION_LEVEL_UT`-guarded null-check at the top of each function, or documenting this invariant prominently in the README and `qnn_unit_test_utils.h`.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L423`
`GetOrtGraph()`'s null guard calls `ORT_CXX_API_THROW`, which — under `ORT_API_MANUAL_INIT` (set by `ORT_UNIT_TEST_BUILD`) — relies on the global ORT API pointer being initialized. If a test reaches this path before the global API is set, `ORT_CXX_API_THROW` itself may null-dereference, making diagnosis harder than intended. Consider replacing with `throw std::runtime_error(...)` to avoid the dependency on the global ORT API pointer.

**[N-2]** `cmake/onnxruntime_providers_qnn.cmake → install()`
When the coverage build switches the target from MODULE to SHARED, the existing `install(TARGETS ... ARCHIVE DESTINATION ...)` rule will also install `libonnxruntime_providers_qnn.a` to `${CMAKE_INSTALL_LIBDIR}`. For a coverage build, this `.a` is an unwanted artifact. Consider gating the ARCHIVE component of the install rule on `NOT ENABLE_COVERAGE`.

yuhuchua-qti:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L120` The `QnnModelWrapper` class now has an interleaved access-specifier layout: `public:` (production ctor) → `public:` (test-only ctor) → `private:` (implementation ctor) → `public:` (DISALLOW macro, destructor, methods). This public→private→public pattern violates QNN EP's convention of grouping access specifiers into contiguous sections, makes the public interface boundary harder to identify, and increases the risk that future contributors miss the constraints of the private implementation constructor.
> 
> **Suggested fix**: Reorganize into two contiguous blocks — all `public:` declarations first (production ctor, test-only ctor, DISALLOW, destructor, methods), then all `private:` members (implementation ctor, data members).
> 
> ```
>  public:
>    // Production constructor (const OrtGraph& / const Ort::Logger& — non-null guaranteed)
>    QnnModelWrapper(const OrtGraph& ort_graph, const ApiPtrs& api_ptrs, const Ort::Logger& logger, ...);
>  #if QNN_EP_FUNCTION_LEVEL_UT
>    // Test-only constructor (nullptr for ort_graph and logger)
>    QnnModelWrapper(std::nullptr_t, const ApiPtrs& api_ptrs, std::nullptr_t, ...);
>  #endif
>    ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(QnnModelWrapper);
>    ~QnnModelWrapper() = default;
>    // ... public methods ...
> 
>  private:
>    // Single implementation constructor; both public overloads delegate here.
>    QnnModelWrapper(const OrtGraph* ort_graph, const ApiPtrs& api_ptrs, const Ort::Logger* logger, ...);
>    // ... data members ...
> ```
> 
> **[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L197` `GetInitializerTensors()` and `FindInitializer()` pass `ort_graph_ptr_` (which is null when constructed via the test-only ctor) directly into `api_ptrs_.ort_api.Graph_GetInitializers(ort_graph_ptr_, ...)`. Current tests are safe because `QnnModelWrapperTestContext` pre-stubs `Graph_GetNumInitializers` and `Graph_GetInitializers` by default. However, any future test that replaces or resets `stub_ort_api` without re-adding those two stubs will forward the null graph pointer to a zero-initialized function pointer → crash with no diagnostic message. Recommend adding a `QNN_EP_FUNCTION_LEVEL_UT`-guarded null-check at the top of each function, or documenting this invariant prominently in the README and `qnn_unit_test_utils.h`.
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L423` `GetOrtGraph()`'s null guard calls `ORT_CXX_API_THROW`, which — under `ORT_API_MANUAL_INIT` (set by `ORT_UNIT_TEST_BUILD`) — relies on the global ORT API pointer being initialized. If a test reaches this path before the global API is set, `ORT_CXX_API_THROW` itself may null-dereference, making diagnosis harder than intended. Consider replacing with `throw std::runtime_error(...)` to avoid the dependency on the global ORT API pointer.
> 
> **[N-2]** `cmake/onnxruntime_providers_qnn.cmake → install()` When the coverage build switches the target from MODULE to SHARED, the existing `install(TARGETS ... ARCHIVE DESTINATION ...)` rule will also install `libonnxruntime_providers_qnn.a` to `${CMAKE_INSTALL_LIBDIR}`. For a coverage build, this `.a` is an unwanted artifact. Consider gating the ARCHIVE component of the install rule on `NOT ENABLE_COVERAGE`.

Hi, below are responses for above comments.

**M-1**

Applied. 
Pulled the impl ctor down to the top of the private: block, leaving one contiguous public: block (production ctor → test-only ctor → DISALLOW → dtor → all public methods) followed by private: (impl ctor → private methods → data members).
The class now has 2 access-specifier transitions instead of 4 in this region. Updated the impl ctor's adjacency comment ("public overloads above" → "Both public overloads delegate here") since "above" no longer holds.

---
**M-2**

Pushing back on Option 1 (production null-check) — the actual crash is on the zero-initialized function pointer, not the null ort_graph_ptr_, and adding QNN_EP_FUNCTION_LEVEL_UT-guarded checks per method doesn't scale (every helper using ort_graph_ptr_ would
need one).

Implemented a tighter alternative — fail at test-helper construction:
```
// In QnnModelWrapperTestContext::CreateWrapper()
assert(stub_ort_api.Graph_GetNumInitializers != nullptr && "...stub missing...");
assert(stub_ort_api.Graph_GetInitializers != nullptr && "...stub missing...");
```

This catches the "test reset stub_ort_api then called CreateWrapper" path — the only risky one, since ApiPtrs is captured by value, so post-CreateWrapper resets don't affect the wrapper.
Production code untouched, clear message at construction instead of silent SIGSEGV later. Also strengthened the stub comment in qnn_unit_test_utils.h with an explicit INVARIANT block.

---
**N-1**

Pushing back — the premise doesn't hold:

1. ORT_CXX_API_THROW doesn't touch the global API pointer. Ort::Exception is just a std::exception wrapper around std::string + OrtErrorCode enum. Ctor / what() / dtor are pure stdlib. The QNN EP has ~75 usages across 20 files today (28 in qnn_def.cc alone) under ORT_API_MANUAL_INIT; if the macro were init-sensitive, the EP would have crashed long ago.
2. The test binary uses auto-init, not manual. ort_api.h:25-27 only defines ORT_API_MANUAL_INIT when ORT_UNIT_TEST_BUILD is not set, and cmake/onnxruntime_unittests.cmake:360 sets ORT_UNIT_TEST_BUILD on onnxruntime_provider_test. So the test binary's TUs run auto-init.

Switching to std::runtime_error would lose the OrtErrorCode and break the codebase pattern with no concrete benefit.

---
**N-2**

Doesn't manifest on Linux — SHARED targets produce only .so, no .a for the ARCHIVE component to install. 
Confirmed against coverage build artifacts (libonnxruntime_providers_qnn.so only; no .a). 
The coverage SHARED branch is also gated on UNIX AND NOT APPLE AND x86_64, so Windows never goes through this path. Adding NOT ENABLE_COVERAGE on ARCHIVE would be a no-op.

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h → CreateWrapper()`

`assert()` is silently stripped by the preprocessor when `NDEBUG` is defined — which CMake's `RelWithDebInfo` build type does by default. The coverage build (`ENABLE_COVERAGE=ON`) uses `RelWithDebInfo`, so the null-stub guards in `CreateWrapper()` are completely disabled in precisely the environment where these unit tests are run. A developer who resets `stub_ort_api` and forgets to re-add the required stubs will get a silent SIGSEGV instead of a clear test failure.

```suggestion
// Replace assert() with a mechanism that fires in Release builds too
if (stub_ort_api.Graph_GetNumInitializers == nullptr ||
    stub_ort_api.Graph_GetInitializers == nullptr) {
  throw std::logic_error(
      "Graph_GetNumInitializers / Graph_GetInitializers stubs missing "
      "— re-add them after resetting stub_ort_api");
}
```

---

**[M-2]** `onnxruntime/test/providers/qnn/unit/README.md`

The README's final paragraph states "Coverage builds run in CI on every PR. Strict regression gating (failing PRs that drop coverage) is being rolled out in stages." — but this PR diff contains no changes to `.github/workflows/`. The claimed CI behavior lacks infrastructure support. If CI integration is planned for a follow-up PR, the README should say so using future tense rather than present tense.

Suggested wording:
> "Coverage builds are planned to run in CI on every PR. CI workflow integration will be added in a follow-up PR."

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/version_script_coverage.lds`

The version script has `global: *;` with no `local:` section. While this is functionally correct for the intent (export all symbols), the omission looks like an oversight rather than a deliberate choice. A brief comment explaining why the `local:` section is intentionally absent would help future maintainers.

```suggestion
/* No local: section — intentional. This coverage script exports ALL symbols
 * (global: *) so the test binary can call any EP-internal function directly.
 * A local: * entry would be contradictory here. */
VERS_1.0 {
  global:
    *;
};
```

---

**[N-2]** `onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc`

Tests use `reinterpret_cast<const OrtValueInfo*>(&g_type_info_sentinel)` where `g_type_info_sentinel` is a `static int`. This works as long as stub functions treat it as an opaque non-null token without dereferencing. The assumption should be documented near the sentinel declaration so future maintainers don't accidentally add a dereference.

```suggestion
// Sentinel lvalue used as a non-null OrtValueInfo* token in unit tests.
// All stubs receiving this pointer MUST NOT dereference it — they only
// check for non-null or pass it through to another stub unchanged.
static int g_type_info_sentinel = 0;
```

yuhuchua-qti:
> ### 🟠 Major
> **[M-1]** `onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h → CreateWrapper()`
> 
> `assert()` is silently stripped by the preprocessor when `NDEBUG` is defined — which CMake's `RelWithDebInfo` build type does by default. The coverage build (`ENABLE_COVERAGE=ON`) uses `RelWithDebInfo`, so the null-stub guards in `CreateWrapper()` are completely disabled in precisely the environment where these unit tests are run. A developer who resets `stub_ort_api` and forgets to re-add the required stubs will get a silent SIGSEGV instead of a clear test failure.
> 
> ```
> // Replace assert() with a mechanism that fires in Release builds too
> if (stub_ort_api.Graph_GetNumInitializers == nullptr ||
>     stub_ort_api.Graph_GetInitializers == nullptr) {
>   throw std::logic_error(
>       "Graph_GetNumInitializers / Graph_GetInitializers stubs missing "
>       "— re-add them after resetting stub_ort_api");
> }
> ```
> 
> **[M-2]** `onnxruntime/test/providers/qnn/unit/README.md`
> 
> The README's final paragraph states "Coverage builds run in CI on every PR. Strict regression gating (failing PRs that drop coverage) is being rolled out in stages." — but this PR diff contains no changes to `.github/workflows/`. The claimed CI behavior lacks infrastructure support. If CI integration is planned for a follow-up PR, the README should say so using future tense rather than present tense.
> 
> Suggested wording:
> 
> > "Coverage builds are planned to run in CI on every PR. CI workflow integration will be added in a follow-up PR."
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/version_script_coverage.lds`
> 
> The version script has `global: *;` with no `local:` section. While this is functionally correct for the intent (export all symbols), the omission looks like an oversight rather than a deliberate choice. A brief comment explaining why the `local:` section is intentionally absent would help future maintainers.
> 
> ```
> /* No local: section — intentional. This coverage script exports ALL symbols
>  * (global: *) so the test binary can call any EP-internal function directly.
>  * A local: * entry would be contradictory here. */
> VERS_1.0 {
>   global:
>     *;
> };
> ```
> 
> **[N-2]** `onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc`
> 
> Tests use `reinterpret_cast<const OrtValueInfo*>(&g_type_info_sentinel)` where `g_type_info_sentinel` is a `static int`. This works as long as stub functions treat it as an opaque non-null token without dereferencing. The assumption should be documented near the sentinel declaration so future maintainers don't accidentally add a dereference.
> 
> ```
> // Sentinel lvalue used as a non-null OrtValueInfo* token in unit tests.
> // All stubs receiving this pointer MUST NOT dereference it — they only
> // check for non-null or pass it through to another stub unchanged.
> static int g_type_info_sentinel = 0;
> ```

Hi, below are responses for above comments.

**M-1:**

Fixed. 
assert() is stripped under NDEBUG, which CMake's RelWithDebInfo (the coverage build's config) defines by default. 
Replaced with throw std::logic_error(...) so the guard fires in every build configuration. 
Added a comment near the throw explaining why assert() is not used here.

----

**M-2:**

Pushing back. 
The coverage CI workflow .github/workflows/qualcomm-internal-coverage.yml was added in #289 and enhanced in #313 / #401 / #442 — coverage builds already run on every PR today, this PR doesn't need to wire it up. 
The "rolled out in stages" wording is for the strict-gating part, which is still a follow-up. 
Keeping the README as-is.

----

**N-1:**

Done — added the comment block you suggested above the VERS_1.0 declaration.

----

**N-2:**

Done — added the no-dereference contract comment, broadened to cover all three sentinels (g_type_info_sentinel, g_shape_info_sentinel, g_initializer_value_sentinel) since the contract applies to each.

----
Thanks.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -58,24 +62,42 @@ class QnnModelWrapper {
...
+
+  // Test-only constructor. Accepts std::nullptr_t for both ort_graph and logger.
+  // Conditionally compiled only when QNN_EP_FUNCTION_LEVEL_UT is defined; absent in
+  // all production builds, so it cannot be called from production code by accident.
+  //
+  // Neither OrtGraph nor Ort::Logger can be mocked or constructed using only public
+  // ORT headers without a running ORT session. Tests that need to query graph data or
+  // log must stub the relevant OrtApi function pointers in the ApiPtrs they provide.
+  // ORT_CXX_LOG_PTR silently skips logging when logger is null.
+#if QNN_EP_FUNCTION_LEVEL_UT
```
minfhong-qti: Not really a fan of doing this. Below are what claude suggests, please evaluate whether possible.
```c
// OrtApi is a public struct of function pointers.
// Take the real one, copy it, override only the Graph_* / GetValueInfo* functions the code path under test calls, and point them at your own test state:

// Test state your stubs read from — this IS your "mock graph"
struct FakeGraph {
  std::vector<const OrtValueInfo*> initializers;
  // ...whatever the code under test queries
};

FakeGraph fake;

// Start from the real, fully-populated API table so unrelated pointers are valid
OrtApi api = *OrtGetApiBase()->GetApi(ORT_API_VERSION);

// Override only what the test exercises:
api.Graph_GetNumInitializers = [](const OrtGraph* g, size_t* out) -> OrtStatusPtr {
  auto* f = reinterpret_cast<const FakeGraph*>(g);
  *out = f->initializers.size();
  return nullptr;
};
// ...similarly Graph_GetInitializers, GetValueInfoName, Graph_GetModelPath, etc.

ApiPtrs api_ptrs{api, /*ep_api*/..., /*model_editor_api*/...};

// Form the OrtGraph& from your fake's address. The wrapper only re-takes &ort_graph_,
// so the bytes are never read except by your own stubs that cast back.
const OrtGraph& g = *reinterpret_cast<const OrtGraph*>(&fake);

QnnModelWrapper w(g, api_ptrs, Ort::Logger{}, /*qnn_interface*/..., ...);
```

tirupath-qti (approved):

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
yuhuchua-qti: Agreed — went further than just removing the #if guard and eliminated all test-infrastructure impact from qnn_model_wrapper.h and .cc:

- Reverted to reference member storage (ort_graph_, logger_); removed the private pointer-based constructor and GetOrtGraph() null-check that were introduced solely to support nullable graph/logger in tests.
- Reverted ORT_CXX_LOG_PTR back to ORT_CXX_LOG throughout .h and .cc.

The test side (qnn_unit_test_utils.h) now uses:
- A fake_graph_sentinel_ (int member) cast to const OrtGraph& — same reinterpret_cast pattern already used for the other sentinels.
- MakeNullLogger(): constructs an Ort::Logger with cached_severity_level_= ORT_LOGGING_LEVEL_FATAL via memcpy. ORT_CXX_LOG checks the cached level before calling LogMessage, so all EP log calls (severity < FATAL) short-circuit without touching the null OrtLogger*. Guarded by static_assert(sizeof(Ort::Logger) == 2*sizeof(void*)) to catch future layout changes at compile time.

Tests call the production constructor directly — no test-only overload needed.

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
```diff
@@ -0,0 +1,235 @@
...
+              &fatal, sizeof(OrtLoggingLevel));
+  return logger;
+}
+
+// Context for constructing a QnnModelWrapper in function-level unit tests.
+//
+//   ctx.input_info.indices  = {{"input0", 0}};   // declare graph inputs
+//   ctx.output_info.indices = {{"output0", 0}};  // declare graph outputs
+//   auto wrapper = ctx.CreateWrapper(settings);
+struct QnnModelWrapperTestContext {
```
minfhong-qti: Try to refactor this struct and decouple `OrtApi`-related stubs and `QnnModelWrapper`-related construction to allow the former one to be reused in other UTs. In addition, prefer to move class-specific struct into corresponding test file (i.e., `QnnModelWrapperTestContext` in `qnn_model_wrapper_test.cc`).

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
```diff
@@ -0,0 +1,2469 @@
...
+// On Linux x86-64 (the unit-test host), HTP supports validation but not graph
+// execution — these tests intentionally exercise only the validation path.
+// They are skipped automatically when the library is unavailable.
+
+// ValidateQnnNode succeeds for a valid Relu op on the HTP backend.
+// Covers ValidateQnnNode → QnnGraphOpValidation → backendValidateOpConfig (success path).
+// Uses UFIXED_POINT_8 with per-tensor quant params — the representative HTP production path.
+TEST(QnnUnit_ModelWrapperTest, ValidateQnnNode_HtpBackend_Relu_Succeeds) {
+  QnnRealHtpBackendContext backend;
+  if (!backend.IsValid()) GTEST_SKIP() << "Could not load libQnnHtp.so or create HTP backend";
```
minfhong-qti: Should we treat backend not valid acceptable in CI? I suppose it should always be valid.

cmake/onnxruntime_unittests.cmake
```diff
@@ -383,6 +384,16 @@ block()
...
   endif()
 
+  # Coverage build: link against the SHARED QNN EP library so tests can call
+  # EP-internal functions directly. Coverage is recorded in the .so's .gcda files and
+  # collected by lcov --directory <build_dir> (recursive search finds them automatically).
+  if(ENABLE_COVERAGE AND UNIX AND NOT APPLE AND CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+    target_link_libraries(onnxruntime_provider_test PRIVATE onnxruntime_providers_qnn)
+    # Enable function-level / component-level unit tests that require direct linkage against
+    # EP-internal symbols (available only when the EP is a SHARED library).
+    target_compile_definitions(onnxruntime_provider_test PRIVATE QNN_EP_FUNCTION_LEVEL_UT=1)
```
minfhong-qti: Since there is no code change in main folder, is this macro still necessary? Should these UTs execute in non-coverage stages as well?

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
yuhuchua-qti: Agreed. Split into two layers:
- OrtApiStubContext (stays in qnn_unit_test_utils.h) — reusable container for the three stub tables + initializer-query defaults + a MakeApiPtrs() invariant guard. Any future test that needs OrtApi stubbing uses this directly.
- QnnModelWrapperTestContext moves into qnn_model_wrapper_test.cc's anonymous namespace, inherits OrtApiStubContext publicly, adds wrapper-specific members + CreateWrapper(). The shared header no longer leaks wrapper concerns. Inheritance over composition keeps ctx.stub_ort_api.X = ... working in 100+ test bodies without churn.

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
yuhuchua-qti: Agreed — libQnnHtp.so ships with the QAIRT SDK that the EP build already depends on, so a missing library is a CI configuration error. 
Both GTEST_SKIP() sites are now ASSERT_TRUE(backend.IsValid()); the helper's docstring and section header updated to match.

cmake/onnxruntime_unittests.cmake
yuhuchua-qti: Right — the original #if QNN_EP_FUNCTION_LEVEL_UT inside qnn_model_wrapper.h is gone, so the macro no longer gates any production code. 
Its remaining role is build-side: it tracks whether the test binary is link-time bound to the SHARED EP. 
When off, the bodies under unit/ compile to empty translation units, so non-coverage builds don't see undefined references to EP-internal symbols. 
Removing it would break non-coverage builds.

Renamed QNN_EP_FUNCTION_LEVEL_UT → QNN_EP_INTERNAL_SYMBOL_ACCESS to reflect what it actually gates (test-side direct linkage to EP internals, not a production-code guard).
The cmake comment now spells this out.

Running these UTs in non-coverage stages is deferred as future work. 
We want to keep them in the Linux x86_64 coverage build only for now to confirm stability of this initial set. 
Once the broader UT migration plan settles and the patterns prove themselves, we'll migrate to other CI stages — the minimum-impact path is an independent
ENABLE_QNN_EP_INTERNAL_SYMBOL_ACCESS cmake option so a regular CI build can opt in without enabling coverage instrumentation.

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -38,7 +38,17 @@
     list(APPEND onnxruntime_providers_qnn_all_srcs "${ONNXRUNTIME_ROOT}/core/providers/qnn/onnxruntime_providers_qnn.rc")
   endif()
 
-  onnxruntime_add_shared_library_module(onnxruntime_providers_qnn ${onnxruntime_providers_qnn_all_srcs})
+  if(ENABLE_COVERAGE AND UNIX AND NOT APPLE AND CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
```
minfhong-qti, line 41: Just curious. Is `APPLE` ever used / set to True is current build pipeline?

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
```diff
@@ -0,0 +1,2524 @@
...
+#if !defined(ORT_MINIMAL_BUILD) && QNN_EP_INTERNAL_SYMBOL_ACCESS
+
+#include "core/providers/qnn/builder/qnn_def.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "test/providers/qnn/unit/qnn_unit_test_utils.h"
+
+using namespace onnxruntime;
+using namespace onnxruntime::qnn;
+
+namespace onnxruntime {
```
minfhong-qti:
```suggestion
namespace onnxruntime {
namespace qnn {
```

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
```diff
@@ -0,0 +1,2524 @@
...
+
+  EXPECT_EQ(wrapper->GetTensorType("unmarked"), QNN_TENSOR_TYPE_NATIVE);
+  wrapper->MarkTensorAsFoldedConstant("folded");
+  EXPECT_EQ(wrapper->GetTensorType("folded"), QNN_TENSOR_TYPE_STATIC);
+}
+
+}  // namespace test
+}  // namespace onnxruntime
+
+#endif  // !defined(ORT_MINIMAL_BUILD) && QNN_EP_INTERNAL_SYMBOL_ACCESS
```
minfhong-qti: Add final new line.

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
```diff
@@ -0,0 +1,2524 @@
...
+//   QnnModelWrapperTestContext ctx;
+//   ctx.input_info.indices  = {{"input0", 0}};   // declare graph inputs
+//   ctx.output_info.indices = {{"output0", 0}};  // declare graph outputs
+//   auto wrapper = ctx.CreateWrapper(settings);
+//
+// Inherits OrtApiStubContext so tests keep using ctx.stub_ort_api directly
+// (e.g. ctx.stub_ort_api.GetTensorData = MyStub) and pick up the initializer-
+// query stubs + MakeApiPtrs() invariant guard for free.
+struct QnnModelWrapperTestContext : public OrtApiStubContext {
+  QNN_INTERFACE_VER_TYPE qnn_interface = QNN_INTERFACE_VER_TYPE_INIT;
```
minfhong-qti, line 33: Prefer explicitly include necessary QNN libs.

cmake/onnxruntime_providers_qnn.cmake
yuhuchua-qti: No — there is no macOS build target in this pipeline. build_and_test.py has no macOS tasks and there are no macOS CI workflows. 
The NOT APPLE condition is inherited boilerplate from the upstream ORT CMake infrastructure.

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
yuhuchua-qti: Done.

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
yuhuchua-qti: Done. 
Moved the test code into namespace onnxruntime::qnn::test and dropped all qnn:: prefixes. 
Since OrtApiStubContext / QnnRealHtpBackendContext / MakeNullLogger live in onnxruntime::test, added explicit using declarations to bridge them in.

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
yuhuchua-qti: Done. 
Added QnnBackend.h, QnnInterface.h, QnnOpDef.h, QnnTypes.h as direct includes rather than relying on transitives.

onnxruntime/test/providers/qnn/unit/qnn_model_wrapper_test.cc
```diff
@@ -0,0 +1,2533 @@
...
+#include "test/providers/qnn/unit/qnn_unit_test_utils.h"
+
+namespace onnxruntime {
+namespace qnn {
+namespace test {
+
+// Bridge helpers defined in onnxruntime::test (qnn_unit_test_utils.h) into this namespace.
+using ::onnxruntime::test::MakeNullLogger;
+using ::onnxruntime::test::OrtApiStubContext;
+using ::onnxruntime::test::QnnRealHtpBackendContext;
```
minfhong-qti: Sorry, I did not make it clear. I prefer not to use `using`, and therefore I prefer to use `QnnModelWrapper` as `qnn::QnnModelWrapper` and remove the `namespace qnn` above.

yuhuchua-qti: Done. Reverted namespace qnn and removed all using declarations. The test code stays in namespace onnxruntime::test; all QNN EP types are now explicitly qualified throughout (qnn::ModelSettings, qnn::QnnModelWrapper, qnn::QnnTensorWrapper, etc.).

minfhong-qti (approved):

PR #295, opened 2026-04-23
fix: propagate QNN tensor ID when override name already registered

huaychou:
### Description

- Change `tensors_created_table` from `unordered_map<string, bool>` to `unordered_map<string, uint32_t>` to store the QNN-assigned tensor ID on first creation. When a duplicate name is found, copy the stored ID to the current tensor via a new `SetQnnTensorID` helper, mirroring the existing `GetQnnTensorID` pattern.
- Add regression test `OffloadGraphIoQuantizationMultipleQDQPairsOnGraphInput` that reproduces the fan-out scenario using the CPU backend.

### Motivation and Context

- When `offload_graph_io_quantization=1` and a graph input is consumed by multiple QDQ pairs, all those pairs share the same override name (e.g. "h0"). `RegisterGraphInputOutputInOrder` calls `CreateTensorInQnnGraph` for each, but the second call finds the name already in `tensors_created_table` and returns early without setting `qnn_tensor.id`, leaving it as 0. Later graph composition fails with `"Failed to compose Qnn graph"`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
@@ -100,6 +100,8 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
       // two APP_READ tensors with the same external name (which reduces the composed
       // QNN graph's input count and causes a null slot in qnn_tensor_infos at runtime).
       if (!qnn_model_wrapper.IsExternalOverrideTarget(node_unit.Outputs()[0].name)) {
+        // The tensor name override is used to align the output name of DLC produced by IRBackend
+        // with the output name of original onnx graph for better consistency.
         qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Inputs()[0].name,
                                                 /*external=*/node_unit.Outputs()[0].name);
       }
@@ -117,10 +119,12 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
         qnn_model_wrapper.IsGraphInput(node_unit.Inputs()[0].name)) {
       // Only register the override for the first Q node that consumes this graph input.
       // If another Q node already maps to the same external name, skip registration so
-      // that the second output becomes a separate APP_WRITE tensor instead of creating
+      // that the second input becomes a separate APP_WRITE tensor instead of creating
       // two APP_WRITE tensors with the same external name (which reduces the composed
       // QNN graph's input count and causes a null slot in qnn_tensor_infos at runtime).
       if (!qnn_model_wrapper.IsExternalOverrideTarget(node_unit.Inputs()[0].name)) {
+        // The tensor name override is used to align the input name of DLC produced by IRBackend
+        // with the input name of original onnx graph for better consistency.
         qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Outputs()[0].name,
                                                 /*external=*/node_unit.Inputs()[0].name);
       }
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
@@ -94,8 +94,15 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
 
     if (qnn_model_wrapper.GetModelSettings().offload_graph_io_quantization &&
         qnn_model_wrapper.IsGraphOutput(node_unit.Outputs()[0].name)) {
-      qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Inputs()[0].name,
-                                              /*external=*/node_unit.Outputs()[0].name);
+      // Only register the override for the first DQ node that consumes this graph output.
+      // If another DQ node already maps to the same external name, skip registration so
+      // that the second output becomes a separate APP_READ tensor instead of creating
+      // two APP_READ tensors with the same external name (which reduces the composed
+      // QNN graph's input count and causes a null slot in qnn_tensor_infos at runtime).
+      if (!qnn_model_wrapper.IsExternalOverrideTarget(node_unit.Outputs()[0].name)) {
+        qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Inputs()[0].name,
+                                                /*external=*/node_unit.Outputs()[0].name);
+      }
       return MAKE_EP_FAIL("QNN EP is configured to not take DQ nodes that generate a graph output.");
     }
   }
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
qti-yuduo: Semantic change for this method after this fix.. This ones is not consumed, we can remove entirely?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -2279,6 +2279,60 @@ TEST_F(QnnCPUBackendTests, PartitionAddedInputRegisteredAsGraphInputOffloadGraph
...
+// Regression test for AISW-174227: a graph input that fans out to multiple QDQ pairs
+// with offload_graph_io_quantization=1 must not fail graph composition. Previously,
+// the second DQ tensor received QNN tensor id=0 (never set) because CreateTensorInQnnGraph
+// returned early when it found the override name already in tensors_created_table.
+TEST_F(QnnCPUBackendTests, OffloadGraphIoQuantizationMultipleQDQPairsOnGraphInput) {
+  std::unique_ptr<ModelAndBuilder> model;
+  CreateModelInMemory(model, BuildGraphInputFanoutQDQModel(), "graph_input_fanout_qdq", 18);
+
+  Ort::SessionOptions so;
+  so.SetGraphOptimizationLevel(ORT_ENABLE_ALL);
```
qti-yuduo: DISABLE all the optimizations otherwise, it may not have the model pattern we're trying to test.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -473,9 +473,9 @@ class QnnModelWrapper {
   // All QnnParamWrapper for the graph
   std::unordered_map<std::string, QnnParamWrapper> model_params_map_;
   std::vector<QnnOpProperty> qnn_op_property_list_;
-  // <tensor_name, qnn_tensor_created> -- true means qnn tensor created in qnn graph
-  // it includs normal qnn_tensors and qnn_tensors inside param_tensors
-  std::unordered_map<std::string, bool> tensor_created_map_;
+  // <tensor_name, qnn_tensor_id> -- stores the QNN-assigned ID once the tensor is created
+  // it includes normal qnn_tensors and qnn_tensors inside param_tensors
+  std::unordered_map<std::string, uint32_t> tensor_created_map_;
```
qti-yuduo: With this semantic change, let's rename it to be more explicit, maybe `qnn_tensor_id_name_map` or `qnn_tensor_id_by_name` ? (also make sure parameters are consistent after renaming)

qti-yuduo (commented):
Thanks for the fix!

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
huaychou: Removed

onnxruntime/test/providers/qnn/qnn_basic_test.cc
huaychou: Done

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
huaychou: Done

chenweng-quic (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
```diff
@@ -94,8 +94,15 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
...
         qnn_model_wrapper.IsGraphOutput(node_unit.Outputs()[0].name)) {
-      qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Inputs()[0].name,
-                                              /*external=*/node_unit.Outputs()[0].name);
+      // Only register the override for the first DQ node that consumes this graph output.
+      // If another DQ node already maps to the same external name, skip registration so
+      // that the second output becomes a separate APP_READ tensor instead of creating
+      // two APP_READ tensors with the same external name (which reduces the composed
+      // QNN graph's input count and causes a null slot in qnn_tensor_infos at runtime).
+      if (!qnn_model_wrapper.IsExternalOverrideTarget(node_unit.Outputs()[0].name)) {
+        qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Inputs()[0].name,
```
chenweng-quic, line 105: It would be nice if you can add comment about why we want to override here.
e.g., 
For DequantizeLinear, if offload_graph_io_quantization is true, then it will be delegated to CPU, and what we see in Qnn graph output is input of DequantizeLinear input, it causes inconvenience for debugging. We map QnnGraph output to onnx graph output here to simplify accuracy debugging.
```
               | (QnnGraph output, CPU input)
      DequantizerLinear
               | (CPU output, onnx graph output)
```

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
```diff
@@ -108,8 +115,15 @@ Ort::Status SimpleOpBuilder::ExplicitOpCheck(QnnModelWrapper& qnn_model_wrapper,
...
         qnn_model_wrapper.IsGraphInput(node_unit.Inputs()[0].name)) {
-      qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Outputs()[0].name,
-                                              /*external=*/node_unit.Inputs()[0].name);
+      // Only register the override for the first Q node that consumes this graph input.
+      // If another Q node already maps to the same external name, skip registration so
+      // that the second output becomes a separate APP_WRITE tensor instead of creating
+      // two APP_WRITE tensors with the same external name (which reduces the composed
+      // QNN graph's input count and causes a null slot in qnn_tensor_infos at runtime).
+      if (!qnn_model_wrapper.IsExternalOverrideTarget(node_unit.Inputs()[0].name)) {
+        qnn_model_wrapper.SetTensorNameOverride(/*internal=*/node_unit.Outputs()[0].name,
```
chenweng-quic, line 128:
```
               | (onnx graph input, CPU input)
     QuantizerLinear
               | (CPU output, QnnGraph input)
```

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
huaychou: Sure.
Hi @qti-yuduo, could you help provide the reason why we want to override the input/output names here? Then, I can add it into the comment. Thanks.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
qti-yuduo: This is required for IRBackend -- where we serialize QNN IR to DLC for later use (AOT mode).

qti-yuduo (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
huaychou: Done

onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
huaychou: Done

PR #296, opened 2026-04-23
[QNN-EP] Handle broadcast for 5D Matmul with 4d input_0.

ankus-qti:
### Description
2D input_0 to 5D Matmul is broadcasted by QNN_EP to 4D input_0.
To make shapes of same dimension, we need to add a reshape before input_0.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -174,6 +174,44 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnMatMul(QnnModelWrapper& qnn_mode
   RETURN_IF_ERROR(ProcessInput0(qnn_model_wrapper, input_info_0, org_input_0_name, input_names,
                                 logger, do_op_validation));
 
+  // If input_0's rank is less than input_1's rank (e.g., input_0 is 2D but input_1 is 5D),
+  // Insert a Reshape node that prepends leading 1s to input_0 to match input_1's rank.
+  const size_t rank_0 = input_info_0.shape.size() < 2 ? 2 : input_info_0.shape.size();  // After ProcessInput0, 1D becomes 2D
+  const size_t rank_1 = input_info_1.shape.size() < 2 ? 2 : input_info_1.shape.size();
+  if (rank_0 < rank_1) {
+    // The input tensor is already in the map from ProcessInput0(); capture its name before popping.
+    const std::string current_input_0_name = input_names.back();
+    const std::string broadcast_input_0_name =
+        utils::UniqueNameGenerator().New(org_input_0_name, "_reshape_broadcast");
+
+    // Determine the actual current shape of input_0 after ProcessInput0 (1D -> 2D, otherwise unchanged).
+    std::vector<uint32_t> current_shape_0 =
+        (input_info_0.shape.size() == 1) ? std::vector<uint32_t>{1, input_info_0.shape[0]} : input_info_0.shape;
+
+    // Prepend (rank_1 - rank_0) ones to match input_1's rank.
+    std::vector<uint32_t> broadcast_shape_0(rank_1 - rank_0, 1);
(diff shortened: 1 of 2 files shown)


qti-chuteng:
**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:207`

This PR only handles the `rank_0 < rank_1` direction. The symmetric `rank_0 > rank_1` case (e.g., 5D × 2D, 4D × 2D where input_0 is the higher-rank tensor) is neither handled nor explicitly rejected — it falls straight into QNN MatMul and either fails QNN validation or produces a wrong-shaped output, with raw QNN noise as the only diagnostic. ONNX MatMul's multi-directional broadcast is symmetric, so users hitting this in real models is realistic.

Suggested fix — pick one:
- (a) Make the fix symmetric: when `rank_1 < rank_0`, insert the same broadcast Reshape on input_1, and rename `broadcast_input_0_rank` to `broadcast_input_rank`.
- (b) If only one direction is intended for this PR, explicitly reject the other in `IsOpSupported()` (or early in the op builder), call out the limitation in the PR description, and add a test (e.g., `5D × 2D`) pinning the rejection.

---

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:226-237`

The broadcast-reshape insertion open-codes `AddTensorWrapper` + `CreateQnnNode(QNN_OP_RESHAPE, …)` instead of using the existing `QnnModelWrapper::AddReshapeNode` helper (used elsewhere in this same file at line 138 / 263). The "input is already registered in the tensor map" comment doesn't justify a separate path — that case appears in other op builders too. Please switch to the helper:

```cpp
RETURN_IF_ERROR(qnn_model_wrapper.AddReshapeNode(
    current_input_0_name, broadcast_input_0_name,
    current_shape_0, broadcast_shape_0,
    input_info_0.qnn_data_type,
    input_info_0.quant_param, broadcast_quant_param_0,
    do_op_validation,
    /*is_for_input=*/false, /*is_for_output=*/false));
```

---

**[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:226-228`

Both `broadcast_quant_param_0.Copy()` and `std::vector<uint32_t>(broadcast_shape_0)` are unused after construction — they can be `std::move`'d, matching the pattern at lines 134-135 of this file:

```cpp
QnnTensorWrapper broadcast_output_tensor(broadcast_input_0_name, QNN_TENSOR_TYPE_NATIVE,
                                         input_info_0.qnn_data_type,
                                         std::move(broadcast_quant_param_0),
                                         std::move(broadcast_shape_0));
```

(Mostly moot if N-1 is taken — `AddReshapeNode` does this internally.)

---

**[N-3]** `onnxruntime/test/providers/qnn/matmul_test.cc:299-323`

The new test hand-rolls `TestInputDef<float>` + `GetFloatDataInRange` + `RunQnnModelTest` instead of using the file's existing `RunMatMulOpTest(...)` helper that every other case in `QnnHTPBackendTests.MatMulOp` uses. Please follow the file convention:

```cpp
RunMatMulOpTest({64, 64}, {1, 192, 33, 64, 1}, false, false,
                ExpectedEPNodeAssignment::All, "htp", 18, 1e-2f);
```

---

**[N-4]** `onnxruntime/test/providers/qnn/matmul_test.cc:299`

A single FP32-dynamic 2D × 5D test is too narrow. Please add:
- A QDQ variant (uint8 and uint16) — extending the existing `MatMulOp_QDQ` block with `RunQDQMatMulOpTest<uint8_t, uint8_t, uint8_t>({64, 64}, {1, 192, 33, 64, 1}, false, false);` (and uint16 equivalent). Per-channel quant goes through `HandleUnsqueeze`, so a QDQ test is necessary to verify the quant axis after prepending 1s.
- A few more shape combinations: 1D × 5D (interacts with `reshape_input_0` + broadcast — most complex case), 3D × 5D, 2D × 4D, 3D × 4D.
- An `input_0` initializer variant.
- A "kill-test fence" note in the PR description: confirm the new test fails when the fix is disabled (checklist [G02]).

PR #300, opened 2026-04-24
[OpBuilder] Fix MatMul uint16 handling.

minfhong-qti:
### Description

MatMul builder inserted a Convert at dynamic input 1 to convert uint16 to asymmetric uint8 in order to pass HTP validation. However, HTP now supports symmetric uint16 for MatMul input 1 and such Convert is no longer required, which in fact may cause accuracy issue. Revise the Convert insertion logic to only insert one to convert asymmetric uint16 to symmetric uint16.
TODO: There is similar logic for static input 1 where a Convert is inserted to convert uint16 into symmetric int16. Since not sure the original motivation, leave modification for such case like dynamic input 1 as future work if encoutering real cases.

### Motivation and Context

MEP PSD8 model suffers accuracy issue due to MatMul opbuilder converting dynamic input 1 from uint16 to uint8.


tirupath-qti (approved):

huaychou (approved):
LGTM :shipit:

PR #302, opened 2026-04-27
[QNN EP]: Fusion of multiply and reciprocal to divide

ankipand-qti:
### Description
The QNN HTP/DSP backend has no native Reciprocal operator. When a standalone Reciprocal feeds directly into a Mul, the two-node sub-graph is mathematically equivalent to a single Div: 
` Mul(a, Reciprocal(b))  ==  Div(a, b)`

I have added ReciprocalMulFusion, an IQnnNodeGroup that detects this pattern and lowers it to a single QNN_OP_ELEMENT_WISE_DIVIDE node, keeping the entire computation on the NPU accelerator and avoiding a CPU fallback.

### Motivation and Context
**Why change is required**: Without the fusion, FP16 and FP32 models with the Reciprocal -> Mul pattern would either fall back to CPU or use an unnecessarily inefficient two-node implementation on the accelerator.
**Problem the change solves**: Reciprocal -> Mul falls back to CPU without the fusion


qti-chuteng:
**[C-1]** `onnxruntime/test/providers/qnn/qnn_node_group/reciprocal_mul_fusion_test.cc:630`

`QnnHTPBackendTests::ShouldSkipIfHtpFp16Unsupported()` does not exist in the test framework — onnxruntime_test_all will fail to link on every platform. This is the only occurrence of that name in the entire repo. Use the same helper that line 588 of this file uses for the U16 case:

```cpp
if (QnnHTPBackendTests::ShouldSkipIfHtpArchIsLessThanOrEqualTo(QNN_HTP_DEVICE_ARCH_V68)) {
  GTEST_SKIP() << "FP16 fusion requires HTP arch > V68";
}
```

## Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/reciprocal_op_builder.cc:41-54`, `onnxruntime/test/providers/qnn/simple_op_test.cc:1098-1122`

The PR's stated goal is "fuse Reciprocal+Mul → Div", but it also (a) makes `ReciprocalOpBuilder::IsOpSupported` reject unquantized Reciprocal on HTP and (b) flips the existing `Reciprocal_Basic_FLOAT` from `ExpectedEPNodeAssignment::All` to `::None`. This is scope creep and a behavior regression: any standalone fp32/fp16 Reciprocal without a downstream Mul (some LayerNorm variants, alternative softmax forms) will now silently fall back to CPU EP, undetectable from the PR description. Also, the inline rationale ("HTP cannot execute ElementWiseDivide(static_1.0, dynamic_x)") is inconsistent with `Reciprocal_QU8` continuing to pass via the same `Div(static_1.0, dynamic_x)` lowering on the QDQ path.

Recommend splitting into a follow-up PR with a minimal HTP-failure reproducer, an explanation of why the QDQ 1.0 case is accepted but the fp32/fp16 case is not (with the exact SDK version), and a "behavior change" section listing the model patterns that move from All to None/Some. If kept here, at least flip the affected tests to `DISABLED_` with a TODO instead of changing the expected EP assignment.

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc:316-348, 411-431`

The Q-output → DQ-output traversal for QDQGroup Reciprocal is implemented twice (in `TryFusion` Step 3 and again in `CreateOrValidateOnQnn`). Same logic, slightly different `RETURN_IF_NOT` strings. Resolve `numerator_def` / `denominator_def` once in `TryFusion`, store them as members on the fusion object, and have `CreateOrValidateOnQnn` consume the cached values. This is the pattern that `hardsigmoid_mul_fusion.cc` follows.

**[M-3]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc:466-501`

In the build path, three `MakeTensorWrapper` calls are unconditional, but the resulting wrappers are dropped (via `std::move`) only inside `if (!IsQnnTensorWrapperExist(...))` blocks. When the tensor is already registered (common: graph inputs touched by upstream ops, denominator shared across LayerNorm-like patterns), the just-built wrapper is silently discarded, wasting a `GetTensorInfo` + shape resolution + quant-param extraction + vector allocation. Wrap each `MakeTensorWrapper` behind the existence check. The same pattern in `gelu_fusion.cc` is also wrong, but as a brand-new file this is the right place to fix it.

## Minor

**[N-1]** `onnxruntime/test/providers/qnn/simple_op_test.cc:1119-1135`

The new `Reciprocal_FP16` test pins `ExpectedEPNodeAssignment::None`. As a "lock current unsupported state" canary it has limited value — best treated together with M-1 (move to follow-up PR or use `DISABLED_` prefix with a TODO).

**[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc:301-305, 346-348`

The `recip_is_input0 && recip_is_input1` "degenerate case" branch is effectively dead code — `GetChildNodeUnitAllowQdq`'s single-consumer guard already prevents the producer-feeds-both-Mul-slots scenario. Keep the branch as defence-in-depth, but reword the comment so future readers don't mistake it for a critical guard.

**[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc:24-27, 81-86`

Alphabetical reorder of `spacetodepth_fusion.h` include and the `Erf` op-type case is bundled with the feature commit. Prefer a separate `[NFC]` commit; if force-pushing again is impractical, just call it out in the PR description.

**[N-4]** `onnxruntime/test/providers/qnn/qnn_node_group/reciprocal_mul_fusion_test.cc` (whole file)

`json_dir` everywhere — existing fusion tests (gelu, channel_shuffle, gather_transpose_reshape) use `json_qnn_graph_dir`. Pure consistency.

**[N-5]** `onnxruntime/test/providers/qnn/qnn_node_group/reciprocal_mul_fusion_test.cc` (every TEST_F)

CWD-relative `json_dir` paths are fragile under `--gtest_repeat` / `--gtest_shuffle` and across CI runners with different CWDs. Tolerated by other fusion tests today; consider a shared RAII helper or `temp_directory_path()` in a future cleanup PR.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.h
```diff
@@ -0,0 +1,155 @@
...
+/// ---------------------
+/// FP32 / FP16 (SingleNode):
+///
+///   [denominator] --> Reciprocal --+
+///                                  v
+///   [numerator]  ----------------> Mul --> [output]
+///
+/// Quantized (QDQGroup):
+///
+///   [denominator] --> DQ --> Reciprocal --> Q --+
```
tirupath-qti: we should avoid fusing this pattern as 1/b is no longer separately quantized.

onnxruntime/core/providers/qnn/builder/opbuilder/reciprocal_op_builder.cc
```diff
@@ -37,7 +38,10 @@ Ort::Status ReciprocalOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrappe
   const auto& outputs = node_unit.Outputs();
   RETURN_IF_NOT(outputs.size() == 1, "Reciprocal operator must have exactly 1 output.");
 
-  // Check input type is float for CPU.
+  // On the QNN CPU backend only float32 is accepted; other backends (HTP, GPU)
```
tirupath-qti, line 41: this is not adding anything. Please avoid this change.

onnxruntime/core/providers/qnn/builder/opbuilder/reciprocal_op_builder.cc
```diff
@@ -3,6 +3,7 @@
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
+#include "core/providers/qnn/builder/qnn_def.h"
```
tirupath-qti, line 6: this is not needed as there are no changes in this file.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc
```diff
@@ -0,0 +1,520 @@
...
+    const std::string dq_output_name = dq_outputs[0].GetName();
+
+    recip_is_mul_input0 = (mul_inputs[0].name == dq_output_name);
+    recip_is_mul_input1 = (mul_inputs[1].name == dq_output_name);
+
+    if (!recip_is_mul_input0 && !recip_is_mul_input1) {
+      return nullptr;
+    }
+    if (recip_is_mul_input0 && recip_is_mul_input1) {
+      // Defence-in-depth: same reasoning as the SingleNode branch above.
```
tirupath-qti: The control can still reach here for edge case: Mul(reciprocal(b), reciprocal(b));
keep a simple comment.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc
```diff
@@ -0,0 +1,520 @@
...
+//
+// The numerator arrives directly; the denominator travels through Reciprocal
+// first.  Neither path is a subset of the other, so the earliest node where
+// BOTH are available is Mul.  Mul is therefore the correct target.
+//
+// Contrast with HardSigmoidMulFusion, which returns node_units_[0]
+// (HardSigmoid) as its target.  That fusion shares a single root tensor x
+// for both branches:
+//
+//   [x] --> HardSigmoid --+
```
tirupath-qti: what is the relation of HardSigmoid in this fusion?

tirupath-qti (commented):
Please note: keep important information short and clear in comments.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc
```diff
@@ -0,0 +1,520 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+// =============================================================================
```
qti-kromero: nit: throughout these comments seem a bit verbose, not sure if that is aligned with other fusion comments

onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc
```diff
@@ -0,0 +1,520 @@
...
+//   3. Confirm the Mul actually consumes the Reciprocal output (sanity check
+//      against malformed graphs where the lookup might return a Mul that is
+//      connected via a different edge).
+//   4. Perform a QNN dry-run validation to ensure the backend can handle the
+//      resulting ElementWiseDivide node.
+//   5. Construct and return the ReciprocalMulFusion object.
+//
+// Note: explicit input/output count guards for Reciprocal (unary) and Mul
+// (binary) are intentionally absent — ONNX spec compliance is assumed per
+// the QNN EP review checklist [T06].  GetChildNodeUnitAllowQdq (Step 2) and
```
qti-kromero: appears to reference some document not publicly available

onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc
```diff
@@ -0,0 +1,520 @@
...
+//   QNN Div input[0]  = numerator
+//   QNN Div input[1]  = denominator
+//   QNN Div output[0] = result
+//
+// =============================================================================
+
+#include "core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.h"
+
+#include <array>
+#include <gsl/gsl>
```
qti-kromero: I think this should be after the standard library includes

onnxruntime/core/providers/qnn/builder/qnn_node_group/reciprocal_mul_fusion.cc
qti-kromero: I think earlier check about single consumer prevents this unless I'm misunderstanding

qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🟠 Concern` `[Arch]` — `reciprocal_mul_fusion.cc:145` (and :178)

The new fusion emits the legacy standalone op `QNN_OP_ELEMENT_WISE_DIVIDE` directly in both branches of `CreateOrValidateOnQnn` (line 145 validate, line 178 build). But the codebase has already converged the "divide" decision onto the unified `QNN_OP_ELEMENT_WISE_BINARY` op plus the `QNN_OP_ELEMENT_WISE_BINARY_OPERATION_DIVIDE` param scalar — verified by grep, every one of the other six divide-emitting sites uses the unified form: `inverse_op_builder.cc:322`, `mean_op_builder.cc:102`, `mod_op_builder.cc:151`, `reciprocal_op_builder.cc:103`, `simple_op_builder.cc:400`, `tan_op_builder.cc:128`.

This fusion is the **only** remaining call site of the legacy `QNN_OP_ELEMENT_WISE_DIVIDE` in the whole repo. It re-splits the "which QNN op represents divide" decision — which had exactly one owner — back into two (⟦C3.1⟧ same decision in >1 place, ⟦C1.1⟧ change amplification). The starkest part: this PR itself edits `reciprocal_op_builder.cc`, whose lines 103–107 use the converged binary+DIVIDE form — so the same PR carries both spellings side by side.

*Better design:* Switch to the unified form — call `CreateQnnNode`/`ValidateQnnNode` with `QNN_OP_ELEMENT_WISE_BINARY` and add the DIVIDE param via `AddQnnScalar<uint32_t>(...)`, mirroring `reciprocal_op_builder.cc:100–112` exactly. If the legacy op was chosen deliberately (e.g. an SDK requirement specific to this fusion), state the reason in the PR description or a code comment — there is none today.

### 🔵 Completeness

**[D-2]** `🟠 Concern` `[Completeness]` — `docs/execution_providers/QNN-ExecutionProvider.md` (Operator decomposition fusions table)

This PR adds an operator-decomposition fusion but does not add a corresponding row to the "Operator decomposition fusions" table. That table is the established convention for this class of fusion: every sibling has a row — Gelu (`Erf → … → Mul` → `QNN_OP_GELU`), HardSwish (`HardSigmoid → Mul` → `QNN_OP_ELEMENT_WISE_NEURON`), LayerNorm, and Softmax are all listed. The new Reciprocal → Mul behavior is invisible to users — a "fix one place, miss three" documentation gap (⟦C6.1⟧).

*Better design:* Add a row matching the sibling format, e.g. `| Reciprocal → Mul (single consumer) | QNN_OP_ELEMENT_WISE_DIVIDE | Reciprocal must be SingleNode (preserves separate quantization of 1/b), single Mul consumer, not a graph output. |`. Keep the Fused-QNN-op column consistent with the final decision on D-1.

**[D-3]** `🔵 Improvement` `[Completeness]` — `reciprocal_mul_fusion_test.cc` (guard at `reciprocal_mul_fusion.cc:140`)

The fusion explicitly declines "both Mul inputs are the same Reciprocal output" at `reciprocal_mul_fusion.cc:140` (`recip_is_mul_input0 && recip_is_mul_input1` → return nullptr, commented "would change semantics"). The test file covers the other decline paths (two-consumers, graph-output, QDQ-wrapped), but no test constructs a `Mul(recip, recip)` graph to verify this branch actually declines. This is a guard the author themselves labeled correctness-relevant; without a test, a later refactor of the input-slot logic could break it undetected.

*Better design:* Add a test case constructing `Reciprocal(b) → recip; Mul(recip, recip) → out` and assert the pattern is NOT fused into a single Divide (e.g. `AssertOpInQnnGraph(dir, "ElementWiseDivide", 0)`), mirroring the existing `*_NoFusion` test style.

PR #280, opened 2026-04-21
[QNN EP] Enable ARM64 Linux Unit Tests

qti-mbadnara:
This PR enables ARM64 Linux Unit Tests. Disables `INT4 / INT16 / FP16 / FP32` tests on `HTP` and `GPU` runtime. Disables File Mapping tests on `v68` architecture.

Co-authored by: [Hua-Yu Chou](https://github.com/huaychou)

PR #303, opened 2026-04-27
[QNN EP] Enable Linux ARM64 tests as part of CI

qti-mbadnara:
This PR does the following for `Linux ARM64` as part of `CI`

- Enables Unit Tests. 
- Enables `onnxruntime_provider_test` and `onnxruntime_plugin_ep_onnx_test`.
- Skips `QDQ Model Test suite` as part of CI due to known issues.
- Skips `MaxPool_Rank3` Test on CPU for ARM64 Linux.
- Skips `QDQ tests` using `INT4 / INT8 / INT16 / INT32` on architectures `<= v68` for `HTP` and `GPU` runtimes.
- Skips `FP16 / FP32` tests on architectures `<= v68` for `HTP` and `GPU` runtimes.
- Skips `File Mapping`, `Fusion Tests`, `Context Binary` tests on architectures `<= v68`.
- Removed `ShouldSkipIfHtpFp16Unsupported` as not used anywhere else in the code base.

Co-authored by: [Hua-Yu Chou](https://github.com/huaychou)

This PR is similar to #280 but acts on reviewer comments to move the skip logic to the header files


onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -276,6 +276,19 @@ void RunQnnModelTest(const GetTestModelFn& build_test_case, ProviderOptions prov
...
+  std::string backend_name = "htp";
+  if (provider_options.find("backend_type") != provider_options.end()) {
+    backend_name = provider_options.at("backend_type");
+  }
+
+  if (backend_name == "htp" || backend_name == "gpu") {
+    if (QnnHTPBackendTests::ShouldSkipIfHtpArchIsLessThanOrEqualTo(QNN_HTP_DEVICE_ARCH_V68)) {
+      std::string backend_upper = backend_name;
+      backend_upper[0] = std::toupper(backend_upper[0]);
+      GTEST_SKIP() << "Test requires " << backend_upper << " FP32/FP16 support (arch > V68)";
```
tirupath-qti: This message for GPU Unit tests is not relevant. Seems we need to check gpu functionality with a different check rather than QnnHTPBackendTests::ShouldSkipIfHtpArchIsLessThanOrEqualTo(QNN_HTP_DEVICE_ARCH_V68)

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -1586,27 +1604,68 @@ class QnnHTPBackendTests : public ::testing::Test {
...
 
+// Helper to check if we should skip QDQ tests on v68
+template <typename QuantType>
+inline bool ShouldSkipQDQTestOnV68(const ProviderOptions& qnn_options) {
+  std::string backend_name = "htp";
+  if (qnn_options.find("backend_type") != qnn_options.end()) {
+    backend_name = qnn_options.at("backend_type");
+  }
+
+  if (backend_name == "htp" || backend_name == "gpu") {
```
tirupath-qti: for GPU backend, we should check different condition. Can we disable GPU backend tests for all linux aarch64 platform?

qcom/scripts/linux/run_tests.sh
```diff
@@ -110,22 +118,39 @@ count_errors ./ctest --verbose --timeout 10800 --stop-on-failure "${exclude_args
...
+    # test strnormalizer_export_monday_casesensintive_upper
+    # test strnormalizer_export_monday_empty_output
+    # Removing them from test suite
+
+    if [ "$(uname -m)" == "aarch64" ]; then
+        rm -rf "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node/test_strnormalizer_export_monday_casesensintive_lower"
+        rm -rf "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node/test_strnormalizer_export_monday_casesensintive_nochangecase"
+        rm -rf "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node/test_strnormalizer_export_monday_casesensintive_upper"
+        rm -rf "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node/test_strnormalizer_export_monday_empty_output"
+    fi
```
kuanyul-qti, line 134: Using `rm -rf` to delete directories from the `cmake/external/onnx` git submodule is destructive and permanently removes test data from the workspace for all subsequent CI runs. Use filter/exclude flags to skip unsupported test cases instead of destroying source tree directories.
Instead of `rm -rf`, use exclude flags if `onnxruntime_plugin_ep_onnx_test` supports them
e.g., add `--exclude` or `--filter` arguments to skip strnormalizer tests on aarch64

```suggestion

exclude_tests=()
if [ "$(uname -m)" == "aarch64" ]; then
    exclude_tests+=(
        "test_strnormalizer_export_monday_casesensintive_lower"
        "test_strnormalizer_export_monday_casesensintive_nochangecase"
        "test_strnormalizer_export_monday_casesensintive_upper"
        "test_strnormalizer_export_monday_empty_output"
    )
fi
# Compose to --exclude or --filter arguments
```

qti-kromero: what is the problem with removing them? the repo should be recreated on each CI run as per my understanding of actions/checkout@v6

kuanyul-qti: Not sure if any tests will run after this. If there aren't any more tests, I'm fine with this.

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -1713,6 +1727,69 @@ bool ReduceOpHasAxesInput(const std::string& op_type, int opset_version);
   } while (0)
 #endif
 
+// Helper to check if we should skip GPU backend tests on Linux ARM64
```
kuanyul-qti: If you want to skip tests on "Linux ARM64", I think `#if defined(__linux__)` is needed.

qti-mbadnara: I think using #if defined(__linux__) will skip the tests on any Linux architecture

Using #if defined(__aarch64__) will skip it only on AArch64 with GCC/Clang (__aarch64__).
It does not skip on MSVC Windows ARM64 (_M_ARM64 / _M_ARM64EC)

onnxruntime/test/providers/qnn/nonzero_test.cc
```diff
@@ -52,6 +52,20 @@ static void RunNonZeroTest(const GetTestModelFn& build_test_case,
   const std::unordered_map<std::string, int> domain_to_version = {{"", opset_version}, {kMSDomain, 1}};
   ModelTestBuilder helper;
   build_test_case(helper);
+
+  std::string backend_name = "htp";
```
qti-kromero: does something prevent us from using ShouldSkipFp16TestOnV68 here?

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -276,6 +276,19 @@ void RunQnnModelTest(const GetTestModelFn& build_test_case, ProviderOptions prov
                      int opset_version, ExpectedEPNodeAssignment expected_ep_assignment,
                      float fp32_abs_err, OrtLoggingLevel log_severity, bool verify_outputs,
                      std::function<void(const Graph&)>* ep_graph_checker) {
+  std::string backend_name = "htp";
```
qti-kromero: similar comment regarding ShouldSkipFp16TestOnV68

qcom/scripts/linux/run_tests.sh
```diff
@@ -78,20 +89,17 @@ orig_build_dir=$(sed -n "s@# Build directory: @@p" CTestTestfile.cmake)
...
 log_info "-=-=-=- Running ctests -=-=-=-"
-# TODO: [AISW-164203] ORT test failures on Rubik Pi
 exclude_args=()
-if [ "$(uname -m)" == "aarch64" ]; then
-    exclude_args+=(--exclude-regex "onnxruntime_provider_test")
-fi
 count_errors ./ctest --verbose --timeout 10800 --stop-on-failure "${exclude_args[@]}"
 
 # TODO: We will support python wheel in linux
+# TODO: [AISW-164203] ORT test failures on Rubik Pi
```
qti-kromero: why is this moved from below location that seemed more relevant to this comment specifically?

qcom/scripts/linux/run_tests.sh
```diff
@@ -32,24 +32,35 @@ function run_model_test() {
     local suite="${2}"
     local test_path="${3:-testdata/${suite}}"
 
-    log_info "-=-=-=- Running onnx/models ${suite} tests with the ABI-stable EP plugin -=-=-=-=-"
+    local model_log="${build_dir}/${suite}_model_tests.log"
+    local model_xml="${build_dir}/${suite}_model_tests.results.xml"
+
+    # Remove old log and XML files
```
qti-kromero, line 38: what was the issue observed here? the workspace wasn't properly cleaning up log files?

qcom/scripts/linux/run_tests.sh
```diff
@@ -32,24 +32,35 @@ function run_model_test() {
...
     set +e
     "${build_dir}/onnxruntime_plugin_ep_onnx_test" \
         -j 1 \
         --plugin_ep_libs "qnn|libonnxruntime_providers_qnn.so" \
         --plugin_eps qnn \
         -i "backend_type|${backend}" \
-        "${test_path}" 2>&1 | tee "${build_dir}/${suite}_model_tests.log"
-    rc=$?
+        "${test_path}" 2>&1 | tee "${model_log}"
+    test_result=$?
```
qti-kromero: nit: rc -> return_code, I'm fine with removing abbreviations but can we make this test_return_code or similar so that it's clear it's a return code when reading. changing it to test_result makes this slightly less obvious IMO

qcom/scripts/linux/run_tests.sh
qti-kromero: I think because each platform tests in it own GitHub action it should be fine - I don't think we would see this change

qcom/scripts/linux/run_tests.sh
```diff
@@ -32,24 +32,35 @@ function run_model_test() {
...
     set -e
 
-    "${python_exe}" "${REPO_ROOT}/qcom/scripts/all/model_test_log_to_junit_xml.py" \
-        "${build_dir}/${suite}_model_tests.log" > "${build_dir}/${suite}_model_tests.results.xml"
+    if [ -f "${model_log}" ]; then
```
qti-kromero, line 58: this is not produced on linux aarch64 because no model tests run? or what's the story here

qcom/scripts/linux/run_tests.sh
```diff
@@ -110,22 +118,33 @@ count_errors ./ctest --verbose --timeout 10800 --stop-on-failure "${exclude_args
...
+        rm -rf "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node/test_strnormalizer_export_monday_empty_output"
+    fi
+    
+    "${runner}" cpu node "${REPO_ROOT}/cmake/external/onnx/onnx/backend/test/data/node"
+
+    #TODO: [AISW-164203] - Known issues with QDQ model suite
+    if [ "$(uname -m)" != "aarch64" ]; then
         "${runner}" cpu float32
         "${runner}" htp qdq
+        log_debug "Scrubbing old context caches"
```
qti-kromero, line 143: nit: can we add a line between 142 and 143 for logical separation of other tests and context cache tests

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -1713,6 +1727,69 @@ bool ReduceOpHasAxesInput(const std::string& op_type, int opset_version);
...
+inline bool ShouldSkipGpuTestOnArm64(const ProviderOptions& qnn_options) {
+  std::string backend_name = "htp";
+  if (qnn_options.find("backend_type") != qnn_options.end()) {
```
qti-kromero: I think instead of adding these functions to qnn_test_utils and having to check this condition elsewhere it might be better to just have a few macros of functions defined: 

SKIP_IF_HTP_VERSION_LESS_THAN_V68 (one line skip for if HTP V68 or less, I think this one will be somewhat common and we can add macros for other versions if they env up being needed)

SKIP_IF_UNSUPPORTED_ON_ARM_64_LINUX (takes in qnn_options, contains the logic for when a general model test is unsupported on ARM 64 Linux) 

then the macros are just one line and the logic and action is taken inside of these functions

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -1713,6 +1727,69 @@ bool ReduceOpHasAxesInput(const std::string& op_type, int opset_version);
...
+template <typename QuantType>
+inline bool ShouldSkipQDQTestOnV68(const ProviderOptions& qnn_options) {
+  std::string backend_name = "htp";
+  if (qnn_options.find("backend_type") != qnn_options.end()) {
+    backend_name = qnn_options.at("backend_type");
+  }
+
+  if (backend_name == "htp") {
+    if (QnnHTPBackendTests::ShouldSkipIfHtpArchIsLessThanOrEqualTo(QNN_HTP_DEVICE_ARCH_V68)) {
+      if (std::is_same_v<QuantType, uint8_t> || std::is_same_v<QuantType, int8_t> ||
```
qti-kromero: the PR description only mentions skipping INT4/16/32 so is it also intentional to skip INT8?

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -628,7 +641,7 @@ void QnnIRBackendTests::SetUp() {
   }
 }
 
-#if defined(_WIN32)
+#if defined(_WIN32) || (defined(__linux__) && defined(__aarch64__))
 // TODO: Remove or set to SUPPORTED once HTP emulation is supported on win arm64.
```
qti-kromero: should update comment for linux arm64

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -1586,22 +1605,17 @@ class QnnHTPBackendTests : public ::testing::Test {
 
   // Returns true if the test should be skipped because HTP architecture is less than or equal to the provided arch.
   // Example: if (QnnHTPBackendTests::ShouldSkipIfHTPArchIsLessThanOrEqualTo(QNN_HTP_DEVICE_ARCH_V68)) { GTEST_SKIP() << "..."; }
-  static bool ShouldSkipIfHtpArchIsLessThanOrEqualTo(QnnHtpDevice_Arch_t arch) {
+  static bool ShouldSkipIfHtpArchIsLessThanOrEqualTo([[maybe_unused]] QnnHtpDevice_Arch_t arch) {
+#if defined(_WIN32) || (defined(__linux__) && defined(__aarch64__))
```
qti-kromero, line 1630: nit: indenting should be fixed here and we should comment where the #end is for this

I might be parsing this wrong but it looks like something was messed up here? or was ShouldSkipIfHtpFp16Unsupported moved and not replaced? unsure exactly

qcom/scripts/linux/run_tests.sh
qti-mbadnara: Noted will fix it with the latest commit

qcom/scripts/linux/run_tests.sh
qti-mbadnara: Noted, will fix with the latest commit

onnxruntime/test/providers/qnn/qnn_test_utils.h
qti-mbadnara: Yes, will update the description

onnxruntime/test/providers/qnn/qnn_test_utils.cc
qti-mbadnara: Noted, will fix it in the latest commit

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -1626,6 +1661,54 @@ class QnnCPUBackendTests : public ::testing::Test {
   static BackendSupport cached_cpu_support_;  // Set by the first test using this fixture.
 };
 
+// Template function implementing the test skip logic for SKIP_TEST_ON_LINUX_ARM64.
+// Placed after the class definitions so QnnHTPBackendTests is fully defined.
+// QuantType defaults to void (no type specified); when provided, skipping is gated on QNN_IS_SUPPORTED_QDQ_TYPE.
+// Returns true if the test should be skipped, and sets skip_reason with the reason.
+template <typename QuantType>
+inline bool CheckAndSkipTest(const ProviderOptions& qnn_options,
```
qti-kromero: this function is named really generally but is specific to Linux ARM?

onnxruntime/test/providers/qnn/qnn_node_group/channel_shuffle_fusion_test.cc
```diff
@@ -99,6 +99,7 @@ ProviderOptions GetProviderOptions() {
 }  // namespace
 
 TEST_F(QnnHTPBackendTests, ChannelShuffleFusion) {
+  SKIP_TEST_ON_LINUX_ARM64(ProviderOptions(), QNN_HTP_DEVICE_ARCH_V68, "FP16");
```
qti-kromero: my issue with this update now is that these are not skipped because it is LINUX_ARM64, the root cause of why this is skipped is the HTP ARCH being <= v68

onnxruntime/test/providers/qnn/maxpool_test.cc
```diff
@@ -102,6 +102,7 @@ TEST_F(QnnCPUBackendTests, MaxPool_Global) {
 }
 
 TEST_F(QnnCPUBackendTests, MaxPool_Rank3) {
+  QNN_SKIP_TEST_ON_AARCH64("Test not supported on Linux ARM64");
```
qti-kromero, line 105: why can't some of the logic for runtime detection of criteria to skip just be moved inside of this macro instead of adding a whole new one?

qti-kromero (approved):

PR #307, opened 2026-04-28
[QNN EP] Convert BQ to LPBQ encodings

qti-ashimaj:
### Description
Convert BQ encodings to htp supported LPBQ encodings

### Motivation and Context
The block quantized ONNX models are currently not supported in QNN-EP. In order to execute them via QNN HTP, we are converting the BQ encodings to LPBQ encodings.

### DIFF
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -197,10 +197,10 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
 |'0'|Disabled. QNN EP will handle quantization and dequantization of graph I/O.|
 |'1'|Default. Enabled. Offload quantization and dequantization of graph I/O to CPU EP.|
 
-|`"convert_bq_to_lpbq"`|Description|
+|`"enable_block_quant_weight_optimization"`|Description|
 |---|---|
-|'0'|Disabled. Block quantized model will run with block quantized weight encodings and float activations.|
-|'1'|Default. Enabled. Block quantized weight encodings will be converted to Low Power Block Quantized encodings.|
+|`"0"`|Default. Disabled. Block-quantized models use the standard compatibility path.|
+|`"1"`|Enabled. Uses an optimized path for block-quantized weights when supported. If the optimized path is not available, QNN EP falls back to the standard compatibility path.|
 
 |`"enable_htp_shared_memory_allocator"`|Description|
 |---|---|
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -197,6 +197,11 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
 |'0'|Disabled. QNN EP will handle quantization and dequantization of graph I/O.|
 |'1'|Default. Enabled. Offload quantization and dequantization of graph I/O to CPU EP.|
 
+|`"convert_bq_to_lpbq"`|Description|
+|---|---|
+|'0'|Disabled. Block quantized model will run with block quantized weight encodings and float activations.|
+|'1'|Default. Enabled. Block quantized weight encodings will be converted to Low Power Block Quantized encodings.|
+
 |`"enable_htp_shared_memory_allocator"`|Description|
 |---|---|
 |'0'|Default. Disabled.|
@@ -903,7 +908,7 @@ session = ort.InferenceSession("model.onnx", sess_options=sess_options)
 
 ### Important Considerations
 #### Feature Disabled if Number of Subgraphs is Less Than 5
-While graph composition is responsible for the majority of the preparation time, asynchronously finalizing the subgraphs cuts the total time down by a considerable amount, depending on the graph. For smaller models or models with only a few subgraphs, the overhead of setting up for parallel graph preparation will negate any possible performance gains and may actually result in worse performance. 
+While graph composition is responsible for the majority of the preparation time, asynchronously finalizing the subgraphs cuts the total time down by a considerable amount, depending on the graph. For smaller models or models with only a few subgraphs, the overhead of setting up for parallel graph preparation will negate any possible performance gains and may actually result in worse performance.
 
 #### Feature Disabled if `num_graph_prepare_threads` is 1
 This defeats the purpose of the feature, and enabling the feature will only add additional overhead from thread pool creation.
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 #include <functional>
@@ -87,10 +87,6 @@ bool IsBQWeight(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnitIODef
   }
   const int64_t num_blocks = scale_shape[k_axis];
   if (num_blocks <= 0 || static_cast<int64_t>(weight_shape[k_axis]) % num_blocks != 0) {
-    return false;
-  }
-  // Go for BQ FP16 only if LPBQ conversion is set to false
-  if (qnn_model_wrapper.GetModelSettings().convert_bq_to_lpbq) {
     return false;
   }
   return true;
@@ -250,16 +246,16 @@ Ort::Status MatMulOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper, c
                                            bool do_op_validation) const {
   const auto& inputs = node_unit.Inputs();
 
-  // Block-quantized weight: translate to a QNN MatMul with a BW_FLOAT_BLOCK weight.
-  if (IsBQWeight(qnn_model_wrapper, inputs[1])) {
-    return ProcessInputsForBQMatMul(qnn_model_wrapper, node_unit, logger, input_names, do_op_validation);
-  }
-
   TensorInfo input_info_0{};
   TensorInfo input_info_1{};
   bool use_fully_connected = false;
   RETURN_IF_ERROR(
       CheckInputs(qnn_model_wrapper, inputs[0], inputs[1], input_info_0, input_info_1, use_fully_connected));
+
+  // Block-quantized weight: translate to a QNN MatMul with a BW_FLOAT_BLOCK weight.
+  if (IsBQWeight(qnn_model_wrapper, inputs[1]) && !input_info_1.quant_param.IsLPBQ()) {
+    return ProcessInputsForBQMatMul(qnn_model_wrapper, node_unit, logger, input_names, do_op_validation);
+  }
 
   if (use_fully_connected) {
     return ProcessInputsForQnnFullyConnected(qnn_model_wrapper,
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 #include <functional>
@@ -130,6 +130,7 @@ Ort::Status CheckInputs(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeU
                                                  !input_info_0.is_initializer &&
                                                  IsQuant16bit(input_info_1.qnn_data_type) &&
                                                  !input_info_1.is_initializer);
+  use_fully_connected = use_fully_connected && !input_info_1.quant_param.IsLPBQ();
 #endif
   return Ort::Status();
 }
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -89,6 +89,10 @@ bool IsBQWeight(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnitIODef
   if (num_blocks <= 0 || static_cast<int64_t>(weight_shape[k_axis]) % num_blocks != 0) {
     return false;
   }
+  // Go for BQ FP16 only if LPBQ conversion is set to false
+  if (qnn_model_wrapper.GetModelSettings().convert_bq_to_lpbq) {
+    return false;
+  }
   return true;
 }
 
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -130,7 +130,6 @@ Ort::Status CheckInputs(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeU
                                                  !input_info_0.is_initializer &&
                                                  IsQuant16bit(input_info_1.qnn_data_type) &&
                                                  !input_info_1.is_initializer);
-  use_fully_connected = use_fully_connected && !input_info_1.quant_param.IsLPBQ();
 #endif
   return Ort::Status();
 }
--- a/onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
@@ -11,6 +11,7 @@
 #include "QnnTypes.h"
 
 #include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
 
 #define ALIGN_PTR_UP(ptr, align, type) \
   reinterpret_cast<type>((reinterpret_cast<std::uintptr_t>(ptr) + (align) - 1) & ~((align) - 1))
@@ -366,6 +367,8 @@ Ort::Status QnnQuantParamsWrapper::Init(const Qnn_QuantizeParams_t& params, cons
       params_.encodingDefinition = params.encodingDefinition;
       params_.quantizationEncoding = params.quantizationEncoding;
 
+      per_channel_scales_size_ = static_cast<uint32_t>(num_scaleoffsets);
+
       // Deep copy the blockwiseExpansion
       const size_t bwe_num_bytes = sizeof(Qnn_BlockwiseExpansion_t);
       constexpr std::uintptr_t bwe_align = alignof(Qnn_BlockwiseExpansion_t);
@@ -399,6 +402,7 @@ Ort::Status QnnQuantParamsWrapper::Init(const Qnn_QuantizeParams_t& params, cons
       params_.encodingDefinition = params.encodingDefinition;
       params_.quantizationEncoding = params.quantizationEncoding;
 
+      num_blocks_ = static_cast<uint32_t>(num_scaleoffsets);
       block_encoding_tensor_rank_ = static_cast<uint32_t>(tensor_rank);
       block_encoding_axis_data_ = std::make_unique<uint32_t[]>(block_encoding_tensor_rank_);
       std::memcpy(block_encoding_axis_data_.get(),
@@ -480,10 +484,13 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
   }
 
   const bool is_per_tensor = scales.size() == 1;
+  const bool is_block_quant = ort_quant_params->block_size.has_value() && ort_quant_params->block_size.value() > 0;
+  const bool is_per_channel = scales.size() > 1 && !is_block_quant;
 
   // QNN uses different structs to represent quantization parameters depending on
   // - per-tensor vs per-channel
   // - int4 vs not int4
+  // - block quantization (LPBQ / BLOCKWISE_EXPANSION)
   if (is_per_tensor && !is_int4_type) {
     params_.encodingDefinition = QNN_DEFINITION_DEFINED;
     params_.quantizationEncoding = QNN_QUANTIZATION_ENCODING_SCALE_OFFSET;
@@ -507,7 +514,7 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
     } else {
       params_.bwScaleOffsetEncoding.offset = 0;
     }
-  } else if (!is_per_tensor && is_int4_type) {
+  } else if (is_per_channel && is_int4_type) {
     std::vector<uint32_t> io_shape;
     RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(io_def.shape, io_shape), "Cannot get shape");
     const int32_t io_rank = static_cast<int32_t>(io_shape.size());
@@ -550,7 +557,7 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
 
     params_.bwAxisScaleOffsetEncoding.scales = scales_span.data();
     params_.bwAxisScaleOffsetEncoding.offsets = zps_span.data();
-  } else if (!is_per_tensor && !is_int4_type) {
+  } else if (is_per_channel && !is_int4_type) {
     std::vector<uint32_t> io_shape;
     RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(io_def.shape, io_shape), "Cannot get shape");
     const int32_t io_rank = static_cast<int32_t>(io_shape.size());
@@ -586,6 +593,71 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
     params_.axisScaleOffsetEncoding.axis = static_cast<int32_t>(axis);
     params_.axisScaleOffsetEncoding.numScaleOffsets = static_cast<uint32_t>(num_elems);
     params_.axisScaleOffsetEncoding.scaleOffset = data_span.data();
+  } else if (is_block_quant) {
+    // ONNX block quantization -> QNN LPBQ (BLOCKWISE_EXPANSION) conversion.
+
+    // Get scale tensor shape to determine block/channel dimensions.
+    const std::vector<int64_t> scale_shape =
+        utils::GetInitializerShape(ort_quant_params->scale, qnn_model_wrapper.GetOrtApi());
+    RETURN_IF_NOT(scale_shape.size() == 2,
+                  "Only 2D block quantization scale tensors are supported for LPBQ conversion");
+
+    // Determine block axis (= ONNX axis attribute, default 0).
+    constexpr int64_t kDefaultBlockAxis = 0;
+    int64_t onnx_axis = ort_quant_params->axis.value_or(kDefaultBlockAxis);
+    if (onnx_axis < 0) onnx_axis += static_cast<int64_t>(scale_shape.size());
+    RETURN_IF_NOT(onnx_axis == 0 || onnx_axis == 1,
+                  "Only axis 0 or 1 is supported for 2D block quantization LPBQ conversion");
+
+    // Scale shape: [num_blocks_per_channel, num_channels] when block_axis=0
+    //              [num_channels, num_blocks_per_channel] when block_axis=1
+    const uint32_t num_blocks_per_channel = static_cast<uint32_t>(scale_shape[onnx_axis]);
+    const uint32_t num_channels = static_cast<uint32_t>(scale_shape[1 - onnx_axis]);
+
+    // The conversion algorithm expects scales in block-major order [num_blocks, num_channels].
+    // If block_axis=1 the raw tensor is channel-major [num_channels, num_blocks]; transpose it.
+    std::vector<float> bq_scales_bm;
+    std::vector<int32_t> bq_offsets_bm;
+    if (onnx_axis == 0) {
+      bq_scales_bm = std::move(scales);
+      bq_offsets_bm = std::move(zero_points);
+    } else {
+      // Transpose [num_channels, num_blocks] -> [num_blocks, num_channels]
+      bq_scales_bm.resize(scales.size());
+      for (uint32_t c = 0; c < num_channels; ++c) {
+        for (uint32_t b = 0; b < num_blocks_per_channel; ++b) {
+          bq_scales_bm[static_cast<size_t>(b) * num_channels + c] =
+              scales[static_cast<size_t>(c) * num_blocks_per_channel + b];
+        }
+      }
+      if (!zero_points.empty()) {
+        bq_offsets_bm.resize(zero_points.size());
+        for (uint32_t c = 0; c < num_channels; ++c) {
+          for (uint32_t b = 0; b < num_blocks_per_channel; ++b) {
+            bq_offsets_bm[static_cast<size_t>(b) * num_channels + c] =
+                zero_points[static_cast<size_t>(c) * num_blocks_per_channel + b];
+          }
+        }
+      }
+    }
+
+    // Apply BQ -> LPBQ algorithm
+    // Use bitwidth=4 for int4 weights and bitwidth=8 for int8 weights.
+    std::vector<float> per_channel_scales;
+    std::vector<uint8_t> per_block_int_scales;
+    std::vector<int32_t> lpbq_offsets;
+    const uint32_t kBitwidth = is_int4_type ? 4u : 8u;
+    RETURN_IF_ERROR(utils::TryConvertBlockQuantScalesToLpbq(
+        bq_scales_bm, bq_offsets_bm,
+        num_blocks_per_channel, num_channels, kBitwidth,
+        per_channel_scales, per_block_int_scales, lpbq_offsets));
+
+    // QNN LPBQ axis = the non-block axis in the weight tensor.
+    // For ONNX axis=0 (block axis=0): QNN axis=1; for axis=1: QNN axis=0.
+    const int64_t qnn_axis = 1 - onnx_axis;
+
+    *this = QnnQuantParamsWrapper(per_channel_scales, per_block_int_scales, lpbq_offsets,
+                                  qnn_axis, ort_quant_params->block_size.value(), is_int4_type);
   } else {
     return MAKE_EP_FAIL("Unexpected tensor kind for QuantParamsWrapper::Init()");
   }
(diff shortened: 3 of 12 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -80,6 +80,7 @@ Ort::Status CheckInputs(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeU
                                                  !input_info_0.is_initializer &&
                                                  IsQuant16bit(input_info_1.qnn_data_type) &&
                                                  !input_info_1.is_initializer);
+  use_fully_connected = use_fully_connected && !input_info_1.quant_param.IsLPBQ();
```
qti-kromero: do we only check one input because the other input having LPBQ encodings is implicit? or should we ensure both inputs don't use LPBQ?

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
qti-kromero: I'm not sure if this is significant but Claude is flagging a byte-order mark that was added to this file and claims it can cause issues on GCC/Clang builds - I suppose we will see about that

regardless seeing as how the change is inconsequential I feel like we should try and remove it to reduce risk, if possible

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -534,6 +541,71 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
...
+
+    // Get scale tensor shape to determine block/channel dimensions.
+    const std::vector<int64_t> scale_shape =
+        utils::GetInitializerShape(ort_quant_params->scale, qnn_model_wrapper.GetOrtApi());
+    RETURN_IF_NOT(scale_shape.size() == 2,
+                  "Only 2D block quantization scale tensors are supported for LPBQ conversion");
+
+    // Determine block axis (= ONNX axis attribute, default 0).
+    constexpr int64_t kDefaultBlockAxis = 0;
+    int64_t onnx_axis = ort_quant_params->axis.value_or(kDefaultBlockAxis);
```
qti-kromero: nit: I'm not sure I like the name onnx_axis - block_axis is maybe more readable? but also not sure if that's the best. onnx_axis just sounds so general

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -534,6 +541,71 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
     params_.axisScaleOffsetEncoding.axis = static_cast<int32_t>(axis);
     params_.axisScaleOffsetEncoding.numScaleOffsets = static_cast<uint32_t>(num_elems);
     params_.axisScaleOffsetEncoding.scaleOffset = data_span.data();
+  } else if (is_block_quant) {
```
qti-kromero, line 597: I feel like this logic and some of the other logic in qnn_utils warrants a new test (or a few new tests)? is that planned to be added later or something we can add in this PR?

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.h
```diff
@@ -81,35 +81,38 @@ class QnnQuantParamsWrapper {
...
       RETURN_IF_NOT(static_cast<size_t>(params_.bwAxisScaleOffsetEncoding.axis) < perm.size(),
                     "Axis value is out of range of the provided permutation");
-      const int32_t new_axis = static_cast<int32_t>(perm[params_.bwAxisScaleOffsetEncoding.axis]);
-      params_.bwAxisScaleOffsetEncoding.axis = new_axis;
+      params_.bwAxisScaleOffsetEncoding.axis = static_cast<int32_t>(perm[params_.bwAxisScaleOffsetEncoding.axis]);
+    } else if (params_.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION &&
+               params_.blockwiseExpansion != nullptr) {
+      RETURN_IF_NOT(static_cast<size_t>(params_.blockwiseExpansion->axis) < perm.size(),
+                    "LPBQ axis value is out of range of the provided permutation");
+      params_.blockwiseExpansion->axis = static_cast<int32_t>(perm[params_.blockwiseExpansion->axis]);
```
qti-kromero, line 129: why is blockwiseExpansion stored as a pointer here whereas other quantization parameters are stored as objects?

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1167,6 +1199,81 @@ Ort::Status DequantizePerChannel(gsl::span<const uint8_t> quant_bytes, gsl::span
...
+                                             /*out*/ std::vector<float>& per_channel_scales,
+                                             /*out*/ std::vector<uint8_t>& per_block_int_scales,
+                                             /*out*/ std::vector<int32_t>& offsets) {
+  RETURN_IF_NOT(bq_scales.size() == static_cast<size_t>(num_blocks_per_channel) * num_channels,
+                "BQ scales size does not match num_blocks_per_channel * num_channels");
+  RETURN_IF_NOT(bq_offsets.empty() || bq_offsets.size() == bq_scales.size(),
+                "BQ offsets size must be empty or equal to bq_scales size");
+  RETURN_IF_NOT(bitwidth > 0 && bitwidth <= 16, "bitwidth must be in range [1, 16]");
+
+  const uint32_t max_int_scale = 1u << bitwidth;  // 2^bitwidth
```
qti-kromero: max_int_scale should be this (1u << bitwidth) - 1 right? i.e. uint8 range is [0, 255], not [1, 256]

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1167,6 +1199,81 @@ Ort::Status DequantizePerChannel(gsl::span<const uint8_t> quant_bytes, gsl::span
...
+  for (uint32_t c = 0; c < num_channels; ++c) {
+    const float pc_scale = per_channel_scales[c];
+    for (uint32_t b = 0; b < num_blocks_per_channel; ++b) {
+      const float raw_scale = bq_scales[static_cast<size_t>(b) * num_channels + c];
+      uint8_t int_scale;
+      if (pc_scale <= 0.0f) {
+        int_scale = 1;
+      } else {
+        const float tentative = std::round(raw_scale / pc_scale);
+        const uint32_t clamped = std::max(1u, std::min(static_cast<uint32_t>(tentative), max_int_scale));
```
qti-kromero, line 1273: I think here is where clamped could be set to 256 if max_int_scale is chosen when it is equal to 256 causing an overflow to 0 on the line below when it is assigned to a uint8

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -374,6 +377,35 @@ std::ostream& operator<<(std::ostream& out, const Qnn_QuantizeParams_t& quantize
         out << quantize_params.bwAxisScaleOffsetEncoding.offsets[i] << (i == num_elems - 1 ? "" : " ");
       }
       out << (truncate ? "...)" : ")");
+    } else if (quantize_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION &&
+               quantize_params.blockwiseExpansion != nullptr) {
+      const Qnn_BlockwiseExpansion_t& lpbq = *quantize_params.blockwiseExpansion;
+      out << " axis=" << lpbq.axis
+          << " numBlocksPerAxis=" << lpbq.numBlocksPerAxis
+          << " blockScaleBitwidth=" << lpbq.blockScaleBitwidth;
+      // For lpbq, num_elems are not present in the quantize_params,
```
qti-kromero: nit: LPBQ

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -534,6 +541,71 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
...
+    // ONNX block quantization -> QNN LPBQ (BLOCKWISE_EXPANSION) conversion.
+
+    // Get scale tensor shape to determine block/channel dimensions.
+    const std::vector<int64_t> scale_shape =
+        utils::GetInitializerShape(ort_quant_params->scale, qnn_model_wrapper.GetOrtApi());
+    RETURN_IF_NOT(scale_shape.size() == 2,
+                  "Only 2D block quantization scale tensors are supported for LPBQ conversion");
+
+    // Determine block axis (= ONNX axis attribute, default 0).
+    constexpr int64_t kDefaultBlockAxis = 0;
```
qti-kromero: this is defined as 0 by default for every op type?

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.h
quic-calvnguy: This is how it is defined in the QNN API in QnnTypes.h (params_ is a Qnn_QuantizeParams_t object)

quic-calvnguy: From QAIRT 2.45.40:

Image

qti-chuteng:
## Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/qnn_utils.cc:1216,1267-1268`
`max_int_scale = 1u << bitwidth` is wrong by one. For int8 the value is 256, then `static_cast<uint8_t>(clamped)` wraps the channel's max-scale block to 0, encoding it as zero-scale; for int4 the value 16 overflows the 4-bit block-scale field. The standard LPBQ convention is `int_scale ∈ [1, 2^bitwidth − 1]`. Use `max_int_scale = (1u << bitwidth) - 1` and divide by it in `per_channel_scales[c] = max_scale / max_int_scale`. This will hit 100% of inputs since the algorithm intentionally maps max_scale to max_int_scale. Pair this fix with M-4.

**[C-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:1`
A UTF-8 BOM (`EF BB BF`) was accidentally inserted at the start of the file. No other `.cc` file in the repo has a BOM, and some toolchains (notably nvcc and older MSVC) treat the BOM as the first source character and break preprocessing. Re-save without BOM, e.g. `sed -i '1s/^\xEF\xBB\xBF//' matmul_op_builder.cc`.

---

## Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:553-555`
`constexpr int64_t kDefaultBlockAxis = 0;` violates the ONNX `QuantizeLinear` / `DequantizeLinear` spec, which has defaulted `axis` to 1 since opset 13. The two per-channel branches in this same function correctly use `DEFAULT_QDQ_AXIS = 1`. Change to `1` (or reuse `DEFAULT_QDQ_AXIS`); otherwise models that omit the `axis` attribute will be processed with the wrong layout, and `qnn_axis = 1 − onnx_axis` will compute the wrong direction.

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:434-442`
The new flag chain has a silent fallback: when `scales.size() == 1 && block_size > 0`, the `is_per_tensor` branch wins and `block_size` is silently ignored. Either set `is_per_tensor = scales.size() == 1 && !is_block_quant`, or add `RETURN_IF_NOT(!(is_per_tensor && is_block_quant), "Per-tensor + block_size>0 is not supported")` before the if-else.

**[M-3]** `onnxruntime/test/providers/qnn/ (no new files)`
~125 lines of new numeric-conversion logic and zero unit-test coverage. Per checklist [G02] every new logic change must have a corresponding test. Add at minimum: (1) a QnnHTPBackendTests end-to-end with opset-21 `MatMul + DequantizeLinear(block_size)` for both int4 and int8 weights asserting `ExpectedEPNodeAssignment::All`; (2) a standalone gtest for `TryConvertBlockQuantScalesToLpbq` covering max-block-scale round-trip (which would catch C-1), axis=0 vs axis=1 transpose, all-zero scales, asymmetric zero-points must fail; (3) a kill-test fence — disable the BQ→LPBQ change and confirm the new tests fail.

**[M-4]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:153`
`lpbq.blockScaleBitwidth = is_int4 ? 4 : 0;` writes `blockScaleBitwidth = 0` for int8. Before this PR the false branch was dead code (only the int4 LPBQ fusions used the ctor); this PR exercises it. Fix: `lpbq.blockScaleBitwidth = is_int4 ? 4 : 8;`. Same root cause as C-1, please fix together.

**[M-5]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:548-563`
After the `scale_shape.size() == 2` check, please also assert `scale_shape[0] > 0 && scale_shape[1] > 0` and `scale_shape[0] * scale_shape[1] == scales.size()` so dynamic-shape scales fail loudly with a clear message instead of overflowing inside `TryConvertBlockQuantScalesToLpbq`.

**[M-6]** `onnxruntime/core/providers/qnn/builder/qnn_utils.h:289 / qnn_utils.cc:1202`
`TryConvertBlockQuantScalesToLpbq` uses the `Try`-prefix idiom (which in ORT — e.g. `TryFusion` — implies bool / non-fatal failure), but it actually returns `Ort::Status` and the caller uses `RETURN_IF_ERROR`. Drop the prefix → `ConvertBlockQuantScalesToLpbq`.

---

## Minor

**[N-1]** `qnn_quant_params_wrapper.cc:581-589` — `bq_offsets` transpose path is dead code: `TryConvertBlockQuantScalesToLpbq` immediately requires every offset to be 0. Validate `zero_points` are all zero up front and drop the transpose.

**[N-2]** `qnn_utils.h:280-284 / qnn_utils.cc:1232-1239` — Doc comment "max_int_scale = 2^bitwidth (256 for int8, 16 for int4)" embeds the same off-by-one as C-1. Update together with C-1.

**[N-3]** `qnn_utils.cc:382-393` — `operator<<` for BLOCKWISE_EXPANSION uses `numBlocksPerAxis` as the print length for `scaleOffsets`, but `QnnTypes.h` documents that array size as `axisSize`. Either pass axisSize through (e.g. via `per_channel_scales_size_`) or update the comment to clarify "first numBlocksPerAxis entries only".

**[N-4]** `qnn_quant_params_wrapper.cc:434-610` — The new `is_block_quant` flag reorders the if-else chain. Add 1-2 lines of comment above mapping (per-tensor / per-channel / block-quant × int4) to each branch and noting the final else as a defensive fallback.

**[N-5]** `qnn_quant_params_wrapper.cc:597` — `const uint32_t kBitwidth = is_int4_type ? 4u : 8u;` is a runtime const, not a constexpr. The `k` prefix is reserved for constexpr / static const in Google C++ style. Rename to `bitwidth`.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
qti-ashimaj: The change has been reverted, matmul op conversion for lpbq is being handled in another PR

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
qti-ashimaj: Removed the BOM token

onnxruntime/core/providers/qnn/builder/qnn_utils.cc

onnxruntime/core/providers/qnn/builder/qnn_utils.cc

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: Updated it to QDQ_DEFAULT_AXIS which is 1 since opset 13

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1167,6 +1199,81 @@ Ort::Status DequantizePerChannel(gsl::span<const uint8_t> quant_bytes, gsl::span
...
+                                          /*out*/ std::vector<float>& per_channel_scales,
+                                          /*out*/ std::vector<uint8_t>& per_block_int_scales,
+                                          /*out*/ std::vector<int32_t>& offsets) {
+  RETURN_IF_NOT(bq_scales.size() == static_cast<size_t>(num_blocks_per_channel) * num_channels,
+                "BQ scales size does not match num_blocks_per_channel * num_channels");
+  RETURN_IF_NOT(bq_offsets.empty() || bq_offsets.size() == bq_scales.size(),
+                "BQ offsets size must be empty or equal to bq_scales size");
+  RETURN_IF_NOT(bitwidth > 0 && bitwidth <= 16, "bitwidth must be in range [1, 16]");
+
+  const uint32_t max_int_scale = (1u << bitwidth) - 1u;
```
quic-ristha: max_int_scale calculation logic differes from what we used in the standard LPBQ workflow

For 4->8 expansion

the int scale will range from [1, 16]

Thus the max_int_sacle is calculated as 2^(decompressed_bw - bq_bw) i.e. 16 in this case


You can refer this class code : https://github.com/qualcomm/aimet/blob/66b2834fcb2711ef3a0bf6cec847090b42da683b/TrainingExtensions/torch/src/python/aimet_torch/quantization/affine/quantizer.py#L1192

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -427,11 +431,15 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
...
+  const bool is_per_channel = scales.size() > 1 && !is_block_quant;
+  const bool is_per_tensor = scales.size() == 1 && !is_block_quant;
 
-  // QNN uses different structs to represent quantization parameters depending on
-  // - per-tensor vs per-channel
-  // - int4 vs not int4
+  // QNN uses different structs to represent quantization parameters depending on:
+  // - per-tensor (scales.size()==1, no block_size): SCALE_OFFSET or BW_SCALE_OFFSET
+  // - per-channel (scales.size()>1, no block_size): AXIS_SCALE_OFFSET or BW_AXIS_SCALE_OFFSET
+  // - block quantization (block_size>0): BLOCKWISE_EXPANSION (LPBQ)
```
quic-pkaul: Specifically, We have both block quantization (QNN_QUANTIZATION_ENCODING_BLOCK) and lpbq (QNN_QUANTIZATION_BLOCKWISE_EXPANSION) supported in QNN.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -498,7 +506,7 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
 
     params_.bwAxisScaleOffsetEncoding.scales = scales_span.data();
     params_.bwAxisScaleOffsetEncoding.offsets = zps_span.data();
-  } else if (!is_per_tensor && !is_int4_type) {
+  } else if (is_per_channel && !is_int4_type) {
```
quic-pkaul, line 561: Not sure if this gets taken elsewhere.
 But before we make BQ->LPBQ transition, besides the two constraints (is_per_channel and is_int4_type) we also need to make sure that the input and output activation of the op is Integer as well. Otherwise better keep in BQ.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -534,6 +542,75 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
...
+    RETURN_IF_NOT(scale_shape.size() >= 2,
+                  "Block quantization scale tensors must have at least rank 2 for LPBQ conversion");
+    RETURN_IF_NOT(scale_shape[0] > 0 && scale_shape[1] > 0,
+                  "Block quantization scale tensor dimensions must be positive");
+    RETURN_IF_NOT(scale_shape[0] * scale_shape[1] == static_cast<int64_t>(scales.size()),
+                  "Block quantization scale tensor shape product must equal number of scales");
+
+    // Determine block axis (= ONNX axis attribute).
+    constexpr int64_t DEFAULT_QDQ_AXIS = 1;
+    int64_t axis = ort_quant_params->axis.value_or(DEFAULT_QDQ_AXIS);
```
quic-pkaul, line 636: The code doesn't seem to take care of transposed weights. Which will swap axis.
Please see transB here.
https://onnx.ai/onnx/operators/onnx__Gemm.html

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1167,6 +1199,81 @@ Ort::Status DequantizePerChannel(gsl::span<const uint8_t> quant_bytes, gsl::span
...
+  }
+
+  // Validate that all scales are non-negative and finite.
+  for (size_t i = 0; i < bq_scales.size(); ++i) {
+    RETURN_IF_NOT(std::isfinite(bq_scales[i]) && bq_scales[i] >= 0.0f,
+                  "BQ scales must be non-negative and finite");
+  }
+
+  // Algorithm:
+  //   max_int_scale             = 2^bitwidth - 1
```
quic-pkaul: Also correct based on rishabh's comment

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -374,6 +377,35 @@ std::ostream& operator<<(std::ostream& out, const Qnn_QuantizeParams_t& quantize
...
       out << (truncate ? "...)" : ")");
+    } else if (quantize_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION &&
+               quantize_params.blockwiseExpansion != nullptr) {
+      const Qnn_BlockwiseExpansion_t& lpbq = *quantize_params.blockwiseExpansion;
+      out << " axis=" << lpbq.axis
+          << " numBlocksPerAxis=" << lpbq.numBlocksPerAxis
+          << " blockScaleBitwidth=" << lpbq.blockScaleBitwidth;
+      // For LPBQ, num_elems are not present in the quantize_params,
+      // we are using numBlocksPerAxis instead to print the first numBlocksPerAxis scale offset values
+      size_t num_elems = lpbq.numBlocksPerAxis;
```
qti-kromero, line 392: is lpbq.scaleOffsets channel-major or block-major? 

am wondering what this does in plain language? prints the first numBlocksPerAxis scaleOffsets so is that just the blocks for the first row of axis?

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -534,6 +542,75 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
...
+    const int32_t io_rank = static_cast<int32_t>(io_shape.size());
+
+    // Get scale tensor shape to determine block/channel dimensions.
+    // Scale tensor may be rank 2 (e.g., MatMul/Gemm) or higher rank (e.g., Conv with rank-4 weights).
+    // Only the first two dimensions (indexed by onnx_axis and 1 - onnx_axis) are used for LPBQ conversion.
+    const std::vector<int64_t> scale_shape =
+        utils::GetInitializerShape(ort_quant_params->scale, qnn_model_wrapper.GetOrtApi());
+    RETURN_IF_NOT(scale_shape.size() >= 2,
+                  "Block quantization scale tensors must have at least rank 2 for LPBQ conversion");
+    RETURN_IF_NOT(scale_shape[0] > 0 && scale_shape[1] > 0,
```
qti-kromero, line 629: does this and below work properly for ranks > 2?

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
qti-kromero: sorry thought I had commented this but I guess it got lost - we need to upgrade the datatype on per_block_int_scales or otherwise adjust for the differing range. otherwise, we will have overflow for uint8 storing 256

cc @quic-calvnguy

qti-ashimaj: Done. The int_scale range will be [1, 16] for 4-bit as suggested. Since the original algorithm is applicable for 4-bits only, we are adding a check to convert bq to lpbq encodings for only 4 bit and so the overflow for 8 bits will not happen with this.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: Done

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: Added a check for `(is_block_quant && is_int4_type)` , the conversion will only happen for these data types else not.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: This can be handled in the separate op builders using the HandleTranspose/HandleUnsqueeze API which will update the axis.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
qti-ashimaj: Updated

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
qti-ashimaj: The scale offsets are of size num_channels only but since this qnn struct doesn't have numScaleOffsets member, we are taking numBlocksPerAxis as a proxy to print the first few scale and offset values.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: Yes, we have tested this on models having MatMul BQ (rank 2) and Conv BQ (rank 4) weights.

qti-kromero:
LGTM aside from missing tests

qti-ashimaj:
https://github.com/onnxruntime/onnxruntime-qnn/pull/307#issuecomment-4663752827

> LGTM aside from missing tests

The test cases are being added in the individual op builder PRs

qti-chuteng:
### 🔴 Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:L597`
The `else if (is_block_quant)` branch only builds LPBQ params when `is_int4_type && convert_bq_to_lpbq`; the inner `if` has no `else`, so a non-int4 (e.g. int8) block-quant weight falls through to `return Ort::Status()` with `params_` left as `QNN_QUANTIZE_PARAMS_INIT` (UNDEFINED). This is reachable under the default `convert_bq_to_lpbq=true`, where an int8 BQ MatMul weight is no longer claimed by `IsBQWeight()` and routes through this generic path — yielding wrong numerics or a QNN compile failure instead of a clean error. Add an `else` that returns `MAKE_EP_FAIL(...)`.

---

### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc:L915`
The new option `convert_bq_to_lpbq` defaults to `true` (both in `ModelSettings` and `ParseBoolOption`). This violates the opt-in convention for new session config keys and silently flips existing int4 BQ models (BW_FLOAT_BLOCK, landed in PR #288) to a lossy LPBQ approximation after upgrade, changing their output. Default to `false`, or document the behavior change explicitly.

**[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:L91`
Only MatMul's `IsBQWeight()` honors `convert_bq_to_lpbq`. The BQ detectors in `conv_op_builder.cc` and `gemm_op_builder.cc` (`IsBQGemmWeight`) ignore the option and always use BW_FLOAT_BLOCK, so under `convert_bq_to_lpbq=1` MatMul uses LPBQ while Conv/Gemm still use BQ — inconsistent and contradicting the doc. Gate Conv/Gemm the same way, or scope the doc to MatMul-only.

**[M-3]** `onnxruntime/test/providers/qnn/matmul_test.cc:L365`
Both new provider-option helpers set `opts["convert_bq_to_lpbq"] = "0"`, which disables the entire new LPBQ conversion path. No test runs it with `"1"` (or the default), so the new algorithm could be fully broken and all tests would still pass. Add an HTP QDQ test (and ideally a direct unit test for `ConvertBlockQuantScalesToLpbq`) that exercises the conversion.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.h → per_channel_scales_size_`
`uint32_t per_channel_scales_size_;` has no in-class initializer (unlike adjacent `block_encoding_tensor_rank_ = 0` / `num_blocks_ = 0`) and is read in the copy ctor for LPBQ. In the degenerate `num_elems==0` LPBQ case it stays indeterminate.
```suggestion
  uint32_t per_channel_scales_size_ = 0;
```

**[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:L655`
`const uint32_t bitwidth = is_int4_type ? 4u : 8u;` sits inside `if (is_int4_type && ...)`, so the `8u` arm is dead code; `ConvertBlockQuantScalesToLpbq` also hard-rejects non-4 bitwidth. Use `const uint32_t bitwidth = 4u;` with a comment to avoid implying int8 support.

**[N-3]** `docs/execution_providers/QNN-ExecutionProvider.md:L200`
The doc says block-quant encodings "will be converted to LPBQ", but conversion is int4-only and (per M-2) MatMul-only. Update to match the actual scope and final default value.

qti-ashimaj:
> ### 🔴 Critical
> **[C-1]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:L597` The `else if (is_block_quant)` branch only builds LPBQ params when `is_int4_type && convert_bq_to_lpbq`; the inner `if` has no `else`, so a non-int4 (e.g. int8) block-quant weight falls through to `return Ort::Status()` with `params_` left as `QNN_QUANTIZE_PARAMS_INIT` (UNDEFINED). This is reachable under the default `convert_bq_to_lpbq=true`, where an int8 BQ MatMul weight is no longer claimed by `IsBQWeight()` and routes through this generic path — yielding wrong numerics or a QNN compile failure instead of a clean error. Add an `else` that returns `MAKE_EP_FAIL(...)`.
> 
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc:L915` The new option `convert_bq_to_lpbq` defaults to `true` (both in `ModelSettings` and `ParseBoolOption`). This violates the opt-in convention for new session config keys and silently flips existing int4 BQ models (BW_FLOAT_BLOCK, landed in PR #288) to a lossy LPBQ approximation after upgrade, changing their output. Default to `false`, or document the behavior change explicitly.
> 
> **[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:L91` Only MatMul's `IsBQWeight()` honors `convert_bq_to_lpbq`. The BQ detectors in `conv_op_builder.cc` and `gemm_op_builder.cc` (`IsBQGemmWeight`) ignore the option and always use BW_FLOAT_BLOCK, so under `convert_bq_to_lpbq=1` MatMul uses LPBQ while Conv/Gemm still use BQ — inconsistent and contradicting the doc. Gate Conv/Gemm the same way, or scope the doc to MatMul-only.
> 
> **[M-3]** `onnxruntime/test/providers/qnn/matmul_test.cc:L365` Both new provider-option helpers set `opts["convert_bq_to_lpbq"] = "0"`, which disables the entire new LPBQ conversion path. No test runs it with `"1"` (or the default), so the new algorithm could be fully broken and all tests would still pass. Add an HTP QDQ test (and ideally a direct unit test for `ConvertBlockQuantScalesToLpbq`) that exercises the conversion.
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.h → per_channel_scales_size_` `uint32_t per_channel_scales_size_;` has no in-class initializer (unlike adjacent `block_encoding_tensor_rank_ = 0` / `num_blocks_ = 0`) and is read in the copy ctor for LPBQ. In the degenerate `num_elems==0` LPBQ case it stays indeterminate.
> 
> ```
>   uint32_t per_channel_scales_size_ = 0;
> ```
> 
> **[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc:L655` `const uint32_t bitwidth = is_int4_type ? 4u : 8u;` sits inside `if (is_int4_type && ...)`, so the `8u` arm is dead code; `ConvertBlockQuantScalesToLpbq` also hard-rejects non-4 bitwidth. Use `const uint32_t bitwidth = 4u;` with a comment to avoid implying int8 support.
> 
> **[N-3]** `docs/execution_providers/QNN-ExecutionProvider.md:L200` The doc says block-quant encodings "will be converted to LPBQ", but conversion is int4-only and (per M-2) MatMul-only. Update to match the actual scope and final default value.

[C-1] If the conditions for BQ->LPBQ conversion are not met, ORT::Status is returned and the respective op builder files will construct the `BW_FLOAT_BLOCK` encodings.

[M-1] Updated the default value to `false`

[M-2] The respective op lpbq PRs will take care of these.

[M-3] The respective op lpbq PRs have the tests added.

[N-1] Done

[N-2] Done

[N-3] Updated the description

qti-kromero (approved):

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -197,6 +197,11 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
 |'0'|Disabled. QNN EP will handle quantization and dequantization of graph I/O.|
 |'1'|Default. Enabled. Offload quantization and dequantization of graph I/O to CPU EP.|
 
+|`"enable_block_quant_weight_optimization"`|Description|
+|---|---|
+|`"0"`|Default. Disabled. Block-quantized models use the standard compatibility path.|
+|`"1"`|Enabled. Uses an optimized path for block-quantized weights when supported. If the optimized path is not available, QNN EP falls back to the standard compatibility path.|
```
huaychou, line 203:
```suggestion
|`0`|Default. Disabled. Block-quantized models use the standard compatibility path.|
|`1`|Enabled. Uses an optimized path for block-quantized weights when supported. If the optimized path is not available, QNN EP falls back to the standard compatibility path.|
```

huaychou (commented):
Could we have unit tests for Conv LPBQ and the BQ->LPBQ conversion?

PR #308, opened 2026-04-28
[QNN EP] Ensure Windows builds wheel using setup.py

qti-mbadnara:
This PR aims to ensure both Linux and Windows wheel builds have a unified point of entry

**Current Behavior**

- Windows: `build.ps1` → `build.bat` → `cmake + setup.py`
- Linux: build.sh → `build.py` → `setup.py` (via `build_python_wheel()` ))

Windows builds CMake prior to wheel building, added arguments which enables wheel-only builds without cmake configure/build steps once again.


qcom/scripts/windows/build.ps1
```diff
@@ -300,26 +300,22 @@ else {
                 }
 
                 if ($BuildWheel) {
-                    $PyNightlyArg = ""
-                    $WheelNameSuffix = ""
+                    $WheelOnlyArgs = @(
+                        "--build_wheel_only",
```
huaychou: Can we use the `--build_wheel` option in ci_build/build.py instead of adding a new argument? Introducing a Windows-only argument could increase the differences between Windows and Linux.

qcom/scripts/windows/build.ps1
```diff
@@ -300,26 +300,22 @@ else {
...
+                        "--build_dir", $BuildDir,
+                        "--config", $Config,
+                        "--cmake_generator", $CMakeGenerator,
+                        "--qnn_home", "$QairtSdkRoot"
+                    )
                     if ($env:ORT_NIGHTLY_BUILD -eq "1") {
-                        $PyNightlyArg = "--nightly_build"
-                        $WheelNameSuffix = "--wheel_name_suffix=qcom_internal"
+                        $WheelOnlyArgs += "--wheel_name_suffix=qcom_internal"
```
huaychou: We should prevent using environment variable to pass argument. Currently, we pass `--nightly_build` to `setup.py` directly in Windows pipeline, but use `NIGHTLY_BUILD` environment variable to set `nightly_build` for `build_python_wheel`. This is the gap we need to eliminate.

huaychou:
@qti-mbadnara, I think you can follow https://github.com/onnxruntime/onnxruntime-qnn/pull/322 to pass the wheel arguments to `build.bat` and remove the `setup.py` part.

kuanyul-qti (approved):
LGTM.
Remember to trigger nightly and RC workflows.

huaychou (approved):
LGTM. I think we need to handle the NIGHTLY_BUILD flag in another PR.

qti-ashwshan (approved):

tirupath-qti (approved):

PR #309, opened 2026-04-28
Fix quantized Gelu <-> Erf fusion

qti-shubham:
### Description
- Add `Div`/`Mul` support as parent to `Erf` as some optimizer replaces Div with Mul.
- Canonicalize duplicated intermediate root tensor names done by ORT.
- Validate against outer quantized I/O when the pattern is wrapped by DQ/Q.
**- Reject pattern matching when there is intermediate QDQGroups**
- Extend tests to cover both float and QDQ `Div`/`Mul` GELU variants.

### Context
- Quantized GELU fusion could fail when `Div(sqrt(2))` before `Erf` is optimized into `Mul(1 / sqrt(2))`.
- ORT generates duplicate tensor for shared root parent QDQ nodes
- SingleNodeUnit for intermediate GELU QDQ nodes instead of QDQGroup causing QNN 3110 errors due to:
-- Mixed Precision,
-- Mismatch scales, etc.
- Intermediate QDQGroups in GELU pattern may have accuracy variation and should not be fused.

### Out of Scope
- Mixed precision support.
- General handling of all mixed or partially grouped quantized GELU topologies.
- actual model: https://github.com/qcom-ai-hub/tetracode/issues/18976, scales are different and Mixed precision Div and Mul nodes
image

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
@@ -414,8 +414,13 @@ std::unique_ptr<IQnnNodeGroup> GeluFusion::TryFusion(
   }
 
   // Build-time GELU should remain float-in/float-out so surrounding DQ/Q nodes can be lowered separately.
-  const OrtNodeUnitIODef* root_input = &div_inputs[0];
-  const OrtNodeUnitIODef* validate_root_input = root_input;
+  if (div_inputs.empty()) {
+    return nullptr;
+  }
+
+  // Copy IODef values to avoid dangling references from temporary vectors returned by Inputs()/Outputs().
+  OrtNodeUnitIODef root_input = div_inputs[0];
+  OrtNodeUnitIODef validate_root_input = root_input;
 
   const OrtNodeUnit* root_input_parent = GetParentOfInput(qnn_model_wrapper,
                                                           *div_node_unit,
@@ -423,32 +428,44 @@ std::unique_ptr<IQnnNodeGroup> GeluFusion::TryFusion(
                                                           node_to_node_unit,
                                                           node_unit_to_qnn_node_group);
   if (root_input_parent != nullptr &&
-      root_input_parent->OpType() == DEQUANTIZE_LINEAR &&
-      !root_input_parent->Inputs().empty()) {
-    validate_root_input = &root_input_parent->Inputs()[0];
+      root_input_parent->OpType() == DEQUANTIZE_LINEAR) {
+    const auto& parent_inputs = root_input_parent->Inputs();
+    if (parent_inputs.empty()) {
+      return nullptr;
+    }
+    validate_root_input = parent_inputs[0];
   }
 
-  if (pattern_match.final_mul_node_unit == nullptr || pattern_match.final_mul_node_unit->Outputs().empty()) {
+  if (pattern_match.final_mul_node_unit == nullptr) {
     return nullptr;
   }
-  const OrtNodeUnitIODef* final_output = &pattern_match.final_mul_node_unit->Outputs()[0];
-  const OrtNodeUnitIODef* validate_final_output = final_output;
+
+  const auto& final_mul_outputs = pattern_match.final_mul_node_unit->Outputs();
+  if (final_mul_outputs.empty()) {
+    return nullptr;
+  }
+
+  OrtNodeUnitIODef final_output = final_mul_outputs[0];
+  OrtNodeUnitIODef validate_final_output = final_output;
 
   const OrtNodeUnit* final_output_child = GetOnlyChildOfOutput(qnn_model_wrapper,
                                                                *pattern_match.final_mul_node_unit,
-                                                               pattern_match.final_mul_node_unit->Outputs()[0],
+                                                               final_mul_outputs[0],
                                                                node_to_node_unit,
                                                                node_unit_to_qnn_node_group);
   if (final_output_child != nullptr &&
-      final_output_child->OpType() == QUANTIZE_LINEAR &&
-      !final_output_child->Outputs().empty()) {
-    validate_final_output = &final_output_child->Outputs()[0];
+      final_output_child->OpType() == QUANTIZE_LINEAR) {
+    const auto& child_outputs = final_output_child->Outputs();
+    if (child_outputs.empty()) {
+      return nullptr;
+    }
+    validate_final_output = child_outputs[0];
   }
 
   Ort::Status status = ValidateOnQnn(qnn_model_wrapper,
                                      pattern_match.node_units,
-                                     *validate_root_input,
-                                     *validate_final_output);
+                                     validate_root_input,
+                                     validate_final_output);
   if (!status.IsOK()) {
     ORT_CXX_LOG(logger,
                 ORT_LOGGING_LEVEL_WARNING,
@@ -460,10 +477,10 @@ std::unique_ptr<IQnnNodeGroup> GeluFusion::TryFusion(
 
   return std::make_unique<GeluFusion>(std::move(pattern_match.node_units),
                                       &erf_node_unit,
-                                      *validate_root_input,
-                                      *validate_final_output,
-                                      *root_input,
-                                      *final_output);
+                                      validate_root_input,
+                                      validate_final_output,
+                                      root_input,
+                                      final_output);
 }
 
 GeluFusion::GeluFusion(std::vector<const OrtNodeUnit*>&& node_units,
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
@@ -319,13 +319,9 @@ std::unique_ptr<IQnnNodeGroup> GeluFusion::TryFusion(
     return nullptr;
   }
 
+  // Reject fusion if QDQ nodes are found inside the GELU topology to possibly avoid accuracy variation.
   if (erf_node_unit.UnitType() == OrtNodeUnit::Type::QDQGroup) {
-    ORT_CXX_LOG(logger,
-                ORT_LOGGING_LEVEL_WARNING,
-                ("[GeluFusion] QDQ nodes were found inside the GELU topology (" + erf_node_unit.Name() +
-                 "). The fused GELU may cause accuracy variation. Re-generate the quantized model without internal "
-                 "GELU QDQ if possible.")
-                    .c_str());
+    return nullptr;
   }
 
   const auto& erf_inputs = erf_node_unit.Inputs();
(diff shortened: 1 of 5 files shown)


onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
```diff
@@ -132,17 +132,12 @@ bool TryMatchErfAddPattern2(
     return false;
   }
 
-  const auto& mul_outputs = mul_after_add_node_unit->Outputs();
-  if (mul_outputs.empty()) {
```
quic-ykota, line 136: Curious why was this check removed?

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
```diff
@@ -293,40 +262,39 @@ std::unique_ptr<IQnnNodeGroup> GeluFusion::TryFusion(
...
-      return nullptr;
+  // Try ErfAdd patterns if ErfMul didn't match (Erf -> Add -> Mul [-> Mul])
+  if (!is_match) {
+    const OrtNodeUnit* erf_child_add = GetChildNodeUnitAllowQdq(qnn_model_wrapper, erf_node_unit, "Add",
+                                                                node_to_node_unit, node_unit_to_qnn_node_group);
+    if (erf_child_add != nullptr) {
+      // ErfAdd Patterns 1 or 2: Erf -> Add -> Mul [-> Mul]
+      // Structure: x * 0.5 * (Erf(x / sqrt2) + 1) [with variations in Mul ordering]
+      const auto& add_inputs = erf_child_add->Inputs();
+      if (add_inputs.size() >= 2) {
```
quic-ykota: When is the size of add node's inputs > 2?

qti-shubham: thanks for pointing! its redundant check. 
removed.

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
qti-shubham: I am now using `GetChildNodeUnitAllowQdq` which already do this check internally.
previously it was `GetOnlyChildOfOutput` which requires this check.

yath1 (approved):
LGTM :)

qti-chuteng:
**[M-1]** `onnxruntime/test/providers/qnn/qnn_node_group/gelu_fusion_test.cc:707-768`

The QDQ U8 tests (`GeluFusionPattern{1,2,3}_QDQ_U8`) do not exercise the PR's main goal — they only run `BuildGeluPattern{1,2,3}TestCase(input_def)` without `use_mul=true`, and the `BuildQDQGeluPattern{1,2,3}TestCase` helpers (lines 232–469) hard-code `Div_sqrt2` with no Mul branch at all. As a result, the new `GetChildNodeUnitAllowQdq` QDQ-traversal logic and the new "Erf parent = Mul" branch have no end-to-end regression coverage under quantization (kill-test fence: revert the fix, the QDQ tests still pass).

**Suggested fix**: add a `use_mul` parameter to all three `BuildQDQGeluPattern{N}TestCase` builders mirroring the float versions, then add a Mul variant to each `_QDQ_U8` test:
```cpp
TEST_F(QnnHTPBackendTests, GeluFusionPattern1_QDQ_U8) {
  auto input_def = TestInputDef<float>({1, 2, 3, 4}, false, -10.0f, 10.0f);
  // Test with Div
  TestQDQModelAccuracy(BuildGeluPattern1TestCase(input_def, false),
                       BuildQDQGeluPattern1TestCase<uint8_t>(input_def, false), ...);
  // Test with Mul (Div replaced by Mul)
  TestQDQModelAccuracy(BuildGeluPattern1TestCase(input_def, true),
                       BuildQDQGeluPattern1TestCase<uint8_t>(input_def, true), ...);
}
```

---

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc:267-297`

The new dispatch ("try `GetChildNodeUnitAllowQdq(..., "Mul", ...)`, then if !is_match try `..."Add"...`") is semantically equivalent to the old "fetch single child once, switch on OpType", but loses the single early "Erf must have one child" decision point. A future fourth GELU pattern is easy to mis-order.

**Suggested fix**: refactor to fetch the single child once (e.g. via a no-OpType-filter overload) then `if (child->OpType() == "Mul") {...} else if (...== "Add") {...}`. At minimum add a comment explaining "Erf has at most one child; sequential calls are equivalent to a switch on OpType".

---

**[N-1]** PR scope

The PR mixes four concerns: (1) QDQ support, (2) Mul-as-Erf-parent support, (3) deletion of `_LargeInput` tests, (4) widening QDQ input range from [-1,1] to [-10,10]. Per checklist [G03], please split or at least document each change in the PR description / commit messages.

---

**[N-2]** `onnxruntime/test/providers/qnn/qnn_node_group/gelu_fusion_test.cc:484-577`

`GeluFusionPattern{1,2,3}_Float32` runs `RunQnnModelTest` twice into the same `json_qnn_graph_dir`, asserting after each. `AssertOpInQnnGraph` iterates `directory_iterator` and takes the first non-`_tensor_log` JSON — after the second run the directory contains two distinct JSON files (graph names differ), and the second assertion may inspect the stale Div-run JSON.

**Suggested fix**: reset the dump dir between sub-runs, or split into two TEST_Fs with separate dump dirs.

---

**[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc:154-160` (and `81-85`, `122-126`)

The ASCII diagrams in `TryMatchErfMulPattern`, `TryMatchErfAddPattern1`, and `TryMatchErfAddPattern2` still say `Div (B=1.4142...)` but the PR now also accepts `Mul (1/sqrt2)`. Mirror the test-side comments: `Div/Mul (1.4142... or 1/1.4142...)`.

yath1:
Based on offline discussions, we decided not to fuse to a single gelu op if the decomposed ops have qdq nodes inbetween right? So, Is it possible to update logic to not have references to qdq? For example: GetChildNodeUnitAllowQdq need not be used to check for children nodes. QDQ would still be supported if its outside the decomposed gelu op.

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
```diff
@@ -36,6 +36,44 @@ static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qnn_model_wrapper,
 
 namespace {
 
+// Quantized GELU subgraphs can route the same logical root tensor through generated
+// duplicated-value names such as `.../duplicated` or `.../duplicated_token_*`.
+std::string_view CanonicalizeRootTensorName(std::string_view tensor_name) {
+  constexpr std::string_view kDuplicatedSuffix = "/duplicated";
+  constexpr std::string_view kGeneratedTokenSuffix = "_token_";
```
yath1: Is there a possibility that this logic fails if tensor name is contains these strings?

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
```diff
@@ -45,30 +83,41 @@ struct GeluPatternMatchContext {
...
-  return GetParentOfInput(ctx.qnn_model_wrapper,
-                          consumer_node_unit,
-                          inputs[input_index],
-                          ctx.node_to_node_unit,
-                          ctx.node_unit_to_qnn_node_group);
+  ORT_CXX_LOG(logger,
+              ORT_LOGGING_LEVEL_WARNING,
+              ("[GeluFusion] Mixed GELU topology detected. Culprit node unit " +
+               GetNodeUnitDescription(*node_unit) +
+               ". GELU fusion expects pattern nodes to be all SingeNodeUnits or QDQGroups"
```
qti-chuteng: single

qti-shubham: fixed.

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
```diff
@@ -45,30 +83,41 @@ struct GeluPatternMatchContext {
...
   const std::unordered_map<const OrtNode*, const OrtNodeUnit*>& node_to_node_unit;
   const std::unordered_map<const OrtNodeUnit*, const IQnnNodeGroup*>& node_unit_to_qnn_node_group;
-  const std::string& root_input_name;
+  const Ort::Logger& logger;
+  std::string_view root_input_name;
 };
 
-/* Checks if the given NodeUnit has an input tensor with the specified name. */
-bool HasInputWithName(const OrtNodeUnit& node_unit, std::string_view input_name) {
+bool HasInputWithEquivalentName(const OrtNodeUnit& node_unit, std::string_view input_name) {
```
yath1, line 94: Nit: Add a short desc explaining the method

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
```diff
@@ -36,6 +36,44 @@ static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qnn_model_wrapper,
 
 namespace {
 
+// Quantized GELU subgraphs can route the same logical root tensor through generated
+// duplicated-value names such as `.../duplicated` or `.../duplicated_token_*`.
+std::string_view CanonicalizeRootTensorName(std::string_view tensor_name) {
```
yath1, line 41: Previous logic did not rely on names to determine the pattern if I recall correctly. Names are generally a bit error prone. Why are we changing to using names instead of op types?

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.h
```diff
@@ -49,7 +49,12 @@ class QnnModelWrapper;
 /// </summary>
 class GeluFusion : public IQnnNodeGroup {
  public:
-  GeluFusion(std::vector<const OrtNodeUnit*>&& node_units, const OrtNodeUnit* target_node_unit);
+  GeluFusion(std::vector<const OrtNodeUnit*>&& node_units,
+             const OrtNodeUnit* target_node_unit,
+             OrtNodeUnitIODef validate_root_input,
+             OrtNodeUnitIODef validate_final_output,
+             OrtNodeUnitIODef root_input,
+             OrtNodeUnitIODef final_output);
```
yath1: Explain these inputs. I guess validate_root_input is the `DQ` input to the Gelu node, validate_final_output is the output of `Q` after the Gelu node. Please rename to better reflect what they are. Same comment for root_input and final_output.

onnxruntime/core/providers/qnn/builder/qnn_node_group/gelu_fusion.cc
qti-shubham: ONNX Tensor names are unique and robust.

op types doesn't work well with QDQ graph, as ROOT node will be DQ node.

qti-chuteng (approved):

PR #311, opened 2026-04-29
Fixes issue ScatterND fallback to CPU due to negative concrete indices

qti-yuduo:
- ONNX `ScatterND` allows negative and INT_64 indices; QNN rejects both, silently dropping the node to CPU and breaking whole-graph capture.
- New `normalize_indices_utils` rewrites static initializer indices to non-negative INT_32 at partition time, and inserts a runtime `Cast(INT_32)` for dynamic INT_64 inputs.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
@@ -1,0 +1,186 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "core/providers/qnn/builder/opbuilder/normalize_indices_utils.h"
+
+#include <cstdint>
+#include <cstddef>
+#include <functional>
+#include <limits>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <gsl/gsl>
+
+#include "core/providers/qnn/builder/qnn_def.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+
+namespace onnxruntime {
+namespace qnn {
+namespace utils {
+
+template <typename SrcType>
+bool NormalizeIndicesBytes(const std::vector<uint8_t>& onnx_bytes,
+                           const std::function<int64_t(size_t)>& axis_dim_for_element,
+                           std::vector<uint8_t>& qnn_bytes,
+                           bool& has_negative_indices) {
+  if (onnx_bytes.size() % sizeof(SrcType) != 0) {
+    return false;
+  }
+
+  const size_t num_elems = onnx_bytes.size() / sizeof(SrcType);
+  gsl::span<const SrcType> onnx_indices{reinterpret_cast<const SrcType*>(onnx_bytes.data()), num_elems};
+
+  qnn_bytes.resize(num_elems * sizeof(int32_t));
+  gsl::span<int32_t> qnn_indices{reinterpret_cast<int32_t*>(qnn_bytes.data()), num_elems};
+
+  for (size_t i = 0; i < num_elems; i++) {
+    const int64_t axis_dim = axis_dim_for_element(i);
+    // int64 arithmetic prevents wraparound when int32 idx + axis_dim >= 2^31.
+    int64_t idx = static_cast<int64_t>(onnx_indices[i]);
+
+    if (idx < 0) {
+      has_negative_indices = true;
+      idx += axis_dim;
+    }
+
+    if (idx < 0 || idx >= axis_dim ||
+        idx > static_cast<int64_t>(std::numeric_limits<int32_t>::max())) {
+      return false;
+    }
+
+    qnn_indices[i] = static_cast<int32_t>(idx);
+  }
+
+  return true;
+}
+
+template bool NormalizeIndicesBytes<int32_t>(
+    const std::vector<uint8_t>&, const std::function<int64_t(size_t)>&,
+    std::vector<uint8_t>&, bool&);
+template bool NormalizeIndicesBytes<int64_t>(
+    const std::vector<uint8_t>&, const std::function<int64_t(size_t)>&,
+    std::vector<uint8_t>&, bool&);
+
+namespace {
+
+// Registers the indices tensor and, for dynamic int64 inputs, inserts a
+// runtime Cast(INT_32).
+Ort::Status AddNormalizedIndicesTensor(QnnModelWrapper& qnn_model_wrapper,
+                                       TensorInfo indices_info,
+                                       const std::string& indices_tensor_name,
+                                       std::vector<uint8_t>&& qnn_indices_bytes,
+                                       const Ort::Logger& logger,
+                                       std::vector<std::string>& input_names,
+                                       bool do_op_validation) {
+  std::vector<uint32_t> cast_output_shape(indices_info.shape);
+
+  if (!qnn_model_wrapper.IsQnnTensorWrapperExist(indices_tensor_name)) {
+    const Qnn_TensorType_t tensor_type = indices_info.is_initializer
+                                             ? QNN_TENSOR_TYPE_STATIC
+                                             : qnn_model_wrapper.GetTensorType(indices_tensor_name);
+    QnnTensorWrapper input_tensorwrapper(indices_tensor_name,
+                                         tensor_type,
+                                         indices_info.qnn_data_type, QnnQuantParamsWrapper(),
+                                         std::move(indices_info.shape),
+                                         std::move(qnn_indices_bytes));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(input_tensorwrapper)),
+                  "Failed to add tensor.");
+  } else {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
+                ("Tensor already added, skip it: " + indices_tensor_name).c_str());
+  }
+
+  auto& input_tensorwrapper = qnn_model_wrapper.GetQnnTensorWrapper(indices_tensor_name);
+  std::string indices_casted_name = indices_tensor_name;
+  if (input_tensorwrapper.GetTensorDataType() == QNN_DATATYPE_INT_64) {
+    // Initializers are converted to INT_32 before registration, so INT_64 here
+    // implies a dynamic input.
+    RETURN_IF_NOT(!indices_info.is_initializer,
+                  "Internal error: static indices tensor registered with INT_64 dtype.");
+    indices_casted_name += "_int32";
+    RETURN_IF_ERROR(qnn_model_wrapper.AddCastNode(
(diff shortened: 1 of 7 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc
```diff
@@ -360,6 +367,36 @@ Ort::Status ProcessReductionAttribute(QnnModelWrapper& qnn_model_wrapper,
   return Ort::Status();
 }
 
+Ort::Status SimpleOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
```
minfhong-qti: How about a separate `scatternd_op_builder.cc`? And probably move `normalize_indices_utils` in it.

qti-yuduo: Moved.

qti-yuduo: util would like we keep separated as other ops like GatherND, ScatterElements they also share same issue just no one has yet hit. Will reactor them as a follow up change.

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
```diff
@@ -0,0 +1,188 @@
...
+                                                            onnx_indices_bytes));
+
+    if (indices_info.qnn_data_type == QNN_DATATYPE_INT_64) {
+      RETURN_IF_NOT((NormalizeIndicesBytes<int64_t>(onnx_indices_bytes, axis_dim_for_element,
+                                                    qnn_indices_bytes, has_negative_indices)),
+                    kOutOfRangeMsg);
+      indices_info.qnn_data_type = QNN_DATATYPE_INT_32;
+      rewrote_bytes = true;
+    } else if (indices_info.qnn_data_type == QNN_DATATYPE_INT_32) {
+      RETURN_IF_NOT((NormalizeIndicesBytes<int32_t>(onnx_indices_bytes, axis_dim_for_element,
```
yath1: Is int32_t unnecessary here? https://onnx.ai/onnx/operators/onnx__ScatterND.html says indices is always int64. Maybe we should just assert if the type is not equal to int64?

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
```diff
@@ -0,0 +1,188 @@
...
+                                         std::vector<std::string>& input_names,
+                                         bool do_op_validation) {
+  std::string indices_tensor_name = indices_input.name;
+
+  TensorInfo indices_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(indices_input, indices_info));
+
+  RETURN_IF_NOT(!indices_info.shape.empty(),
+                "ScatterND-style indices tensor must have rank >= 1.");
+  const uint32_t k = indices_info.shape.back();
```
yath1: Nit: maybe a better variable name instead of k?

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
```diff
@@ -0,0 +1,188 @@
...
+  qnn_bytes.resize(num_elems * sizeof(int32_t));
+  const auto qnn_indices = gsl::span<int32_t>{
+      reinterpret_cast<int32_t*>(qnn_bytes.data()), num_elems};
+
+  for (size_t i = 0; i < num_elems; ++i) {
+    const int64_t axis_dim = axis_dim_for_element(i);
+    // int64 prevents wraparound on int32 idx + axis_dim >= 2^31.
+    int64_t idx = static_cast<int64_t>(onnx_indices[i]);
+
+    if (idx < 0) {
```
yath1, line 40: Should this be a while statement instead? Suppose shape has 3, and idx = -7, we should bring it to positive by adding 3 three times

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
```diff
@@ -0,0 +1,188 @@
...
+      has_negative_indices = true;
+      idx += axis_dim;
+    }
+
+    if (idx < 0 || idx >= axis_dim ||
+        idx > static_cast<int64_t>(std::numeric_limits<int32_t>::max())) {
+      return false;
+    }
```
yath1, line 48: This can be avoided with the above logic maybe?

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
```diff
@@ -0,0 +1,188 @@
...
+#include "core/providers/qnn/builder/qnn_def.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+
+namespace onnxruntime {
+namespace qnn {
+namespace utils {
+
+template <typename SrcType>
+bool NormalizeIndicesBytes(gsl::span<const uint8_t> onnx_bytes,
```
yath1, line 25: Nice util :)

onnxruntime/test/providers/qnn/scatternd_test.cc
```diff
@@ -0,0 +1,198 @@
...
+                    {test::MakeAttribute("to",
+                                         static_cast<int64_t>(ONNX_NAMESPACE::TensorProto_DataType_FLOAT))});
+    builder.MakeOutput("Y");
+  };
+
+  RunQnnModelTest(build_model, MakeHtpProviderOptions(), 17,
+                  ExpectedEPNodeAssignment::All);
+}
+
+// Regression for an AI Hub compile job: ScatterND(-1) between producer/consumer
```
yath1: Should we remove this line: "Regression for an AI Hub compile job"

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
qti-yuduo: Good catch, removed some other redundant checks as well.

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
qti-yuduo: ORT core does not gate on value -- so this is still safer I feel?

yath1: Interesting, I thought the max value of indices would be the num_elements in that particular dimension - 1. If it can be any arbitrary positive value then better to error out here.

qti-yuduo: Not saying it's valid.. but user can make mistakes on runtime tensor values.. this is safer we just defer to CPU kernel where it might eventually fail but hopefully clear error than HTP.

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
qti-yuduo: Not entirely following... could you explain a bit more?

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
yath1: Cool, makes sense

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
yath1: Lets say the input array is [2, 3, 4, 5] (size= 4). If idx = -1, we choose index = 3 which is (-1 + axis_dim=4), the output will be [2, 3, 4, new-val]. But if the value (say, -7) is lesser than the -1 * axis_dim, ie -4 here, in this implementation it would error out. Do you see the value in treating -7 idx as -7 + 4 + 4 = 1 and change the output to this [2, new-val, 4, 5] instead?

qti-yuduo: Got it.... In that case, it's usually violating semantics. I think it consistent with ONNX spec also other similar languages: 

- **Python lists**: `[2,3,4,5][-5]` → `IndexError: list index out of range`
- **NumPy**: `np.array([2,3,4,5])[-5]` → `IndexError: index -5 is out of bounds for axis 0 with size 4`
- **PyTorch indexing**: same `IndexError`

Wrapping would diverge QNN EP output from the CPU EP reference for malformed-but-compiling models — a silent-correctness bug.

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
```diff
@@ -0,0 +1,153 @@
...
+    const int64_t axis_dim = axis_dim_for_element(i);
+    // int64 prevents wraparound on int32 idx + axis_dim >= 2^31.
+    int64_t idx = static_cast<int64_t>(onnx_indices[i]);
+    if (idx < 0) {
+      idx += axis_dim;
+    }
+    // ORT core validates shapes/dtypes; value-range is the actual safeguard
+    // because ONNX allows out-of-range ints to ship in user models.
+    if (idx < 0 || idx >= axis_dim ||
+        idx > static_cast<int64_t>(std::numeric_limits<int32_t>::max())) {
```
yath1, line 46: Minor comment, you can address later on: returning false from here would error out eventually but the actual error message is missed. Maybe a log would help narrow down the issue.

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.h
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+// ONNX Gather/Scatter allow negative and INT_64 indices; QNN rejects both.
+// A single negative static index otherwise silently drops the node to CPU.
+// Dynamic INT_64 indices get a runtime Cast; negative runtime values are
```
yath1, line 6: What does this statement mean? Are you talking about `data` attr. If so how do their values matter? 

> negative runtime values are NOT corrected.

yath1 (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.h
qti-yuduo: This is for `indices`, basically means compile-time we actually fix the negative indices (change to positive values), but we'd give up and not "correct" the indices in dynamic case. Does it make sense?

onnxruntime/core/providers/qnn/builder/opbuilder/normalize_indices_utils.cc
qti-yuduo: Good point, will do a follow up.

PR #313, opened 2026-04-29
Add diff coverage (patch coverage) report to CI

yuhuchua-qti:
## Summary                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                             
  Introduces `diff-cover` to the CI pipeline to report **patch coverage** — the percentage of PR-changed lines that are covered by tests — alongside the existing full coverage report.                                                                                      
                                                                                                                                                                                                                                                                             
  ## Changes                                                                                                                                                                                                                                                                 
                            
  - **`qcom/requirements.txt`** — Add `diff-cover` and `lcov_cobertura` pip packages                                                                                                                                                                                         
  - **`generate_coverage.sh`** — Convert filtered lcov `.info` to Cobertura XML via `lcov_cobertura`; generate `README.md` artifact; UT failures are now blocking
  - **`generate_diff_coverage.sh`** — New script: runs `diff-cover` to produce `diff-coverage.html` and `diff-coverage.txt`                                                                                                                                                  
  - **`ep_build/tasks/build.py`** — Add `GenerateDiffCoverageTask` (`git fetch` + `git diff` + `generate_diff_coverage.sh`)                                                                                                                                                  
  - **`build_and_test.py`** — Add `diff_coverage_linux_x86_64` public task (depends on `coverage_linux_x86_64`)                                                                                                                                                              
  - **`qualcomm-internal-coverage.yml`** — Reusable workflow (`workflow_call` + `workflow_dispatch`); path gate skips build if no QNN sources changed; post diff coverage summary to Job Summary and PR comment                                                              
  - **`qualcomm-internal-ci.yml`** — Call coverage workflow as part of PR Workflow check suite (`if: pull_request`); add `pull-requests: write` permission required by the called workflow                                                                                   
                                                                                                                                                                                                                                                                             
  ## Architecture                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                             
  Coverage workflow is a **reusable workflow** called from `qualcomm-internal-ci.yml` via `workflow_call`. This ensures `coverage-linux-x86_64` appears in the same **Qualcomm Internal PR Workflow** check suite rather than a separate one.                                
                            
  A path gate inside `qualcomm-internal-coverage.yml` skips the build automatically when no `onnxruntime/core/providers/qnn/` sources are touched.                                                                                                                           
                            
  ## PR comment                                                                                                                                                                                                                                                              
                            
  On each run, the diff coverage summary is posted (or updated) as a PR comment:                                                                                                                                                                                             
                            
  ---Diff Coverage                                                                                                                                                                                                                                                           
  Diff: origin/main...HEAD, staged and unstaged changes
                                                                                                                                                                                                                                                                             
  <changed files with per-file coverage %>             
  ---Total:   N lines
  Missing: N lines
  Coverage: N%


  ## How to run locally                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                             
`
python3 qcom/build_and_test.py diff_coverage_linux_x86_64
`


.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -17,20 +19,91 @@ jobs:
...
           # 3.10 matches DEFAULT_PYTHON_LINUX in qcom/ep_build/util.py.
           python3 qcom/build_and_test.py \
             --target-py-version 3.10 \
-            coverage_linux_x86_64
+            diff_coverage_linux_x86_64
 
       - name: Upload Coverage Report
         uses: actions/upload-artifact@v7
-        if: always()
+        if: always() && steps.gate.outputs.run == 'true'
```
huaychou:
```suggestion
        if: steps.gate.outputs.run == 'true'
```

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -17,20 +19,91 @@ jobs:
...
+          # submodules omitted here; initialized below only when coverage is needed.
 
-      - name: Run Coverage (Linux x86_64, always RelWithDebInfo)
+      - name: Check if coverage is needed
+        id: gate
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "run=true" >> "${GITHUB_OUTPUT}"
+          elif git diff --name-only origin/${{ github.base_ref }}...HEAD \
+              | grep -q 'onnxruntime/core/providers/qnn/'; then
```
huaychou: If we only modify UT, steps.gate.outputs.run will be false?

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -6,8 +6,10 @@ name: Qualcomm Internal Coverage
 permissions:
   contents: read
   checks: write
+  pull-requests: write
 
 on:
+  workflow_call:
   workflow_dispatch:
```
huaychou, line 13: I think we need to check-admin-permission for workflow_dispatch.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: I've fixed it. Thanks.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Agree, I've updated the logic here to also monitor the UT modification.
Thanks.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Other workflows with workflow_dispatch (e.g. qualcomm-internal-ci.yml, qualcomm-internal-release.yml) don't have the check either, so I finally decide not to add the admin check for consistency.
Thanks.

qti-chuteng (approved):

.github/workflows/qualcomm-internal-ci.yml
```diff
@@ -31,3 +32,8 @@ jobs:
     secrets: inherit
     with:
       do_all: true
+
+  coverage:
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/qualcomm-internal-coverage.yml
+    secrets: inherit
```
qti-hungjuiw, line 39: Currently, there are **three** jobs under `.github/workflows/qualcomm-internal-build-and-test-single-os.yml`: `build`, `test` and `test-wheel`.

Could we move this `coverage` job to there to become the fourth job?
We can use `if:` to ensure the `coverage` job is only executed on `linux-x86_64` and `uses` to import the `qualcomm-internal-coverage.yml‎`

Then, we can leverage the `env.BUILD_CONFIG_DIR` there

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -17,20 +19,91 @@ jobs:
...
         with:
           name: "coverage-linux-x86_64-${{ github.sha }}"
-          retention-days: 7
-          path: build/linux-x86_64/RelWithDebInfo/coverage/
+          retention-days: 3
+          path: |
+            build/linux-x86_64/RelWithDebInfo/coverage/coverage_lcov_filtered.info
+            build/linux-x86_64/RelWithDebInfo/coverage/README.md
+            build/linux-x86_64/RelWithDebInfo/coverage/diff-coverage.html
+            build/linux-x86_64/RelWithDebInfo/coverage/diff-coverage.txt
```
qti-hungjuiw: If we move the job to `qualcomm-internal-build-and-test-single-os.yml`, we can use something like `${{ github.workspace }}/build/${{inputs.target_os}}-${{inputs.target_arch}}/${{ env.BUILD_CONFIG_DIR }}`  to avoid the hardcode path

.github/workflows/qualcomm-internal-ci.yml
yuhuchua-qti: Thanks for the suggestion, but I think keeping coverage as a separate workflow works better here:

1. BUILD_CONFIG_DIR doesn't apply — it's hardcoded to Release, but coverage builds with RelWithDebInfo + gcov instrumentation. We'd need a coverage-only path variable anyway, so no DRY benefit.
2. Different compilation, not just a different job. Coverage produces an instrumented build that can't share artifacts with the Release build used for distribution.
3. single-os.yml is matrix-style, invoked across Linux/Windows/Android × multiple archs. Coverage only applies to Linux x86_64, so we'd need a job-level if: filter that produces skipped jobs in every other invocation's PR checks.
4. Different lifecycle — coverage has its own file-path gate, artifact format (lcov/cobertura/diff-cover), and PR-comment posting, none of which map onto build/test/test-wheel.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Good point on avoiding hardcoded paths. Reasons for not moving the job are in my reply above.

To address the hardcoded-path concern within the coverage workflow itself, I introduced a job-level COVERAGE_DIR env var. All 6 previous hardcoded references (4 artifact paths, 1 bash txt=, 1 github-script txtPath) now resolve through it — ${{ env.COVERAGE_DIR }} in YAML, ${COVERAGE_DIR} in bash, process.env.COVERAGE_DIR in the Node.js step. Future build-dir/config changes only need a one-line update.

qcom/scripts/linux/generate_coverage.sh
```diff
@@ -202,7 +202,86 @@ genhtml "${build_dir}/${config}/coverage_lcov_filtered.info" \
...
+git diff origin/main...HEAD > patch.diff
+
+# Run diff-cover (output goes to same directory as coverage.xml by default)
+bash qcom/scripts/linux/generate_diff_coverage.sh \
+    --coverage-xml=coverage.xml \
+    --diff-file=patch.diff
+
+# View results
+cat diff-coverage.txt
+```
```
kuanyul-qti: Is it possible to keep the `README.md` file in the code base instead of generating it each time the `generate_coverage.sh` script is run?

qcom/scripts/linux/generate_coverage.sh
kuanyul-qti, line 20: Add `set -euo pipefail`

qcom/scripts/linux/generate_diff_coverage.sh
```diff
@@ -0,0 +1,131 @@
...
+#     Activate the project venv before calling this script:
+#       source venv/bin/activate
+#
+# Usage:
+#   bash generate_diff_coverage.sh \
+#       --coverage-xml=/path/to/coverage/coverage.xml \
+#       --diff-file=/path/to/patch.diff \
+#       [--output-dir=/path/to/report] \
+#       [--fail-under=N]
+
```
kuanyul-qti, line 21: Add `set -euo pipefail`

qcom/scripts/linux/generate_coverage.sh
yuhuchua-qti: Thanks for the suggestion. We're already enabling strict mode via set_strict_mode on line 26, which qcom/scripts/linux/common.sh defines as set -euo pipefail.
This is the convention used by other scripts in qcom/scripts/linux/ (build.sh, run_tests.sh), so I'd prefer to keep it consistent here.

qcom/scripts/linux/generate_diff_coverage.sh
yuhuchua-qti: Thanks for the suggestion. We're already enabling strict mode via set_strict_mode on line 26, which qcom/scripts/linux/common.sh defines as set -euo pipefail.
This is the convention used by other scripts in qcom/scripts/linux/ (build.sh, run_tests.sh), so I'd prefer to keep it consistent here.

qcom/scripts/linux/generate_coverage.sh
yuhuchua-qti: Agreed. The heredoc was using <<'EOF' (no variable expansion), so the content was fully static — no reason to generate it at runtime. Moved it to qcom/scripts/linux/coverage_artifact_README.md and the script now just cps it. Verified locally with diff_coverage_linux_x86_64: produced README is identical.

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -6,31 +6,106 @@ name: Qualcomm Internal Coverage
...
+          fetch-depth: 0
+          # submodules omitted here; initialized below only when coverage is needed.
 
-      - name: Run Coverage (Linux x86_64, always RelWithDebInfo)
+      - name: Check if coverage is needed
+        id: gate
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "run=true" >> "${GITHUB_OUTPUT}"
+          elif git diff --name-only origin/${{ github.base_ref }}...HEAD \
```
kuanyul-qti: When this workflow is triggered via `workflow_call` (from `qualcomm-internal-ci.yml`), `github.event_name` is `"workflow_call"`, not `"pull_request"`. The gate's first branch checks for `"workflow_dispatch"` — it will miss — and falls into the `elif` that uses `github.base_ref`. In a `workflow_call` context, `github.base_ref` is documented as inherited from the caller but has been reported empty in some configurations. If it is empty, `git diff --name-only origin/...HEAD` fails silently and the gate emits `run=false`, skipping the entire coverage job without any error.

```suggestion
          elif [ -n "${{ github.base_ref }}" ] && git diff --name-only origin/${{ github.base_ref }}...HEAD \
```

yuhuchua-qti: Thanks for the careful read. One clarification on the trigger: when invoked via workflow_call, the reusable workflow inherits the caller's github context — github.event_name is the caller's event, not "workflow_call". Since the caller qualcomm-internal-ci.yml gates the coverage job with if: github.event_name == 'pull_request', this workflow only ever sees event_name == 'pull_request' (with non-empty base_ref) or event_name == 'workflow_dispatch' (manual run from the Actions tab).

That said, your underlying concern about silent failure is fair — relying on the caller's if gate is a non-local invariant that future changes could break. Hardened the gate to fail explicitly: set -euo pipefail, validate base_ref is non-empty, treat git diff errors as hard failures (with ::error:: annotations), and use early returns to flatten the branches. PTAL.

kuanyul-qti (approved):

PR #319, opened 2026-04-30
[ModelCompatibility] Fix checking HNRD on x86 platform.

minfhong-qti:
### Description

For checking whether HNRD path is adopted for model compatibility, add new logic at the beginning to check whether HNRD library is at the expected path. If not, then no need to take HNRD into consideration.
Test: Move one of existing UTs for x86 platform.

### Motivation and Context

Fail to generate model compatibility info on x86 platform as prepare/stub/skel libs are not packaged and thus incorrectly considered adopting HNRD path, which is not the case.

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
@@ -3149,8 +3149,8 @@ TEST_F(QnnHTPBackendTests, CompileApi_InitializerHandler_ReuseExternalInitialize
 
   ASSERT_EQ(num_reused_ext_initializers, 2);  // Reused external conv weight and bias.
 }
+#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
 
-#ifdef _WIN32
 // Utility class to help create enviornment using HNRD for testing.
 // Expected usage is used along with smart pointer to automatically restore temporarily moved libraries.
 class HnrdTestHandle {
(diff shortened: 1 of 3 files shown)


onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -234,7 +234,7 @@ void RegisterQnnEpLibrary(RegisteredEpDeviceUniquePtr& registered_ep_device,
                                                                   "libQnnHtp.so"
 #endif
        )) {
-#if defined(__linux__)
+#if defined(__linux__) || (defined(_WIN32) && defined(_M_X64))
```
huaychou, line 237: Just wondering, do we have testcases using HTP emulator?

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -3606,9 +3607,7 @@ TEST_F(QnnHTPBackendTests, ModelCompatibility_ApiValidate_CbNewHtpArch) {
 
   TestModelCompatibilityApiValidate(test_info, OrtCompiledModelCompatibility_EP_UNSUPPORTED);
 }
-#endif  // _WIN32
-
-#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
+#endif  // defined(_WIN32) && defined(_M_ARM64)
```
huaychou, line 3610: Nice change 👍

huaychou (approved):
LGTM :shipit: Just wondering, do we have Linux or Android HNRD use cases now?

onnxruntime/test/providers/qnn/qnn_test_utils.cc
minfhong-qti: If `QnnHTPBackendTests` executing on Linux x86 is using HTP emulator, then I think so?

minfhong-qti:
> LGTM :shipit: Just wondering, do we have Linux or Android HNRD use cases now?

No, I believe not.

qti-chuteng (approved):

onnxruntime/test/providers/qnn/qnn_test_utils.cc
huaychou: If using HTP emulator, should the `target_hw_device_type` be CPU or NPU?

minfhong-qti: `target_hw_device_type` should be CPU since on Linux x86 there is no actual NPU device.

PR #320, opened 2026-04-30
Wire build_nuget and build_archive flags for all Windows architectures

huaychou:
### Description

- Wire build_nuget and build_archive flags for all Windows architectures to reduce gaps

### Motivation and Context

- The --build-nuget and --build-archive CLI flags were only forwarded to the arm64 build task. arm64ec, arm64x, and x86_64 targets silently ignored them, preventing artifact generation for those architectures.

### DIFF
--- a/qcom/build_and_test.py
+++ b/qcom/build_and_test.py
@@ -456,6 +456,7 @@ class TaskLibrary:
                         self.__qairt_sdk_root,
                         "build",
                         extra_args=extra_args,
+                        build_archive=self.__build_archive,
                     )
                 )
             else:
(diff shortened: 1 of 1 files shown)


qti-chuteng (approved):

qcom/build_and_test.py
kuanyul-qti, line 623:
```suggestion
                            build_nuget=self.__build_nuget,
                            build_archive=self.__build_archive,
                        ),
```

huaychou: This is not valid. We build `arm64-x-slice` in the arm64 build task, and then `arm64x` in the arm64ec build task, so we only need to wire to arm64ec build task.

qti-mbadnara: Yes this is agreed. The NuGet should be built only for the ARM64ec task

qti-mbadnara:
LGTM, merging it!

PR #322, opened 2026-04-30
Standardize archive asset invocation across platforms

huaychou:
### Description

- Pass --build_archive_asset to the existing build.bat
- Remove the direct pkg_assets.py call
- Fix build.ps1 for better maintainability

### Motivation and Context

- Linux already invokes build_archive_asset() via build.py using `--build_archive_asset`. Windows called pkg_assets.py directly, bypassing build.py

### DIFF
--- a/qcom/scripts/windows/build.ps1
+++ b/qcom/scripts/windows/build.ps1
@@ -212,6 +212,11 @@ if ($env:ORT_VERSION_SUFFIX) {
     $VersionSuffixArg += "--version_suffix", "$env:ORT_VERSION_SUFFIX"
 }
 
+$BuildArchiveArgs = @()
+if ($BuildArchive) {
+    $BuildArchiveArgs += "--build_archive_asset"
+}
+
 switch ($Mode) {
     "build" {
         if ($BuildIsDirty) {
@@ -272,6 +277,7 @@ else {
     if ($GenerateBuild -or $DoBuild) {
         try {
             python.exe "$RepoRoot\qcom\scripts\all\fetch_cmake_deps.py"
+            $BuildBatPath = (Join-Path $RepoRoot "build.bat")
 
             if ($GenerateBuild) {
                 if (-not (Test-Path $BuildVEnv)) {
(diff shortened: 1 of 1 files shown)


qcom/scripts/windows/build.ps1
```diff
@@ -284,15 +290,17 @@ else {
...
                 }
             }
 
             if ($DoBuild) {
                 $BuildOutputDir = (Join-Path $BuildDir $Config)
-                Assert-Success -ErrorMessage "Failed to build" {
-                    & cmake --build $BuildOutputDir --config $Config
+                Use-PyVenv -PyVenv $BuildVEnv {
+                    Assert-Success -ErrorMessage "Failed to build" {
+                        & $BuildBatPath --build --skip_tests $ArchArgs $CommonArgs $QnnArgs $PlatformArgs $VersionSuffixArg $BuildArchiveArgs
```
kuanyul-qti: Is there a reason for adding `--skip_tests`?

huaychou: Tests should be handled by `run_tests.ps1`

kuanyul-qti: Not all jobs will run `run_tests.ps1`. I believe `--skip_tests` should be removed here, unless its removal leads to errors. This will help ensure that the code does not introduce regressions at the frontline.

huaychou: I don't believe it's necessary to run tests during the build stage, as it would add to the CI workload.

huaychou: After checking, we found that `--skip_tests` is unnecessary in this context, so it should be removed.

huaychou:
@qti-jkilpatrick Could you please review and let me know if there is any impact on the AI hub workflow?

huaychou:
> @qti-jkilpatrick Could you please review and let me know if there is any impact on the AI hub workflow?

After our offline discussion, for the AI Hub workflow, please ensure that publishing the Python 3.10 wheel for x86_64 Linux to ort-ep-win-01 is not interrupted.

kuanyul-qti (approved):

qti-chuteng (approved):

PR #257, opened 2026-04-16
Revert "[QNN EP] Add monolithic lstm (#111)"

qti-ashwshan:
### Description
- This reverts commit 8870e71abfe03edd6a5e4a874e67d6c7e0abffff.

PR #323, opened 2026-04-30
Reapply "[QNN EP] Add monolithic lstm (#111)" (#257)

chunghow-qti:
This reverts commit 30ffaf51a16f577bac11e13e060a006bf9375798 (the revert of #111).

Reapplies [QNN EP] Add monolithic LSTM support.

### Description

The monolithic LSTM feature was originally added in #111 and later reverted in #257 due to an apparent 7x inference time regression observed with `onnxruntime_perf_test` compared to `disable_htp_monolithic_lstm`. Further investigation has confirmed the regression is **not** caused by the monolithic LSTM kernel itself or by the QNN EP integration, but by differing default QNN log levels between the two tools:

| Tool | Default QNN log level | Observed time |
|---|---|---|
| `onnxruntime_perf_test` | `WARN` | ~580 ms |
| `qnn-net-run` | `ERROR` | ~80 ms |
| `qnn-net-run --log_level=verbose` | `VERBOSE` | ~580 ms |

The HTP monolithic LSTM skel emits a large number of `WARN`-level log messages on the DSP. When the caller's log level is set to `WARN` or higher, these messages are printed, which significantly increases the reported inference time. When log level is `ERROR`, these messages are suppressed and performance returns to expected levels.

Verification:
- Confirmed `onnxruntime_perf_test` with `ORT_LOGGING_LEVEL_ERROR` produces the same ~80 ms inference time as `qnn-net-run` default.
- Confirmed `qnn-net-run` with `--log_level=verbose` reproduces the ~580 ms slow path seen in ORT.
- Confirmed via DSP skel log capture that the excess time correlates with log output volume, not with compute.
- After the fix in the HTP backend introduced in QAIRT SDK 2.47 RC2, all models with performance regression are now resolved.

The upcoming QAIRT SDK release will remove these `WARN`-level log messages from the monolithic LSTM skel, eliminating the regression entirely. With that fix in flight, it is safe to reapply the monolithic LSTM support in QNN EP — no code change is needed on the ORT side for correctness or performance.

### Motivation and Context

Restore the monolithic LSTM code path in QNN EP, which provides a large performance improvement for LSTM-heavy models on HTP.

### IMPORTANT: This PR should only be merged after the CI is upgraded to QAIRT 2.47, as it depends on an HTP fix introduced in that release.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -730,7 +730,7 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     enable_HTP_FP16_precision_ = false;
   } else {
     ORT_CXX_LOG(logger_,
-                ORT_LOGGING_LEVEL_VERBOSE,
+                ORT_LOGGING_LEVEL_ERROR,
                 ("Invalid enable_htp_fp16_precision: " +
                  enable_htp_fp16_precision_str +
                  " only 0 or 1 allowed. Set to 0.")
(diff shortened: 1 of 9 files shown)


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -765,6 +753,29 @@ QnnEp::QnnEp(QnnEpFactory& factory,
               ORT_LOGGING_LEVEL_VERBOSE,
               ("User specified enable_htp_fp16_precision: " + enable_htp_fp16_precision_str).c_str());
 
+  // HTP monolithic lstm
+  std::string disable_htp_monolithic_lstm_str;
+  GetSessionConfigEntryOrDefault(ort_api,
+                                 session_options_,
+                                 FormatEPConfigKey("disable_htp_monolithic_lstm"),
```
tirupath-qti, line 747: Let us remove this option and keep it enable always as it improves performance, and session creation time and also reduces context binary size for majority representative models.

chunghow-qti: Hi @tirupath-qti ,  think we should retain the option since it’s not guaranteed to always improve performance, keeping it might be safer. And the option will default enable monolithic lstm. 

There is a use case for monolithic LSTM provided in the qnn sdk doc. 

https://docs.qualcomm.com/doc/80-63442-10/topic/htp_backend.html#recommended-use-cases
image

huaychou:
Since we can't merge this until QAIRT is updated to 2.47, could we add something like [DO NOT MERGE] to the title to help prevent an accidental merge?

chunghow-qti:
Since ORT has uplevel to QAIRT 2.47 by #471. This PR can be merged now.

chenweng-quic (approved):

qti-chuteng (approved):

PR #21, opened 2026-02-02
Replace Non-ABI implementation with ABI

qti-hungjuiw:
### Description

- This PR migrates the ABI implementation from `core/providers/qnn-abi` to `core/providers/qnn` and from `test/providers/qnn-abi` to `test/providers/qnn`, effectively replacing the existing QNN provider with the enhanced ABI implementation. The migration includes:
- Removed legacy `qnn_provider_factory_creator.h` and associated **provider bridge implementations**
- Removed corresponding **legacy test cases** as they will be replaced by the **plugin EP library** architecture

### Motivation and Context

- The primary motivation for this migration approach is to preserve the valuable commit history of the existing `core/providers/qnn` and `test/providers/qnn` folder while upgrading to the enhanced ABI implementation. This ensures that all historical development context, bug fixes, and feature evolution remain accessible for future maintenance and debugging.

### Known Risks and Limitations
1. **Python Wheel Compatibility**: Due to different use cases and user guidelines for Python Wheel in the plugin EP library, the testcases require careful validation and potential updates to accommodate the new ABI provider's Python integration patterns.
    - `test_ort_windows_arm64_pysmoke`
    - `test_ort_windows_arm64_pygpu`
2. **Missing Recent Operations**: The following recent operation builders and their test cases are not yet included in this migration:
    - `quick_gelu_op_builder.cc` and `quick_gelu_op_test.cc`
    - `fused_matmul_op_builder.cc` and `fused_matmul_op_test.cc`
    - `matmulnbits_op_builder.cc`
    - `is_nan_op_builder.cc‎`

PR #328, opened 2026-05-01
[QNN EP] Restore main_context_pos_list indirection in TryGetMaxSpillFillSize

qti-ashwshan:
### Description
- Fix: ABI rewrite (#21) dropped 'main_context_pos_list[idx]' indirection, causing to read non-main EPContext subgraphs leading to mis-size the shared spill-fill biffer for multi-split weight shared ctx models.
- Added unit test to verify fix


onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -24,6 +24,16 @@
 #include "HTP/QnnHtpCommon.h"
 #include "QnnSdkBuildId.h"
 
+// Needed by the NonConsecutiveMainContexts regression test below.
```
minfhong-qti: As announced in the offline group chat, we should not include these private headers.
In addition, which testcases are they for? I did not see `NonConsecutiveMainContexts`.

minfhong-qti (commented):
Thanks @qti-ashwshan for fixing this!

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
qti-ashwshan: Thats my bad. Was supposed to remove the whole test file

tirupath-qti (approved):

PR #331, opened 2026-05-04
[QNN EP] Add Range op support

qti-chuteng:
ONNX Range has no native QNN op. The QNN EP now constant-folds Range at graph-build time: when all three inputs (start, limit, delta) are initializers, values are computed host-side, emitted as a STATIC tensor, and wired to the graph output via an identity-permutation Transpose. Range with any dynamic input is rejected in IsOpSupported so ORT falls back to CPU, preserving the ONNX contract.

Implementation
--------------
- New range_op_builder.cc:
  - IsOpSupported: requires start/limit/delta to be constant initializers with matching dtype; rejects delta == 0 and int64 values outside the int32 range.
  - ComputeRangeValues: host-side computation for FLOAT32, INT32, and INT64. INT64 is narrowed to INT32 for the STATIC tensor (QNN NPU Transpose does not accept INT_64) and restored to INT64 via a trailing Cast when the ONNX output demands it.
  - ProcessAttributesAndOutputs: emits the STATIC values tensor + 1D identity Transpose, plus the optional Cast for int64.
- Registered in op_builder_factory.{h,cc}.

Tests
-----
New range_op_test.cc adds six Range scenarios (float32 ascending/fractional-delta/descending, int32 ascending/step, int64 ascending) on both QnnCPUBackendTests and QnnHTPBackendTests (HTP block guarded by __aarch64__/_M_ARM64/__linux__, matching the repo convention used by cumsum_test.cc).

Validation
----------
- QnnCPUBackendTests.Range*: 6/6 pass on x86 and arm64 (hamoa).
- QnnHTPBackendTests.Range*: 6/6 pass on arm64 hamoa (HTP arch 73).
- x86 QnnCPU* regression: 205/205 pass (1 unrelated pre-existing IR-backend failure).


onnxruntime/core/providers/qnn/builder/opbuilder/range_op_builder.cc
```diff
@@ -0,0 +1,274 @@
...
+  }
+  // Reject early on unsupported dtypes by running ComputeRangeValues, which
+  // also validates delta != 0 and int64 overflow.
+  Qnn_DataType_t onnx_dtype = QNN_DATATYPE_UNDEFINED;
+  Qnn_DataType_t static_dtype = QNN_DATATYPE_UNDEFINED;
+  std::vector<uint8_t> bytes;
+  uint32_t count = 0;
+  RETURN_IF_ERROR(ComputeRangeValues(qnn_model_wrapper, node_unit,
+                                     onnx_dtype, static_dtype, bytes, count));
+  return AddToModelBuilder(qnn_model_wrapper, node_unit, logger, true);
```
qti-yuduo: Why do we invoke `AddToModelBuilder` from `IsOpSupported`? This would double build range op? This function should only perform validations.

onnxruntime/core/providers/qnn/builder/opbuilder/range_op_builder.cc
```diff
@@ -0,0 +1,274 @@
...
+};
+
+namespace {
+
+template <typename T>
+void ComputeRangeTyped(T start, T limit, T delta, std::vector<uint8_t>& out_bytes, uint32_t& count) {
+  int64_t n = 0;
+  if constexpr (std::is_floating_point_v<T>) {
+    const double diff = static_cast<double>(limit) - static_cast<double>(start);
+    const double d = static_cast<double>(delta);
```
qti-yuduo: if `(limit-start)/delta` is NaN/±Inf or overflows int64, `static_cast<int64_t>(v)` is UB. Better we should check overflow and reject on `IsOpSupported`. Re-use same compute logics to capture overflow is desired (re-purpose this function to both validate and compute)

onnxruntime/test/providers/qnn/range_op_test.cc
```diff
@@ -0,0 +1,112 @@
...
+  RunRangeOpTest<int32_t>(0, 8, 1, 11, ExpectedEPNodeAssignment::All, "htp");
+}
+
+TEST_F(QnnHTPBackendTests, Range_int32_step) {
+  RunRangeOpTest<int32_t>(2, 20, 3, 11, ExpectedEPNodeAssignment::All, "htp");
+}
+
+TEST_F(QnnHTPBackendTests, Range_int64_ascending) {
+  RunRangeOpTest<int64_t>(0, 16, 1, 11, ExpectedEPNodeAssignment::All, "htp");
+}
```
qti-yuduo, line 148: Can we test also `start == limit` (empty range)

onnxruntime/test/providers/qnn/range_op_test.cc
```diff
@@ -0,0 +1,112 @@
...
+
```
qti-yuduo, line 149: - Test dynamic case -> reject. 
- delta == 0 -> reject
- mixed-dtype -> reject
- `int64` expect cast.
- negative delta.

onnxruntime/test/providers/qnn/range_op_test.cc
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#if !defined(ORT_MINIMAL_BUILD)
+
+#include <string>
+
+#include "core/graph/graph.h"
```
minfhong-qti: This header could no longer be used. Please update branch and make sure CI passed.

qti-chuteng:
Hi @qti-yuduo @minfhong-qti , could you help review this PR?

qti-yuduo (approved):
LGTM.

qti-yuduo:
> Hi @qti-yuduo @minfhong-qti , could you help review this PR?

Just click on re-request next time :) That'll send out notifications.

minfhong-qti (approved):

PR #337, opened 2026-05-05
fix: prevent QnnSignal_trigger TerminateProcess in file-mapped context loading

huaychou:
### Description

Two defensive fixes for file-mapped weights on Windows ARM64:

1. **MapDmaData**: when RPCMEM's to_fd() returns -1, deregister the partially-registered buffer and set file_mapped_weights_enabled_=false. Previously the stale FASTRPC registration could corrupt DSP session state on teardown, and subsequent MapDmaData calls would retry the same failing path for every weight tensor instead of failing fast.

2. **SetupBackend probe**: before creating the WindowsFileMapper, probe RPCMEM with a real mmap'd temp file at an unaligned offset (+1). This mirrors what MapDmaData does for each weight tensor. On devices where only page-aligned heap succeeds but unaligned file-mapped memory does not, the probe correctly disables file mapping early — preventing contextCreateFromBinaryWithCallback from being called and avoiding the QnnSignal_trigger that schedules a background TerminateProcess(0). A heap-only probe gives false positives on such devices because RPCMEM can import heap but cannot import file-mapped sub-ranges.

### Motivation and Context

- fix: prevent QnnSignal_trigger TerminateProcess in file-mapped context loading

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -4,6 +4,7 @@
 
 #include "core/providers/qnn/builder/qnn_backend_manager.h"
 
+#include <atomic>
 #include <filesystem>
 #include <fstream>
 #include <functional>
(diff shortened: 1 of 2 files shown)


huaychou:
Not an issue

PR #339, opened 2026-05-05
Add constant folding for standalone post-partition Q/DQ

qti-yuduo:
Introduce constant folding for standalone Q/DQ ops Track folded constants in QnnModelWrapper.

For model such as, Qwen3-0.6B, some constant weights follow this pattern:

`initializer -> DequantizeLinear -> QuantizeLinear -> DequantizeLinear -> op`

QNN EP folded only the first hop. After that, the intermediate Q/DQ output was no longer recognized as effectively constant, so later folded weight tensors were emitted as QNN `APP_WRITE` graph inputs instead of `STATIC` tensors. In the reproduced Qwen DLC, the ONNX graph had 60 real runtime inputs, but the generated QNN graph had 284 inputs: 224 leaked `v_proj_sha.*.weight_qdq` tensors.

Our fix helped by teaching QNN EP to remember tensors produced by successful constant Q/DQ folding. QnnModelWrapper now marks folded outputs as folded constants, and simple_op_builder/the extracted QDQ folding helper treat those outputs as constant inputs for the next Q/DQ hop. That lets the whole initializer → DQ → Q → DQ chain keep folding through to the final consumer.

Result: those chained constant weights become QNN STATIC tensors instead of runtime APP_WRITE graph inputs, so the generated DLC exposes only the real model inputs and context-binary linking no longer sees a bloated input signature.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/qdq_constant_folding.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/qdq_constant_folding.cc
@@ -20,7 +20,7 @@ namespace qnn {
 
 namespace {
 
-// Fetch raw bytes from either a real initializer or a previously-folded STATIC tensor.
+// Accepts either a real initializer or a previously-folded STATIC tensor.
 Ort::Status GetConstantTensorBytes(QnnModelWrapper& qnn_model_wrapper,
                                    const std::string& tensor_name,
                                    /*out*/ std::vector<uint8_t>& bytes) {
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/qdq_constant_folding.cc
```diff
@@ -0,0 +1,208 @@
...
+  std::vector<float> scales;
+  RETURN_IF_ERROR(qnn_model_wrapper.UnpackScales(output_def.quant_param->scale, scales));
+
+  std::vector<int32_t> offsets;
+  if (output_def.quant_param->zero_point != nullptr) {
+    ONNXTensorElementDataType zp_type;
+    RETURN_IF_ERROR(qnn_model_wrapper.UnpackZeroPoints(output_def.quant_param->zero_point, offsets, zp_type));
+  } else {
+    offsets.assign(scales.size(), 0);
+  }
```
yath1: Repeated code - I understand you are doing quantize here and dequant in the other method. But, if possible can you refactor repeated code into separate helper method?

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -934,6 +934,78 @@ TEST_F(QnnCPUBackendTests, ConvTranspose1Df32_DynamicWeights_DefaultBias) {
                 ExpectedEPNodeAssignment::All);
 }
 
+// Builds: weight_q0 (int8 init) -> DQ -> Q -> DQ -> Conv.
+// Used to regression-test chained folding; differing scale0/scale1 exercises real requant
```
yath1, line 938: Curious what would happen if weight tensor is shared between multiple layers. Is the flow then, ORT core would add separate QDQ for each of them and then this new logic kicks in to fold those weights to respective layers?

onnxruntime/core/providers/qnn/builder/opbuilder/qdq_constant_folding.cc
qti-yuduo: Done.

onnxruntime/test/providers/qnn/conv_test.cc
qti-yuduo: Yes. We'd have different chains and each would have distinct tensor names. So chain folds independently into STATIC tensor here. And if two folds ever produce the same output tensor name, registration is idempotent.

yath1: I see, so when the first dequant layer is processed in the graph, we would memcpy the weight, use the scale/offset available to the dequant node to create an fp32 tensor which can be used by consumer Q/DQ nodes. Because of the memcpy we don't have any issue with the QDQ nodes in each branch having different encodings. Could you please confirm if my understanding is correct?

yath1 (approved):

onnxruntime/test/providers/qnn/conv_test.cc
qti-yuduo: Correct and the copy happens at every fold hop, not the first DQ. It decouples the fold from the lifetime of the prior tensor's buffer.

PR #342, opened 2026-05-05
Workaround QNN rank-5 matmul rejection

qti-yuduo:
QNN Rejects rank 5 matmul: 

```Failed to validate op /model/blocks.5/proj/MatMul incorrect Rank 5. ```

Workaround by flatten input 0.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#include <functional>
+#include <limits>
 #include <numeric>
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by qti-chuteng** [comment]
---

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc` — `ProcessInput0` batch accumulate

The new batch computation uses `uint32_t` as the `std::accumulate` accumulator, so the product silently wraps on overflow for large batch dims (e.g., `{65536, 65536, k}` → wraps instead of erroring). The same pre-existing issue exists in `ProcessAttributesAndOutputs` (~line 442). Please accumulate in `int64_t` first and add a range check before narrowing:

```cpp
int64_t batch_i64 = std::accumulate(
    input_0_info.shape.begin(), input_0_info.shape.end() - 1,
    static_cast<int64_t>(1), std::multiplies<int64_t>());
RETURN_IF(batch_i64 <= 0 || batch_i64 > static_cast<int64_t>(std::numeric_limits<uint32_t>::max()),
          "MatMul: batch dimension product overflows uint32_t");
shape_2d = {static_cast<uint32_t>(batch_i64), input_0_info.shape.back()};
```

Consider fixing `ProcessAttributesAndOutputs` in the same pass.

---

**[N-1]** `onnxruntime/test/providers/qnn/matmul_test.cc` — `QnnHTPBackendTests::MatMulOp`

The new rank-3 and rank-5 FullyConnected cases are only added to `QnnCPUBackendTests`. Since the workaround targets QNN HTP's rejection of rank > 4 FC inputs, coverage on HTP is more important. Please add corresponding HTP cases, e.g.:

```cpp
RunMatMulOpTest({2, 3, 4}, {4, 2}, false, true, ExpectedEPNodeAssignment::All, "htp", 18, 1e-2f);
RunMatMulOpTest({1, 1, 2, 2, 4}, {4, 2}, false, true, ExpectedEPNodeAssignment::All, "htp", 18, 1e-2f);
```

If rank-5 is not yet accepted by HTP, add an `ExpectedEPNodeAssignment::None` case with a comment explaining the expected behavior.

---

**[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc` — duplicated batch flatten

The `std::accumulate` batch-dim product pattern now appears in both `ProcessInput0` and `ProcessAttributesAndOutputs`. Consider extracting a small file-static helper (e.g., `FlattenBatchDims`) — especially worth doing alongside the M-1 overflow fix to ensure both sites are corrected consistently.

qti-yuduo:
@qti-chuteng thanks for the feedback, address all of them.

qti-shubham (approved):

yath1 (approved):
lgtm

PR #344, opened 2026-05-06
fix: suppress graphFinalize stdout via SetStdHandle to avoid QNN DLL freopen corruption

huaychou:
### Description

- Fix: on Windows use SetStdHandle(STD_OUTPUT_HANDLE, null_handle) for the redirect instead of _dup2. fd 1 is never touched, so _get_osfhandle(1) returns the same value the DLL cached at startup — no freopen triggered, gtest output preserved. On Linux the existing dup/dup2 approach is retained as the QNN DLL behaviour described above is Windows/UCRT-specific.

### Motivation and Context

- The QNN HTP backend DLL monitors both GetStdHandle(STD_OUTPUT_HANDLE) and _get_osfhandle(1) during contextLoadFromBinary. When either value differs from what the DLL cached at startup, it internally calls freopen("NUL","w",stdout) to suppress its own output. freopen replaces the FILE* internal state (buffer pointers, flags, error bits), causing all subsequent C stdio writes — including gtest's [  PASSED  ] lines — to be silently lost.

- The previous fd-level approach (_dup2) triggered this: UCRT's _dup2(x,1) always calls SetStdHandle(STD_OUTPUT_HANDLE, new_dup_handle) internally, producing a different HANDLE integer value that the QNN DLL detects via _get_osfhandle(1) even after we restore STD_OUTPUT_HANDLE manually.


huaychou:
No longer required after https://github.com/onnxruntime/onnxruntime-qnn/pull/366 merged

PR #345, opened 2026-05-06
[QNN EP] Add framework op tracing feature

kuanyul-qti:
### Description

- New session options: qnn.enable_framework_op_trace, qnn.framework_op_trace_dir
- Collect unsupported node info with rejection reasons in GetSupportedNodes
- Record op/tensor mappings via OpTraceCollector hook in CreateQnnNode during ComposeGraph, and tensor mappings from QnnModelWrapper after composition
- Serialize trace to JSON file in CompileImpl
- Add unit tests

### Motivation and Context
Add framework op tracing feature

### DIFF
--- a/cmake/onnxruntime_unittests.cmake
+++ b/cmake/onnxruntime_unittests.cmake
@@ -212,6 +212,19 @@ file(GLOB onnxruntime_test_framework_src CONFIGURE_DEPENDS
 list(REMOVE_ITEM onnxruntime_test_framework_src
      "${TEST_SRC_DIR}/providers/qnn/optimizer/transpose_optimizer_test.cc")
 
+# qnn_op_tracing_serialization.cc is compiled directly into the test binary so
+# that QnnFrameworkOpTraceUnit tests can call ComputeTraceSummary and
+# SerializeFrameworkOpTrace without linking against the EP library.
+# This translation unit intentionally does NOT include ort_api.h, avoiding
+# the ORT_API_MANUAL_INIT mismatch linker error on Windows.  It is needed in
+# both shared-lib builds (hidden symbol visibility prevents linking) and
+# static-lib builds (test binary has build-order dependency only, not a linker
+# dependency, on the EP).
+if(onnxruntime_USE_QNN AND NOT onnxruntime_MINIMAL_BUILD AND NOT onnxruntime_REDUCED_OPS_BUILD)
+  list(APPEND onnxruntime_test_framework_src
+       "${ONNXRUNTIME_ROOT}/core/providers/qnn/builder/op_tracing/qnn_op_tracing_serialization.cc")
+endif()
+
 #This is a small wrapper library that shouldn't use any onnxruntime internal symbols(except onnxruntime_common).
 #Because it could dynamically link to onnxruntime. Otherwise you will have two copies of onnxruntime in the same
 #process and you won't know which one you are testing.
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -217,6 +217,15 @@ Alternatively to setting profiling_level at compile time, profiling can be enabl
 |---|---|
 |Directory path (string)|Directory path for dumping QNN IR DLC files. Only effective when `dump_qnn_ir_dlc` is enabled.|
 
+|`"enable_framework_op_trace"`|Description|
+|---|---|
+|'0'|Default. Disabled.|
+|'1'|Enable framework op tracing. Writes a JSON file that maps each ONNX operator (or fused group) to its corresponding QNN operator(s), records unsupported operators, and includes summary statistics. See `framework_op_trace_dir` to control the output location. The trace is produced from the QNN graph composition path: it is emitted on a fresh JIT compile and on AOT context-binary generation (when `ep.context_enable=1` writes a new EPContext model). It is **not** emitted when loading an existing EPContext model (cached run), because no graph composition occurs in that path; in that case an INFO log records that no trace file is written.|
+
+|`"framework_op_trace_dir"`|Description|
+|---|---|
+|Directory path (string)|Directory path for framework op trace JSON files. Only effective when `enable_framework_op_trace` is `'1'`. Defaults to the current working directory if not set. The directory is created automatically if it does not exist. If the directory cannot be created or written to (e.g. read-only mount, missing permissions), framework op tracing is disabled with a WARNING log and no trace file is produced. Trace files use the pattern `qnn_op_trace.json`.|
+
 |`"qnn_ir_backend_path"`|Description|
 |---|---|
 |Backend path (string)|Path to the QNN IR backend library. Defaults to 'libQnnIr.so' or 'QnnIr.dll'. Only effective when `dump_qnn_ir_dlc` is enabled.|
--- a/onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
+++ b/onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
@@ -1,0 +1,241 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "core/providers/qnn/builder/op_tracing/qnn_op_tracing.h"
+
+#include <algorithm>
+#include <fstream>
+
+#include "nlohmann/json.hpp"
+
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_node_group/qnn_node_group.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+// ---------------------------------------------------------------------------
+// OpTraceCollector
+// ---------------------------------------------------------------------------
+
+void OpTraceCollector::SetCurrentNodeGroup(const IQnnNodeGroup* node_group) {
+  current_node_group_ = node_group;
+}
+
+void OpTraceCollector::ClearCurrentNodeGroup() {
+  current_node_group_ = nullptr;
+}
+
+void OpTraceCollector::RecordOpMapping(const std::string& qnn_op_name, const std::string& qnn_op_type,
+                                       gsl::span<const std::string> output_tensor_names) {
+  // Guard: node group context is set by NodeGroupGuard around each AddToModelBuilder call.
+  // If nullptr, this QNN op was created outside that bracket (e.g. BF16
+  // conversion ops Cast added during ComposeQnnGraph postprocessing),
+  // which has no corresponding ONNX node group - skip it.
+  if (current_node_group_ == nullptr) {
+    return;
+  }
+
+  TraceMapping mapping;
+  mapping.dst_name = qnn_op_name;
+  mapping.dst_qnn_op_type = qnn_op_type;
+  mapping.node_group_type = std::string(current_node_group_->Type());
+
+  // Collect all raw nodes into a flat list. Inline capacity 8 covers the
+  // typical case (most node groups have <= 8 raw nodes; see
+  // qnn_execution_provider.cc).
+  InlinedVector<const OrtNode*, 8> all_nodes;
+  for (const OrtNodeUnit* node_unit : current_node_group_->GetNodeUnits()) {
+    for (const OrtNode* node : node_unit->GetAllNodesInGroup()) {
+      all_nodes.push_back(node);
+    }
+  }
+
+  // Build srcInfo chain:
+  //   op1, tensor_out1, op2, tensor_out2, ..., opN
+  // Each source op is followed by the output tensor that actually flows into
+  // the next op in the chain (found by intersecting node[i]'s outputs with
+  // node[i+1]'s inputs). Falls back to outputs[0] if no match is found.
+  for (size_t i = 0; i < all_nodes.size(); ++i) {
+    Ort::ConstNode const_node(all_nodes[i]);
+    // Node names are optional per ONNX spec. Fall back to OpType+position,
+    // mirroring UniqueNameGenerator::New(node_unit)'s OpType+Index fallback
+    // so that source names are structurally consistent with QNN op names
+    // in the 1:1 case. The position index `i` differs from the graph-global
+    // NodeUnit index used by UniqueNameGenerator, so the names may not match
+    // exactly for unnamed nodes — but they remain unique within this group and
+    // are typed as OP (not a tensor name masquerading as an OP source).
+    std::string node_name = std::string(const_node.GetName());
+    if (node_name.empty()) {
+      node_name = std::string(const_node.GetOperatorType()) + std::to_string(i);
+    }
+    mapping.sources.push_back({std::move(node_name), TraceTargetType::kOp});
+
+    // Add linking tensor for all nodes except the last.
+    if (i + 1 < all_nodes.size()) {
+      std::vector<Ort::ConstValueInfo> outputs = const_node.GetOutputs();
+      if (!outputs.empty()) {
+        // Collect this node's output names for O(1) lookup.
+        InlinedHashSet<std::string> output_names;
+        for (const Ort::ConstValueInfo& out : outputs) {
+          output_names.insert(std::string(out.GetName()));
+        }
+
+        // Find all outputs consumed by the next node in the chain.
+        // Note: Ort::ConstValueInfo wraps a nullable OrtValueInfo pointer;
+        // there is no operator bool(), so we extract the raw pointer to check.
+        InlinedVector<std::string, 8> linking_tensors;
+        Ort::ConstNode next_node(all_nodes[i + 1]);
+        for (const Ort::ConstValueInfo& in : next_node.GetInputs()) {
+          const OrtValueInfo* raw_in = in;
+          if (!raw_in) continue;  // optional input not provided
+          std::string in_name = std::string(in.GetName());
+          if (output_names.count(in_name)) {
+            linking_tensors.push_back(std::move(in_name));
+          }
+        }
+
+        // Fall back to outputs[0] if the intersection is empty (should not
+        // happen for a well-formed fusion, but avoids a missing entry).
+        if (linking_tensors.empty()) {
+          linking_tensors.push_back(std::string(outputs[0].GetName()));
+        }
+
+        for (std::string& t : linking_tensors) {
+          mapping.sources.push_back({std::move(t), TraceTargetType::kTensor});
+        }
+      }
+    }
+  }
+
+  if (mapping.sources.empty()) {
+    return;
+  }
+
+  // Record tensor mappings for each output tensor of this QNN op.
+  // Each output tensor carries the same srcInfo chain as the op
+  // (N:1 or N:M — same sources, different dst_name).
+  for (const auto& out_name : output_tensor_names) {
+    TraceMapping t;
+    t.dst_name = out_name;
+    t.sources = mapping.sources;  // shared sources, copy intentional
+    tensor_mappings_.push_back(std::move(t));
+    recorded_tensor_names_.insert(out_name);
+  }
+
+  op_mappings_.push_back(std::move(mapping));
+}
+
+void OpTraceCollector::Finalize(const std::string& graph_name, const QnnModelWrapper& wrapper, OpTraceInfo& out) {
+  // Records the direct ONNX<->QNN tensor correspondence when a tensor was
+  // renamed by an op builder (onnx_name != qnn_name).
+  //
+  // Identity mappings (onnx_name == qnn_name, e.g. graph inputs and
+  // initializers that flow through unchanged) are NOT recorded — they carry
+  // no information for downstream tools and would inflate the trace with
+  // zero-information entries for every graph input and weight tensor.
+  //
+  // Output tensors of QNN ops are recorded by RecordOpMapping (with the op's
+  // full srcInfo chain) and are therefore also excluded here to avoid duplicates.
+  //
+  // Two emission rules ensure each qnn_name appears at most once in
+  // tensor_mappings_:
+  //   * Option A (merge): when several wrappers resolve to the same qnn_name
+  //     for non-canonical reasons, all distinct internal source names are
+  //     merged into one entry's sources list.
+  //   * Option B (canonical): when qnn_name is an external override target
+  //     (i.e. produced by `offload_graph_io_quantization` aliasing a QNN
+  //     internal tensor back to its original ONNX graph-input/output name),
+  //     the trace records the canonical mapping {dst: qnn_name, sources:
+  //     [qnn_name]}. The intermediate internal names (q_*_out / dq_*_out)
+  //     are an artefact of the alias and carry no information for downstream
+  //     tools, so they are dropped.
+  const auto& tensors_map = wrapper.GetModelTensorsMap();
+
+  // First pass: group unique source names by qnn_name to deduplicate.
+  InlinedHashMap<std::string, InlinedVector<std::string, 8>> grouped;
+  for (const auto& [key, tensor_wrapper] : tensors_map) {
+    const std::string& onnx_name = tensor_wrapper.GetName();
+    const std::string& qnn_name = tensor_wrapper.GetResolvedTensorName();
+    if (onnx_name == qnn_name || recorded_tensor_names_.count(qnn_name)) {
+      continue;  // identity or already recorded by RecordOpMapping
+    }
+    auto& sources = grouped[qnn_name];
+    if (std::find(sources.begin(), sources.end(), onnx_name) == sources.end()) {
+      sources.push_back(onnx_name);
+    }
+  }
+
+  // Second pass: emit one entry per unique qnn_name, applying B or A.
+  for (auto& [qnn_name, sources] : grouped) {
+    TraceMapping mapping;
+    mapping.dst_name = qnn_name;
+    if (wrapper.IsExternalOverrideTarget(qnn_name)) {
+      // Option B: canonical mapping.
+      mapping.sources.push_back({qnn_name, TraceTargetType::kTensor});
+    } else {
+      // Option A: merged sources.
+      for (auto& src : sources) {
+        mapping.sources.push_back({std::move(src), TraceTargetType::kTensor});
+      }
+    }
+    tensor_mappings_.push_back(std::move(mapping));
+    recorded_tensor_names_.insert(qnn_name);
+  }
+
+  out.graph_name = graph_name;
+  out.op_mappings = std::move(op_mappings_);
+  out.tensor_mappings = std::move(tensor_mappings_);
+}
+
+bool WriteTraceToFile(const FrameworkOpTrace& trace,
+                      const std::filesystem::path& output_path,
+                      const Ort::Logger& logger) {
+  std::error_code ec;
+  auto parent = output_path.parent_path();
+  // parent is empty when output_path is a filename-only path (write to CWD).
+  if (!parent.empty()) {
+    std::filesystem::create_directories(parent, ec);
+    if (ec) {
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                  ("Failed to create trace output directory: " + parent.string() +
+                   " error: " + ec.message())
+                      .c_str());
+      return false;
+    }
+  }
+
+  nlohmann::json j = SerializeFrameworkOpTrace(trace);
+
+  // Intentional overwrite: the trace file is a diagnostic artefact and the
+  // most recent run's data is always the most useful.  Log a warning so the
+  // developer is aware, but do not treat it as an error.
+  if (std::filesystem::exists(output_path)) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                ("Overwriting existing framework op trace file: " + output_path.string()).c_str());
+  }
+
+  std::ofstream ofs(output_path);
+  if (!ofs.is_open()) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                ("Could not open framework op trace file: " + output_path.string()).c_str());
+    return false;
+  }
+
+  ofs << j.dump(2);  // pretty-print with 2-space indent
+  ofs.close();
+  if (!ofs) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                ("Framework op trace write failed (disk full / I/O error): " + output_path.string()).c_str());
+    std::error_code rm_ec;
+    std::filesystem::remove(output_path, rm_ec);
+    return false;
+  }
+
+  ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_INFO,
+              ("Framework op trace written to: " + output_path.string()).c_str());
+  return true;
+}
+
+}  // namespace qnn
+}  // namespace onnxruntime
--- a/onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.h
+++ b/onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.h
@@ -1,0 +1,84 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+// EP-internal header: OpTraceCollector class definition and WriteTraceToFile.
+// Test TUs should include qnn_op_tracing_types.h (which also declares
+// SerializeFrameworkOpTrace via nlohmann/json_fwd.hpp) and add
+// nlohmann/json.hpp directly if they use the return value.
+
+#pragma once
+
+#include <filesystem>
+#include <string>
+#include <vector>
+
+#include "core/providers/qnn/common/inlined_containers.h"
+#include "core/providers/qnn/builder/op_tracing/qnn_op_tracing_types.h"
+#include "core/providers/qnn/ort_api.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+class IQnnNodeGroup;
+class QnnModelWrapper;
+
+// Collects ONNX-to-QNN op/tensor mappings during a single ComposeGraph() call.
+// Lifetime: stack-allocated in QnnModel::ComposeGraph(), pointer passed to QnnModelWrapper.
+class OpTraceCollector {
+ public:
+  OpTraceCollector() = default;
+  ORT_DISALLOW_COPY_AND_ASSIGNMENT(OpTraceCollector);
+
+  // Called from QnnModelWrapper::CreateQnnNode() when a QNN op is created.
+  // Also records a tensor mapping for each output tensor using the same sources.
+  void RecordOpMapping(const std::string& qnn_op_name, const std::string& qnn_op_type,
+                       gsl::span<const std::string> output_tensor_names);
+
+  // Finalize the subgraph trace: record tensor mappings from the wrapper and
+  // populate `out` with the collected op_mappings, tensor_mappings, and the
+  // graph name. Call once after ComposeQnnGraph().
+  void Finalize(const std::string& graph_name, const QnnModelWrapper& wrapper, OpTraceInfo& out);
+
+ private:
+  // Set/clear the current node group context. Only NodeGroupGuard may call these
+  // to preserve the set/clear pairing and exception-safety invariant.
+  friend class NodeGroupGuard;
+  void SetCurrentNodeGroup(const IQnnNodeGroup* node_group);
+  void ClearCurrentNodeGroup();
+
+  const IQnnNodeGroup* current_node_group_ = nullptr;
+  // Real subgraphs frequently exceed any small inline capacity (LLM subgraphs
+  // can have hundreds of ops/tensors), so use std::vector and let it grow on
+  // the heap. The trace path is debug-only, so the extra allocation is
+  // immaterial.
+  std::vector<TraceMapping> op_mappings_;
+  std::vector<TraceMapping> tensor_mappings_;
+  // Tracks QNN tensor names already recorded by RecordOpMapping so that
+  // Finalize can skip them and avoid duplicate entries.
+  InlinedHashSet<std::string> recorded_tensor_names_;
+};
+
+// RAII guard that brackets SetCurrentNodeGroup / ClearCurrentNodeGroup around one
+// AddToModelBuilder call so that ClearCurrentNodeGroup is always called even if
+// AddToModelBuilder returns early via an exception.
+class NodeGroupGuard {
+ public:
+  NodeGroupGuard(OpTraceCollector* c, const IQnnNodeGroup* g) : c_(c) {
+    if (c_) c_->SetCurrentNodeGroup(g);
+  }
+  ~NodeGroupGuard() {
+    if (c_) c_->ClearCurrentNodeGroup();
+  }
+  ORT_DISALLOW_COPY_AND_ASSIGNMENT(NodeGroupGuard);
+
+ private:
+  OpTraceCollector* c_;
+};
+
+// Write trace JSON to file. Returns true on success.
+bool WriteTraceToFile(const FrameworkOpTrace& trace,
+                      const std::filesystem::path& output_path,
+                      const Ort::Logger& logger);
+
+}  // namespace qnn
+}  // namespace onnxruntime
--- a/onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing_serialization.cc
+++ b/onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing_serialization.cc
@@ -1,0 +1,144 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+// Implementations of ComputeTraceSummary and SerializeFrameworkOpTrace.
+//
+// This translation unit (.cc file) is compiled directly into the test binary
+// (via cmake) so that unit tests can call these functions without linking
+// against the EP library. It intentionally does NOT include ort_api.h (or any
+// header that pulls it in) to avoid the ORT_API_MANUAL_INIT mismatch linker
+// error on Windows.
+
+#include "core/providers/qnn/builder/op_tracing/qnn_op_tracing_types.h"
+
+#include <unordered_set>
+
+#include "nlohmann/json.hpp"
+
+namespace onnxruntime {
+namespace qnn {
+
+void ComputeTraceSummary(FrameworkOpTrace& trace) {
+  TraceSummary& summary = trace.summary;
+  summary.unsupported_nodes = trace.unsupported_nodes.size();
+  summary.qnn_subgraphs = trace.subgraph_traces.size();
+  summary.total_qnn_ops = 0;
+  summary.fusion_count.clear();
+
+  // An ONNX node is "supported" if it appears as an OP-typed source on any QNN
+  // op_mapping. N:M fusions emit the same ONNX source set from multiple QNN op
+  // entries (e.g. ReshapeEinsumReshape: 3 ONNX ops referenced by 3 QNN op
+  // entries = 9 OP source occurrences but only 3 unique ONNX nodes), so we
+  // dedup by source name before counting.
+  std::unordered_set<std::string> supported_onnx_node_names;
+  for (const auto& subgraph : trace.subgraph_traces) {
+    summary.total_qnn_ops += subgraph.op_mappings.size();
+    for (const auto& mapping : subgraph.op_mappings) {
+      summary.fusion_count[mapping.node_group_type]++;
+      for (const auto& src : mapping.sources) {
+        if (src.type == TraceTargetType::kOp) {
+          supported_onnx_node_names.insert(src.name);
+        }
+      }
+    }
+  }
+
+  summary.supported_nodes = supported_onnx_node_names.size();
+  summary.total_onnx_nodes = summary.supported_nodes + summary.unsupported_nodes;
+}
+
+namespace detail {
+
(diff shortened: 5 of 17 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -254,6 +254,8 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
     return Ort::Status();
   }
 
+  uint32_t GetSocModel() const { return soc_model_; }
```
minfhong-qti, line 259: FYI. SoC model now cannot be queried even on QCOM devices, and therefore this value is only useful if user explicitly sets it. Not sure whether it would impact the design.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -204,6 +206,19 @@ class QnnEp : public OrtEp, public ApiPtrs {
...
+  std::atomic<uint32_t> framework_op_trace_count_{0};
+  // Accumulates unsupported nodes across all GetSupportedNodes() partitions
+  // until CompileImpl consumes them (via std::move, which also resets the
+  // vector). Protected by unsupported_nodes_mutex_ because ORT may invoke the
+  // GetCapability C-API callback concurrently for different graphs within the
+  // same session (each invocation calls through to GetSupportedNodes()).
+  // mutable: GetSupportedNodes() is const (required by the OrtEp contract) but
+  // accumulates state — this is the canonical ORT pattern for logically-const
+  // operations that require internal synchronisation.
+  mutable std::mutex unsupported_nodes_mutex_;
```
minfhong-qti: Just curious is there any reference of how ORT will concurrently call EP's `GetCapability`?

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Prefer to collect these new tracing-related files in a new folder.

onnxruntime/test/providers/qnn/framework_op_trace_test.cc
```diff
@@ -0,0 +1,1376 @@
...
+
+    // Unregister so subsequent RegisterQnnEpLibrary calls can re-register cleanly.
+    unreg();
+    return functional;
+  }();
+  return kResult;
+}
+
+// Skip the current HTP test if deviceCreate fails (backend non-functional).
+#define FWK_OP_TRACE_SKIP_IF_HTP_NOT_FUNCTIONAL()                          \
```
minfhong-qti: Why define this? Isn't `QnnHTPBackendTests` already handles it in setup?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2145,6 +2192,56 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
   ep->onnx_graph_io_names_.reset();
   ep->tensor_name_overrides_.clear();
 
+  // Framework op trace: serialize and write JSON.
+  // When multiple graphs are compiled, all subgraph traces are collected into a single file.
+  if (ep->enable_framework_op_trace_) {
```
minfhong-qti, line 2288: Separate as another function.

onnxruntime/core/providers/qnn/builder/qnn_op_tracing_serialization.cc
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// Uses static_cast to uint8_t so only the forward declaration of QnnBackendType
+// (which includes the underlying type) is needed — avoids pulling qnn_def.h
+// (and therefore ort_api.h) into this test-safe TU.
+//
+// KEEP IN SYNC WITH QnnBackendType enum in qnn_def.h.
+// The static_asserts in qnn_op_tracing.cc (which includes qnn_def.h) verify
+// that these integer literals match the actual enum values at compile time.
+std::string QnnBackendTypeToString(QnnBackendType backend_type) {
```
minfhong-qti: Not getting why this function must be here. It seems only `qnn_execution_provider` is using it. Could it be placed elsewhere and avoid all these troubles?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Acknowledged. When the SoC model cannot be queried from hardware (as is now the case), `GetSocModel()` returns 0 — the trace JSON will show "soc_model": 0. This is acceptable; the field is still meaningful when the user explicitly sets it via `device_id`.

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
kuanyul-qti: All four tracing files moved to `builder/op_tracing/` subfolder. Updated include paths in 7 consumer files and the cmake test TU path.

onnxruntime/test/providers/qnn/framework_op_trace_test.cc
kuanyul-qti: `QnnHTPBackendTests::SetUp()` calls `GetHTPSupport()` which currently unconditionally returns SUPPORTED (there's a TODO to fix this). It does NOT perform an actual functional compile check. `FWK_OP_TRACE_SKIP_IF_HTP_NOT_FUNCTIONAL()` is therefore needed: it runs a minimal Add model through the plugin EP interface to confirm HTP can actually compile graphs, which is the prerequisite for all framework op trace tests.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: The `if (ep->enable_framework_op_trace_)` block in `CompileImpl` extracted to `QnnEp::CollectAndWriteFrameworkOpTrace(QnnEp*, const OrtGraph* const*)`, declared as a private static method in the header.

onnxruntime/test/providers/qnn/framework_op_trace_test.cc
kuanyul-qti: Removed.

onnxruntime/core/providers/qnn/builder/qnn_op_tracing_serialization.cc
kuanyul-qti: Moved to `qnn_def.cc`.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -195,6 +195,14 @@ void TryEnableQNNSaver(ProviderOptions& qnn_options) {
   }
 }
 
+OrtHardwareDeviceType GetQnnHtpDeviceType() {
```
minfhong-qti: It seems the change for this can be removed.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -65,6 +67,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
                                              _In_ size_t count,
                                              _Out_writes_all_(count) OrtNodeComputeInfo** node_compute_infos,
                                              _Out_writes_(count) OrtNode** ep_context_nodes) noexcept;
+  static void CollectAndWriteFrameworkOpTrace(QnnEp* ep, const OrtGraph* const* graphs);
```
minfhong-qti: Move to the bottom of these functions with comment separating it with existing functions since they are required by `OrtEp` class.
In addition, I think it does not have to be `static` nor passing the ep as parameter.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Removed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -33,6 +33,7 @@
 #include "core/providers/qnn/builder/qnn_node_group/qnn_node_group.h"
 #include "core/providers/qnn/builder/qnn_thread_pool.h"
 #include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/providers/qnn/builder/op_tracing/qnn_op_tracing.h"
```
minfhong-qti: Please sort it.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -204,6 +207,19 @@ class QnnEp : public OrtEp, public ApiPtrs {
...
+  // Accumulates unsupported nodes across all GetSupportedNodes() partitions
+  // until CompileImpl consumes them (via std::move, which also resets the
+  // vector). Protected by unsupported_nodes_mutex_ because ORT may invoke the
+  // GetCapability C-API callback concurrently for different graphs within the
+  // same session (each invocation calls through to GetSupportedNodes()).
+  // mutable: GetSupportedNodes() is const (required by the OrtEp contract) but
+  // accumulates state — this is the canonical ORT pattern for logically-const
+  // operations that require internal synchronisation.
+  mutable std::mutex unsupported_nodes_mutex_;
+  mutable InlinedVector<qnn::UnsupportedNodeInfo, 16> unsupported_nodes_;
```
minfhong-qti: Can we pass `unsupported_nodes_` into `GetSupportedNodes` and no need to define as mutable?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1220,6 +1247,25 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
           supported_nodes.push_back(node);
         }
       }
+    } else if (enable_framework_op_trace_) {
+      InlinedVector<qnn::UnsupportedNodeInfo, 8> batch;
```
minfhong-qti: Is 8 a magic number that the maximum nodes in a node unit?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -247,6 +264,13 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
                                                       *context.model_settings,
                                                       context.tensor_name_overrides);
 
+  // Framework op trace: create collector if tracing is enabled
+  std::unique_ptr<OpTraceCollector> trace_collector;
+  if (context.enable_op_trace) {
+    trace_collector = std::make_unique<OpTraceCollector>();
+    qnn_model_wrapper.SetOpTraceCollector(trace_collector.get());
```
minfhong-qti: How about pass the collector during the `QnnModelWrapper` constructor? With the collector set, it will know that tracing is on and need to record.

onnxruntime/core/providers/qnn/builder/qnn_model.h
```diff
@@ -141,6 +144,8 @@ class QnnModel {
 
   const std::string& Name() const { return graph_info_->Name(); }
 
+  std::optional<OpTraceInfo> TakeOpTraceInfo() { return std::move(op_trace_info_); }
```
minfhong-qti: Taking the ownership of a class memeber is a bit weird. Could the `trace.subgraph_traces` take a const reference referring to the trace info owned by each `qnn_model`?
If concerning the memory usage, how about output the trace info through function parameter or the context?

onnxruntime/test/providers/qnn/qnn_test_utils.cc
kuanyul-qti: Removed.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Done. `CollectAndWriteFrameworkOpTrace` is a non-static instance method at the bottom of the private section.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: Done. `SetOpTraceCollector()` removed. Constructor now accepts `op_trace_collector` in `qnn_model_wrapper.h`. In `ComposeGraph`, the collector is created before the wrapper and passed at construction.

onnxruntime/core/providers/qnn/builder/qnn_model.h
kuanyul-qti: Done. Trace info output through context.
`op_trace_info_` member and `TakeOpTraceInfo()` are removed from `QnnModel`. `QnnModelContext` now has `op_trace_output`. No model member holds trace state between compose and compile.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Add comment. most node groups have <= 8 raw nodes (e.g. a QDQ OrtNodeUnit has 3–4 nodes; complex fusions like GeluFusion span 5–6 OrtNodeUnits of 1 node each).
8 is a stack-allocation hint, not a capacity limit — the vector can hold more.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Done.
An `unsupported_nodes` pointer is added as the last parameter — nullptr means "don't collect", non-null means "write here". This replaces the `else if (enable_framework_op_trace_)` member check inside the function body with `else if (unsupported_nodes)`, making the function a pure function of its parameters with no hidden member access.
`unsupported_nodes_` remains a member variable because it is a cross-call accumulator: `GetCapabilityImpl` is called once per subgraph, each call appending to the same member via `&ep->unsupported_nodes_`. `CollectAndWriteFrameworkOpTrace` then consumes and clears it in one shot after all subgraphs are compiled. Removing the member would leave nowhere to accumulate across those calls.

cmake/onnxruntime_unittests.cmake
```diff
@@ -214,6 +214,19 @@ file(GLOB onnxruntime_test_framework_src CONFIGURE_DEPENDS
...
+# that QnnFrameworkOpTraceUnit tests can call ComputeTraceSummary and
+# SerializeFrameworkOpTrace without linking against the EP library.
+# This TU intentionally does NOT include ort_api.h, avoiding
+# the ORT_API_MANUAL_INIT mismatch linker error on Windows.  It is needed in
+# both shared-lib builds (hidden symbol visibility prevents linking) and
+# static-lib builds (test binary has build-order dependency only, not a linker
+# dependency, on the EP).
+if(onnxruntime_USE_QNN AND NOT onnxruntime_MINIMAL_BUILD AND NOT onnxruntime_REDUCED_OPS_BUILD)
+  list(APPEND onnxruntime_test_framework_src
+       "${ONNXRUNTIME_ROOT}/core/providers/qnn/builder/op_tracing/qnn_op_tracing_serialization.cc")
```
huaychou, line 225: Could we expose the symbol in shared library instead of including core in unittest?

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing_serialization.cc
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+// Implementations of ComputeTraceSummary and SerializeFrameworkOpTrace.
+//
+// This TU is compiled directly into the test binary (via cmake) so that unit
```
huaychou: What is TU meaning?

onnxruntime/core/providers/qnn/builder/qnn_def.cc
```diff
@@ -647,5 +647,24 @@ bool IsQpuBackend(QnnBackendType backend_type) {
...
+  switch (backend_type) {
+    case QnnBackendType::CPU:
+      return "cpu";
+    case QnnBackendType::GPU:
+      return "gpu";
+    case QnnBackendType::DSP:
+      return "dsp";
+    case QnnBackendType::HTP:
+      return "htp";
+    case QnnBackendType::HTP_FP16:
```
huaychou, line 660: Do we have HTP_FP16 specific backend?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1153,7 +1180,8 @@ static void LogNodeSupport(const Ort::Logger& logger,
 OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
                                     const std::unordered_map<const OrtNode*, const OrtNodeUnit*>& node_unit_map,
                                     const size_t node_unit_size,
-                                    std::vector<const OrtNode*>& supported_nodes) const {
+                                    std::vector<const OrtNode*>& supported_nodes,
+                                    InlinedVector<qnn::UnsupportedNodeInfo, 16>* unsupported_nodes) const {
```
huaychou: Just wondering, how much memory can InlinedVector save?

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
```diff
@@ -0,0 +1,201 @@
...
+  // initializers that flow through unchanged) are NOT recorded — they carry
+  // no information for downstream tools and would inflate the trace with
+  // zero-information entries for every graph input and weight tensor.
+  //
+  // Output tensors of QNN ops are recorded by RecordOpMapping (with the op's
+  // full srcInfo chain) and are therefore also excluded here to avoid duplicates.
+  const auto& tensors_map = wrapper.GetModelTensorsMap();
+  for (const auto& [key, tensor_wrapper] : tensors_map) {
+    const std::string& onnx_name = tensor_wrapper.GetName();
+    const std::string& qnn_name = tensor_wrapper.GetResolvedTensorName();
```
huaychou, line 159: Do we have an UT for offload_io_quantization with multiple inputs like the corner case in https://github.com/onnxruntime/onnxruntime-qnn/pull/295?

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -227,6 +236,9 @@ class QnnEp : public OrtEp, public ApiPtrs {
   mutable std::shared_ptr<GenieApiLoader> genie_api_loader_;
   GenieLog_Level_t genie_log_level_ = GENIE_LOG_LEVEL_ERROR;
   mutable std::atomic<uint64_t> genie_kv_cache_rewind_{1};
+
+  void CollectAndWriteFrameworkOpTrace(const OrtGraph* const* graphs,
```
minfhong-qti: Move to the section at LN118.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1973,6 +2022,47 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
...
+  // Only write when there are actual QNN subgraph traces.
+  // An empty trace occurs during EPContext loading (Phase 2) or any path
+  // that produces no compiled subgraphs — skip the file write but continue.
+  if (!trace.subgraph_traces.empty()) {
+    namespace fs = std::filesystem;
+    fs::path trace_dir = framework_op_trace_dir_.empty()
+                             ? fs::current_path()
+                             : fs::path(framework_op_trace_dir_);
+
+    std::string filename = "qnn_op_trace_" + std::to_string(framework_op_trace_count_++) + ".json";
```
minfhong-qti: Is this counter still needed considering there is no concurrent `GetCapability`?

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -204,6 +207,12 @@ class QnnEp : public OrtEp, public ApiPtrs {
 
   bool dump_json_qnn_graph_ = false;
   std::string json_qnn_graph_dir_ = "";
+  bool enable_framework_op_trace_ = false;
```
minfhong-qti, line 217: How about move these to a separate section commented with op trace-related.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2074,13 +2168,21 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
         /*tensor_name_overrides=*/&ep->tensor_name_overrides_,
         /*json_qnn_graph_path=*/std::string{}};
 
+    qnn::OpTraceInfo op_trace;
+    if (ep->enable_framework_op_trace_) {
+      context.op_trace_output = &op_trace;
+    }
```
```suggestion
    if (ep->enable_framework_op_trace_) {
      subgraph_traces.push_back(qnn::OpTraceInfo{});
      context.op_trace_output = &subgraph_traces.back();
    }
```
And no need LN2183-2185.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1973,6 +2022,47 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
   return nullptr;
 }
 
+void QnnEp::CollectAndWriteFrameworkOpTrace(const OrtGraph* const* graphs,
+                                            std::vector<qnn::OpTraceInfo> subgraph_traces) {
```
```suggestion
                                            std::vector<qnn::OpTraceInfo>&& subgraph_traces) {
```
Make it explicit.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1973,6 +2022,47 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
...
+  if (trace.model_name.empty()) {
+    trace.model_name = "<in-memory>";
+  }
+  trace.backend_type = qnn::QnnBackendTypeToString(qnn_backend_manager_->GetQnnBackendType());
+  trace.compilation_target.device_id = device_id_;
+  trace.compilation_target.soc_model = qnn_backend_manager_->GetSocModel();
+
+  if (qnn::IsNpuBackend(qnn_backend_manager_->GetQnnBackendType())) {
+    QnnHtpDevice_Arch_t htp_arch = qnn_backend_manager_->GetHtpArch();
+    if (htp_arch != QNN_HTP_DEVICE_ARCH_NONE) {
```
minfhong-qti, line 2088: Just curious. Should this check as well apply to `soc_model`? Or should the check be removed?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1973,6 +2022,47 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
...
+
+  trace.subgraph_traces = std::move(subgraph_traces);
+
+  qnn::ComputeTraceSummary(trace);
+
+  // Only write when there are actual QNN subgraph traces.
+  // An empty trace occurs during EPContext loading (Phase 2) or any path
+  // that produces no compiled subgraphs — skip the file write but continue.
+  if (!trace.subgraph_traces.empty()) {
+    namespace fs = std::filesystem;
```
minfhong-qti: Remove this and use `std::filesystem`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1973,6 +2022,47 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
   return nullptr;
 }
 
+void QnnEp::CollectAndWriteFrameworkOpTrace(const OrtGraph* const* graphs,
+                                            std::vector<qnn::OpTraceInfo> subgraph_traces) {
+  qnn::FrameworkOpTrace trace;
```
minfhong-qti: Considering that there is no concurrent usecase and this EP is dedicated to one model, how about set `qnn::FrameworkOpTrace trace` as EP member and collect necessary info along the way? And there is no need to record `unsupported_nodes` nor pass `subgraph_traces` in this function.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -61,6 +62,22 @@ std::vector<std::string> ResolveGraphInputOutputOrder(
   return resolved_onnx_order;
 }
 
+// RAII guard that brackets SetCurrentNodeGroup / ClearCurrentNodeGroup around one
+// AddToModelBuilder call so that ClearCurrentNodeGroup is always called even if
+// AddToModelBuilder returns early via an exception.
+struct NodeGroupGuard {
```
minfhong-qti: If this is required by tracing, should it be defined in tracing files?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -295,6 +322,12 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
   const bool build_json_graph = !context.json_qnn_graph_path.empty();
   RETURN_IF_NOT(qnn_model_wrapper.ComposeQnnGraph(build_json_graph), "Failed to compose Qnn graph.");
 
+  // Collect framework op trace after graph composition
+  if (trace_collector) {
+    trace_collector->RecordTensorMappings(qnn_model_wrapper);
+    *context.op_trace_output = trace_collector->GetOpTraceInfo(graph_name);
```
minfhong-qti:
```suggestion
    trace_collector->GetOpTraceInfo(graprh_name, qnn_model_wrapper, *context.op_trace_output);
```
How about combine these two functions as one as I don't think they can be separately used?

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
```diff
@@ -0,0 +1,201 @@
...
+}
+
+void OpTraceCollector::ClearCurrentNodeGroup() {
+  current_node_group_ = nullptr;
+}
+
+void OpTraceCollector::RecordOpMapping(const std::string& qnn_op_name, const std::string& qnn_op_type,
+                                       const std::vector<std::string>& output_tensor_names) {
+  // Guard: node group context is set by NodeGroupGuard around each AddToModelBuilder call.
+  // If nullptr, this QNN op was created outside that bracket (e.g. I/O tensor creation),
```
minfhong-qti: Is there really QNN op created during I/O tensor creation?

cmake/onnxruntime_unittests.cmake
kuanyul-qti: It would expand the EP DLL's public ABI just for tests. Current exports are limited to CreateEpFactories / ReleaseEpFactory (onnxruntime/core/providers/qnn/version_script.lds, symbols.def). The cmake comment documents this rationale.

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing_serialization.cc
kuanyul-qti: Translation unit. In C++, a translation unit is the actual unit of compilation: a single .cpp / .cc source file after the preprocessor has run — meaning #includes expanded, macros substituted, conditional #if branches resolved, and comments stripped. The compiler then translates that single combined token stream into one object file (.o / .obj).

onnxruntime/core/providers/qnn/builder/qnn_def.cc
kuanyul-qti: HTP_FP16 is in the enum (`qnn_def.h`) but never assigned anywhere in this codebase. The case `QnnBackendType::HTP_FP16` exists only to satisfy -Wswitch; it is effectively unreachable.

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
kuanyul-qti: Updated comment.

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.cc
kuanyul-qti: Yes, the test name is `FrameworkOpTrace_OffloadGraphIoQuantization_FanoutQDQ`.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: Moved to `qnn_op_tracing.h`.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: keep parameter as `std::vector`. The InlinedVector saving (~one heap alloc per call when <= N unsupported nodes) is below noise floor on a debug-only path.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: `subgraph_traces` is removed in the latest commit.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Add `soc_model` record.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Removed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1222,6 +1256,28 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
           supported_nodes.push_back(node);
         }
       }
+    } else if (unsupported_nodes) {
+      // Most node groups have <= 8 raw nodes (a QDQ OrtNodeUnit has 3-4 nodes;
+      // complex fusions like GeluFusion span 5-6 OrtNodeUnits of 1 node each),
+      // and tracing-only paths run rarely, so a plain std::vector batch is
+      // sufficient.
+      std::vector<qnn::UnsupportedNodeInfo> batch;
```
```suggestion
      // Most node groups have <= 8 raw nodes (a QDQ OrtNodeUnit has 3-4 nodes;
      // complex fusions like GeluFusion span 5-6 OrtNodeUnits of 1 node each),
      // and tracing-only paths run rarely, so a plain std::vector batch is
      // sufficient.
      std::vector<qnn::UnsupportedNodeInfo> batch;
      batch.reserve(8);
```
Not sure the comment here still means anything. If it would like to emphasize the number of nodes in a nodeunit is small enough, would it be better to add this to address the comment?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1222,6 +1256,28 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
           supported_nodes.push_back(node);
         }
       }
+    } else if (unsupported_nodes) {
```
```suggestion
    } else if (enable_framework_op_trace_) {
```
And `unsupported_nodes` can be passed by reference. and no need to check at the caller.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1973,6 +2051,77 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
   return nullptr;
 }
 
+void QnnEp::ProbeFrameworkOpTraceDir() {
```
minfhong-qti: Should this be check during parsing option value instead of here? And in that case, I think this can be static function instead of class member function.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1650,6 +1707,16 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
 
   // If no supported nodes, return empty
   if (supported_nodes.empty()) {
+    // ORT does not invoke CompileImpl when no nodes are supported, so the
+    // unsupported list collected in GetSupportedNodes would otherwise be
+    // discarded. Flush a trace here so the diagnostic record (why every node
+    // was rejected) is preserved.
+    if (ep->enable_framework_op_trace_ && !ep->trace_.unsupported_nodes.empty()) {
+      ep->ProbeFrameworkOpTraceDir();
+      if (ep->enable_framework_op_trace_) {
```
minfhong-qti: This check is redundant.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1973,6 +2051,77 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
...
+  }
+
+  // A non-writable path (read-only Android mount, read-only network share,
+  // missing parent permissions) would otherwise only surface as a WARNING
+  // after the entire compile + in-memory trace build finishes. Probing here
+  // (before any compile work has begun) lets us disable tracing early so the
+  // work is not wasted.
+  std::filesystem::path probe_dir = framework_op_trace_dir_.empty()
+                                        ? std::filesystem::current_path()
+                                        : std::filesystem::path(framework_op_trace_dir_);
```
minfhong-qti: It seems this logic presents here and below during writing. Should it be directly determined during paring option value?

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.h
```diff
@@ -0,0 +1,79 @@
...
+ private:
+  // Set/clear the current node group context. Only NodeGroupGuard may call these
+  // to preserve the set/clear pairing and exception-safety invariant.
+  friend class NodeGroupGuard;
+  void SetCurrentNodeGroup(const IQnnNodeGroup* node_group);
+  void ClearCurrentNodeGroup();
+
+  const IQnnNodeGroup* current_node_group_ = nullptr;
+  InlinedVector<TraceMapping, 32> op_mappings_;
+  InlinedVector<TraceMapping, 32> tensor_mappings_;
```
minfhong-qti: Just curious what 32 here mean? If it is number of subgraphs, then I think it may be too small as I've seen 99 subgraphs case.

kuanyul-qti: 32 is the inline capacity for op/tensor mappings within a single subgraph, not a capacity on the number of subgraphs. Each `OpTraceCollector` instance is constructed fresh in a single `QnnModel::ComposeGraph()` call (one per subgraph). The total subgraph count is held by `FrameworkOpTrace::subgraph_traces`, which is `std::vector` with no fixed limit — 99 subgraphs is fine.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Remove the outdated comment.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Removed. And move `ProbeFrameworkOpTraceDir` check to the value parsing stage.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Moved `ProbeFrameworkOpTraceDir` check to the value parsing stage. Since this validation now occurs at the initial stage, the function is no longer necessary.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Moved to value parsing stage.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: Reopen this thread.
I mean should we only set `soc_model` if the returned one is not `QNN_SOC_MODEL_UNKNOWN`?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1941,6 +2033,11 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
                                          size_t count,
                                          OrtNodeComputeInfo** node_compute_infos) {
   QnnEp* ep = static_cast<QnnEp*>(this_ptr);
+  if (ep->enable_framework_op_trace_) {
+    ORT_CXX_LOG(ep->logger_, ORT_LOGGING_LEVEL_INFO,
+                "Framework op tracing requested but graph is loaded from EPContext "
+                "(no fresh ComposeGraph) - no trace file will be written.");
+  }
```
minfhong-qti, line 2042:
```suggestion
  }

```
Nit.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -217,6 +217,15 @@ Alternatively to setting profiling_level at compile time, profiling can be enabl
...
 
+|`"enable_framework_op_trace"`|Description|
+|---|---|
+|'0'|Default. Disabled.|
+|'1'|Enable framework op tracing. Writes a JSON file that maps each ONNX operator (or fused group) to its corresponding QNN operator(s), records unsupported operators, and includes summary statistics. See `framework_op_trace_dir` to control the output location. The trace is produced from the QNN graph composition path: it is emitted on a fresh JIT compile and on AOT context-binary generation (Phase 1, when `ep.context_enable=1` writes a new EPContext model). It is **not** emitted when loading an existing EPContext model (Phase 2 / cached run), because no graph composition occurs in that path; in that case an INFO log records that no trace file is written.|
+
+|`"framework_op_trace_dir"`|Description|
+|---|---|
+|Directory path (string)|Directory path for framework op trace JSON files. Only effective when `enable_framework_op_trace` is `'1'`. Defaults to the current working directory if not set. The directory is created automatically if it does not exist. Trace files use the pattern `qnn_op_trace.json`.|
```
minfhong-qti: Should we add a description regarding that if the provided directory is not writtable, trace will be disabled?

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -217,6 +217,15 @@ Alternatively to setting profiling_level at compile time, profiling can be enabl
 |---|---|
 |Directory path (string)|Directory path for dumping QNN IR DLC files. Only effective when `dump_qnn_ir_dlc` is enabled.|
 
+|`"enable_framework_op_trace"`|Description|
+|---|---|
+|'0'|Default. Disabled.|
+|'1'|Enable framework op tracing. Writes a JSON file that maps each ONNX operator (or fused group) to its corresponding QNN operator(s), records unsupported operators, and includes summary statistics. See `framework_op_trace_dir` to control the output location. The trace is produced from the QNN graph composition path: it is emitted on a fresh JIT compile and on AOT context-binary generation (Phase 1, when `ep.context_enable=1` writes a new EPContext model). It is **not** emitted when loading an existing EPContext model (Phase 2 / cached run), because no graph composition occurs in that path; in that case an INFO log records that no trace file is written.|
```
minfhong-qti: Is the term `Phase 1` and `Phase 2` widely used in the doc? If not, I prefer remove these two.

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.h
minfhong-qti: Wouldn't 32 be too small? It seems op/tensor mappings can easily exceed 32 in a subgraph.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Got it. Fixed. Guard the assignment behind `soc_model != QNN_SOC_MODEL_UNKNOWN`.

onnxruntime/core/providers/qnn/builder/op_tracing/qnn_op_tracing.h
kuanyul-qti: Switched both `op_mappings_` and `tensor_mappings_` from `InlinedVector<TraceMapping, 32>` to plain `std::vector<TraceMapping>`. Real subgraphs (LLMs, etc.) blow past any small inline capacity, so the inline buffer rarely helps and just inflates the on-stack `OpTraceCollector` footprint. Tracing is debug-only, so the extra heap allocation is below noise floor.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done

docs/execution_providers/QNN-ExecutionProvider.md
kuanyul-qti: Appended the description to the framework_op_trace_dir row.

docs/execution_providers/QNN-ExecutionProvider.md
kuanyul-qti: Removed "Phase 1" / "Phase 2" string.

minfhong-qti (approved):

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
```diff
@@ -2245,7 +2247,8 @@ index 71f9050730..1e879bf5df 100644
 +                             "qnn_saver_path", "htp_graph_finalization_optimization_mode", "qnn_context_priority",
 +                             "htp_arch", "enable_htp_fp16_precision", "offload_graph_io_quantization",
 +                             "enable_htp_spill_fill_buffer", "enable_htp_shared_memory_allocator", "dump_json_qnn_graph",
-+                             "json_qnn_graph_dir", "disable_file_mapped_weights", "htp_bf16_enable", "enable_vtcm_backup_buffer_sharing", "extended_udma", "htp_share_resource_optimization"});
++                             "json_qnn_graph_dir", "disable_file_mapped_weights", "htp_bf16_enable", "enable_vtcm_backup_buffer_sharing", "extended_udma", "htp_share_resource_optimization",
++                             "enable_framework_op_trace", "framework_op_trace_dir"});
```
tirupath-qti: onnxrunitme_perf_test can accept arbitrary key value arguments defined in plugin EP. We don't need to make changes to onnxruntime_perf_test app.

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
```diff
@@ -1135,7 +1136,8 @@ index 0000000000..4727e56596
 +              "Wrong key type entered. Choose from options: ['backend_type', 'backend_path', "
 +              "'profiling_level', 'profiling_file_path', 'rpc_control_latency', 'vtcm_mb', 'htp_performance_mode', "
 +              "'qnn_saver_path', 'htp_graph_finalization_optimization_mode', 'op_packages', 'qnn_context_priority', "
-+              "'soc_model', 'htp_arch', 'device_id', 'enable_htp_fp16_precision', 'offload_graph_io_quantization']");
++              "'soc_model', 'htp_arch', 'device_id', 'enable_htp_fp16_precision', 'offload_graph_io_quantization', "
++              "'enable_framework_op_trace', 'framework_op_trace_dir']");
```
tirupath-qti: do we still need changes to this file?

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
kuanyul-qti: Removed.

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
kuanyul-qti: Removed.

qti-chuteng (approved):

PR #346, opened 2026-05-06
Implement graph checker with public GetEpGraphAssignmentInfo API

qti-hungjuiw:
### Description

- `EPVerificationParams::graph_verifier` in `test_utils.h` now accepts const `Ort::Session&` instead of `const Graph&`
- `RunQnnModelTest`, `TestQDQModelAccuracy`, and `InferenceModel` signatures updated             accordingly
- The previously dead-code checker invocation (commented out with a TODO) is now active
- TestModelLoad's check_graph callback similarly updated
- All three test-side checkers (AveragePool1DFusedQnnNodePresent,  MaxPool1D_ReshapeNodesPresent, EPOffloadsGraphIOQuantDequant, EPRejectsDynamicShapesF32) rewritten using `Session::GetEpGraphAssignmentInfo() → ConstEpAssignedSubgraph / ConstEpAssignedNode`
- Removed `#include "core/graph/graph.h"` from `qnn_test_utils.cc`

### Motivation and Context

- The QNN EP tests were depending on `core/graph/graph.h` (an ORT-internal header) via `std::function<void(const Graph&)> callbacks`. This creates a coupling to internal ORT types that should be avoided as the EP API moves toward a fully public interface.                                                                                                    
- The checker callbacks were also entirely non-functional — the call site in
  `RunAndVerifyOutputsWithEP` had been commented out with a TODO: Implement graph_checker      once public API for ep partition is ready. With `Session::GetEpGraphAssignmentInfo()` now available in the public C++ API, we have everything needed to implement the same verification without internal headers.

### DIFF
--- a/onnxruntime/test/providers/qnn/averagepool_test.cc
+++ b/onnxruntime/test/providers/qnn/averagepool_test.cc
@@ -275,9 +275,16 @@ TEST_F(QnnHTPBackendTests, AveragePool1DFusedQnnNodePresent) {
   ProviderOptions options;
   options["backend_type"] = "htp";
 
-  std::function<void(const Graph&)> check_num_nodes = [](const Graph& graph) {
-    int number_of_nodes = graph.NumberOfNodes();
-    EXPECT_EQ(number_of_nodes, 1) << "Expected 1 fused QNN node for AveragePool rank-3 input.";
+  std::function<void(const Ort::Session&)> check_num_nodes = [](const Ort::Session& session) {
+    // The Reshape -> Pool -> Reshape gets fused to a single QNN node, so there should be
+    // exactly 1 QNN EP subgraph.
+    size_t num_qnn_subgraphs = 0;
+    for (const auto& subgraph : session.GetEpGraphAssignmentInfo()) {
+      if (subgraph.GetEpName() == kQnnExecutionProvider) {
+        num_qnn_subgraphs++;
+      }
+    }
+    EXPECT_EQ(num_qnn_subgraphs, 1u) << "Expected 1 fused QNN node for AveragePool rank-3 input.";
   };
 
   RunQnnModelTest(build_test_case,
--- a/onnxruntime/test/providers/qnn/maxpool_test.cc
+++ b/onnxruntime/test/providers/qnn/maxpool_test.cc
@@ -243,10 +243,16 @@ TEST_F(QnnHTPBackendTests, MaxPool1D_ReshapeNodesPresent) {
   ProviderOptions options;
   options["backend_type"] = "htp";
 
-  std::function<void(const Graph&)> check_num_nodes = [](const Graph& graph) {
-    int number_of_nodes = graph.NumberOfNodes();
-    // The Reshape -> Pool -> Reshape gets fused to a single QNN node
-    EXPECT_EQ(number_of_nodes, 1) << "Expected 1 QNN fused node for MaxPool rank-3 input.";
+  std::function<void(const Ort::Session&)> check_num_nodes = [](const Ort::Session& session) {
+    // The Reshape -> Pool -> Reshape gets fused to a single QNN node, so there should be
+    // exactly 1 QNN EP subgraph.
+    size_t num_qnn_subgraphs = 0;
+    for (const auto& subgraph : session.GetEpGraphAssignmentInfo()) {
+      if (subgraph.GetEpName() == kQnnExecutionProvider) {
+        num_qnn_subgraphs++;
+      }
+    }
+    EXPECT_EQ(num_qnn_subgraphs, 1u) << "Expected 1 QNN fused node for MaxPool rank-3 input.";
   };
 
   RunQnnModelTest(build_test_case,
--- a/onnxruntime/test/providers/qnn/qnn_basic_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_basic_test.cc
@@ -1350,11 +1350,9 @@ TEST_F(QnnHTPBackendTests, EPRejectsDynamicShapesF32) {
       for (const auto& node : subgraph.GetNodes()) {
         std::string op_type = node.GetOperatorType();
         if (op_type == "Reshape" || op_type == "Softmax") {
-          EXPECT_NE(ep_name, kQnnExecutionProvider)
-              << op_type << " should NOT be assigned to QNN EP";
+          EXPECT_EQ(ep_name, kCpuExecutionProvider) << op_type << " should be assigned to CPU EP";
         } else {
-          EXPECT_EQ(ep_name, kQnnExecutionProvider)
-              << op_type << " should be assigned to QNN EP";
+          EXPECT_EQ(ep_name, kQnnExecutionProvider) << op_type << " should be assigned to QNN EP";
         }
       }
     }
@@ -1441,10 +1439,22 @@ TEST_F(QnnHTPBackendTests, EPOffloadsGraphIOQuantDequant) {
     return [offload_graph_io_quantization](const Ort::Session& session) {
       // The public API returns pre-fusion nodes grouped by EP subgraph.
       // We verify:
-      //   offload=0: exactly 1 QNN subgraph (all nodes fused), 0 CPU nodes
-      //   offload=1: exactly 1 QNN subgraph (inner DQ+Op+Q fused), 2 CPU nodes (boundary Q + DQ)
+      //   offload=0: exactly 1 QNN subgraph (all QNN-eligible ops grouped), 0 CPU nodes
+      //   offload=1: exactly 1 QNN subgraph (all QNN-eligible ops grouped), 2 CPU nodes (boundary Q + DQ)
       size_t num_qnn_subgraphs = 0;
       size_t num_cpu_nodes = 0;
+      std::vector<std::string> cpu_op_types;
+
+      // For offload=1 verify the CPU Q/DQ are the graph-IO boundary nodes, not interior ones.
+      // ConstEpAssignedNode exposes only GetName/GetDomain/GetOperatorType — no input/output
+      // tensor name accessors — so we proxy the boundary check via the node names assigned by
+      // AddQDQNodePair / AddQDQNodePairWithOutputAsGraphOutput in BuildQDQOpTestCase:
+      //   input-side boundary Q   → "qdq_in0_q"  (consumes graph input "quant_input_defs_0")
+      //   output-side boundary DQ → "qdq_out_dq" (produces graph output "qdq_out_dq_out")
+      // Interior counterparts are "qdq_in0_dq" and "qdq_out_q"; ending up on CPU would mean
+      // the partitioner sent the wrong nodes to the CPU EP.
+      bool found_boundary_q = false;
+      bool found_boundary_dq = false;
 
       std::vector<Ort::ConstEpAssignedSubgraph> subgraphs = session.GetEpGraphAssignmentInfo();
       for (const auto& subgraph : subgraphs) {
@@ -1454,21 +1464,37 @@ TEST_F(QnnHTPBackendTests, EPOffloadsGraphIOQuantDequant) {
         } else {
           // CPU EP should only receive Q/DQ boundary nodes when offloading is enabled.
           for (const auto& node : subgraph.GetNodes()) {
+            std::string op_type = node.GetOperatorType();
+            std::string node_name = node.GetName();
+            cpu_op_types.push_back(op_type);
             if (offload_graph_io_quantization) {
-              std::string op_type = node.GetOperatorType();
               EXPECT_TRUE(op_type == "QuantizeLinear" || op_type == "DequantizeLinear")
                   << op_type << " should not be on CPU EP when IO quantization offloading is enabled";
+              if (op_type == "QuantizeLinear") {
+                found_boundary_q = (node_name == "qdq_in0_q");
+              } else if (op_type == "DequantizeLinear") {
+                found_boundary_dq = (node_name == "qdq_out_dq");
+              }
             }
             num_cpu_nodes++;
           }
         }
       }
 
-      EXPECT_EQ(num_qnn_subgraphs, 1u) << "Expected all QNN-assigned nodes fused into 1 subgraph";
+      EXPECT_EQ(num_qnn_subgraphs, 1u) << "Expected all QNN-assigned nodes grouped into 1 subgraph";
       if (offload_graph_io_quantization) {
+        const std::vector<std::string> graph_inputs = session.GetInputNames();
+        const std::vector<std::string> graph_outputs = session.GetOutputNames();
         EXPECT_EQ(num_cpu_nodes, 2u) << "Expected 2 boundary Q/DQ nodes offloaded to CPU EP";
+        EXPECT_TRUE(found_boundary_q)
+            << "Expected input-side boundary Q (qdq_in0_q) on CPU EP consuming a graph input; "
+            << "graph inputs: " << testing::PrintToString(graph_inputs);
+        EXPECT_TRUE(found_boundary_dq)
+            << "Expected output-side boundary DQ (qdq_out_dq) on CPU EP producing a graph output; "
+            << "graph outputs: " << testing::PrintToString(graph_outputs);
       } else {
-        EXPECT_EQ(num_cpu_nodes, 0u) << "Expected no CPU nodes when IO quantization offloading is disabled";
+        EXPECT_EQ(num_cpu_nodes, 0u) << "Expected no CPU nodes when IO quantization offloading is disabled, "
+                                     << "got: " << testing::PrintToString(cpu_op_types);
       }
     };
   };
--- a/onnxruntime/test/providers/qnn/qnn_basic_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_basic_test.cc
@@ -1343,14 +1343,19 @@ TEST_F(QnnHTPBackendTests, EPRejectsDynamicShapesF32) {
   };
 
   // Local function that checks that the nodes with dynamic shape I/O were assigned to CPU EP.
-  std::function<void(const Graph&)> ep_graph_checker = [](const Graph& graph) {
-    for (const Node& node : graph.Nodes()) {
-      const std::string& ep_name = node.GetExecutionProviderType();
-      const std::string& op_type = node.OpType();
-      if (op_type == "Reshape" || op_type == "Softmax") {
-        EXPECT_EQ(ep_name, kCpuExecutionProvider);
-      } else {
-        EXPECT_TRUE((ep_name == kQnnExecutionProvider) || (ep_name == onnxruntime::kQnnExecutionProvider));
+  std::function<void(const Ort::Session&)> ep_graph_checker = [](const Ort::Session& session) {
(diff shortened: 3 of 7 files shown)


onnxruntime/test/providers/qnn/averagepool_test.cc
```diff
@@ -275,9 +275,16 @@ TEST_F(QnnHTPBackendTests, AveragePool1DFusedQnnNodePresent) {
...
+  std::function<void(const Ort::Session&)> check_num_nodes = [](const Ort::Session& session) {
+    // The Reshape -> Pool -> Reshape gets fused to a single QNN node, so there should be
+    // exactly 1 QNN EP subgraph.
+    size_t num_qnn_subgraphs = 0;
+    for (const auto& subgraph : session.GetEpGraphAssignmentInfo()) {
+      if (subgraph.GetEpName() == kQnnExecutionProvider) {
+        num_qnn_subgraphs++;
+      }
+    }
+    EXPECT_EQ(num_qnn_subgraphs, 1u) << "Expected 1 fused QNN node for AveragePool rank-3 input.";
```
minfhong-qti, line 287: Just curious why this test checks this.

onnxruntime/test/providers/qnn/maxpool_test.cc
```diff
@@ -242,10 +242,16 @@ TEST_F(QnnHTPBackendTests, MaxPool1D_ReshapeNodesPresent) {
   ProviderOptions options;
   options["backend_type"] = "htp";
 
-  std::function<void(const Graph&)> check_num_nodes = [](const Graph& graph) {
-    int number_of_nodes = graph.NumberOfNodes();
-    // The Reshape -> Pool -> Reshape gets fused to a single QNN node
-    EXPECT_EQ(number_of_nodes, 1) << "Expected 1 QNN fused node for MaxPool rank-3 input.";
+  std::function<void(const Ort::Session&)> check_num_nodes = [](const Ort::Session& session) {
+    // The Reshape -> Pool -> Reshape gets fused to a single QNN node, so there should be
```
minfhong-qti, line 247: Is there really Reshape node? I do not see above.

qti-hungjuiw: I think this is introduced by the op builder in this PR:
https://github.com/microsoft/onnxruntime/pull/24735

minfhong-qti: Got it.

onnxruntime/test/providers/qnn/averagepool_test.cc
minfhong-qti: Keep it for now.

minfhong-qti (approved):

qti-chuteng:
---

**[M-1]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1345-1360`

`EPRejectsDynamicShapesF32` weakens the assertion from `EXPECT_EQ(ep_name, kCpuExecutionProvider)` to `EXPECT_NE(ep_name, kQnnExecutionProvider)`. If the runtime omits the fallback CPU subgraph from the returned list (which is plausible — the API name is "EpGraphAssignmentInfo" and fallback CPU may not be a true EP subgraph), the inner loop runs zero times and the test passes vacuously. Suggest counting `reshape_seen` and `softmax_seen` and asserting both `>= 1u` after the loop, and keeping the strong `EXPECT_EQ(ep_name, kCpuExecutionProvider)` for those branches:

```cpp
size_t reshape_seen = 0, softmax_seen = 0;
for (const auto& subgraph : subgraphs) {
  std::string ep_name = subgraph.GetEpName();
  for (const auto& node : subgraph.GetNodes()) {
    std::string op_type = node.GetOperatorType();
    if (op_type == "Reshape")  { reshape_seen++; EXPECT_EQ(ep_name, kCpuExecutionProvider); }
    else if (op_type == "Softmax") { softmax_seen++; EXPECT_EQ(ep_name, kCpuExecutionProvider); }
    else { EXPECT_EQ(ep_name, kQnnExecutionProvider); }
  }
}
EXPECT_EQ(reshape_seen, 1u);
EXPECT_EQ(softmax_seen, 1u);
```

---

**[M-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1436-1474`

`EPOffloadsGraphIOQuantDequant` previously verified via `graph.IsInputsIncludingInitializers()` / `graph.IsOutput()` that the offloaded Q consumes a graph input and the offloaded DQ produces a graph output — i.e. the literal "GraphIO" semantics in the test name. The new version drops that and only count-checks 2 CPU nodes. If a partitioner regression sends *interior* Q/DQ to CPU instead of boundary Q/DQ, the count would still be 2 and the test would still pass while the contract is broken. Recommend collecting graph IO names from the `Ort::Session` (`GetInputCount`/`GetOutputCount` + `GetInputNameAllocated`/`GetOutputNameAllocated`) and asserting the CPU-side Q's input is in `graph_inputs` and the DQ's output is in `graph_outputs`.

---

**[M-3]** `onnxruntime/test/util/test_utils.cc:284-301`

`RunAndVerifyOutputsWithEP` and `TestModelLoad` are cross-EP shared entry points. They now invoke `params.graph_verifier(ort_session)` / `check_graph(session_object)`, but they do not enforce that the caller's `SessionOptions` carry `kOrtSessionOptionsRecordEpGraphAssignmentInfo`. Non-QNN callers that pass a `graph_verifier` without the flag will receive an empty subgraph list and the verifier silently becomes a no-op. The QNN path sets the flag in `qnn_test_utils.cc:338, 438`, so QNN tests are safe — but the cross-EP utility should either auto-set the flag when `params.graph_verifier != nullptr`, or document the precondition in the header so non-QNN callers aren't caught off guard. Suggested:

```cpp
if (params.graph_verifier) {
  ort_so.AddConfigEntry(kOrtSessionOptionsRecordEpGraphAssignmentInfo, "1");
}
```

---

**[N-1]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1457-1463`

When `offload_graph_io_quantization=false`, the new callback only counts CPU nodes and doesn't record their op_types. If `num_cpu_nodes != 0`, the failure message just says "Expected no CPU nodes" without naming the misassigned op. Consider collecting the op_type list and including it in the failure message via `testing::PrintToString(cpu_op_types)`.

---

**[N-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1442-1445`

The callback comment states `// offload=1: exactly 1 QNN subgraph (inner DQ+Op+Q fused), 2 CPU nodes (boundary Q + DQ)`, but the assertions only check the QNN subgraph count and CPU node count — they don't verify the inner DQ+Op+Q is actually fused. Either add a node-count assertion on the QNN subgraph (pre-fusion should be 3 nodes: DQ, Op, Q — note that `GetEpGraphAssignmentInfo` returns pre-fusion nodes), or soften the comment to "all QNN-eligible ops grouped into 1 subgraph" so it doesn't imply fusion is verified.

---

**[T-1]** `onnxruntime/test/providers/qnn/averagepool_test.cc:278-285`, `maxpool_test.cc:245-252`

The two callbacks are nearly identical. Consider extracting a helper into `qnn_test_utils.h` in a follow-up:

```cpp
size_t CountAssignedSubgraphs(const Ort::Session& session, const std::string& ep_name);
```

---

**[T-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1352-1357`

The original code accepted both `ep_name == kQnnExecutionProvider` and `ep_name == onnxruntime::kQnnExecutionProvider`. The new code only compares against `kQnnExecutionProvider`. If both ABI and non-ABI EP can coexist with different namespaced string values, the simplification could misclassify. Could you confirm in the PR description that on the public-API path `subgraph.GetEpName()` always matches `kQnnExecutionProvider` exactly?

---

**[T-3]** `onnxruntime/test/util/include/test_utils.h`

Observation only — with the migration complete, `test_utils.h` no longer needs the `class Graph;` forward declaration (correctly removed), but the file still wraps everything in `namespace onnxruntime { namespace test { ... } }` which is now somewhat redundant. Out of scope for this PR; flagging as a follow-up cleanup candidate.

qti-hungjuiw:
> **[M-1]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1345-1360`
> 
> `EPRejectsDynamicShapesF32` weakens the assertion from `EXPECT_EQ(ep_name, kCpuExecutionProvider)` to `EXPECT_NE(ep_name, kQnnExecutionProvider)`. If the runtime omits the fallback CPU subgraph from the returned list (which is plausible — the API name is "EpGraphAssignmentInfo" and fallback CPU may not be a true EP subgraph), the inner loop runs zero times and the test passes vacuously. Suggest counting `reshape_seen` and `softmax_seen` and asserting both `>= 1u` after the loop, and keeping the strong `EXPECT_EQ(ep_name, kCpuExecutionProvider)` for those branches:
> 
> ```c++
> size_t reshape_seen = 0, softmax_seen = 0;
> for (const auto& subgraph : subgraphs) {
>   std::string ep_name = subgraph.GetEpName();
>   for (const auto& node : subgraph.GetNodes()) {
>     std::string op_type = node.GetOperatorType();
>     if (op_type == "Reshape")  { reshape_seen++; EXPECT_EQ(ep_name, kCpuExecutionProvider); }
>     else if (op_type == "Softmax") { softmax_seen++; EXPECT_EQ(ep_name, kCpuExecutionProvider); }
>     else { EXPECT_EQ(ep_name, kQnnExecutionProvider); }
>   }
> }
> EXPECT_EQ(reshape_seen, 1u);
> EXPECT_EQ(softmax_seen, 1u);
> ```
> 
> **[M-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1436-1474`
> 
> `EPOffloadsGraphIOQuantDequant` previously verified via `graph.IsInputsIncludingInitializers()` / `graph.IsOutput()` that the offloaded Q consumes a graph input and the offloaded DQ produces a graph output — i.e. the literal "GraphIO" semantics in the test name. The new version drops that and only count-checks 2 CPU nodes. If a partitioner regression sends _interior_ Q/DQ to CPU instead of boundary Q/DQ, the count would still be 2 and the test would still pass while the contract is broken. Recommend collecting graph IO names from the `Ort::Session` (`GetInputCount`/`GetOutputCount` + `GetInputNameAllocated`/`GetOutputNameAllocated`) and asserting the CPU-side Q's input is in `graph_inputs` and the DQ's output is in `graph_outputs`.
> 
> **[M-3]** `onnxruntime/test/util/test_utils.cc:284-301`
> 
> `RunAndVerifyOutputsWithEP` and `TestModelLoad` are cross-EP shared entry points. They now invoke `params.graph_verifier(ort_session)` / `check_graph(session_object)`, but they do not enforce that the caller's `SessionOptions` carry `kOrtSessionOptionsRecordEpGraphAssignmentInfo`. Non-QNN callers that pass a `graph_verifier` without the flag will receive an empty subgraph list and the verifier silently becomes a no-op. The QNN path sets the flag in `qnn_test_utils.cc:338, 438`, so QNN tests are safe — but the cross-EP utility should either auto-set the flag when `params.graph_verifier != nullptr`, or document the precondition in the header so non-QNN callers aren't caught off guard. Suggested:
> 
> ```c++
> if (params.graph_verifier) {
>   ort_so.AddConfigEntry(kOrtSessionOptionsRecordEpGraphAssignmentInfo, "1");
> }
> ```
> 
> **[N-1]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1457-1463`
> 
> When `offload_graph_io_quantization=false`, the new callback only counts CPU nodes and doesn't record their op_types. If `num_cpu_nodes != 0`, the failure message just says "Expected no CPU nodes" without naming the misassigned op. Consider collecting the op_type list and including it in the failure message via `testing::PrintToString(cpu_op_types)`.
> 
> **[N-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1442-1445`
> 
> The callback comment states `// offload=1: exactly 1 QNN subgraph (inner DQ+Op+Q fused), 2 CPU nodes (boundary Q + DQ)`, but the assertions only check the QNN subgraph count and CPU node count — they don't verify the inner DQ+Op+Q is actually fused. Either add a node-count assertion on the QNN subgraph (pre-fusion should be 3 nodes: DQ, Op, Q — note that `GetEpGraphAssignmentInfo` returns pre-fusion nodes), or soften the comment to "all QNN-eligible ops grouped into 1 subgraph" so it doesn't imply fusion is verified.
> 
> **[T-1]** `onnxruntime/test/providers/qnn/averagepool_test.cc:278-285`, `maxpool_test.cc:245-252`
> 
> The two callbacks are nearly identical. Consider extracting a helper into `qnn_test_utils.h` in a follow-up:
> 
> ```c++
> size_t CountAssignedSubgraphs(const Ort::Session& session, const std::string& ep_name);
> ```
> 
> **[T-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1352-1357`
> 
> The original code accepted both `ep_name == kQnnExecutionProvider` and `ep_name == onnxruntime::kQnnExecutionProvider`. The new code only compares against `kQnnExecutionProvider`. If both ABI and non-ABI EP can coexist with different namespaced string values, the simplification could misclassify. Could you confirm in the PR description that on the public-API path `subgraph.GetEpName()` always matches `kQnnExecutionProvider` exactly?
> 
> **[T-3]** `onnxruntime/test/util/include/test_utils.h`
> 
> Observation only — with the migration complete, `test_utils.h` no longer needs the `class Graph;` forward declaration (correctly removed), but the file still wraps everything in `namespace onnxruntime { namespace test { ... } }` which is now somewhat redundant. Out of scope for this PR; flagging as a follow-up cleanup candidate.



> **[M-1]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1345-1360`
> 
> `EPRejectsDynamicShapesF32` weakens the assertion from `EXPECT_EQ(ep_name, kCpuExecutionProvider)` to `EXPECT_NE(ep_name, kQnnExecutionProvider)`. If the runtime omits the fallback CPU subgraph from the returned list (which is plausible — the API name is "EpGraphAssignmentInfo" and fallback CPU may not be a true EP subgraph), the inner loop runs zero times and the test passes vacuously. Suggest counting `reshape_seen` and `softmax_seen` and asserting both `>= 1u` after the loop, and keeping the strong `EXPECT_EQ(ep_name, kCpuExecutionProvider)` for those branches:
> 
> ```c++
> size_t reshape_seen = 0, softmax_seen = 0;
> for (const auto& subgraph : subgraphs) {
>   std::string ep_name = subgraph.GetEpName();
>   for (const auto& node : subgraph.GetNodes()) {
>     std::string op_type = node.GetOperatorType();
>     if (op_type == "Reshape")  { reshape_seen++; EXPECT_EQ(ep_name, kCpuExecutionProvider); }
>     else if (op_type == "Softmax") { softmax_seen++; EXPECT_EQ(ep_name, kCpuExecutionProvider); }
>     else { EXPECT_EQ(ep_name, kQnnExecutionProvider); }
>   }
> }
> EXPECT_EQ(reshape_seen, 1u);
> EXPECT_EQ(softmax_seen, 1u);
> ```
> 
> **[M-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1436-1474`
> 
> `EPOffloadsGraphIOQuantDequant` previously verified via `graph.IsInputsIncludingInitializers()` / `graph.IsOutput()` that the offloaded Q consumes a graph input and the offloaded DQ produces a graph output — i.e. the literal "GraphIO" semantics in the test name. The new version drops that and only count-checks 2 CPU nodes. If a partitioner regression sends _interior_ Q/DQ to CPU instead of boundary Q/DQ, the count would still be 2 and the test would still pass while the contract is broken. Recommend collecting graph IO names from the `Ort::Session` (`GetInputCount`/`GetOutputCount` + `GetInputNameAllocated`/`GetOutputNameAllocated`) and asserting the CPU-side Q's input is in `graph_inputs` and the DQ's output is in `graph_outputs`.
> 
> **[M-3]** `onnxruntime/test/util/test_utils.cc:284-301`
> 
> `RunAndVerifyOutputsWithEP` and `TestModelLoad` are cross-EP shared entry points. They now invoke `params.graph_verifier(ort_session)` / `check_graph(session_object)`, but they do not enforce that the caller's `SessionOptions` carry `kOrtSessionOptionsRecordEpGraphAssignmentInfo`. Non-QNN callers that pass a `graph_verifier` without the flag will receive an empty subgraph list and the verifier silently becomes a no-op. The QNN path sets the flag in `qnn_test_utils.cc:338, 438`, so QNN tests are safe — but the cross-EP utility should either auto-set the flag when `params.graph_verifier != nullptr`, or document the precondition in the header so non-QNN callers aren't caught off guard. Suggested:
> 
> ```c++
> if (params.graph_verifier) {
>   ort_so.AddConfigEntry(kOrtSessionOptionsRecordEpGraphAssignmentInfo, "1");
> }
> ```
> 
> **[N-1]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1457-1463`
> 
> When `offload_graph_io_quantization=false`, the new callback only counts CPU nodes and doesn't record their op_types. If `num_cpu_nodes != 0`, the failure message just says "Expected no CPU nodes" without naming the misassigned op. Consider collecting the op_type list and including it in the failure message via `testing::PrintToString(cpu_op_types)`.
> 
> **[N-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1442-1445`
> 
> The callback comment states `// offload=1: exactly 1 QNN subgraph (inner DQ+Op+Q fused), 2 CPU nodes (boundary Q + DQ)`, but the assertions only check the QNN subgraph count and CPU node count — they don't verify the inner DQ+Op+Q is actually fused. Either add a node-count assertion on the QNN subgraph (pre-fusion should be 3 nodes: DQ, Op, Q — note that `GetEpGraphAssignmentInfo` returns pre-fusion nodes), or soften the comment to "all QNN-eligible ops grouped into 1 subgraph" so it doesn't imply fusion is verified.
> 
> **[T-1]** `onnxruntime/test/providers/qnn/averagepool_test.cc:278-285`, `maxpool_test.cc:245-252`
> 
> The two callbacks are nearly identical. Consider extracting a helper into `qnn_test_utils.h` in a follow-up:
> 
> ```c++
> size_t CountAssignedSubgraphs(const Ort::Session& session, const std::string& ep_name);
> ```
> 
> **[T-2]** `onnxruntime/test/providers/qnn/qnn_basic_test.cc:1352-1357`
> 
> The original code accepted both `ep_name == kQnnExecutionProvider` and `ep_name == onnxruntime::kQnnExecutionProvider`. The new code only compares against `kQnnExecutionProvider`. If both ABI and non-ABI EP can coexist with different namespaced string values, the simplification could misclassify. Could you confirm in the PR description that on the public-API path `subgraph.GetEpName()` always matches `kQnnExecutionProvider` exactly?
> 
> **[T-3]** `onnxruntime/test/util/include/test_utils.h`
> 
> Observation only — with the migration complete, `test_utils.h` no longer needs the `class Graph;` forward declaration (correctly removed), but the file still wraps everything in `namespace onnxruntime { namespace test { ... } }` which is now somewhat redundant. Out of scope for this PR; flagging as a follow-up cleanup candidate.

Fixed **[M-1]**, **[M-2]**, **[M-3]**, **[N-1]**, **[N-2]**. The **[T-1]**, **[T-2]**, **[T-3]** tend to be observation and might not require real fix.

qti-chuteng (approved):

PR #347, opened 2026-05-06
[QNN-EP] Add prepare_only mode for AOT-only context generation

qti-ashwshan:
Adds session config option ep.context_prepare_only=1 that compiles the QNN graph, writes the _ctx.onnx, then immediately releases the QNN backend and shuts down the device — freeing ~300 MB of HTP memory that would otherwise stay resident waiting for inference.

Use case: customers doing on-device AOT compilation that don't want a JIT-ready session afterward. They generate the ctx model and create a separate session to load it for inference.

### Description
Changes:
- New bool prepare_only_ member parsed from "ep.context_prepare_only"
- Skip SetupQnnInputOutput in CompileImpl (single-threaded + ARM64 MT)
- After CreateEPContextNodes: clear qnn_models_ + reset qnn_backend_manager_
- Run() returns ORT_EP_FAIL with clear "prepare_only mode" message
- Auto-enables context_cache_enabled_ with warning if user sets prepare_only=1 but forgets context_enable=1


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1537,9 +1555,10 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
     return ep->ort_api.CreateStatus(ORT_EP_FAIL, message.c_str());
   }
 
-  if (qnn::IsNpuBackend(ep->qnn_backend_manager_->GetQnnBackendType())) {
+  if (qnn::IsNpuBackend(ep->qnn_backend_manager_->GetQnnBackendType()) && !ep->prepare_only_) {
     // Set the power config id and the default power mode from provider option for main thread,
     // otherwise it will mess up the power mode if user just create session without run it.
+    // Skipped in prepare_only mode: the session will not run inference, so no power config is needed.
```
minfhong-qti: With PR #38, this is required as well for finalize graph.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -489,6 +489,24 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    prepare_only_ = prepare_only_str == "1";
+    ORT_CXX_LOG(logger_,
+                ORT_LOGGING_LEVEL_VERBOSE,
+                ("User specified option - enable_htp_prepare_only: " + prepare_only_str).c_str());
+
+    if (prepare_only_ && !context_cache_enabled_) {
+      ORT_CXX_LOG(logger_,
+                  ORT_LOGGING_LEVEL_WARNING,
+                  "enable_htp_prepare_only=1 requires context cache; auto-enabling ep.context_enable.");
+      context_cache_enabled_ = true;
```
minfhong-qti: Should it be the opposite? If `enable_htp_prepare_only` is set but not `context_enable`, set `prepare_only_ ` to `false`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2061,7 +2080,9 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
       total_finalize_time += std::chrono::duration_cast<std::chrono::milliseconds>(end - finalize_start);
 #endif
 
-      RETURN_IF_NOT_OK(qnn_model->SetupQnnInputOutput(ep->logger_));
+      if (!ep->prepare_only_) {
+        RETURN_IF_NOT_OK(qnn_model->SetupQnnInputOutput(ep->logger_));
```
minfhong-qti, line 2086: Could you elaborate more on why this help? It seems there is no interaction with backend in this function.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2107,6 +2130,13 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
   if (ep->context_cache_enabled_) {
     RETURN_IF_NOT_NULL(ep->CreateEPContextNodes(graphs[0], fused_nodes, count, ep_context_nodes));
   }
+
+  if (ep->prepare_only_) {
```
minfhong-qti: Prefer to move to the end of this function, i.e., right above LN2152.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2219,6 +2249,10 @@ bool QnnEp::GetPerThreadHtpPowerConfigs(qnn::PerThreadHtpPowerConfigs_t& per_thr
 OrtStatus* ORT_API_CALL QnnEp::OnRunStartImpl(_In_ OrtEp* this_ptr, _In_ const ::OrtRunOptions* run_options) noexcept {
   QnnEp* ep = static_cast<QnnEp*>(this_ptr);
 
+  if (ep->prepare_only_) {
+    return nullptr;
```
minfhong-qti, line 2258: Should we log something here? As well for others.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2373,6 +2415,12 @@ OrtStatus* QnnEp::ValidateCompiledModelCompatibilityInfo(const OrtHardwareDevice
                                                          size_t /*num_devices*/,
                                                          const char* compatibility_info,
                                                          OrtCompiledModelCompatibility* model_compatibility) noexcept {
+  if (qnn_backend_manager_ == nullptr) {
```
minfhong-qti: Validation only happens when loading a context model. The check here should be removed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-ashwshan: Prepare only is a mode explicitly set by the user. But if setting context enable is missed then we set it for them as prepare only mode is only for AOT on device. Or we can either let the uesr know that both context enable and prepare only have to be set.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-ashwshan: we don't want I/O setup in prepare_only mode as the qnn input output tensors is only set for inference. So technically this is just an optimization where we dont want to setup the QNn input output tensors as we would not be doing inference in the session

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-ashwshan: Yes

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-ashwshan: Agreed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
tirupath-qti: Yes. Context_enable is old argument and has larger scope hence it should drive the saving of context functionality. Using enable_htp_prepare_only without context_enable should result in an error i believe.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
tirupath-qti: finalize graph happens on ARM side, why we need Htp power config?

qti-ashwshan: Thanks @minfhong-qti

minfhong-qti (approved):

tirupath-qti (approved):
LGTM

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1452,6 +1463,13 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
                                                  OrtEpGraphSupportInfo* graph_support_info) noexcept {
   QnnEp* ep = static_cast<QnnEp*>(this_ptr);
 
+  if (ep->prepare_only_ && !ep->context_cache_enabled_) {
+    ORT_CXX_LOG(ep->logger_, ORT_LOGGING_LEVEL_WARNING,
+                "enable_htp_prepare_only=1 requires ep.context_enable=1. "
```
qti-ashwshan, line 1468: Error out when prepare only is enabled but context enable is not enabled.

PR #349, opened 2026-05-07
Fix nupkg built twice and isolate --build_nuget arg

huaychou:
### Description

- Remove the redundant build.bat --skip_tests call from the $BuildNuget block; keep only the dist-copy logic.
- Move --build_nuget out of $CommonArgs into a new $BuildNugetArgs, mirroring the existing $BuildArchiveArgs pattern, so it can never silently propagate to unintended build.bat invocations.

### Motivation and Context

- After 8eb617e, the main --build invocation already passes --build_nuget (via $CommonArgs), so the separate $BuildNuget block's --skip_tests call triggered a redundant second build_nuget_package() run.

### DIFF
--- a/qcom/scripts/windows/build.ps1
+++ b/qcom/scripts/windows/build.ps1
@@ -191,6 +191,7 @@ if ($BuildAsX) {
     $CommonArgs += "--buildasx"
 }
 
+$BuildNugetArgs = @()
 if ($BuildNuget) {
     $TargetNugetDir = (Get-NugetBinDir)
     $env:Path = "$TargetNugetDir;" + $env:Path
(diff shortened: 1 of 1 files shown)


qti-mbadnara:
The code looks good to me, can you trigger a workflow for this and validate that the build isn't happening twice, especially during the nightly flow?

huaychou:
> The code looks good to me, can you trigger a workflow for this and validate that the build isn't happening twice, especially during the nightly flow?

https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/25476893767

kuanyul-qti (approved):
LGTM if nightly test is passed.

qti-mbadnara (approved):
LGTM, please merge it once the nightly workflow is completed

qti-chuteng (approved):

PR #350, opened 2026-05-07
add reuse_sparse_indices attribute for Conv2d and Conv3d

xiha0704:
### Description

Added a new parameter reuse_sparse_indices to Conv2D and Conv3D to align with native QNN.

### Motivation and Context

In the QNN code, there is a special translation path for SparseConv. It is also mapped to the IR Conv op, but in that case the reuse_sparse_indices parameter is set to true, while for a regular Conv the parameter is false, for consistency.
In the QNN EP, we currently only support the standard Conv, which corresponds to the native ONNX Conv operator. This Conv does not have a reuse_sparse_indices parameter, so I simply set it to false.


onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -999,6 +999,20 @@ Ort::Status ConvOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
                 ("Using DepthWiseConv2d instead of Conv2d for node " + node_unit.Name()).c_str());
   }
 
+  // reuse_sparse_indices parameter: only for Conv2d and Conv3d (not DepthWiseConv2d, not ConvTranspose).
```
qti-yuduo, line 1214: Thanks for the change. Could you also add corresponding unit tests?

qti-chuteng:
**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc:1004-1014`

The PR title says "add reuse_sparse_indices **attribute**" and the new comment claims "Set to true when ...", but `bool8Value` is hardcoded to `0`. There is no `node_helper.Get("reuse_sparse_indices", ...)`, no session config key, and no domain attribute — nothing in the codebase can flip it to true. Please clarify the intent: either complete the feature (read an ONNX attribute under a `com.qti` domain or expose a session config key) and add tests for both default-false and true paths, or, if the explicit-default-false is the goal, document the motivation in the PR description and rewrite the misleading comment. The PR title should also drop the word "attribute" if no attribute is actually read.

```cpp
// Example for the read-attribute approach:
const uint8_t reuse_sparse = static_cast<uint8_t>(
    node_helper.Get("reuse_sparse_indices", static_cast<int64_t>(0)) != 0);
reuse_sparse_scalar.bool8Value = reuse_sparse;
```

---

**[M-2]** `onnxruntime/test/providers/qnn/conv_test.cc` (missing)

No unit tests cover the new control flow. Please add tests in `conv_test.cc` for: (1) Conv2d / Conv3d basic graph build (CPU + HTP) to confirm the change does not break existing functionality; (2) ConvTranspose and DepthWiseConv2d regression cases that verify the parameter is **not** attached on those paths; (3) attribute=true / attribute=false coverage if the M-1 read-mechanism approach is taken. Including the kill-test fence result (temporarily disable the change, confirm the new test fails) in the PR description is recommended.

---

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc:1002-1003`

The comment "Defaults to false. Set to true when the same sparse weight indices are reused across inference calls." describes a capability that the PR does not provide (no switch exists anywhere). Please rewrite to match the actual behaviour, e.g.:

```cpp
// QNN's reuse_sparse_indices is set to false (the default). The QNN EP
// currently does not expose a way to enable sparse-index reuse — to add
// it, plumb an ONNX attribute or session config key through here.
```

xiha0704:
Hi @qti-chuteng,
In the QNN code, there is a special translation path for SparseConv. It is also mapped to the IR Conv op, but in that case the reuse_sparse_indices parameter is set to true, while for a regular Conv the parameter is false, for consistency.
In the QNN EP, we currently only support the standard Conv, which corresponds to the native ONNX Conv operator. This Conv does not have a reuse_sparse_indices parameter, so I simply set it to false.

qti-chuteng:
**[C-1]** `onnxruntime/test/providers/qnn/conv_test.cc:L991`
`QnnCPUBackendTests::ConvTranspose2D_NoReuseSparseIndices` is missing the closing `}` after `ExpectedEPNodeAssignment::All);` — the following static helper `BuildPerChannelQDQChainConstWeightConvTestCase` is parsed as part of the test body, causing a hard syntax error. `conv_test.cc` will not compile.

```suggestion
                ExpectedEPNodeAssignment::All);
}

// Builds: weight_q0 (int8 init) -> DQ -> Q -> DQ -> Conv.
```

---

**[C-2]** `onnxruntime/test/providers/qnn/conv_test.cc:L2998`
Same issue: `QnnHTPBackendTests::ConvTranspose2D_NoReuseSparseIndices` is missing the closing `}` after `ExpectedEPNodeAssignment::All);`. The Block-Quantized Conv section that follows is parsed inside the unclosed TEST_F body.

```suggestion
                                     ExpectedEPNodeAssignment::All);
}

// ─────────────────────────────────────────────────────────────────────────────
```

qti-chuteng (approved):

PR #356, opened 2026-05-08
gather block quantize support for qnn gpu

sachjang-qti:
### Description
Added support for gather block quantize for QNN GPU

### Motivation and Context
It allows to add the quantization support for gather blocks in llms and run them on qualcomm gpu through EpContextBinary.


qti-chuteng:
## Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc:15-17`

The file-header op-resolution comment block was committed as a placeholder (`/* [COMMENT BLOCK UNCHANGED – omitted here for brevity] */`). This is the actual checked-in content, not a tooling truncation. Please restore the full ONNX-to-QNN op-resolution table (attributes / inputs / outputs / shape mapping). Use `matmulnbits_op_builder.cc` lines 12-46 as the reference format. Without it, the weight-shape rewrite at lines 210-211 and the INT4-packed reinterpretation are undocumented.

---

**[C-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc:251`

```cpp
RETURN_IF_NOT(is_gpu_backend, "MatMulNBits Op Supported Only for Qnn Gpu Backend");
```

Copy-paste leak from `matmulnbits_op_builder.cc`. Update the message to:

```cpp
RETURN_IF_NOT(is_gpu_backend, "GatherBlockQuantized Op Supported Only for QNN GPU Backend");
```

---

## Major

**[M-1]** `gatherBlockQuantized_op_builder.cc:91-132` (IsOpSupported)

Optional `zero_points` input (4th input of `com.microsoft.GatherBlockQuantized`) is never validated. `ToSignedFixedPoint4` hard-codes `zero_point = 8` and `QnnQuantParamsWrapper` offsets are zeroed — yet the op is still accepted when a model supplies non-zero zero_points. Result: silent wrong output for asymmetric-quant models. Either reject explicitly:

```cpp
RETURN_IF(inputs.size() > 3 && inputs[3].Exists(),
          "GatherBlockQuantized: zero_points input not supported (symmetric quant only)");
```

or follow `matmulnbits_op_builder.cc` lines 166-189 and validate the zero_points content.

---

**[M-2]** `gatherBlockQuantized_op_builder.cc:74-134` (IsOpSupported)

`quantize_axis` attribute is never read or validated. The implementation hard-codes `block_sizes = {1, block_size}` and rewrites `weight_shape[1]`, both assuming `quantize_axis == 1`. A model with `quantize_axis == 0` will be accepted silently and produce wrong output. Add:

```cpp
const int64_t quantize_axis = helper.Get("quantize_axis", static_cast<int64_t>(1));
RETURN_IF_NOT(quantize_axis == 1,
              "GatherBlockQuantized: only quantize_axis == 1 supported on QNN GPU");
```

---

**[M-3]** `qnn_model_wrapper.cc:817-829` and `qnn_utils.cc:837-905`, `qnn_utils.h:145-146`

The new `bool is_backend_gpu` flag on `GetTensorInfo` / `GetQnnDataType` is wrong-layered: the backend identity is already known to `QnnModelWrapper` via `GetQnnBackendType()`. Threading the flag through every caller is a footgun — any future op that forgets to pass `true` will silently fall back to the SFIXED_POINT_8 mapping for INT4 on GPU. Please:

1. Drop the flag from the public API.
2. Have `QnnModelWrapper::GetTensorInfo` compute `IsGpuBackend(GetQnnBackendType())` internally.
3. Keep `CreateMap` / `CreateMapQuantize` as `static` / anonymous-namespace helpers, do not expose them in `qnn_utils.h` (the names are too generic for the public namespace).
4. Use function-local `static const` so the maps are built once instead of per call.

---

**[M-4]** `gatherBlockQuantized_op_builder.cc:74-134` (IsOpSupported)

`IsOpSupported` does not invoke `ProcessInputs` / `ProcessAttributesAndOutputs` with `do_op_validation=true`. This means the QNN backend never validates the op at partition time — backend rejections will surface as compile-time fatals instead of partition-time fallbacks. Match the pattern in `matmulnbits_op_builder.cc:192-194`:

```cpp
std::vector<std::string> input_names;
RETURN_IF_ERROR(ProcessInputs(qnn_model_wrapper, node_unit, logger, input_names, true));
RETURN_IF_ERROR(ProcessAttributesAndOutputs(qnn_model_wrapper, node_unit, std::move(input_names), logger, true));
```

---

**[M-5]** `gatherBlockQuantized_op_builder.cc:160-211`

No rank-2 guard on `weight_shape` / `scale_shape`. `weight_shape[1] = scale_shape[1] * block_size` and `block_sizes = {1, block_size}` both assume rank == 2. ONNX spec allows ND. 1D input would be UB; 3D+ would crash QNN at compile. Add explicit checks:

```cpp
RETURN_IF_NOT(weights_info.shape.size() == 2,
              "GatherBlockQuantized: only rank-2 weights supported");
RETURN_IF_NOT(scales_info.shape.size() == 2,
              "GatherBlockQuantized: only rank-2 scales supported");
```

---

**[M-6]** `gatherBlockQuantized_op_builder.cc:210-211`

```cpp
weight_shape[1] = static_cast<uint32_t>(scale_shape[1] * block_size);                 // dead assignment
std::vector<uint32_t> weight_shape_ = {static_cast<uint32_t>(weight_shape[0]),
                                       static_cast<uint32_t>(scale_shape[1] * block_size)};
```

Line (a) is a dead assignment — `weight_shape[1]` is never read after this. Remove it. Also missing a consistency check between packed-byte weights and unpacked-element shape:

```cpp
RETURN_IF_NOT(static_cast<int64_t>(weights_info.shape[1]) * 8 ==
                  static_cast<int64_t>(scales_info.shape[1]) * block_size,
              "GatherBlockQuantized: weight packed bytes mismatch with scales * block_size");
```

---

**[M-7]** `gatherBlockQuantized_op_builder.cc:25-40`

Missing `ORT_MUST_USE_RESULT` on the three overrides (`IsOpSupported`, `ProcessInputs`, `ProcessAttributesAndOutputs`). Compare `matmulnbits_op_builder.cc:53-68`. Without this, callers that forget `RETURN_IF_ERROR` get no compiler warning ([C45]).

---

**[M-8]** `gatherBlockQuantized_op_builder.cc:1`, `gatherblockquantized_test.cc:1`

New files use Microsoft copyright. Per [B1][C10], new QNN EP files must use the Qualcomm copyright header plus `SPDX-License-Identifier: MIT`.

---

**[M-9]** `onnxruntime/test/providers/qnn/gatherblockquantized_test.cc:14, 188`

- The test file's top-level guard `#if defined(_M_ARM64)` excludes Linux ARM64. Should be `#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)`.
- No CPU/HTP backend negative tests — the `IsOpSupported` rejection path is unverified.
- `GatherBlockQuantized_LargerIndices` uses `vocab=201088, hidden=2880`, which allocates ~2.3 GB for the float reference vector and ~70 MB for packed weights. Confirm CI runners can host it, or shrink (e.g. vocab≈4096, hidden≈512).

---

## Minor

**[N-1]** `gatherBlockQuantized_op_builder.cc:51-69`

`ToSignedFixedPoint4` uses implicit narrowing (`int8_t hi = (int_expr) - zero_point;`). Match `MatMulNBitsOpBuilder::DQQToSignedFixedPoint4` and use `SafeInt<int64_t>` + `gsl::narrow_cast<int8_t>` ([B4]).

---

**[N-2]** `gatherBlockQuantized_op_builder.cc:226-237`

Indices block uses generic local names `name`, `info`, `wrapper`. Rename to `indices_tensor_name`, `indices_info`, `indices_tensor_wrapper` for consistency with the weight block above.

---

**[N-3]** `gatherBlockQuantized_op_builder.cc:229-236`

Indices `IsQnnTensorWrapperExist == true` branch is silently empty. The weight (179-180) and output (263-264) blocks log `"Tensor already added, skip it: ..."`. Mirror that pattern for consistency.

---

**[N-4]** `gatherBlockQuantized_op_builder.cc:51-69`

`block_size / 2` (2 nibbles per byte) and `zero_point = 8` (`= 2^(bits-1)`) implicitly assume `bits == 4`. Add a header comment to the helper noting the bits=4-only assumption, or parameterize `bits` and derive `zero_point`.

---

**[N-5]** `qnn_utils.cc:837-905`, `qnn_utils.h:145-146`

`CreateMap` / `CreateMapQuantize` rebuild `unordered_map`s on every call (perf regression on the partition-time hot path) and the names are too generic for the public namespace. See M-3 for the recommended fix.

onnxruntime/core/providers/qnn/builder/opbuilder/gatherblockquantized_op_builder.cc
qti-mattsinc, line 1: nit. snake_case for file name (gather_block_quantized_op_builder.cc)

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
qti-mattsinc, line 1: nit. snake_case for file name (gather_block_quantized_test.cc)

onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,341 @@
...
+#include "core/providers/qnn/builder/qnn_utils.h"
+
+#include <numeric>
+
+namespace onnxruntime {
+namespace qnn {
+
+/*
+  [COMMENT BLOCK UNCHANGED – omitted here for brevity]
+*/
```
qti-mattsinc: Comment needs to be updated?

onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,341 @@
...
+  OrtNodeAttrHelper helper(node_unit);
+  const int64_t bits = helper.Get("bits", static_cast<int64_t>(4));
+  const int64_t block_size = helper.Get("block_size", static_cast<int64_t>(32));
+
+  RETURN_IF_NOT(bits == 4,
+                "GatherBlockQuantized: only INT4 (bits == 4) supported");
+  RETURN_IF_NOT(block_size >= 16 && ((block_size & (block_size - 1)) == 0),
+                "GatherBlockQuantized: block_size must be power of 2 and >= 16");
+
+  // Validate scales datatype (must be float32)
```
qti-mattsinc: This comment should be moved lower, closer to the check for scales. And fp16 is also allowed

onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,341 @@
...
+// ProcessAttributesAndOutputs
+// ================================================================
+Ort::Status GatherBlockQuantizedOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                                                       const OrtNodeUnit& node_unit,
+                                                                       std::vector<std::string>&& input_names,
+                                                                       const Ort::Logger& logger,
+                                                                       bool do_op_validation) const {
+  if (do_op_validation) {
+    bool is_gpu_backend = IsGpuBackend(qnn_model_wrapper.GetQnnBackendType());
+    RETURN_IF_NOT(is_gpu_backend, "MatMulNBits Op Supported Only for Qnn Gpu Backend");
```
qti-mattsinc: Should be GatherBlockQuantized, not MatMulNBits. Also, this check is already handled in `IsOpSupported`, so we can remove it here.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -482,9 +482,9 @@ bool QnnModelWrapper::CreateQnnNode(const std::string& qnn_node_name,
 
     using namespace onnxruntime::qnn::utils;
 
-    std::ostringstream oss;
-    oss << op_config_wrapper;
-    ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, oss.str().c_str());
+    // std::ostringstream oss;
+    // oss << op_config_wrapper;
+    // ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, oss.str().c_str());
```
qti-mattsinc: Looks like this was commented for debugging? Should be restored

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -666,9 +666,9 @@ bool QnnModelWrapper::ComposeQnnGraph(bool build_json_qnn_graph) {
 
     using namespace onnxruntime::qnn::utils;
 
-    std::ostringstream oss;
-    oss << op_config_wrapper;
-    ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, oss.str().c_str());
+    // std::ostringstream oss;
```
qti-mattsinc: Same comment as above

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -860,7 +858,16 @@ bool OnnxDataTypeToQnnDataType(const ONNXTensorElementDataType onnx_data_type,
       {ONNX_TENSOR_ELEMENT_DATA_TYPE_BOOL, QNN_DATATYPE_BOOL_8},
   };
 
-  const std::unordered_map<ONNXTensorElementDataType, Qnn_DataType_t> onnx_to_qnn_data_type_quantized = {
+  if (is_backend_gpu) {
+    onnx_to_qnn_data_type[ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4] = QNN_DATATYPE_SFIXED_POINT_4;
+    onnx_to_qnn_data_type[ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4] = QNN_DATATYPE_UFIXED_POINT_4;
+  }
```
qti-mattsinc: Why is this only for GPU? Do we really have to add GPU-specific branches to all these util functions?

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -877,6 +884,21 @@ bool OnnxDataTypeToQnnDataType(const ONNXTensorElementDataType onnx_data_type,
       {ONNX_TENSOR_ELEMENT_DATA_TYPE_BOOL, QNN_DATATYPE_BOOL_8},
   };
 
+  if (is_backend_gpu) {
+    onnx_to_qnn_data_type_quantized[ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4] = QNN_DATATYPE_SFIXED_POINT_4;
+    onnx_to_qnn_data_type_quantized[ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4] = QNN_DATATYPE_UFIXED_POINT_4;
+  }
```
qti-mattsinc: Same comment as above

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
```diff
@@ -0,0 +1,193 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
qti-mattsinc: New files should use QC copyright:

```
```

onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,341 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
qti-mattsinc: New files should use QC copyright:

```
```

tools/nuget/generate_nuspec_for_native_nuget.py
qti-mattsinc, line 175: Why are the changes to this file needed?

onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,341 @@
...
+ private:
+  void ToSignedFixedPoint4(std::vector<uint8_t>& quant_data,
+                           int64_t num_blocks,
+                           int64_t block_size) const;
+};
+
+// ================================================================
+// uint4 → signed int4
+// ================================================================
+void GatherBlockQuantizedOpBuilder::ToSignedFixedPoint4(
```
qti-mattsinc: Can this be moved out of the op builder into a util function? Or perhaps this op builder can use the existing `utils::TransformUnsignedToSignedFixedPoint` function?

onnxruntime/core/providers/qnn/builder/opbuilder/gatherBlockQuantized_op_builder.cc
```diff
@@ -0,0 +1,341 @@
...
+      float_scale = std::vector<float>(float_scale_ptr, float_scale_ptr + num_blocks);
+    } else {
+      Ort::Float16_t* fp16_scale_ptr = reinterpret_cast<Ort::Float16_t*>(uint8_scale.data());
+      float_scale.reserve(num_blocks);
+      for (int64_t i = 0; i < num_blocks; i++) {
+        float_scale.emplace_back(static_cast<float>(fp16_scale_ptr[i]));
+      }
+    }
+
+    // Quantization Offsets : QNN Support only symmetric quantization with default value of 0
```
qti-mattsinc: Is this a statement about the QNN API or just the current GPU backend support?

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
sachjang-qti: Done

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
sachjang-qti: I am following file naming convention as per other files, matmulnbits_test.cc, randomnormallike_test.cc

tools/nuget/generate_nuspec_for_native_nuget.py
sachjang-qti: Done

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
sachjang-qti: Done

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
sachjang-qti: Done

onnxruntime/core/providers/qnn/builder/opbuilder/gatherblockquantized_op_builder.cc
sachjang-qti: Done

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
sachjang-qti: Initially i kept it for gpu only so it won't disturb the other code

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
sachjang-qti: Initially i kept it for gpu only so it won't disturb the other code

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
skadaver-qti: QNN NPU uses INT4 in a different way. They pack a single INT4 value into INT8 variable and use SFIXED_POINT_8 (not sure why), whereas we pack 2int4 values in a single int8 variable use SFIXED_POINT_4

onnxruntime/core/providers/qnn/builder/opbuilder/gatherblockquantized_op_builder.cc
```diff
@@ -0,0 +1,312 @@
...
+    TensorInfo indices_info{};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(indices_tensor, indices_info));
+    RETURN_IF_ERROR(utils::GetQnnDataType(
+        indices_tensor.quant_param.has_value(),
+        indices_tensor.type,
+        indices_datatype,
+        true));
+    RETURN_IF(indices_datatype != QNN_DATATYPE_INT_64,
+              "GatherBlockQuantized: indices must be INT_64");
+  }
```
qti-mattsinc: Where is this int64 restriction coming from? The QNN OpDef for gather says:

> in[1]
> 
> Indices in in[0] to extract based on axis.
> 
>    Mandatory: true
>    Data type: QNN_DATATYPE_INT_32, QNN_DATATYPE_UINT_32, backend specific
>    ...

And it looks like the GPU backend does support uint32/int32 for this tensor (as well as fp32). So do we need this check here?

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
```diff
@@ -0,0 +1,192 @@
...
+
+#include "gtest/gtest.h"
+
+#include "test/providers/qnn/qnn_test_utils.h"
+#include "test/unittest_util/qdq_test_utils.h"
+
+namespace onnxruntime {
+namespace test {
+
+#if defined(_M_ARM64)
```
qti-mattsinc: nit. The  `#if defined(_M_ARM64)` guard typically only surrounds the backend test cases instead of the whole file, since each backend uses a different guard (for HTP, it's `#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)`; for CPU, no guard is required). But since only GPU tests are present, it's not really an issue.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -844,7 +844,7 @@ Ort::Status QnnModelWrapper::IsPerChannelQuantized(const OrtNodeUnitIODef& io_de
   return Ort::Status();
 }
 
-Ort::Status QnnModelWrapper::GetTensorInfo(const OrtNodeUnitIODef& tensor, TensorInfo& tensor_info) const {
+Ort::Status QnnModelWrapper::GetTensorInfo(const OrtNodeUnitIODef& tensor, TensorInfo& tensor_info, bool is_backend_gpu) const {
```
qti-mattsinc: We shouldn't need to change this function signature, because `QnnModelWrapper` already has the `qnn_backend_type_` member variable that it can use at the call-site to `utils::GetQnnDataType`.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
qti-mattsinc: Oh I see. In that case we do have to change these util functions then. It might be a little cleaner to pass a `QnnBackendType qnn_backend_type` as the new argument instead of `bool is_backend_gpu`.

qti-mattsinc (approved):
LGTM. @sachjang-qti, I think you need to sign the CLA for CI to be fully green (https://github.com/onnxruntime/onnxruntime-qnn/pull/356#issuecomment-4403705270)

skadaver-qti (approved):

qti-chuteng:
### 🔴 Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gatherblockquantized_op_builder.cc:L202`
`ProcessInputs` calls `GetConstantTensor(weight_tensor_name)` which returns `nullptr` for any non-constant-initializer weight, and then passes that `nullptr` directly into `UnpackInitializerData` — the first ORT API call inside that function will null-dereference and crash. `IsOpSupported` never validates `weights_info.is_initializer`, so the partition phase accepts the op and the crash happens at compile time.

Fix: add `RETURN_IF_NOT(weights_info.is_initializer, "GatherBlockQuantized: weight must be a constant initializer")` (and the same for `scales`) inside `IsOpSupported`, or add an explicit null guard after `GetConstantTensor` in `ProcessInputs`.

---

### 🟠 Major

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_utils.h:L820`
`TransformUnsignedToSignedFixedPoint4` (per-block nibble subtract loop) duplicates the already-existing `TransformUnsignedToSignedFixedPoint(bits=4)` (XOR-based, constant-time, ~line 414). Per pattern [C64], the XOR version should be used; the new function also drops error reporting by returning `void` instead of `Ort::Status`.

Fix: remove the new function and replace the call site in `ProcessInputs` with `RETURN_IF_ERROR(utils::TransformUnsignedToSignedFixedPoint(quant_data, 4))`.

**[M-3]** `onnxruntime/core/providers/qnn/builder/opbuilder/gatherblockquantized_op_builder.cc:L190`
`IsOpSupported` calls the full `ProcessInputs` and `ProcessAttributesAndOutputs` at the end (lines 190–191), which internally call `AddTensorWrapper` and `AddCastNode` on `QnnModelWrapper` — violating checklist [C40] that `IsOpSupported` must be a pure query. This can pollute `model_tensors_map_` during `GetCapability`, causing `AddToModelBuilder` to skip initialization and crash with I/O name mismatch during compilation.

Fix: remove the `ProcessInputs` / `ProcessAttributesAndOutputs` calls from `IsOpSupported`. If QNN backend validation is needed, use `QnnModelWrapper::ValidateQnnNode` instead. See `gather_op_builder.cc` for the correct pattern.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gatherblockquantized_op_builder.cc:L192`
Multiple `std::vector` containers are used where QNN EP conventions ([B3]) require `InlinedVector<T, N>` for small/fixed-size containers on the partition-time hot path: `block_sizes` (2 elements), `input_names` (2 inputs), `param_tensor_names` (1 param), `weight_shape_` (2 elements).

**[N-2]** `onnxruntime/test/providers/qnn/gatherblockquantized_test.cc:L130`
The existing tests do not clearly exercise both the static-initializer (element-wise recast) and dynamic-input (Cast-node insertion) branches of the int64 → int32 indices handling. Add an explicit test with `is_initializer=true` for int64 indices and a separate one with `is_initializer=false`.

tirupath-qti (approved):

PR #363, opened 2026-05-11
[QnnNodeGroup] Support Sub->Cast->Pow variant in LayerNormFusion

kuanyul-qti:
### Description
Extend TryFusion() to follow an optional Cast→Pow path: if Sub's Pow-side consumer is Cast, get the Pow child of Cast and include the Cast node in node_units_. CreateOrValidateOnQnn is unchanged.
Add test: LayerNormFusion_SubCastPow_3D.

### Motivation and Context
Some models (e.g. opset-10 transformers) insert a Cast(to=float32) between Sub and Pow for numerical precision, producing Sub consumers {Cast, Div} instead of {Pow, Div}. TryFusion() returned nullptr because it expected Pow as a direct child of Sub.


kuanyul-qti:
No longer needed.

PR #364, opened 2026-05-11
[QNN EP] Fuse DynamicQuantizeLinear + ConvInteger pattern

tirupath-qti:
### Description
Introduces DQConvIntegerFusion, a new IQnnNodeGroup fusion that rewrites the dynamic-quantize ConvInteger subgraph into a floating-point QNN Conv2d node, as QNN HTP doesn't support Dynamic Quantization.

  x --> DynamicQuantizeLinear --> (a_q, a_scale, a_zp)
           a_q, B_int8, a_zp, B_zp_int8 --> ConvInteger --> Cast(FLOAT)
           a_scale, B_scale_init             --> Mul
           Cast.out, parallel_Mul.out   --> Mul
           requant_Mul.out, bias_init   --> Add [optional]

  x --> Transpose(NCHW->NHWC) -------+
                                                                   |   (activation input)
                                                                   v
  B --> [Dequantize(B_scale,B_zp)] --> Conv2d --> Transpose(NHWC->NCHW) --> [Add(bias)] --> out

  * Per-channel B_scale: HTP Dequantize does not accept per-channel quant inputs, so the int8 weight is pre-dequantized to float32 offline and emitted as a STATIC float tensor without a Dequantize op.


qti-chuteng:
**[M-1]** `dq_conv_integer_fusion.cc:L1`, `dq_conv_integer_fusion.h:L1`, `dq_conv_integer_fusion_test.cc:L1`

All three new files use the Microsoft copyright header instead of the Qualcomm header. Per convention B1, new files created by QTI engineers must use:
```
```

---

**[M-2]** `dq_conv_integer_fusion_test.cc:L35`

The test file platform guard is missing `defined(__linux__)`:
```cpp
// Current (incomplete):
#if defined(__aarch64__) || defined(_M_ARM64)

// Should be:
#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
```
Without `defined(__linux__)`, the Linux ARM64 CI pipeline cannot compile or run these fusion tests.

---

**[N-1]** `dq_conv_integer_fusion_test.cc` (multiple `AddNode` calls)

Several `AddNode` calls use `/*domain=*/""` instead of `kOnnxDomain`. Per the convention from PR #200, `AddNode` calls in `test/providers/qnn/qnn_node_group/` should consistently use `kOnnxDomain`.

---

**[N-2]** `dq_conv_integer_fusion.cc:L714-L722`

Small op attribute containers in `CreateOrValidateOnQnn` use `std::vector<uint32_t>` instead of `InlinedVector`. These have fixed element counts (2 or 4):
```cpp
// Should use:
InlinedVector<uint32_t, 2> strides = ...;
InlinedVector<uint32_t, 2> dilations = ...;
InlinedVector<uint32_t, 4> pads = ...;
InlinedVector<uint32_t, 4> pad_amount = ...;
```

onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc
```diff
@@ -0,0 +1,862 @@
...
+  out.resize(bytes.size() / sizeof(float));
+  std::memcpy(out.data(), bytes.data(), bytes.size());
+  return Ort::Status();
+}
+
+// Reads a zero-point initializer (INT8 or UINT8) as int32 values. Empty `name` returns an
+// empty vector. Per ONNX spec the zero-point dtype matches its corresponding tensor's dtype;
+// callers should already have validated the weight dtype before calling this.
+Ort::Status ReadZeroPointAsInt32(const QnnModelWrapper& qmw,
+                                 const std::string& name,
```
minfhong-qti: `OrtNodeUnitIODef` already records data type in member `type`. Prefer passing `OrtNodeUnitIODef` instead of the name and no need the logic of acquiring data type.
As well for other similar functions.

tirupath-qti:
> **[M-1]** `dq_conv_integer_fusion.cc:L1`, `dq_conv_integer_fusion.h:L1`, `dq_conv_integer_fusion_test.cc:L1`
> 
> All three new files use the Microsoft copyright header instead of the Qualcomm header. Per convention B1, new files created by QTI engineers must use:
> 
> ```
> ```
> 
> **[M-2]** `dq_conv_integer_fusion_test.cc:L35`
> 
> The test file platform guard is missing `defined(__linux__)`:
> 
> ```c++
> // Current (incomplete):
> #if defined(__aarch64__) || defined(_M_ARM64)
> 
> // Should be:
> #if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
> ```
> 
> Without `defined(__linux__)`, the Linux ARM64 CI pipeline cannot compile or run these fusion tests.
> 
> **[N-1]** `dq_conv_integer_fusion_test.cc` (multiple `AddNode` calls)
> 
> Several `AddNode` calls use `/*domain=*/""` instead of `kOnnxDomain`. Per the convention from PR #200, `AddNode` calls in `test/providers/qnn/qnn_node_group/` should consistently use `kOnnxDomain`.
> 
> **[N-2]** `dq_conv_integer_fusion.cc:L714-L722`
> 
> Small op attribute containers in `CreateOrValidateOnQnn` use `std::vector<uint32_t>` instead of `InlinedVector`. These have fixed element counts (2 or 4):
> 
> ```c++
> // Should use:
> InlinedVector<uint32_t, 2> strides = ...;
> InlinedVector<uint32_t, 2> dilations = ...;
> InlinedVector<uint32_t, 4> pads = ...;
> InlinedVector<uint32_t, 4> pad_amount = ...;
> ```

addressed the comments.

qti-chuteng:
**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc:91-107, 285-548`

`ConsumersAreAllConvIntegers()` checks that DQL's a_q/a_zp consumers are all ConvIntegers, but does not guarantee that those ConvIntegers will all successfully fuse. When two ConvIntegers (X, Y) share one DQL: if X fuses and Y is rejected by `CreateOrValidateOnQnn(validate=true)`, Y falls back to CPU EP as a standalone ConvInteger, but DQL has been absorbed into X's fusion — Y can no longer access a_q / a_zp. The new `DQConvIntegerFusion_TwoConvIntegersShareDQL` test does not cover this gap because both ConvIntegers fuse successfully.

Suggested fix: pre-validate every sibling ConvInteger's pattern shape (kernel/strides/dilations/pads/auto_pad/rank-4) before claiming DQL, and reject if any sibling cannot fuse. Add a test where one of the two siblings has a rejected shape (e.g., pads with 5 elements), and assert the partition does not end up with "DQL on QNN but ConvInteger on CPU".

---

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc:725-734`

`CreateOrValidateOnQnn` reads the terminator NCHW shape from the ONNX graph but does not verify that all dims are static at TryFusion time. Dynamic-shape models (unknown N/H/W) pass TryFusion and only fail later — either inside the validate-time `GetOnnxShape` (after sibling-claim accounting is already done, see M-1), or deferred to the AddToModelBuilder call in the second GetCapability pass.

Suggested fix: validate `terminator_nchw_shape.size() == 4` and every dim > 0 inside TryFusion **before** claiming DQL; reject with `reject("Conv output shape is not fully static")` on violation.

---

**[N-1]** `onnxruntime/test/providers/qnn/qnn_node_group/dq_conv_integer_fusion_test.cc:1206-1248, 1311-1366`

`HasQnnJsonGraph()` returning false silently bypasses every `AssertOpInQnnGraph` check. On HTP-unavailable platforms, the test reports green without verifying that the fusion actually fired. Loses the kill-test fence — broken fusion logic would still appear to pass in CI.

Suggested fix: gate the test entry with explicit `SKIP_IF`/`GTEST_SKIP()` based on architecture; when the test executes, the asserts must always run. Or convert the missing-JSON case to `GTEST_SKIP() << "..."` so CI sees the skip.

---

**[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc:40` and `.h:51`

`kFusionType` (`"DQConvIntegerFusion"`) is declared in the .cc anonymous namespace, while `Type()` in the .h returns the same literal directly. Two literals must stay manually in sync — a future rename can silently break the sibling-claim comparison (`claim_it->second->Type() != kFusionType`).

Suggested fix:
```cpp
// In .h
class DQConvIntegerFusion : public IQnnNodeGroup {
 public:
  static constexpr std::string_view kType = "DQConvIntegerFusion";
  std::string_view Type() const override { return kType; }
  ...
};
// In .cc
constexpr auto kFusionType = DQConvIntegerFusion::kType;
```

onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc
```diff
@@ -0,0 +1,844 @@
...
+      floats[i] = scales[c] * static_cast<float>(static_cast<int32_t>(src[i]) - zps_onnx[c]);
+    }
+  }
+
+  out_float_bytes.resize(num_elems * sizeof(float));
+  std::memcpy(out_float_bytes.data(), floats.data(), out_float_bytes.size());
+  return Ort::Status();
+}
+
+}  // namespace
```
huaychou: Duplicated with https://github.com/onnxruntime/onnxruntime-qnn/pull/367

onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc
```diff
@@ -0,0 +1,844 @@
...
+  Qnn_Scalar_t group_scalar = QNN_SCALAR_INIT;
+  group_scalar.dataType = QNN_DATATYPE_UINT_32;
+  group_scalar.uint32Value = group;
+  QnnParamWrapper group_param(conv_integer.Index(), conv_node_name, QNN_OP_CONV_2D_PARAM_GROUP,
+                              group_scalar);
+
+  const uint32_t in_channels = nchw_in_shape[1];
+  const uint32_t out_channels = b_info.shape[0];
+  const bool is_depthwise = (group == in_channels) && (group == out_channels);
+  const char* conv_qnn_op = is_depthwise ? QNN_OP_DEPTH_WISE_CONV_2D : QNN_OP_CONV_2D;
```
huaychou, line 550: Could we have UTs for QNN_OP_DEPTH_WISE_CONV_2D?

tirupath-qti:
> **[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc:91-107, 285-548`
> 
> `ConsumersAreAllConvIntegers()` checks that DQL's a_q/a_zp consumers are all ConvIntegers, but does not guarantee that those ConvIntegers will all successfully fuse. When two ConvIntegers (X, Y) share one DQL: if X fuses and Y is rejected by `CreateOrValidateOnQnn(validate=true)`, Y falls back to CPU EP as a standalone ConvInteger, but DQL has been absorbed into X's fusion — Y can no longer access a_q / a_zp. The new `DQConvIntegerFusion_TwoConvIntegersShareDQL` test does not cover this gap because both ConvIntegers fuse successfully.
> 
> Suggested fix: pre-validate every sibling ConvInteger's pattern shape (kernel/strides/dilations/pads/auto_pad/rank-4) before claiming DQL, and reject if any sibling cannot fuse. Add a test where one of the two siblings has a rejected shape (e.g., pads with 5 elements), and assert the partition does not end up with "DQL on QNN but ConvInteger on CPU".
> 
> **[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc:725-734`
> 
> `CreateOrValidateOnQnn` reads the terminator NCHW shape from the ONNX graph but does not verify that all dims are static at TryFusion time. Dynamic-shape models (unknown N/H/W) pass TryFusion and only fail later — either inside the validate-time `GetOnnxShape` (after sibling-claim accounting is already done, see M-1), or deferred to the AddToModelBuilder call in the second GetCapability pass.
> 
> Suggested fix: validate `terminator_nchw_shape.size() == 4` and every dim > 0 inside TryFusion **before** claiming DQL; reject with `reject("Conv output shape is not fully static")` on violation.
> 
> **[N-1]** `onnxruntime/test/providers/qnn/qnn_node_group/dq_conv_integer_fusion_test.cc:1206-1248, 1311-1366`
> 
> `HasQnnJsonGraph()` returning false silently bypasses every `AssertOpInQnnGraph` check. On HTP-unavailable platforms, the test reports green without verifying that the fusion actually fired. Loses the kill-test fence — broken fusion logic would still appear to pass in CI.
> 
> Suggested fix: gate the test entry with explicit `SKIP_IF`/`GTEST_SKIP()` based on architecture; when the test executes, the asserts must always run. Or convert the missing-JSON case to `GTEST_SKIP() << "..."` so CI sees the skip.
> 
> **[N-2]** `onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc:40` and `.h:51`
> 
> `kFusionType` (`"DQConvIntegerFusion"`) is declared in the .cc anonymous namespace, while `Type()` in the .h returns the same literal directly. Two literals must stay manually in sync — a future rename can silently break the sibling-claim comparison (`claim_it->second->Type() != kFusionType`).
> 
> Suggested fix:
> 
> ```c++
> // In .h
> class DQConvIntegerFusion : public IQnnNodeGroup {
>  public:
>   static constexpr std::string_view kType = "DQConvIntegerFusion";
>   std::string_view Type() const override { return kType; }
>   ...
> };
> // In .cc
> constexpr auto kFusionType = DQConvIntegerFusion::kType;
> ```

Addressed in new commit pushed in this PR

qti-chuteng (approved):

huaychou (approved):
LGTM

PR #365, opened 2026-05-11
Add provider options for GPU custom config settings

qti-mattsinc:
### Description
* Add  `"gpu_precision_mode"`, `"disable_gpu_memory_optimizations"`,
  `"disable_gpu_node_optimizations"`, and `"disable_gpu_queue_recording"` provider options.

### Motivation and Context
* Make GPU custom config options accessible through ORT.
* `"gpu_precision_mode"` is useful for e.g. running in hybrid precision or testing out fp16 precision without having to reexport model weights to fp16. The other options can be helpful when debugging.

### DIFF
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -244,6 +244,28 @@ Warning: Enabling HTP Monolithic LSTM may improve session creation time, but thi
 |'0'|Default. Disabled.|
 |'1'|Enable HTP extended UDMA mode for better performance on supported hardware.|
 
+|`"gpu_precision_mode"`|Description|
+|---|---|
+|'0'|Sets the precision mode to floating point 32-bit (FP32).|
+|'1'|Sets the precision mode to floating point 16-bit (FP16).|
+|'2'|Sets the precision mode to FP16 for storage and FP32 for calculations.|
+|'3'|Default. Uses the tensor data type provided by the user.|
+
+|`"disable_gpu_memory_optimizations"`|Description|
+|---|---|
+|'0'|Default. QNN GPU memory optimizations enabled.|
+|'1'|Disable QNN GPU memory optimizations.|
+
+|`"disable_gpu_node_optimizations"`|Description|
+|---|---|
+|'0'|Default. QNN GPU node optimizations enabled.|
+|'1'|Disable QNN GPU node optimizations.|
+
+|`"disable_gpu_queue_recording"`|Description|
+|---|---|
+|'0'|Default. QNN GPU recordable command queue (RCQ) enabled.|
+|'1'|Disable the QNN GPU recordable command queue (RCQ).|
+
 |`"op_packages"`|Description|
 |---|---|
 |Op package config (string)|Register custom op packages. Format: `<OpType>:<PackagePath>:<InterfaceSymbolName>[:<Target>]`. Multiple packages can be separated by commas.|
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -24,6 +24,7 @@
 #endif
 
 #include "HTP/QnnHtpGraph.h"
+#include "GPU/QnnGpuGraph.h"
 
 #include "core/providers/qnn/common/qnn_graph_utils.h"
 #include "core/providers/qnn/ort_api.h"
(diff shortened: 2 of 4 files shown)


qti-chuteng:
## [M-1] `docs/execution_providers/QNN-ExecutionProvider.md`

The three new `disable_gpu_*` doc entries say `'0'` = "Disable …" and `'1'` = "Default. Enabled.", but `qnn_execution_provider.cc` initializes the default string to `"0"` and treats `"1"` as setting `gpu_disable_*_ = true`. So the meaning of `'0'` and `'1'` is inverted in the docs — every reader who follows the table will configure the opposite of what they want.

**Suggested fix** — swap the rows for all three tables, e.g.:

```
|`"disable_gpu_memory_optimizations"`|Description|
|---|---|
|'0'|Default. QNN GPU memory optimizations enabled.|
|'1'|Disable QNN GPU memory optimizations.|
```

Also worth aligning `disable_gpu_queue_recording` doc text with the inline code comment (mention "recordable command queue (RCQ)").

---

## [M-2] `onnxruntime/core/providers/qnn/qnn_execution_provider.cc`

Within the same PR, log severity for invalid values is inconsistent:

- `gpu_precision_mode` invalid → `ORT_LOGGING_LEVEL_ERROR` (matches `enable_htp_fp16_precision`, `htp_share_resource_optimization`).
- `disable_gpu_memory_optimizations` / `disable_gpu_node_optimizations` / `disable_gpu_queue_recording` invalid → `ORT_LOGGING_LEVEL_VERBOSE`.

VERBOSE is suppressed under the default logger, so a user typo like `"true"` silently falls back. The error string also says "Set to 0" while the real member stays at its initialized `false` — misleading even if the log were visible.

**Suggested fix** — bump the three `disable_gpu_*` invalid-value branches to `ORT_LOGGING_LEVEL_ERROR` and rephrase the message to describe the actual fallback (e.g., "Falling back to default (optimizations enabled).").

---

## [N-1] `onnxruntime/core/providers/qnn/qnn_execution_provider.cc` — duplicated parser boilerplate

The three `disable_gpu_*` parsers are ~24 lines each, differing only in key and target member (~72 lines of near-identical boilerplate). Checklist item C01 calls this out explicitly, and M-2 is exactly the kind of regression duplicated boilerplate produces.

**Suggested fix** — extract a file-static helper:

```cpp
void ParseGpuDisableFlag(const OrtApi& ort_api,
                         const SessionOptions& opts,
                         const std::string& key,
                         bool& out_flag,
                         const Ort::Logger& logger);
```

…that handles `GetSessionConfigEntryOrDefault`, the `"0"`/`"1"` mapping, and the ERROR log in one place.

---

## [N-2] No unit test added under `onnxruntime/test/providers/qnn/`

The PR adds 4 session config keys but no test changes. Even if GPU-backend CI is not always available, please add at least a set-only test in `qnn_basic_test.cc` that:

1. Sets each new key with a valid value and confirms session creation succeeds.
2. Sets each `disable_gpu_*` key with an invalid value and confirms the error is observed (round-trip).
3. Confirms `gpu_precision_mode="3"` maps to `QNN_GPU_PRECISION_USER_PROVIDED`.

Without this, an SDK enum renumbering would not be caught locally.

qti-mattsinc:
@qti-chuteng, I have addressed your previous comments after revisiting this PR. PTAL

> ## [M-1] `docs/execution_providers/QNN-ExecutionProvider.md`
> 
> The three new `disable_gpu_*` doc entries say `'0'` = "Disable …" and `'1'` = "Default. Enabled.", but `qnn_execution_provider.cc` initializes the default string to `"0"` and treats `"1"` as setting `gpu_disable_*_ = true`. [...]

Fixed

> ## [M-2] `onnxruntime/core/providers/qnn/qnn_execution_provider.cc`
> 
> Within the same PR, log severity for invalid values is inconsistent: [...]

Switched to `ParseBoolOption`, which uses VERBOSE. Seems like we might want to change it to ERROR?
https://github.com/onnxruntime/onnxruntime-qnn/blob/527934ab01a0ea1bf13bb95d7b71c750c47ded3e/onnxruntime/core/providers/qnn/qnn_execution_provider.cc#L299-L301

But that's not specific to this change.

> ## [N-1] `onnxruntime/core/providers/qnn/qnn_execution_provider.cc` — duplicated parser boilerplate
> [...]

Switched to `ParseBoolOption`

> ## [N-2] No unit test added under `onnxruntime/test/providers/qnn/`
> [...]

Added code coverage tests

qti-mattsinc:
gentle bump for review @qti-chuteng

PR #366, opened 2026-05-12
[ModelCompatibility] Reduce peak memory in compatibility validation.

minfhong-qti:
### Description

Fix this memory issue by removing the context blob version from compatibility info and thus no need to create the fake binary. Since the combination of a new context binary generated from an old SDK is impossible, backend API version or SDK version is enough to cover the scope of checking the context blob version, and thus the removal is safe.
Test:
 - Remove context blob version related UTs
 - Add UTs validating old API/SDK.

### Motivation and Context

During validating the model compatibility, a fake context binary is created to acquire the context blob version. However, even though the single layer graph used to create the binary is small, loading the prepare library introduces higher peak memory usage to ~200MB than without such validation where the peak memory usage is only ~50MB.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -246,13 +246,11 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   void SetQnnBackendType(uint32_t backend_id);
   QnnBackendType GetQnnBackendType() { return qnn_backend_type_; }
 
+  Qnn_Version_t GetBackendApiVersion() { return backend_api_version_; }
+
   const std::string& GetSdkVersion() { return sdk_build_version_; }
 
-  Ort::Status GetHtpArch(QnnHtpDevice_Arch_t& htp_arch) {
-    RETURN_IF_ERROR(GetPlatformInfo());
-    htp_arch = htp_arch_internal_;
-    return Ort::Status();
-  }
+  QnnHtpDevice_Arch_t GetHtpArch() { return htp_arch_internal_; }
 
   // Get backend library directory by adopting identical logic as in LoadLib.
   std::string GetBackendLibDir() {
@@ -587,6 +585,7 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   int htp_share_resource_optimization_ = -1;
 
   uint32_t backend_id_ = QNN_BACKEND_ID_CPU;
+  Qnn_Version_t backend_api_version_ = QNN_VERSION_INIT;
   bool file_mapped_weights_enabled_ = false;
 
 #ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
--- a/onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
@@ -1,17 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-#ifdef _WIN32
-#include <io.h>
-#else
-#include <unistd.h>
-#endif
-
 #include <string>
-#include <type_traits>
 
-#include "QnnCommon.h"
-#include "System/QnnSystemContext.h"
+#include "HTP/QnnHtpDevice.h"
 
 #include "core/providers/qnn/builder/qnn_cache_compatibility_manager.h"
 #include "core/providers/qnn/builder/qnn_def.h"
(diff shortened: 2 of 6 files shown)


onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
huaychou: Do we still need to check the context_blob_version?

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
```diff
@@ -262,12 +66,19 @@ Ort::Status QnnCacheCompatibilityManager::ValidateCompatibilityInfo(const QnnCom
   // The comparison order is as below:
   //   1. backend ID
   //   2. backend API version (if no HNRD involved) / SDK version (if HNRD involved)
-  //   3. context blob bersion
+  //   3. (Deprecated) context blob bersion
```
huaychou:
```suggestion
  //   3. (Deprecated) context blob version
```

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
huaychou: ditto

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
```diff
@@ -262,12 +66,19 @@ Ort::Status QnnCacheCompatibilityManager::ValidateCompatibilityInfo(const QnnCom
...
-  // SDK and HNRD separately in HNRD paths. The fake context binary has context blob version tagged with the smaller
-  // one between SDK and HNRD, and thus comparing to it directly reflect comparing to both SDK and HNRD.
+  // Notes:
+  //   - Context blob version is deprecated from compatibility info due to the peak memory concern in the previous
+  //     approach which acquires the runtime context blob version through creating a fake context binary. Nevertheless,
+  //     the removal in fact does not impact the compatibility validation since the combination of a new context binary
+  //     generated by an old SDK is impossible. Thus, comparing backend API version / SDK version is enough to cover
+  //     context blob version.
+
+  // Override expected context blob version to achieve backward compatibility with those generated before deprecating
```
huaychou: Do we have plan to deprecate the context blob version?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -246,13 +246,11 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+  Qnn_Version_t GetBackendApiVersion() { return backend_api_version_; }
+
   const std::string& GetSdkVersion() { return sdk_build_version_; }
 
-  Ort::Status GetHtpArch(QnnHtpDevice_Arch_t& htp_arch) {
-    RETURN_IF_ERROR(GetPlatformInfo());
-    htp_arch = htp_arch_internal_;
-    return Ort::Status();
-  }
+  QnnHtpDevice_Arch_t GetHtpArch() { return GetPlatformInfo().IsOK() ? htp_arch_internal_ : QNN_HTP_DEVICE_ARCH_NONE; }
```
huaychou: Could we use the htp_arch_internal_ once GetPlatformInfo is called?

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
minfhong-qti: Removed.

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
minfhong-qti: Revised.

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
minfhong-qti: Removed.

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.cc
minfhong-qti: My plan is when FCB is introduced, model compatibility will as well be modified to accommodate which I'll introduce versioning and mark as previous version.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
minfhong-qti: `GetPlatformInfo` handles it internally.

huaychou: Currently, we have to call GetPlatformInfo every time we use GetHtpArch. Is there a way to avoid calling GetPlatformInfo repeatedly after it has already been called?

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.h
kuanyul-qti: Do we need to remove this?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Is there a reason to get rid of `Ort::Status`? Removing it may result in callers losing access to important diagnostics, such as identifying a missing driver or an SDK mismatch.

minfhong-qti: My thought is that the caller is using this function to trivially acquire the HTP arch and original design required passing output variable and handling error status. But the previous design existed is only due to `GetPlatformInfo` may fail, which should not affect how caller uses it. Therefore, I prefer to wrap the error handling of `GetPlatformInfo` inside `GetHtpArch` and let caller to handle the received HTP arch should not be None.

onnxruntime/core/providers/qnn/builder/qnn_cache_compatibility_manager.h
minfhong-qti: I prefer to keep it for now. I'll revise related design including versioning during FCB implementation.

huaychou: Can add a comment as deprecated

minfhong-qti: Added

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
minfhong-qti: `GetPlatformInfo` is now invoked in `SetupBackend` once only. Please check again.

huaychou (approved):
LGTM. We can extend the GetPlatformInfo to other backends in the future.

qti-chuteng (approved):

PR #368, opened 2026-05-12
[QNN EP] Introduce ScopedGenieSession RAII helper in Genie integration tests

huaychou:
### Description

- Replace the fragile bracket-scoping workaround with a ScopedGenieSession class that jointly owns the RegisteredEpDeviceUniquePtr and Ort::Session. C++ reverse-member-destruction order enforces that the session is always destroyed before the EP device handle is unregistered, making the constraint structural rather than editorial.


qti-chuteng (approved):

kuanyul-qti (approved):

huaychou:
Integrated into https://github.com/onnxruntime/onnxruntime-qnn/pull/376

PR #369, opened 2026-05-12
[QNN EP] Ensure ORT Core is 1.24.4

qti-mbadnara:
This PR ensures we always build ORT QNN EP with the desired version of ORT Core for Python wheels


tirupath-qti (approved):
This impacts only python wheel dependency.

PR #387, opened 2026-05-14
[QNN EP] Fix build failure on Ubuntu 24.04 (GCC 13)

huaychou:
### Description

- Extend the -Wno-maybe-uninitialized suppression to cover GCC 13+.

### Motivation and Context

- The previous condition (VERSION_LESS 13 AND VERSION_GREATER_EQUAL 12) only applied to GCC 12, so Ubuntu 24.04's default GCC 13 hit a false positive -Wmaybe-uninitialized in std::variant internals (triggered in qnn_basic_test.cc) as a -Werror compilation failure.


qti-chuteng (approved):

cmake/CMakeLists.txt
```diff
@@ -634,7 +634,7 @@ function(onnxruntime_set_compile_flags target_name)
     foreach(FLAG ${ORT_WARNING_FLAGS})
       target_compile_options(${target_name} PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${FLAG}>")
     endforeach()
-    if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS 13 AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 12)
+    if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 12)
```
qti-hungjuiw: How about narrow the change?
```suggestion
if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION_LESS_EQUAL 14 AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 12)
```

huaychou: Done

qti-hungjuiw (approved):

PR #388, opened 2026-05-14
Fix build fail caused by system language

chunghow-qti:
### Description
`Win32_OperatingSystem.OSArchitecture` returns localized strings, causing Unknown OS Architecture on non-English systems. Switch to language-neutral `RuntimeInformation::OSArchitecture` enum

### Motivation and Context
When system language is set to Chinese `(Get-CimInstance Win32_operatingsystem).OSArchitecture` will return `ARM 64 位元處理器`, which cause build fail.

### DIFF
--- a/qcom/scripts/windows/run_tests.ps1
+++ b/qcom/scripts/windows/run_tests.ps1
@@ -12,6 +12,8 @@ param(
 $RootDir = (Resolve-Path -Path "$(Split-Path -Parent $MyInvocation.MyCommand.Definition)").Path
 $RepoRoot= (Resolve-Path -Path ("$RootDir\..\..\.."))
 
+. (Join-Path $RootDir "utils.ps1")
+
 if (-not $OnnxModelsRoot) {
     $OnnxModelsRoot = (Join-Path $RootDir (Join-Path "model_tests" "onnx_models"))
 }
@@ -51,7 +53,7 @@ $NewBuildDirectoryBackslashes = ($NewBuildDirectory -replace "/", "\\")
     Out-File -Encoding ascii $CTestTestFile
 
 # Figure out if HTP is available
-if ((Get-CimInstance Win32_operatingsystem).OSArchitecture -eq "ARM 64-bit Processor") {
+if ((Get-HostArch) -eq "arm64") {
     $QdqBackend = "htp"
 } else {
     $QdqBackend = "cpu"
--- a/qcom/scripts/windows/utils.ps1
+++ b/qcom/scripts/windows/utils.ps1
@@ -113,10 +113,13 @@ function Get-DefaultCMakeGenerator() {
 }
 
 function Get-HostArch() {
-    switch ([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture) {
-        ([System.Runtime.InteropServices.Architecture]::Arm64) { "arm64" }
-        ([System.Runtime.InteropServices.Architecture]::X64)   { "x86_64" }
-        default { throw "Unknown OS Architecture $_." }
+    # Read from machine-level registry entry set by the OS installer — locale-independent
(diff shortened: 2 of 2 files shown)


huaychou (approved):
LGTM

kuanyul-qti (approved):

qti-chuteng:
## Major

**[M-1]** `qcom/scripts/windows/run_tests.ps1:54`

The same locale-dependent `OSArchitecture` check that this PR fixes in `utils.ps1` still exists in `run_tests.ps1`:

```powershell
if ((Get-CimInstance Win32_operatingsystem).OSArchitecture -eq "ARM 64-bit Processor") {
    $QdqBackend = "htp"
} else {
    $QdqBackend = "cpu"
}
```

On non-English Windows installs (Traditional Chinese, Japanese, etc.), the comparison silently fails and tests on an ARM64 host get misrouted to the CPU backend instead of HTP — silent test-matrix degradation that won't fail CI loudly. Please fix in this PR. Suggested replacement:

```powershell
$hostArch = [System.Environment]::GetEnvironmentVariable("PROCESSOR_ARCHITECTURE", "Machine")
if ($hostArch -eq "ARM64") { $QdqBackend = "htp" } else { $QdqBackend = "cpu" }
```

Better still: dot-source `utils.ps1` and reuse `Get-HostArch`. (L81 in the same file uses numeric `Win32_Processor.Architecture -ne 12`, which is locale-safe — leave as-is.)

---

## Minor

**[N-1]** `qcom/scripts/windows/utils.ps1:115` (`Get-HostArch`)

Reading `PROCESSOR_ARCHITECTURE` at `Machine` scope returns `AMD64` for an x64-emulated PowerShell process running on ARM64 (depending on OEM image), causing a misclassification of the host as `x86_64`. The existing helper `qcom/scripts/upleveling/platform_loader.py::_detect_arch()` already handles this by also checking `PROCESSOR_ARCHITEW6432`. Suggest aligning:

```powershell
$arch = [System.Environment]::GetEnvironmentVariable("PROCESSOR_ARCHITEW6432", "Process")
if ([string]::IsNullOrEmpty($arch)) {
    $arch = [System.Environment]::GetEnvironmentVariable("PROCESSOR_ARCHITECTURE", "Machine")
}
```

This matters because `Get-DefaultCMakeGenerator` (utils.ps1 L98) uses `Get-HostArch` to decide Ninja vs Visual Studio generator — a misclassified host can produce binaries that won't run natively on the device.

---

**[N-2]** `qcom/scripts/windows/utils.ps1:119` (switch in `Get-HostArch`)

`PROCESSOR_ARCHITECTURE` returns `x86` on 32-bit Windows, which falls through to `throw "Unknown OS Architecture x86."`. Reads like the locale bug regressed. Make the unsupported-platform intent explicit:

```powershell
"X86"   { throw "32-bit Windows is not supported. Use a 64-bit PowerShell." }
```

chunghow-qti:
Thanks @qti-chuteng for the review.

**[M-1]** Fixed in `0d2baa8`. Dot-sourced `utils.ps1` and replaced the locale-dependent check with `Get-HostArch`. Left the numeric `Win32_Processor.Architecture -ne 12` check at L81 as-is per your note.

**[N-1]** Fixed in `0d2baa8`. `Get-HostArch` now reads `PROCESSOR_ARCHITEW6432` (process scope) first and falls back to machine-scope `PROCESSOR_ARCHITECTURE`, matching the defensive pattern in `platform_loader.py::_detect_arch()`.

**[N-2]** Skipping. The current `default` branch already throws with the offending value (`"Unknown OS Architecture x86."`), so it's not a silent failure — just a less specific message. 32-bit Windows isn't a supported host for this repo and adding a dedicated case felt like overhead for a path no one hits. Happy to add it if you feel strongly.

qti-chuteng (approved):

PR #391, opened 2026-05-15
[QNN EP] Support NonZero with dynamic output shape

huaychou:
### Description

- Remove the static output-shape requirement in IsOpSupported and instead verify the input has a static shape (required to compute total_elements). ProcessAttributesAndOutputs already derives the QNN output shape from the input dims, so no model-building logic changes are needed.

### Motivation and Context

- QNN HTP NonZero always outputs [rank, total_elements] with actual non-zero indices in the leading columns and -1 fill in the remainder.


huaychou:
Nonzero dynamic shape support is no longer required.

PR #393, opened 2026-05-15
pytorch_half_pixel Route rank-4 linear Resize to ResizeBilinear

qti-yuduo:
^ fixes QNN error:

```
Error occurred due to op : node_upsample_bilinear2d_1_token_1050_2;
reason : Wrong number of Parameters 6.
Op specific validation failed.
Failed to validate op node_upsample_bilinear2d_1_token_1050_2 with error 0xc26 : QNN_OP_PACKAGE_ERROR_VALIDATION_FAILURE
```

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc
@@ -93,13 +93,19 @@ const OnnxAttrInfo<int64_t> ResizeOpBuilder::onnx_antialias_attr = {"antialias",
 const OnnxAttrInfo<int64_t> ResizeOpBuilder::onnx_exclude_outside_attr = {"exclude_outside", 0};
 const OnnxAttrInfo<float> ResizeOpBuilder::onnx_cubic_coeff_a_attr = {"cubic_coeff_a", -0.75f};
 
-// PYTORCH_HALF_PIXEL diverges from HALF_PIXEL only when an output spatial dim == 1
-// (then it pins the coord to 0). Elsewhere the two are identical.
-static bool IsEquivalentPyTorchHalfPixel(const OrtNodeUnit& node_unit) {
+// Returns true when ONNX 'pytorch_half_pixel' is bit-identical to 'half_pixel'
+// for this rank-4 Resize. Per ONNX spec, pytorch_half_pixel only diverges from
+// half_pixel on output axes whose length == 1 (it pins the source coord to 0
+// instead of evaluating (x + 0.5) / scale - 0.5). When both spatial output dims
+// are > 1, the two modes are mathematically equivalent and the node can be
+// lowered to QNN ResizeBilinear with half_pixel_centers=true.
+//
+// Caller contract: input_rank == 4 must already be verified by the gate, which
+// (per ONNX Resize: output_rank == input_rank) implies output rank == 4.
+// IsOpSupported has already validated that output shape is present.
+static bool IsPyTorchHalfPixelEquivalentToHalfPixel(const OrtNodeUnit& node_unit) {
   const auto& output_shape_opt = node_unit.Outputs()[0].shape;
-  if (!output_shape_opt.has_value() || output_shape_opt->size() != 4) {
-    return false;
-  }
+  assert(output_shape_opt.has_value() && output_shape_opt->size() == 4);
   const auto& output_shape = *output_shape_opt;
   const bool is_nhwc = node_unit.Domain() == kMSInternalNHWCDomain;
   const size_t h_axis = is_nhwc ? 1 : 2;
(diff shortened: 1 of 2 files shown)


qti-shubham (approved):

qti-chuteng:
### Major

**[M-1]** `onnxruntime/test/providers/qnn/resize_test.cc → ResizeU8_2xLinearPytorchHalfPixel_EmitsResizeBilinear`
The new test rolls its own string-grep over JSON dumps via a `qnn_ep_graph_checker` lambda, but the codebase already has `AssertOpInQnnGraph(dump_dir, op, count)` in `qnn_node_group/qnn_graph_checker.cc` (used by `reshape_einsum_reshape_test.cc`, `scale_softmax_fusion_test.cc`, etc.) which parses JSON via `nlohmann::json` and matches `graph.nodes[*].type` properly. Substring search on `"ResizeBilinear"` is fragile (collides with comments, tensor names, future op renames like a hypothetical `ResizeBilinearGrad`), and `qnn_ep_graph_checker` is meant to inspect `Ort::Session` for EP-assignment via public API — not to do filesystem walks. Replace the lambda with two `AssertOpInQnnGraph` calls (`"ResizeBilinear", 1` + `"Resize", 0`) after `TestQDQModelAccuracy`. This also subsumes N-1 and T-1.

---

### Minor

**[N-1]** `onnxruntime/test/providers/qnn/resize_test.cc → ResizeU8_2xLinearPytorchHalfPixel_EmitsResizeBilinear`
Dump dir uses `fs::temp_directory_path() / "resize_pytorch_half_pixel_qnn_graph"`, which is shared across all parallel processes; the `remove_all` at the top can wipe another in-flight process's dump. Sibling fusion tests use a cwd-relative path (e.g., `"ReshapeEinsumReshape_Float"`) for this reason. Use a named cwd-relative dir + `gsl::finally` cleanup.

**[N-2]** `onnxruntime/test/providers/qnn/resize_test.cc`
The new logic forks two paths, but the new test only covers the `H,W > 1` branch. Add a case with a unit spatial dim (e.g., `{1,3,4,4}` -> `{1,3,1,8}`) and assert `AssertOpInQnnGraph(graph_dir, "Resize", 1)` + `AssertOpInQnnGraph(graph_dir, "ResizeBilinear", 0)` so a future QAIRT uplevel that breaks the fallback path is caught by CI.

**[N-3]** PR description (commit `db752d508d`)
Per `[G02]`, please record a kill-test fence in the PR description: with `IsEquivalentPyTorchHalfPixel` short-circuit removed, `ResizeU8_2xLinearPytorchHalfPixel_EmitsResizeBilinear` must fail because the dumped QNN graph emits `Resize`, not `ResizeBilinear` (rather than failing via an HTP op-validation early abort).

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/test/providers/qnn/resize_test.cc:L556`
This PR routes `pytorch_half_pixel` rank-4 linear Resize to `ResizeBilinear` whenever both output spatial dims > 1, which means the existing `ResizeU8_2xLinearPytorchHalfPixel` test (input `{1,3,4,4}`, output `{1,3,8,8}`) now also lowers to ResizeBilinear (its comment was correctly updated). After this PR, **no test** exercises the `transformation_mode == "pytorch_half_pixel"` else branch (`H_out == 1` / `W_out == 1` falling back to QNN Resize). Please add an explicit fallback-path test, e.g. a downsample to a 1-pixel spatial dim, so both exits of `IsEquivalentPyTorchHalfPixel` have regression guards.

```suggestion
// pytorch_half_pixel with H_out == 1 must fall back to QNN Resize.
TEST_F(QnnHTPBackendTests, ResizeU8_DownsampleToHeight1_LinearPytorchHalfPixel) {
  std::vector<float> input_data = GetFloatDataInRange(-10.0f, 10.0f, 24);
  RunQDQResizeOpTest<uint8_t>(TestInputDef<float>({1, 3, 4, 2}, false, input_data),
                              {1, 3, 1, 2}, "linear", "pytorch_half_pixel", "",
                              ExpectedEPNodeAssignment::All, 19);
}
```

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc:L98`
`IsEquivalentPyTorchHalfPixel` is only invoked from the `is_npu_backend && input_rank == 4 && interp_mode == "linear"` branch in `ProcessAttributesAndOutputs`, where the layout transformer has already rewritten the node to NHWC. So `is_nhwc` is always `true` at this call site, and the NCHW indexing branch (`h_axis = 2, w_axis = 3`) is dead code. Either assert the NHWC invariant, or add a comment explaining that the NCHW path is kept defensive for future reuse.

**[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc:L98`
When the output shape is unknown (dynamic-shape models), `IsEquivalentPyTorchHalfPixel` silently returns `false` and falls back to the slower QNN Resize op even when the runtime spatial dims would be > 1. Consider emitting `LOGS_DEFAULT(VERBOSE)` to surface that the fast path was missed because shape inference data was unavailable, or document this limitation in the PR description.

qti-yuduo:
@qti-chuteng thanks for the review, address feedback.

qti-chuteng (approved):

PR #394, opened 2026-05-15
Fix Docker volume mounts failing on symlinked host paths

qti-kromero:
### Description
* Uses resolve() instead of absolute() to get the path of volume

### Motivation and Context
* absolute() is incompatible if people happen to use a symlink for this cache, whereas resolve() follows the symlink


qti-mbadnara (approved):
LGTM

qti-ashwshan (approved):
LGTM. 
Nit - Missing PR description.

PR #395, opened 2026-05-16
[QNN EP] Minor CI Fixes

qti-mbadnara:
This PR fixes the following as part of the CI workflows

- Upload all the libs extracted from each RC candidate under the `unsigned_libs` folder. Before it was uploaded under the `unsigned` folder.
- Renamed the folder of the wheel artifacts from `wheels.zip` to `wheel.zip`
- When the zip is uploaded to the artifactory, the SHA and other metadata information is displayed. This change forwards the output to > dev/null to prevent it from being visible to external actors.
- Fix indentation errors

### DIFF
--- a/.github/workflows/qualcomm-internal-build-and-test.yml
+++ b/.github/workflows/qualcomm-internal-build-and-test.yml
@@ -51,26 +51,26 @@ jobs:
       target_py_vsn: "3.10"
 
   build-linux-aarch64-oe-gcc11_2:
-      if: ${{ inputs.do_all }}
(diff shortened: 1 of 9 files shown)


qti-chuteng (approved):

kuanyul-qti (commented):
Just curious, why the folder name should be changed to `unsigned_libs`?

qti-mbadnara:
> Just curious, why the folder name should be changed to `unsigned_libs`?

We are just extracting the DLLs to prepare them for signing, therefore for better readability would prefer if we renamed them to `unsigned_libs`

kuanyul-qti (approved):

PR #399, opened 2026-05-18
Add tier-based test architecture (component/snapshot/accuracy) and migrate Clip

yuhuchua-qti:
## Summary

Migrate the `Clip` op test suite into the new tier-based QNN EP unit-test
architecture, add the numerical **accuracy** tier, and land the supporting
CI/coverage infrastructure that runs and gates these layers.

This PR is additive on the production side: the legacy integration tests remain
in place while the new component/snapshot/accuracy coverage is introduced.

The work spans:

1. **Migrate Clip to component + snapshot layers** — translate the
   integration-tier `Clip` tests into function-level UTs under
   `component/builder/opbuilder/`.
2. **Add accuracy UT tier + finish Clip quantized-scalar cases** — add the
   end-to-end numerical layer and the remaining Q/DQ-wrapped min/max cases.
3. **Add two-pass snapshot+accuracy runner** — `run_snapshot_accuracy.sh`.
4. **Split coverage tests into component + verify phases** — `generate_coverage.sh`.
5. **Close the folded-constant (Path A) coverage gap** — add a
   `..._FoldedConstTest` suite (snapshot + accuracy) so the two
   `Clip_U8/U16_FloatData_QDQConstMinMax` integration cases have a migrated
   equivalent; they were previously assumed redundant with the
   quantized-scalar cases, which gcov disproved (see Coverage impact below).
6. **New migrated tiers target HTP** — the new component/snapshot/accuracy
   suites exercise the current HTP-backed QNN EP test path. `Clip_4D_f32_DefaultMinMax`
   / `Clip_5D_f32` run on HTP with no code change (HtpOpDefSupplement confirms
   `RELUMINMAX` supports FP32 up to rank 5). The legacy integration-tier
   `QnnCPUBackendTests.*` cases in `clip_test.cc` are unaffected by this PR.

## Test layers (Clip)

Every layer shares per-case specs from `clip_specs.h`, so a single spec literal
drives the component, snapshot, and accuracy cases for one scenario.

Suite names are op-first: `QnnUnit_<Op>_<Tier>[_<Variant>]Test` — the op
segment is always between `QnnUnit_` and the first tier token, so
`--gtest_filter="QnnUnit_Clip_*"` / `--filter=Clip` cleanly select every Clip
case regardless of tier.

| Layer | Suites | Count | File | Mechanism |
|---|---|---|---|---|
| Component | `QnnUnit_Clip_ComponentTest` | 19 | `component/builder/opbuilder/clip_test.cc` | `AddToModelBuilder` + mock `NodeUnit`: dtype-dispatch arms (`do_op_validation=false`) + partition-reject arm (`do_op_validation=true`). No production code touched. |
| Op-builder snapshot | `QnnUnit_Clip_Snapshot[_{QDQFloat,QDQQuant,FoldedConst}]Test` | 14 | `snapshot/builder/opbuilder/clip_test.cc` | `ComposeQnnGraph` + `QnnJSONGraph` byte-compare vs frozen goldens under `$QNN_UT_SNAPSHOT_GOLDEN_DIR/snapshot/builder/opbuilder/clip/`. `..._FoldedConstTest` additionally replays the production `qdq_constant_folding` end-state via `AddTensorWrapper` + `MarkTensorAsFoldedConstant`, since the real ORT-side constant-folding pass can't run against a mock `NodeUnit`. |
| Session snapshot | `QnnUnit_Clip_SessionSnapshot_QDQFloatTest` | 2 | `session_snapshot/builder/opbuilder/clip_test.cc` | Same JSON compare, but through the session/partition path; goldens under `$QNN_UT_SNAPSHOT_GOLDEN_DIR/session_snapshot/builder/opbuilder/clip/`. |
| Accuracy | `QnnUnit_Clip_Accuracy[_{QDQFloat,QDQQuant,FoldedConst}]Test` | 16 | `accuracy/builder/opbuilder/clip_test.cc` | End-to-end run via public ORT API, diff outputs against the ORT-CPU EP reference within tolerance. |

51 cases total. Snapshot answers "did the QNN graph structure change?";
accuracy answers "if it changed, is it still numerically correct?". The two
form the verify tier.

## Coverage impact (clip_op_builder.cc)

Isolated runs on the same coverage-instrumented binary (121 lines / 6 fns /
158 branches), all cases executed, none skipped:

| Test set | Cases | Lines | Functions | Branches |
|---|---|---|---|---|
| Old integration (`Qnn*BackendTests.Clip_*`) | 17 | 62.8% (76/121) | 100% (6/6) | 52.5% (83/158) |
| New component + snapshot (no accuracy) | 35 | 98.3% (119/121) | 100% (6/6) | 64.6% (102/158) |
| New component + snapshot + accuracy | 51 | 98.3% (119/121) | 100% (6/6) | 64.6% (102/158) |

The +35.5pp line gain and +12.1pp branch gain over old integration come
entirely from the component/snapshot tiers, including the new
`..._FoldedConstTest` suite closing the Path A gap described above. Adding
the 16 accuracy cases moves builder lines and branches by **0** — the
snapshot tier's manual replay of the folded-constant end-state already
covers what accuracy would additionally exercise on `clip_op_builder.cc`.
Accuracy verifies output correctness, not builder code paths, so builder
coverage is attributed to component/snapshot, and accuracy is not used to
close coverage gaps.

## Supporting infra

All UT-only additions are gated and compile to nothing in production builds.

- **Gate macros** (`cmake/onnxruntime_unittests.cmake`):
  - `QNN_EP_INTERNAL_SYMBOL_ACCESS` — gates every UT layer (component/snapshot/
    accuracy). Auto-defined under coverage builds on Linux x86_64 with the
    shared QNN EP.
  - `QNN_EP_ACCURACY_UT` — gates the accuracy tier. Same prereqs; enabled
    together with coverage rather than as a separate opt-in.
- **`qnn_utils.cc`** — `QnnJSONGraph` `quant_params` now dispatches by encoding
  so QDQ snapshot goldens serialize the correct scale/offset variant.
- **`qcom/scripts/linux/run_snapshot_accuracy.sh`** — two-pass local runner:
  run snapshots first; only run accuracy for ops whose snapshot drifted.
  `--generate-goldens` accepts a new structure (then verifies accuracy);
  `--force-accuracy` / `--filter` scope the run. Exit 0 = OK (or drift verified),
  1 = accuracy regression, 99 = setup error. If snapshot drift requires Pass 2
  but no matching `QnnUnit_*_Accuracy*` tests are compiled in, the script now
  fails setup instead of silently treating the run as successful.
- **`qcom/scripts/linux/generate_coverage.sh`** — coverage run split into two
  separately-tracked phases so component-UT and verify (snapshot + accuracy)
  fail/skip independently. The two gtest filters union to the previous `*Qnn*`
  with no gap/overlap; `.gcda` accumulates across both and one `lcov --capture`
  yields a single merged report. Adds `--skip-verify`.

## Test plan

- [x] `python qcom/build_and_test.py lint_and_fix`
- [x] `python qcom/build_and_test.py lint`
- [x] `python qcom/build_and_test.py coverage_linux_x86_64 --target-py-version 3.12`
  - Provider/component/snapshot/accuracy tests completed successfully.
  - Coverage report generated successfully: overall lines 80.9%, functions 86.1%, branches 42.8%.
- [x] `onnxruntime_provider_test --gtest_filter='QnnUnit_Clip_*'` — component 19/19, op-builder snapshot 14/14, session snapshot 2/2, accuracy 16/16 PASS on Linux x86_64 (CPU + HTP emulator).
- [x] Snapshot byte-determinism: independent process runs produce byte-identical JSON.
- [x] Two-pass runner exercised locally: corrupt a golden → Pass 2 accuracy triggers → `--generate-goldens` restores; clean git diff after.
- [x] `generate_coverage.sh` two-phase run verified (component phase + merged coverage report); filter union proven equal to `*Qnn*` (18+... suites, no gap/overlap).
- [ ] CI green.

## Out of scope / follow-up

- Removing the original integration-tier `Clip` tests — the coverage table above shows the new component/snapshot tiers already dominate the old integration suite on `clip_op_builder.cc` (98.3% vs 62.8% lines, same 100% functions), so removal is safe with no coverage regression. Deferred to a follow-up to keep this PR additive on the production side.
- GPU variants (`Clip_fp32`, `DISABLED_Clip_int32`) — remain at the integration tier until a host-compatible GPU emulation path exists.
- Multi-partition snapshot golden support — extended when an op with real partition behavior (e.g. Conv) is migrated.

## Migration pattern for follow-up ops (Conv, MatMul, …)

1. Translate integration → snapshot 1:1, including partition/fallback assertions.
2. gcov against the op-builder; identify uncovered switch arms / error branches. Don't assume two cases hit the same branch just because their graphs look similar — verify with gcov, not just code reading (see Path A above).
3. Add component dispatch tests; add the accuracy tier sharing the same spec literals.
4. **Do not** modify production op-builder code to expose internals — drive dispatch via `AddToModelBuilder` + mock `NodeUnit` + `g_mock_init_reg`.


onnxruntime/test/providers/qnn/component/onnx_ctx_model_helper_test.cc
```diff
@@ -25,8 +25,8 @@
 #include "core/providers/qnn/builder/onnx_ctx_model_helper.h"
 #include "core/providers/qnn/builder/qnn_backend_manager.h"
 #include "core/providers/qnn/builder/qnn_model.h"
-#include "test/providers/qnn/unit/qnn_fake_ort_graph.h"
-#include "test/providers/qnn/unit/qnn_unit_test_utils.h"
+#include "test/providers/qnn/test_infra/qnn_fake_ort_graph.h"
```
qti-chuteng: would it better to use /infra/ or /utils/ ?

onnxruntime/test/providers/qnn/integration/README.md
```diff
@@ -10,16 +10,20 @@ operator pipelines rather than targeting specific internal code paths.
 The `integration/` subdirectory introduces a separate tier: **targeted pipeline
 integration tests** that exercise specific internal EP code paths (e.g. `ort_api.cc`,
 `qnn_model_wrapper.cc`) by building minimal models inline via the ORT model editor API
-and running them through a real backend. Tests skip automatically when the required
-backend is unavailable.
+and running them through a real backend. The tests are Linux-only (guarded by
```
qti-chuteng, line 13: Just curious, if there are some op UT, like GQA will always skip in v68, would them be skipped?

yuhuchua-qti: Existing UTs use SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO to detect SoC arch and skip. We'd carry that forward in the new tiers when an arch-gated op (like GQA) gets migrated — open to a more general design if you have one in mind.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -737,10 +737,50 @@ static nlohmann::json GetQnnTensorJSON(const Qnn_Tensor_t& tensor, bool include_
...
+        scale_offsets.push_back({{"scale", enc.scales[i]},
+                                 {"offset", enc.offsets ? enc.offsets[i] : 0}});
+      }
+      quant_params_json["bw_axis_scale_offset"] = {{"bitwidth", enc.bitwidth},
+                                                   {"axis", enc.axis},
+                                                   {"num_elements", enc.numElements},
+                                                   {"scale_offsets", std::move(scale_offsets)}};
+      break;
+    }
+    default:
```
huaychou: **[M-1] (Major)** GetQnnTensorJSON switch omits BLOCKWISE_EXPANSION encoding while the comment claims full QAIRT schema compliance

**Description:** The new `GetQnnTensorJSON` switch (lines 742–781) has no case for `QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION`. When this encoding is encountered, the `default: break` path executes and the resulting JSON `quant_params` object contains only `definition`/`encoding` — the actual quantization coefficients are silently dropped. The same file already handles BLOCKWISE_EXPANSION at lines 355, 408, and 1911, confirming this is an omission rather than an intentional choice.

**Impact:** Line 740's comment claims "quant_params JSON layout matches the QAIRT official qairt-dlc-to-json schema", but the missing BLOCKWISE_EXPANSION case makes that claim inaccurate. Clip op does not use blockwise quantization, so existing Clip snapshot tests are unaffected. However, `GetQnnTensorJSON` is a shared serialization utility — when future op snapshot tests (e.g., BQ ops) use it, golden JSON files will silently be missing quantization data, causing hard-to-detect golden inconsistencies.

**Suggested fix:**

Add an explicit case for BLOCKWISE_EXPANSION (even if only with a TODO comment), and update the comment at line 740 to accurately describe which encodings are currently serialized.

```cpp
case QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION: {
  // TODO: serialize blockwiseExpansion fields when BQ-op snapshot tests are added.
  // For now, only definition/encoding are emitted (no coefficient data).
  break;
}
```

Update the line 740 comment from "matches the QAIRT official qairt-dlc-to-json schema" to:
"Serializes scale_offset / bw_scale_offset / axis_scale_offset / bw_axis_scale_offset encodings.
QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION is not yet serialized (see TODO above)."

onnxruntime/test/providers/qnn/test_infra/backend_contexts.h
```diff
@@ -0,0 +1,199 @@
...
+    context_handle = manager_->GetQnnContext(0);
+    initialized_ = true;
+#endif
+  }
+
+  ~QnnRealHtpBackendManagerContext() = default;
+
+  bool IsValid() const { return initialized_; }
+
+  QnnRealHtpBackendManagerContext(const QnnRealHtpBackendManagerContext&) = delete;
```
huaychou: **[N-1] (Minor)** QnnRealHtpBackendManagerContext uses manual = delete instead of ORT_DISALLOW_COPY_AND_ASSIGNMENT macro

**Description:** Lines 182–183 manually spell out `= delete` for the copy constructor and copy-assignment operator of `QnnRealHtpBackendManagerContext`, whereas the codebase convention (and `QnnRealHtpBackendContext` in the same PR's `qnn_unit_test_utils.h`) uses the `ORT_DISALLOW_COPY_AND_ASSIGNMENT` macro.

**Impact:** Functionally equivalent, but inconsistent with the B5 codebase convention and with the sibling class added in the same PR, creating a style inconsistency.

**Suggested fix:**

Replace lines 182–183 with the macro form, placed in the public section at the top of the class body (the conventional location):

```cpp
ORT_DISALLOW_COPY_AND_ASSIGNMENT(QnnRealHtpBackendManagerContext);
```

qcom/scripts/linux/generate_coverage.sh
```diff
@@ -31,7 +31,49 @@ set_strict_mode
...
+# clip_op_builder.cc: component+snapshot 94.2% == with-accuracy 94.2%). Its .gcda
+# is still captured — that is harmless because snapshot already covers those
+# lines. "Accuracy is not a coverage patch" is a migration-completeness criterion
+# (don't close coverage gaps with accuracy), not a data-exclusion rule.
+#
+# gtest filter grammar: a single '-' separates the positive section from the
+# negative section; ':'-joined patterns after that '-' are ALL negative (do NOT
+# prefix each with its own '-', or they become literal, never-matching patterns).
+component_filter="*Qnn*:-QnnUnit_*_Snapshot*:QnnUnit_*_SessionSnapshot*:QnnUnit_*_Accuracy*"
+snapshot_filter="QnnUnit_*_Snapshot*:QnnUnit_*_SessionSnapshot*"
```
huaychou: **[T-1] (Nit)** component_filter's negative exclusion list has a redundant QnnUnit_*_SessionSnapshot* entry

**Description:** Line 73's `component_filter`:
`"*Qnn*:-QnnUnit_*_Snapshot*:QnnUnit_*_SessionSnapshot*:QnnUnit_*_Accuracy*"`
The `QnnUnit_*_SessionSnapshot*` entry in the negative section is redundant: `QnnUnit_*_Snapshot*` already matches test suite names that contain the substring SessionSnapshot (e.g. `QnnUnit_Clip_SessionSnapshot_QDQFloatTest`) through gtest's glob matching.

**Impact:** No functional impact. However, the redundant entry may mislead a future maintainer into thinking `QnnUnit_*_Snapshot*` and `QnnUnit_*_SessionSnapshot*` are disjoint patterns, potentially leading to incorrect future edits.

**Suggested fix:**

Remove the redundant entry, simplifying to:
```bash
component_filter="*Qnn*:-QnnUnit_*_Snapshot*:QnnUnit_*_Accuracy*"
```
Optional: add a comment noting that QnnUnit_*_Snapshot* already covers SessionSnapshot suites.

onnxruntime/test/providers/qnn/component/onnx_ctx_model_helper_test.cc
yuhuchua-qti: After comparing with the sibling tier directories, I think `infra` is the better name — it matches the naming convention of the existing tier directories (`component/`, `snapshot/`, `session_snapshot/`, `accuracy/`, `integration/`), none of which use a `test_` prefix even though they all live under `test/providers/qnn/` and are test-only. I've renamed `test_infra/` to `infra/` accordingly.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -737,10 +737,50 @@ static nlohmann::json GetQnnTensorJSON(const Qnn_Tensor_t& tensor, bool include_
   tensor_json["axis_format"] = "NOT_YET_DEFINED";
 
   const Qnn_QuantizeParams_t& quant_params = GetQnnTensorQParams(tensor);
-  tensor_json["quant_params"] = {
-      {"definition", quant_params.encodingDefinition},
-      {"encoding", quant_params.quantizationEncoding},
-      {"scale_offset", {{"scale", quant_params.scaleOffsetEncoding.scale}, {"offset", quant_params.scaleOffsetEncoding.offset}}}};
+  // quant_params JSON layout matches the QAIRT official qairt-dlc-to-json schema.
```
huaychou, line 740: Do we need to support the following quant params?
1. `QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION`
2. `QNN_QUANTIZATION_ENCODING_BLOCK`
3. `QNN_QUANTIZATION_ENCODING_BW_FLOAT_BLOCK`

onnxruntime/test/providers/qnn/infra/backend_contexts.h
```diff
@@ -0,0 +1,199 @@
...
+    initialized_ = true;
+#endif
+  }
+
+  ~QnnRealHtpBackendManagerContext() = default;
+
+  bool IsValid() const { return initialized_; }
+
+  QnnRealHtpBackendManagerContext(const QnnRealHtpBackendManagerContext&) = delete;
+  QnnRealHtpBackendManagerContext& operator=(const QnnRealHtpBackendManagerContext&) = delete;
```
huaychou:
```suggestion
  ORT_DISALLOW_COPY_AND_ASSIGNMENT(QnnRealHtpBackendManagerContext);
```

onnxruntime/test/providers/qnn/infra/qnn_fake_ort_graph.h
```diff
@@ -510,6 +537,51 @@ inline void InstallFakeGraphApiStubs(OrtApi& api) {
...
+  explicit OrtGlobalApiOverride(const OrtApi* new_api) {
+    original_ = OrtGetApiBase()->GetApi(ORT_API_VERSION);
+    Ort::detail::Global::Api(new_api);
+  }
+  ~OrtGlobalApiOverride() { Ort::detail::Global::Api(original_); }
+
+  OrtGlobalApiOverride(const OrtGlobalApiOverride&) = delete;
+  OrtGlobalApiOverride& operator=(const OrtGlobalApiOverride&) = delete;
+  OrtGlobalApiOverride(OrtGlobalApiOverride&&) = delete;
+  OrtGlobalApiOverride& operator=(OrtGlobalApiOverride&&) = delete;
```
huaychou:
```suggestion
  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(OrtGlobalApiOverride);
```

onnxruntime/test/providers/qnn/infra/backend_contexts.h
yuhuchua-qti: Applied in af00312b1a (pushed). Thanks!

onnxruntime/test/providers/qnn/infra/qnn_fake_ort_graph.h
yuhuchua-qti: Applied in af00312b1a (pushed). Thanks!

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
yuhuchua-qti: Good point — to be precise, it's not that Clip doesn't use these (it doesn't use `BW_SCALE_OFFSET`/`AXIS_SCALE_OFFSET` either, yet #399 decodes those). The real difference: those four share one flat `{scale, offset}` shape, so decoding them just mirrors the already-verified `SCALE_OFFSET` case. `BLOCKWISE_EXPANSION`/`BLOCK`/`BW_FLOAT_BLOCK` are nested (extra `blockSize` arrays, separate per-block scale tables) — writing that schema blind with no real producer to check against risks a silently-wrong dump.
They are real today via Conv/MatMul/MatMulNBits block-quant weights, currently falling into `default: break` (safe no-op, not UB). I'd like to add their decode logic alongside the Conv/MatMul tier migration, where a real golden JSON can verify the schema — will call that out in that PR.

onnxruntime/test/providers/qnn/component/README.md
```diff
@@ -33,6 +35,19 @@ All test code in this directory is guarded by `#if !defined(ORT_MINIMAL_BUILD) &
 | `onnx_ctx_model_helper_test.cc` | `QnnUnit_OnnxCtxModelHelperTest` | `builder/onnx_ctx_model_helper.cc` |
 | `qnn_execution_provider_test.cc` | `QnnUnit_ExecutionProviderTest` | `qnn_execution_provider.cc` |
 | `qnn_execution_provider_test.cc` | `QnnUnit_ExecutionProviderHtpTest` | `qnn_execution_provider.cc` (real-`libQnnHtp.so` paths) |
+| `builder/opbuilder/<op>_test.cc` | `QnnUnit_<Op>_ComponentTest` | `builder/opbuilder/<op>_op_builder.cc` |
+
+The last row is the **op-builder component tier**: unlike the core-component
```
qti-chuteng: nit: it looks okay on markdown, but a format looks a little bit weird here.

onnxruntime/test/providers/qnn/component/README.md
```diff
@@ -1,10 +1,12 @@
...
+# QNN EP Component-Tier Unit Tests
 
 ## Why this directory exists
 
 Tests in `onnxruntime/test/providers/qnn/` have historically been integration tests — they require a QNN SDK runtime, physical hardware, and a fully compiled EP stack. This makes them expensive to run and impossible to execute in most developer and CI environments.
 
-The `unit/` subdirectory introduces a separate testing tier: **function-level and component-level unit tests** that target the internal logic of the QNN EP. No on-device hardware is required — all tests run on a Linux x86-64 host. Tests that exercise op validation load `libQnnHtp.so` locally on the host (validation only, not graph execution); those tests are automatically skipped if the SDK is unavailable.
+The `component/` subdirectory is the **component tier** in the tier-first test layout: **function-level and component-level white-box tests** that target the internal logic of the QNN EP. No on-device hardware is required — all tests run on a Linux x86-64 host. Tests that exercise op validation load `libQnnHtp.so` locally on the host (validation only, not graph execution); those tests are automatically skipped if the SDK is unavailable.
```
qti-chuteng: Probablly need to explain why Windows Arm64 and Linux Arm64 can be skipped in Component-Tier UT.

onnxruntime/test/providers/qnn/integration/README.md
```diff
@@ -10,16 +10,20 @@ operator pipelines rather than targeting specific internal code paths.
...
+`defined(__linux__)`); on the Linux CI host the QNN SDK is always present, so a missing
+backend is treated as a configuration error (hard failure) rather than a skip.
 
 ## How this differs from the other test tiers
 
 | Tier | Directory | Backend required | What it targets |
 |---|---|---|---|
-| Unit | `unit/` | Optional (`libQnnHtp.so` for op validation only, no session) | Pure function logic; no graph execution |
+| Component | `component/` | Optional (`libQnnHtp.so` for op validation only, no session) | Pure function / op-builder logic; no graph execution |
+| Op-builder snapshot | `snapshot/` | Real backend, no finalize | QNN JSON graph structure produced by the op builder |
```
qti-chuteng: no graph finalization (?)

onnxruntime/test/providers/qnn/integration/README.md
```diff
@@ -48,15 +53,21 @@ fixtures there).
 
 ## Platform availability
 
+These tests are **Linux-only** — the file guard `defined(__linux__)` compiles them out
```
qti-chuteng: I think that we still need to give the reason why we don't provide Windows test by compiling the binary for windows.

qcom/scripts/linux/run_snapshot_accuracy.sh
```diff
@@ -0,0 +1,284 @@
...
+    if [ "${generate_goldens}" = true ]; then
+        log_info "Goldens updated. Verifying accuracy for: ${target_ops}"
+    else
+        log_info "Force-accuracy mode. Running accuracy for: ${target_ops}"
+    fi
+else
+    # Normal mode: run accuracy only for groups whose snapshot FAILED (drift):
+    # the builder changed but the golden was not updated, so the new graph
+    # structure is unverified and must be numerically checked.
+    if [ ${snapshot_exit} -eq 0 ]; then
```
tzuhsuanwei: When the golden store/file absent, the snapshot helper uses GTEST_SKIP(), so the test process exits 0. This branch then exits without running accuracy.

That gives a green result with neither snapshot comparison nor numerical verification on a fresh/missing golden store.

Suggested fix: record/parse the QNN_GOLDEN_ABSENT cases and add their op groups to target_ops, so accuracy runs for those groups instead of treating the snapshot phase as fully passed.

qcom/scripts/linux/generate_coverage.sh
```diff
@@ -227,8 +352,26 @@ cp "${REPO_ROOT}/qcom/scripts/linux/coverage_artifact_README.md" \
...
+# The component and accuracy phases GATE (non-zero exit fails this script). The
+# snapshot phase is NON-gating: a golden byte-mismatch means the graph
+# structure drifted, which is allowed to land (goldens may go stale on main; a
+# nightly job reconciles them). Drift is a routing signal for accuracy, and
+# numerical correctness is enforced by the accuracy phase above.
 # ---------------------------------------------------------------------------
-if [ "${test_exit}" -ne 0 ]; then
-    die "Tests failed with exit code ${test_exit}. Coverage report was still generated at ${output_dir}."
+if [ "${snapshot_exit}" -ne 0 ]; then
+    log_warn "snapshot phase exited ${snapshot_exit} — graph-structure drift detected."
```
tzuhsuanwei: The snapshot phase currently treats every non-zero exit as non-gating drift.

Snapshot tests also have ASSERT/setup paths such as graph creation, ComposeQnnGraph, session JSON discovery, and JSON parsing. Those failures would be masked if the accuracy phase passes.

Suggested fix: only allow failures marked QNN_SNAPSHOT_DRIFT to be non-gating. Parse the snapshot result/log and fail the job for any other snapshot failure.

qcom/scripts/linux/extract_snapshot_groups.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
```
qti-chuteng: is this file neccessary? could we put this inside the script only to avoid maintaining extra files?

qcom/scripts/linux/run_snapshot_accuracy.sh
```diff
@@ -0,0 +1,284 @@
+#!/usr/bin/env bash
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+#
+# Two-pass snapshot+accuracy test runner for QNN EP unit tests.
+#
+# Pass 1: Run all snapshot tests (QnnUnit_<Op>_Snapshot* + QnnUnit_<Op>_SessionSnapshot*).
+#          If all pass -> done (exit 0). Graph structure unchanged -> accuracy is redundant.
```
qti-chuteng: accuracy tests are redundant.

qcom/scripts/linux/run_snapshot_accuracy.sh
```diff
@@ -0,0 +1,284 @@
...
+
+log_info "Accuracy targets: ${target_ops}"
+
+# ---------------------------------------------------------------------------
+# Pass 2: Accuracy tests for target groups
+# ---------------------------------------------------------------------------
+
+# Probe whether accuracy tests are compiled in.
+accuracy_probe=$("${binary}" --gtest_list_tests --gtest_filter="QnnUnit_*_Accuracy*" 2>/dev/null || true)
+if [ -z "${accuracy_probe}" ]; then
```
qti-chuteng: We may need to change the warn to error when ACCURACY_UT enabled in long-term.

onnxruntime/test/providers/qnn/infra/backend_contexts.h
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+//
+// Backend / wrapper test contexts for QNN EP function-level unit tests.
+//
+//   OpBuilderTestContext       — stub-backed wrapper context (no live backend)
+//   QnnRealHtpBackendManagerContext  — full HTP backend via QnnBackendManager (has context handle)
+//
+// QnnRealHtpBackendManagerContext gives a usable Qnn_ContextHandle_t suitable
+// for CreateQnnGraph + ComposeQnnGraph (Path E1 JSON snapshot tests).
```
qti-chuteng: What is Path E1?

onnxruntime/test/providers/qnn/infra/backend_contexts.h
```diff
@@ -0,0 +1,198 @@
...
+
+OrtStatus* DefaultStubGetInitializersEmpty(const OrtGraph*, const OrtValueInfo**, size_t) noexcept {
+  return nullptr;
+}
+
+}  // namespace
+
+// Context for constructing a QnnModelWrapper in function-level unit tests.
+//
+//   ctx.input_info.indices  = {{"input0", 0}};   // declare graph inputs
```
qti-chuteng: are line#65-67 neccessary?

onnxruntime/test/providers/qnn/infra/backend_contexts.h
```diff
@@ -0,0 +1,198 @@
...
+        output_info,
+        settings);
+  }
+};
+
+// Context for tests that need a real QNN HTP backend with a live context handle
+// suitable for ComposeQnnGraph. This is the only real-backend context in this
+// header — QnnCpu (libQnnCpu.so) is no longer shipped with the QNN EP wheel.
+//
+// Note: HTP triggers `ProcessBF16Conversions` for FP32 tensors, which inserts
```
qti-chuteng: what's this about?

onnxruntime/test/providers/qnn/infra/qnn_unit_test_utils.h
```diff
@@ -252,6 +158,73 @@ class StubBackendManager {
...
+}  // namespace test
+}  // namespace onnxruntime
+
+// Component-tier helpers split out into focused headers. Included AFTER
+// MakeNullLogger so backend_contexts.h can use it in member initializers.
+// Declared above so backend_contexts.h can build a StubBackendManager.
+// Existing test sources that pull only qnn_unit_test_utils.h keep access to
+// MakeMockIODef / MakeMockNodeUnit / MockInitRegistry / OpBuilderTestContext
+// without extra explicit includes. The snapshot harness (snapshot.h) is NOT
+// pulled here — snapshot-tier TUs include it directly.
```
qti-chuteng: Could we place these header in the top of the file?

onnxruntime/test/providers/qnn/infra/qnn_unit_test_utils.h
```diff
@@ -252,6 +158,73 @@ class StubBackendManager {
...
+#include "test/providers/qnn/infra/backend_contexts.h"
+#include "test/providers/qnn/infra/mock_init_registry.h"
+#include "test/providers/qnn/infra/mock_node_unit.h"
+
+namespace onnxruntime {
```
qti-chuteng, line 26: do we need two section of the same namespace?

onnxruntime/test/providers/qnn/infra/snapshot_golden_utils.h
```diff
@@ -0,0 +1,162 @@
...
+#include <string>
+#include <string_view>
+
+#include "gtest/gtest.h"
+#include "nlohmann/json.hpp"
+
+namespace onnxruntime {
+namespace test {
+
+// Golden-tree root, taken from $QNN_UT_SNAPSHOT_GOLDEN_DIR (absolute path, no
```
qti-chuteng: code can explain itself.
Leave the neat and neccessary comment, or remove all of them.

onnxruntime/test/providers/qnn/infra/qnn_unit_test_utils.h
yuhuchua-qti: Thanks, I think the issue was that `qnn_unit_test_utils.h` was doing too many things at once: it owned logger helpers, backend stubs, real-backend contexts, and then reopened namespaces around additional includes.

I split the focused helpers into `qnn_test_logger.h` and `stub_backend_manager.h`, and kept `qnn_unit_test_utils.h` as the compatibility umbrella plus the remaining shared contexts. This should make the include/namespace ownership clearer without forcing
all existing tests to update their includes in this PR.

onnxruntime/test/providers/qnn/infra/snapshot_golden_utils.h
yuhuchua-qti: Done. I trimmed the long comments and removed stale/internal wording so the headers focus on ownership and usage only.

qcom/scripts/linux/run_snapshot_accuracy.sh
yuhuchua-qti: Comment is updated. Thanks.

onnxruntime/test/providers/qnn/component/README.md
yuhuchua-qti: Updated. I clarified that the component tier is currently enabled only in the Linux x86-64 coverage/export-symbol build, because that is the current build lane with `QNN_EP_INTERNAL_SYMBOL_ACCESS`. Windows x86-64, Windows ARM64, and Linux ARM64 enablement are described as follow-up work.

onnxruntime/test/providers/qnn/component/README.md
yuhuchua-qti: Updated. Thanks.

onnxruntime/test/providers/qnn/integration/README.md
yuhuchua-qti: Updated. I replaced “real backend, no finalize” with “real backend + QNN graph creation, no ORT session execution” to make the boundary clearer.

onnxruntime/test/providers/qnn/integration/README.md
yuhuchua-qti: Updated. I changed the stale `unit/` reference to `component/`, and clarified that integration tests are Linux-only because they require the current QNN HTP execution environment. Windows builds compile this tier out until there is an equivalent Windows HTP execution lane.

qcom/scripts/linux/extract_snapshot_groups.py
yuhuchua-qti: Done. Since `extract_snapshot_groups.py` was only used by `run_snapshot_accuracy.sh`, I inlined the small parser into the shell script to avoid maintaining a separate one-use script.

qcom/scripts/linux/run_snapshot_accuracy.sh
yuhuchua-qti: Done. If snapshot drift requires corresponding accuracy verification but no `QnnUnit_*_Accuracy*` tests are compiled in, the script now fails setup instead of warning and exiting successfully. This prevents snapshot drift from silently bypassing numerical verification.

onnxruntime/test/providers/qnn/infra/backend_contexts.h
yuhuchua-qti: Done. I removed the internal “Path E1” reference and changed the comment to describe the actual purpose directly: this context gives snapshot tests a usable `Qnn_ContextHandle_t` for `CreateQnnGraph` + `ComposeQnnGraph`.

onnxruntime/test/providers/qnn/infra/backend_contexts.h
yuhuchua-qti: Done. `OpBuilderTestContext` only needs a stable `OrtGraph*` sentinel and does not use the fake graph primitives directly, so I removed the `qnn_fake_ort_graph.h` include and replaced it with the focused infra headers it actually depends on.

onnxruntime/test/providers/qnn/infra/backend_contexts.h
yuhuchua-qti: Good point. The QnnCpu note was meant to explain why this helper only keeps the HTP real-backend context: the docs say `libQnnCpu.so` / `QnnCpu.dll` are not shipped in the onnxruntime-qnn release package and must come from the QAIRT SDK. I also removed the unrelated BF16/golden note so the comment stays focused on this helper’s scope.

onnxruntime/test/providers/qnn/infra/qnn_unit_test_utils.h
yuhuchua-qti: Done. I split the dependencies into `qnn_test_logger.h` and `stub_backend_manager.h`, then moved the infra includes back to the top of `qnn_unit_test_utils.h`. This removes the late include / namespace reopening pattern.

cmake/onnxruntime_unittests.cmake
```diff
@@ -200,8 +200,18 @@ if(onnxruntime_USE_QNN AND NOT onnxruntime_MINIMAL_BUILD AND NOT onnxruntime_RED
...
   list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/ssr/qnn_ssr_test.cc)
   include(onnxruntime_unittests_udo.cmake)
-  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/unit/*)
   list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/integration/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/infra/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/infra/specs/builder/opbuilder/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/component/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/component/builder/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/component/builder/opbuilder/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/component/builder/qnn_node_group/*)
```
kuanyul-qti, line 209: I don't see this folder.

cmake/onnxruntime_unittests.cmake
```diff
@@ -200,8 +200,18 @@ if(onnxruntime_USE_QNN AND NOT onnxruntime_MINIMAL_BUILD AND NOT onnxruntime_RED
...
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/snapshot/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/snapshot/builder/opbuilder/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/session_snapshot/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/session_snapshot/builder/opbuilder/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/accuracy/builder/opbuilder/*)
   list(APPEND onnxruntime_test_providers_dependencies onnxruntime_providers_qnn)
```
kuanyul-qti, line 215: Nit: In in Alphabet order.

qcom/scripts/linux/generate_coverage.sh
```diff
@@ -132,17 +213,61 @@ rm -f "${build_dir}/${config}/coverage_lcov.info" \
...
+# Holds the QnnUnit_*_Snapshot* / QnnUnit_*_SessionSnapshot* per-case results.
+snapshot_json="${build_dir}/${config}/snapshot_results.json"
+
+comp_exit=0
+snapshot_exit=0
+accuracy_exit=0
 
-if [ "${test_exit}" -ne 0 ]; then
-    log_warn "Tests exited with ${test_exit}; continuing to collect coverage from .gcda written so far."
+if [ -n "${test_filter}" ]; then
```
kuanyul-qti, line 359: Just want to confirm, is current CI routed to this branch instead of the three-phase testing process?

qcom/scripts/linux/generate_coverage.sh
```diff
@@ -132,17 +213,61 @@ rm -f "${build_dir}/${config}/coverage_lcov.info" \
...
+    # Legacy single-phase override.
+    run_test_phase "filtered" "${test_filter}" || comp_exit=$?
```
kuanyul-qti, line 361: Should it be `comp_exit` or another variable?

qcom/scripts/linux/run_snapshot_accuracy.sh
```diff
@@ -0,0 +1,308 @@
+#!/usr/bin/env bash
```
kuanyul-qti, line 1: It appears that no one is calling the file. Is that expected?

onnxruntime/test/providers/qnn/integration/README.md
```diff
@@ -132,15 +146,16 @@ Example: `QnnInteg_OrtApiTest.QDQGroup_CoversGetQDQIODefs`
 # Run only pipeline integration tests
 ./onnxruntime_provider_test --gtest_filter="QnnInteg_*"
 
-# Run all QNN tests (unit + integration + op-level)
+# Run all QNN tests (component + snapshot + accuracy + integration + op-level)
 ./onnxruntime_provider_test --gtest_filter="Qnn*"
```
kuanyul-qti, line 150: We don't need the `Qnn` string.

onnxruntime/test/providers/qnn/infra/backend_contexts.h
```diff
@@ -0,0 +1,186 @@
...
+//   QnnRealHtpBackendManagerContext htp;
+//   if (!htp.IsValid()) GTEST_SKIP() << "libQnnHtp.so not available";
+//   auto wrapper = MakeSnapshotWrapperHtpJson(ctx, htp, {"in"}, {"out"});
+struct QnnRealHtpBackendManagerContext {
+  QNN_INTERFACE_VER_TYPE qnn_interface = QNN_INTERFACE_VER_TYPE_INIT;
+  Qnn_BackendHandle_t backend_handle = nullptr;
+  Qnn_ContextHandle_t context_handle = nullptr;
+
+  QnnRealHtpBackendManagerContext() {
+#ifndef _WIN32
```
kuanyul-qti, line 131: Does it support linux aarch64?

minfhong-qti (commented):
Suggest to separate the reorganization into another PR as there are too many files in this PR and developers are keep updating existing files.

qcom/scripts/linux/run_snapshot_accuracy.sh
yuhuchua-qti: Thanks, updated. The runner now treats `[QNN_GOLDEN_ABSENT]` as an unverified snapshot result and routes the affected op group to its matching `QnnUnit_<Op>_Accuracy*` tests instead of skipping accuracy.

For example, if Clip snapshot goldens are missing, it runs `QnnUnit_Clip_Accuracy*`. It does not force all accuracy tests unless all in-scope snapshot groups are unverified or no snapshot JSON is produced.

qcom/scripts/linux/generate_coverage.sh
yuhuchua-qti: Updated. Snapshot is now treated as an accuracy-skip signal, not the correctness gate.

A clean snapshot pass can skip accuracy. Any unverified snapshot state, including drift, missing golden, snapshot failure/error, or missing snapshot JSON, falls back to matching accuracy tests. It only fails setup if the affected snapshot group has no matching QnnUnit_<Op>_Accuracy* test, because then numerical correctness would not be gated.

yuhuchua-qti:
> Suggest to separate the reorganization into another PR as there are too many files in this PR and developers are keep updating existing files.

PR#399 has been reorganized into 3 PRs (PR#840, PR#841, PR#842)
Thanks

yuhuchua-qti:
Thanks for the feedback. To make the review easier and avoid blocking developers who are editing the existing QNN test files, I split this large PR into three smaller stacked PRs:

  - #840: QNN test tier infra reorganization (`unit/` -> `component/`, shared helpers under `infra/`, base snapshot/session_snapshot helpers, README/CMake updates)
  - #841: snapshot/accuracy coverage runner and CI workflow updates
  - #842: first op migration example: Clip migration into component/snapshot/session_snapshot/accuracy tiers

I will use #840-#842 as the active review path. Most of the comments in this PR are either addressed there or moved to the corresponding smaller PR for focused review.

cmake/onnxruntime_unittests.cmake
yuhuchua-qti: Addressed in the split PRs. The CMake source patterns are now ordered by tier/folder name in the final stacked diff.

cmake/onnxruntime_unittests.cmake
yuhuchua-qti: Good catch. `component/builder/qnn_node_group/` was an obsolete path from an earlier revision. It has been removed in the split stack; the latest CMake only keeps the existing `providers/qnn/qnn_node_group/*` glob.

qcom/scripts/linux/generate_coverage.sh
yuhuchua-qti: Yes, this is the intended CI path for coverage. In #841, the coverage workflow still enters through `generate_coverage.sh`; the script is updated so that the default coverage run is split into component, snapshot, and accuracy phases.

qcom/scripts/linux/generate_coverage.sh
yuhuchua-qti: Addressed in #841. The latest version uses separate `comp_exit`, `snapshot_exit`, and `accuracy_exit` variables so each phase has its own gating result.

qcom/scripts/linux/run_snapshot_accuracy.sh
yuhuchua-qti: Yes, that is expected. `run_snapshot_accuracy.sh` is a standalone/local helper for manually exercising the snapshot -> accuracy fallback flow.

The CI entrypoint remains `generate_coverage.sh`; in #841 that script now contains the coverage flow that runs component, snapshot, and accuracy phases separately.

onnxruntime/test/providers/qnn/integration/README.md
yuhuchua-qti: I kept `QnnInteg_` for now so these suites are easy to select from the shared provider test binary with `--gtest_filter="QnnInteg_*"`. This is now moved to #840 for focused review.

onnxruntime/test/providers/qnn/infra/backend_contexts.h
yuhuchua-qti: The code path is not x86-specific; it is enabled for non-Windows builds and should work on Linux aarch64 when `libQnnHtp.so` is available. This PR does not enable the Linux aarch64 CI lane yet; that is planned as follow-up platform enablement.

PR #403, opened 2026-05-18
feat: add IsInf op builder to QNN EP

qti-yuduo:
Maps ONNX IsInf -> QNN_OP_IS_INF, wiring the detect_positive and detect_negative ONNX attributes (default 1) to the corresponding QNN_DATATYPE_BOOL_8 scalar params. Without this, models containing IsInf (typically emitted by torch.nan_to_num decompositions) fall back to CPU and shatter the single-graph requirement of qnn_context_binary compilation.

Validates input is FP32/FP16 and output is BOOL_8 per the QNN op contract.

### DIFF
--- a/onnxruntime/test/providers/qnn/isinf_test.cc
+++ b/onnxruntime/test/providers/qnn/isinf_test.cc
@@ -1,0 +1,121 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#if !defined(ORT_MINIMAL_BUILD)
+
+#include <limits>
+#include <string>
+#include <vector>
+
+#include "test/providers/qnn/qnn_test_utils.h"
+
+#include "gtest/gtest.h"
+
+namespace onnxruntime {
+namespace test {
+
+template <typename DataType>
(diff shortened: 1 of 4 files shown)


onnxruntime/test/providers/qnn/isinf_test.cc
```diff
@@ -0,0 +1,122 @@
...
+
+  RunIsInfTest<float>({TestInputDef<float>(input_shape, false, input_data)},
+                      attrs,
+                      ExpectedEPNodeAssignment::All,
+                      0.0f);
+}
+
+#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
+
+// HTP supports IsInf only in FP16 mode (per QAIRT HtpOpDefSupplement). The
```
yath1: Nit: HTP supports both in FP16 and FP32. Maybe we can add the same test for FP32 as well?

yath1 (approved):
lgtm

onnxruntime/test/providers/qnn/isinf_test.cc
qti-yuduo: Adding CPU only support for now... Will figure out HTP issues as follow up.

PR #405, opened 2026-05-19
[QNN-EP] mem leak issue found by ASan

chunghow-qti:
## Summary

Fix three issues found by AddressSanitizer on Linux x86_64:
1. **Heap UB** — `new-delete-type-mismatch` on every session destruction (`qnn_execution_provider.cc`)
2. **Memory leak** — output name `char*` leaked in test utility (`test_utils.cc`)
3. **Memory leak** — metadata lookup `char*` leaked in EP context test (`qnn_ep_context_test.cc`)

---

### ASan finding
```
==XXXX==ERROR: AddressSanitizer: new-delete-type-mismatch
  allocated:   40 bytes;
  deallocated: 32 bytes.
    #1 onnxruntime::QnnEp::ReleaseNodeComputeInfosImpl
       qnn_execution_provider.cc:2177
```
Trigger: `./onnxruntime_provider_test --gtest_filter='QnnHTPBackendTests.ConvU16S4S32_PerChannel'`

### Root cause
`CompileImpl` may allocate either `QnnNodeComputeInfo` (40B, regular path) or `GenieNodeComputeInfo` (56B, DLC context model path), and stores both as `OrtNodeComputeInfo*`. The release callback deleted through the C-API base pointer (`OrtNodeComputeInfo*`, 32B), which is a POD with no virtual destructor — the wrong destructor runs and the wrong size is passed to `operator delete`. For `GenieNodeComputeInfo` this also leaks the embedded `std::shared_ptr<GenieNodeBuilder>`.

### Fix
`OrtNodeComputeInfo` is a public C-API struct and cannot get a virtual destructor (ABI break). Introduce an internal C++ base `QnnNodeComputeInfoBase` with a virtual destructor; both `QnnNodeComputeInfo` and `GenieNodeComputeInfo` inherit from it. Release casts via the base so virtual dispatch picks the correct derived destructor.

```cpp
// New: qnn_node_compute_info_base.h
struct QnnNodeComputeInfoBase : OrtNodeComputeInfo {
  virtual ~QnnNodeComputeInfoBase() = default;
};

// Release callback
delete static_cast<QnnNodeComputeInfoBase*>(node_compute_infos[idx]);
```

---

### ASan finding
```
Direct leak of N byte(s) in 1 object(s) allocated from:
    #0 operator new
    #5 onnxruntime::StrDup
    #8 OrtApis::SessionGetOutputName
    #9 Ort::ConstSessionImpl::GetOutputNameAllocated
    #10 onnxruntime::test::RunAndVerifyOutputsWithEP
        test_utils.cc:260
```
Trigger: any test that goes through `RunQnnModelTest`, e.g. `QnnHTPBackendTests.RMSNorm1D_LastAxis`.

### Root cause
`GetOutputNameAllocated` returns `AllocatedStringPtr` (a `unique_ptr` with the ORT allocator's deleter). Calling `.release()` transferred ownership to the caller, but the raw `char*` was never freed.

### Fix
`output_names` is `std::vector<std::string>`. Switch to `.get()` so `std::string` copies the content and the smart ptr frees the original on scope exit.

```cpp
// Before
output_names.push_back(output_name.release());  // leaks char*

// After
output_names.push_back(output_name.get());      // std::string copies; smart ptr frees
```

---

### ASan finding
```
Direct leak of 28 byte(s) in 1 object(s) allocated from:
    #0 __interceptor_malloc
    #1 onnxruntime::test::MallocAllocator::Alloc
        qnn_ep_context_test.cc:3598
    #6 OrtApis::ModelMetadataLookupCustomMetadataMap
    #7 QnnHTPBackendTests_ModelCompatibility_GetCompatibility_Test::TestBody
        qnn_ep_context_test.cc:3652
```
Trigger: `./onnxruntime_provider_test --gtest_filter='QnnHTPBackendTests.ModelCompatibility_GetCompatibility'`

### Root cause
`ModelMetadataLookupCustomMetadataMap` allocates `char* val` via the caller-provided `MallocAllocator`. The test used `val` but never freed it (C-API ownership pattern: ORT allocates, caller frees).

### Fix
```cpp
ASSERT_TRUE(val != nullptr && expected_info.ToString() == val);
free(val);  // added
```

---

## Reproduce on Linux x86_64

```bash
export LD_LIBRARY_PATH=$PWD:<qairt>/lib/x86_64-linux-clang:$LD_LIBRARY_PATH
export ASAN_OPTIONS="halt_on_error=0:detect_leaks=1"
export LSAN_OPTIONS="exitcode=0"
./onnxruntime_provider_test --gtest_filter='*'
```

## Test plan

- [x] Linux x86_64 `--enable_address_sanitizer` build — all 969 tests pass (`--gtest_filter='*'`), zero ORT-side ASan findings
- [x] Remaining leaks are QAIRT SDK backend libraries only: `libQnnHtp.so` (5600B) and `libQnnCpu.so` (32B), suppressed via #419
- [x] CI green

### DIFF
--- a/onnxruntime/core/providers/qnn/genie/genie_node_compute_info.h
+++ b/onnxruntime/core/providers/qnn/genie/genie_node_compute_info.h
@@ -7,12 +7,13 @@
 
 #include "core/providers/qnn/genie/genie_node.h"
 #include "core/providers/qnn/ort_api.h"
+#include "core/providers/qnn/qnn_node_compute_info_base.h"
 
 namespace onnxruntime {
 
 class QnnEp;
 
-struct GenieNodeComputeInfo : OrtNodeComputeInfo {
+struct GenieNodeComputeInfo : QnnNodeComputeInfoBase {
   GenieNodeComputeInfo(QnnEp& ep, std::shared_ptr<GenieNodeBuilder> builder);
 
   static OrtStatus* ORT_API_CALL CreateStateImpl(OrtNodeComputeInfo* this_ptr,
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -2174,7 +2174,9 @@ void ORT_API_CALL QnnEp::ReleaseNodeComputeInfosImpl(OrtEp* this_ptr,
                                                      size_t num_node_compute_infos) noexcept {
   ORT_UNUSED_PARAMETER(this_ptr);
   for (size_t idx = 0; idx < num_node_compute_infos; ++idx) {
-    delete node_compute_infos[idx];
+    // All derived types share QnnNodeComputeInfoBase, which provides the virtual destructor
+    // required to safely delete through the OrtNodeComputeInfo C-API base.
+    delete static_cast<QnnNodeComputeInfoBase*>(node_compute_infos[idx]);
   }
 }
 
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.h
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.h
@@ -21,6 +21,7 @@
 #include "core/providers/qnn/builder/onnx_ctx_model_helper.h"
 #include "core/providers/qnn/qnn_telemetry.h"
 #include "core/providers/qnn/rpcmem_library.h"
+#include "core/providers/qnn/qnn_node_compute_info_base.h"
 #include "core/providers/qnn/genie/genie_api_loader.h"
 #include "core/providers/qnn/genie/genie_node.h"
 #include "core/providers/qnn/genie/genie_node_compute_info.h"
@@ -128,7 +129,7 @@ class QnnEp : public OrtEp, public ApiPtrs {
     return GetProviderOptionPrefix(name_) + key;
   }
 
-  struct QnnNodeComputeInfo : OrtNodeComputeInfo {
+  struct QnnNodeComputeInfo : QnnNodeComputeInfoBase {
     explicit QnnNodeComputeInfo(QnnEp& ep);
 
     static OrtStatus* ORT_API_CALL CreateStateImpl(OrtNodeComputeInfo* this_ptr,
--- a/onnxruntime/core/providers/qnn/qnn_node_compute_info_base.h
+++ b/onnxruntime/core/providers/qnn/qnn_node_compute_info_base.h
@@ -1,0 +1,19 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "core/providers/qnn/ort_api.h"
+
+namespace onnxruntime {
+
+// Internal C++ base class for all QNN EP node compute info objects.
+// Provides a virtual destructor so derived objects stored as
+// `OrtNodeComputeInfo*` can be safely deleted in `ReleaseNodeComputeInfosImpl`.
+// The C-API base `OrtNodeComputeInfo` is a POD struct with no virtual
+// destructor and cannot get one without breaking ABI.
+struct QnnNodeComputeInfoBase : OrtNodeComputeInfo {
+  virtual ~QnnNodeComputeInfoBase() = default;
+};
+
+}  // namespace onnxruntime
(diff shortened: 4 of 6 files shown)


qti-yuduo (approved):

minfhong-qti (approved):

kuanyul-qti (approved):

qti-chuteng:
### 🔴 Critical

**[C-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc:L2178`
`delete static_cast<QnnNodeComputeInfo*>(...)` is a wrong-type delete when the array element was produced by `CompileDlcContextModel`, which writes `GenieNodeComputeInfo` objects (see line 1968). The `~GenieNodeComputeInfo()` is never called, so its `std::shared_ptr<GenieNodeBuilder> builder` member leaks together with the underlying `GenieApi` handles, and the call remains C++ undefined behavior. The original ASan `new-delete type mismatch` warning was telling us that the C-API base `OrtNodeComputeInfo` has no virtual destructor — hard-coding the cast to one of the derived types only moves the UB from one branch to another; it does not eliminate it. Please introduce a small internal base class (e.g. `QnnNodeComputeInfoBase`) that derives from `OrtNodeComputeInfo` and provides a `virtual ~QnnNodeComputeInfoBase() = default;`, have both `QnnNodeComputeInfo` and `GenieNodeComputeInfo` derive from it, and `delete static_cast<QnnNodeComputeInfoBase*>(...)`. Add a unit test that exercises `CompileDlcContextModel` under ASan to prevent regression.

---

### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc:L2177`
The new comment "Cast to derived type before delete: base OrtNodeComputeInfo is a C-API POD with no virtual destructor" implies the design has been thought through, but it is only correct when every element is actually `QnnNodeComputeInfo`. With the existing `GenieNodeComputeInfo` path it actively misleads future readers. Either fix the dispatch (preferred, see C-1) or make the comment explicit: "Every object written into this array must currently be `QnnNodeComputeInfo`; new derived types must update the dispatch here." A matching `TODO` next to `CompileDlcContextModel` is recommended.

---

### 🔵 Minor

**[N-1]** PR #405 (overall)
This PR mixes three unrelated fixes across `qnn_execution_provider.cc`, `test/util/test_utils.cc`, and `test/providers/qnn/qnn_ep_context_test.cc`, and the PR description does not document each leak's ASan stack trace / triggering test or the chosen fix's rationale. Commit message also has a typo (`tyep` → `type`). Please expand the PR description, fix the typo, and ideally split the two pure-test fixes into a separate `[NFC]` / `[Test]` PR for clean cherry-pick / bisect.

chunghow-qti:
Thanks for the careful review @qti-chuteng — C-1 was a real bug (the `GenieNodeComputeInfo` shared_ptr leak), good catch.

### [C-1] — adopted Option A, pushed in `87619b6692`

Introduced an internal base `QnnNodeComputeInfoBase` with a virtual destructor; both `QnnNodeComputeInfo` and `GenieNodeComputeInfo` now derive from it. Release callback casts via the base so the correct derived destructor runs:

```cpp
// onnxruntime/core/providers/qnn/qnn_node_compute_info_base.h (new)
struct QnnNodeComputeInfoBase : OrtNodeComputeInfo {
  virtual ~QnnNodeComputeInfoBase() = default;
};

// qnn_execution_provider.cc:2177
delete static_cast<QnnNodeComputeInfoBase*>(node_compute_infos[idx]);
```

`OrtNodeComputeInfo` itself can't get a virtual destructor without breaking the public C-API ABI, hence the intermediate base.

> Add a unit test that exercises `CompileDlcContextModel` under ASan to prevent regression.

Filed as a follow-up — happy to add it in a separate PR once #419 (CI ASan workflow) lands so the test can actually run under sanitizer in CI.

### [M-1] — comment rewritten in the same commit

```cpp
// All derived types share QnnNodeComputeInfoBase, which provides the virtual destructor
// required to safely delete through the OrtNodeComputeInfo C-API base.
delete static_cast<QnnNodeComputeInfoBase*>(node_compute_infos[idx]);
```

Now points to the structural reason rather than implying a one-off cast is sufficient.

### [N-1] — description expanded; respectfully pushing back on typo + split

- **PR description**: updated. Each fix now has its ASan stack trace summary, the gtest filter that triggers it, and the rationale for the chosen fix.
- **Typo `tyep → type`**: leaving as-is. Editing the merged commit's message would require force-push and detach existing review threads; the cost outweighs the value of a cosmetic typo fix.
- **Split into 3 PRs**: leaving as one PR. The three fixes were all surfaced by the same single ASan run; splitting now means dragging this thread plus all `#419` (CI ASan) context into three places, and bisect/cherry-pick of these specific commits is unlikely in practice. The expanded description gives the same per-fix traceability that splitting would.

PTAL.

qti-chuteng (approved):

PR #397, opened 2026-05-16
[QNN EP] Add support to build ARM64 (ARM64x) Zip

qti-mbadnara:
This PR adds support to build `Windows ARM64 (ARM64x)` Zip as part of each RC and Release.

PR #406, opened 2026-05-19
[QNN EP] Fix Windows arm64x archive suffix

huaychou:
### Description

- Fix it as arm64x

### Motivation and Context

- The arm64x archive suffix is arm64ec now


qti-mbadnara:
This change has been incorporated as part of #397. This PR can be closed

PR #409, opened 2026-05-19
[QNN EP] Add support for If Op

qti-niscmami:
### Description

  Adds ONNX `If` operator support in the QNN Execution Provider through inline-both-branches translation that lowers `If` into a single `ElementWiseSelect`.

  ### Motivation & Context

  `If` is a control-flow op that selects between two subgraphs based on a runtime boolean condition. It appears in models with conditional execution paths (input-dependent processing branches, optional post-processing, fallback paths). Previously unsupported in QNN EP.

  https://onnx.ai/onnx/operators/onnx__If.html

  | Opset | Type Constraint V (output types) |
  |---|---|
  | 1, 11, 13, 16, 19, 21, 23, 24, 25 | tensor (sequence/optional rejected) |

### Translation

Both branches are inlined into the parent QNN graph and a single `ElementWiseSelect` picks between their outputs at runtime:

**ONNX (control flow — one branch runs):**

```
                 ┌─── If node ────────────────┐
                 │                            │
  outer ────────►│  then_branch: ... ► then_o │
                 │                            │ ──► if_out
  outer ────────►│  else_branch: ... ► else_o │
                 │                            │
                 └────────────────▲───────────┘
                                  │
  cond ───────────────────────────┘
```

**QNN (dataflow — both branches inlined, select picks):**

```
                 ┌──► [then_branch ops inlined] ──► then_out ───┐
                 │                                              │
  outer ─────────┤                                              │
                 │                                              ├──► ElementWiseSelect ──► if_out
                 │                                              │             ▲
                 └──► [else_branch ops inlined] ──► else_out ───┘             │
                                                                              │
  cond ───────────────────────────────────────────────────────────────────────┘
```

 For each branch, op-builders are dispatched recursively on the branch's nodes (`OrtNodeUnit` per node). Both branches always execute on-device — this is the trade-off vs. native `QNN_OP_IF`'s real conditional dispatch.

  ### Restrictions enforced in `IsOpSupported`

  - Condition is scalar — empty shape, or every dim equals 1
  - Each branch produces exactly 1 output
  - Branch outputs have identical shape and dtype
  - Branch outputs are plain tensors (sequence / optional rejected)
  - Every op inside both branches has a registered QNN op-builder
  - Branch terminus name does not collide with the `If` output name
  - Each implicit input is a graph input of the QNN partition (cross-partition implicit inputs declined; ORT's partitioner doesn't promote them to partition inputs, leaving the branch consumer unconnected in QNN — error `6007 UNCONNECTED_NODE`)

  ### Files Changed

  | File | Change |
  |---|---|
  | `core/providers/qnn/builder/opbuilder/if_op_builder.cc` | **New** `IfOpBuilder` inline-both-branches translation |
  | `core/providers/qnn/builder/op_builder_factory.h` | Added `CreateIfOpBuilder` declaration |
  | `core/providers/qnn/builder/op_builder_factory.cc` | Registered `CreateIfOpBuilder("If", *this)` |
  | `core/providers/qnn/builder/qnn_model_wrapper.h` | Added `Push/PopBranchGraphScope` and scope-aware `FindInitializer` so branch initializers resolve as `QNN_TENSOR_TYPE_STATIC` |
  | `test/providers/qnn/if_test.cc` | **New** test file with CPU, HTP, GPU sections |

  ### Implementation Notes

  - **Branch graph scope stack.** ORT folds `Constant` ops inside subgraphs into branch-graph initializers. `FindInitializer` now walks a stack of branch graphs (pushed/popped during branch translation) before falling back to the parent graph. Without this, `IsConstantInput` returns false for folded constants
  and the resulting tensor is classified as `NATIVE`, expecting a producer node that no longer exists (QNN error 6007: `Input tensor doesn't have any producers`).

  - **Implicit inputs.** Outer-scope tensors consumed inside branch subgraphs are not surfaced as `If` node inputs by ORT's partitioner. `ProcessInputs` calls `Node_GetImplicitInputs` and registers each as a QNN tensor before branch translation, so branch op-builders can resolve them. If an implicit input is
  not already produced by an upstream QNN op in the partition and is not a partition graph input, the `If` is declined so the partitioner falls it back to CPU.

  - **Pure-constant branches.** When ORT folds an entire branch away (0 compute nodes, 1 initializer), `EnsureBranchOutputRegistered` explicitly registers the branch output from its initializer table after `TranslateBranch`. Covers the `Then=Constant`, `Else=Constant`, and `Both=Constant` test cases.

  - **Tensor name collision.** ONNX permits a branch terminus to share its name with the parent `If` output. QNN cannot represent two tensors with the same name. The check rejects this case in `IsOpSupported`; the `If` falls to CPU EP. Renaming is deferred.

  - **Single registration covers all opsets.** The op-builder claims `"If"` by op name; ORT routes opset variants 1, 11, 13, 16, 19, 21, 23, 24, 25 through it. The structural spec is unchanged across versions; recent opsets only add new dtypes which are filtered by the per-op-builder dtype allowlist.

  ### Test coverage

  | Backend | Tests |
  |---|---|
  | CPU | `BasicBranches`, `NameCollision_DeclinesFusion`, `ThenConstant_ElseDynamic`, `ThenDynamic_ElseConstant`, `BothBranchesConstant`, `CrossPartitionImplicitInput` (declined) |
  | HTP (FP32-as-FP16) | `BasicBranches`, `ThenConstant_ElseDynamic`, `ThenDynamic_ElseConstant`, `BothBranchesConstant`, `CrossPartitionImplicitInput` (declined); `DISABLED_If_QDQ_U8_BranchesWrapMul` (QDQ-in-branch fusion not yet supported) |
  | GPU | Same four positive cases as HTP, gated by `_M_ARM64` |


qti-chuteng:
### 🔴 Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → TranslateBranch()`
Branch ops are dispatched directly through `AddToModelBuilder`, bypassing `BaseOpBuilder::IsOpSupported` (which runs `ProcessDataTypes`). The EP-side `CheckCpuDataTypes` / `CheckHtpDataTypes` / `CheckGpuDataTypes` filter therefore never fires for ops inside a branch, so a branch containing an HTP-disallowed dtype (e.g., FP64) is silently accepted at partition time and crashes later in QNN. Branch on `do_op_validation` and call `IsOpSupported` instead when `do_op_validation == true`.
```suggestion
    Ort::Status s = do_op_validation
        ? builder->IsOpSupported(qmw, unit, logger)
        : builder->AddToModelBuilder(qmw, unit, logger, false);
```

---

### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → GetBranchGraph()`
`Node_GetNumSubgraphs` and `Node_GetSubgraphs` return `OrtStatus*` that is compared against `nullptr` and dropped — `OrtStatus` is heap-allocated and must be released. This both leaks memory and collapses three distinct failure modes (API failure / zero subgraphs / attribute name not found) into a single `nullptr` return, hiding the real root cause from the caller's `RETURN_IF_NOT` message. Refactor to return `Ort::Status` and use `ORT_CXX_RETURN_ON_API_FAIL`.

**[M-2]** `onnxruntime/test/providers/qnn/if_test.cc → If_Fp32_DynamicCond_ShapeMismatch_DeclinesFusion / NameCollision_DeclinesFusion`
Both negative tests assert `ExpectedEPNodeAssignment::Some`, but the model has a `Sigmoid` upstream of the `If` that is always partitioned to QNN. So `::Some` is trivially satisfied even if the If's rejection logic regresses and the If is wrongly partitioned. Use `ExpectedEPNodeAssignment::None` on a simplified `If`-only model, or use `QnnGraphChecker` to assert the If specifically is on the CPU side.

**[M-3]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h`
`git rebase origin/main pr-409` produces a conflict in this header — recent main commits (e.g., #339 standalone Q/DQ folding, #295 QNN tensor-ID propagation) overlap with the PR's `FindInitializer` refactor. Rebase onto latest main and double-check the interaction between the new branch scope stack and `folded_constant_tensors_`.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → EnsureBranchOutputRegistered()`
`GetBranchSingleOutput` already retrieved the branch's outputs and validated `n == 1`; this function fetches them again for `outs[0]`. Have `GetBranchSingleOutput` return the `const OrtValueInfo*` so this re-fetch can be eliminated and the single-output invariant is owned by one function.

**[N-2]** `onnxruntime/test/providers/qnn/if_test.cc`
No QDQ test path. Real models commonly wrap QDQ subgraphs (DQ→Op→Q) inside If branches; add at least one HTP QDQ test exercising that shape.

**[N-3]** `onnxruntime/test/providers/qnn/if_test.cc → BuildIfTestCase()`
Tests use `MakeInputBool("cond", {1})` (rank-1, shape `[1]`), but the ONNX `If` spec defines `cond` as scalar (rank-0). Add a scalar-cond case. Also add a case where the If output feeds a downstream op (e.g., `If → Add`) to verify the registered output tensor is consumable.

qti-niscmami:
> ### 🔴 Critical
> **[C-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → TranslateBranch()` Branch ops are dispatched directly through `AddToModelBuilder`, bypassing `BaseOpBuilder::IsOpSupported` (which runs `ProcessDataTypes`). The EP-side `CheckCpuDataTypes` / `CheckHtpDataTypes` / `CheckGpuDataTypes` filter therefore never fires for ops inside a branch, so a branch containing an HTP-disallowed dtype (e.g., FP64) is silently accepted at partition time and crashes later in QNN. Branch on `do_op_validation` and call `IsOpSupported` instead when `do_op_validation == true`.
> 
> ```
>     Ort::Status s = do_op_validation
>         ? builder->IsOpSupported(qmw, unit, logger)
>         : builder->AddToModelBuilder(qmw, unit, logger, false);
> ```
> 
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → GetBranchGraph()` `Node_GetNumSubgraphs` and `Node_GetSubgraphs` return `OrtStatus*` that is compared against `nullptr` and dropped — `OrtStatus` is heap-allocated and must be released. This both leaks memory and collapses three distinct failure modes (API failure / zero subgraphs / attribute name not found) into a single `nullptr` return, hiding the real root cause from the caller's `RETURN_IF_NOT` message. Refactor to return `Ort::Status` and use `ORT_CXX_RETURN_ON_API_FAIL`.
> 
> **[M-2]** `onnxruntime/test/providers/qnn/if_test.cc → If_Fp32_DynamicCond_ShapeMismatch_DeclinesFusion / NameCollision_DeclinesFusion` Both negative tests assert `ExpectedEPNodeAssignment::Some`, but the model has a `Sigmoid` upstream of the `If` that is always partitioned to QNN. So `::Some` is trivially satisfied even if the If's rejection logic regresses and the If is wrongly partitioned. Use `ExpectedEPNodeAssignment::None` on a simplified `If`-only model, or use `QnnGraphChecker` to assert the If specifically is on the CPU side.
> 
> **[M-3]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h` `git rebase origin/main pr-409` produces a conflict in this header — recent main commits (e.g., #339 standalone Q/DQ folding, #295 QNN tensor-ID propagation) overlap with the PR's `FindInitializer` refactor. Rebase onto latest main and double-check the interaction between the new branch scope stack and `folded_constant_tensors_`.
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → EnsureBranchOutputRegistered()` `GetBranchSingleOutput` already retrieved the branch's outputs and validated `n == 1`; this function fetches them again for `outs[0]`. Have `GetBranchSingleOutput` return the `const OrtValueInfo*` so this re-fetch can be eliminated and the single-output invariant is owned by one function.
> 
> **[N-2]** `onnxruntime/test/providers/qnn/if_test.cc` No QDQ test path. Real models commonly wrap QDQ subgraphs (DQ→Op→Q) inside If branches; add at least one HTP QDQ test exercising that shape.
> 
> **[N-3]** `onnxruntime/test/providers/qnn/if_test.cc → BuildIfTestCase()` Tests use `MakeInputBool("cond", {1})` (rank-1, shape `[1]`), but the ONNX `If` spec defines `cond` as scalar (rank-0). Add a scalar-cond case. Also add a case where the If output feeds a downstream op (e.g., `If → Add`) to verify the registered output tensor is consumable.

Addressed all feedback.

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → IfOpBuilder::ProcessInputs / ProcessAttributesAndOutputs`
`If` is translated to a single `ElementWiseSelect(cond, then_out, else_out)`, so **both branches always execute** (diverges from ONNX `If` control-flow semantics — only correct under the PR's single-output / identical-shape-dtype / scalar-cond constraints). More importantly, implicit inputs (outer-scope tensors) are manually registered in `ProcessInputs`, but every positive test feeds the implicit input `x` from an upstream `Sigmoid` that is itself QNN-eligible, so it stays partition-internal. No test covers an implicit input produced by a CPU-partition node, which is exactly where the manual cross-partition wiring could break (SetupTensors I/O name mismatch / silent miscompute). Add that test and document the supported `If` subset + the both-branches-always-run limitation.

**[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:86`
`GetBranchSingleOutput` does `out_name = name;` without null-checking the result of `GetValueInfoName`. `std::string(nullptr)` is UB. The sibling `MakeIODefFromValueInfo` at line 51 already guards this.
```suggestion
  out_name = name ? name : "";
```

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → ProcessAttributesAndOutputs`
`cond` is fed directly as the first operand of `ElementWiseSelect` with no check that its dtype is BOOL or that it broadcasts to the output shape. ONNX spec'd `cond` as a scalar boolean. Validate dtype/shape early and reject with NotImplemented; also confirm HTP `Select` accepts a bool cond (BOOL_8 is generally unsupported on HTP — see Where op behavior).

**[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → TranslateBranch / EnsureBranchOutputRegistered`
`PushBranchGraphScope` / `PopBranchGraphScope` are manually paired around calls that can throw `Ort::Exception` (`MakeTensorWrapper`, op-builder dispatch). On throw the Pop is skipped, leaving a stale branch graph on the scope stack and corrupting later `FindInitializer` lookups. Use a RAII scope guard instead.

**[N-3]** `onnxruntime/test/providers/qnn/if_test.cc:860`
The HTP path only tests FP32-as-FP16; the sole real-quantized test `DISABLED_If_QDQ_U8_BranchesWrapMul` is disabled (branch DQ/Q not wrapped in OrtNodeUnits → HTP error 3110). Per [C06] a new op should have HTP u8/u16 coverage. Document this limitation in the PR description and open a follow-up.

qti-yuduo (approved):

qti-niscmami:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → IfOpBuilder::ProcessInputs / ProcessAttributesAndOutputs` `If` is translated to a single `ElementWiseSelect(cond, then_out, else_out)`, so **both branches always execute** (diverges from ONNX `If` control-flow semantics — only correct under the PR's single-output / identical-shape-dtype / scalar-cond constraints). More importantly, implicit inputs (outer-scope tensors) are manually registered in `ProcessInputs`, but every positive test feeds the implicit input `x` from an upstream `Sigmoid` that is itself QNN-eligible, so it stays partition-internal. No test covers an implicit input produced by a CPU-partition node, which is exactly where the manual cross-partition wiring could break (SetupTensors I/O name mismatch / silent miscompute). Add that test and document the supported `If` subset + the both-branches-always-run limitation.
> 
> **[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:86` `GetBranchSingleOutput` does `out_name = name;` without null-checking the result of `GetValueInfoName`. `std::string(nullptr)` is UB. The sibling `MakeIODefFromValueInfo` at line 51 already guards this.
> 
> ```
>   out_name = name ? name : "";
> ```
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → ProcessAttributesAndOutputs` `cond` is fed directly as the first operand of `ElementWiseSelect` with no check that its dtype is BOOL or that it broadcasts to the output shape. ONNX spec'd `cond` as a scalar boolean. Validate dtype/shape early and reject with NotImplemented; also confirm HTP `Select` accepts a bool cond (BOOL_8 is generally unsupported on HTP — see Where op behavior).
> 
> **[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc → TranslateBranch / EnsureBranchOutputRegistered` `PushBranchGraphScope` / `PopBranchGraphScope` are manually paired around calls that can throw `Ort::Exception` (`MakeTensorWrapper`, op-builder dispatch). On throw the Pop is skipped, leaving a stale branch graph on the scope stack and corrupting later `FindInitializer` lookups. Use a RAII scope guard instead.
> 
> **[N-3]** `onnxruntime/test/providers/qnn/if_test.cc:860` The HTP path only tests FP32-as-FP16; the sole real-quantized test `DISABLED_If_QDQ_U8_BranchesWrapMul` is disabled (branch DQ/Q not wrapped in OrtNodeUnits → HTP error 3110). Per [C06] a new op should have HTP u8/u16 coverage. Document this limitation in the PR description and open a follow-up.

Addressed all comments.

qti-chuteng:
### 🟠 Major

**[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L165`
`IfOpBuilder` has no `IsOpSupported` override, so `BaseOpBuilder::IsOpSupported` redirects to `AddToModelBuilder(do_op_validation=true)` and the validation phase runs the full `ProcessInputs`/`ProcessAttributesAndOutputs`, including `AddTensorWrapper` and `TranslateBranch` on both branches. Per [C40], `IsOpSupported` must not mutate `QnnModelWrapper` state. Please confirm validate→compose produces no `SetupTensors` name mismatch (esp. shared-wrapper / weight-sharing) and add a comment explaining why this is safe.

**[M-3]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L127`
`TranslateBranch` flattens both branches' internal tensors into one QNN graph using original ONNX names. ONNX subgraphs are independent namespaces, so two branches may legally reuse an internal name (e.g. `mul_out`); `AddTensorWrapper` is idempotent on a name clash and silently mis-wires the second branch onto the first's tensor. The branch-output collision is checked (L298) but branch-internal tensors are not. The tests prefix names with `then_/else_`, masking this. Add branch-unique prefixing (or decline on internal-name intersection) and a regression test with a shared internal name.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L273`
`cond_is_scalar` accepts any all-dims-equal-1 tensor, looser than the ONNX `If` spec (rank-0 scalar). Mostly safe after Select broadcasting; document the intentional relaxation or tighten it.

**[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L32`
`MakeIODefFromValueInfo` leaves `quant_param` unset, so implicit inputs / pure-constant branch outputs are treated as non-quantized. This matches the disabled QDQ test limitation; add a comment noting QDQ is not yet supported here.


**[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L162`
The `PushBranchGraphScope` comment says "Used by IfFusion", but this is an op builder (`IfOpBuilder`), not an `IQnnNodeGroup` fusion. Rename to avoid sending readers after a non-existent fusion class.

qti-niscmami:
> ### 🟠 Major
> **[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L165` `IfOpBuilder` has no `IsOpSupported` override, so `BaseOpBuilder::IsOpSupported` redirects to `AddToModelBuilder(do_op_validation=true)` and the validation phase runs the full `ProcessInputs`/`ProcessAttributesAndOutputs`, including `AddTensorWrapper` and `TranslateBranch` on both branches. Per [C40], `IsOpSupported` must not mutate `QnnModelWrapper` state. Please confirm validate→compose produces no `SetupTensors` name mismatch (esp. shared-wrapper / weight-sharing) and add a comment explaining why this is safe.
> 
> **[M-3]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L127` `TranslateBranch` flattens both branches' internal tensors into one QNN graph using original ONNX names. ONNX subgraphs are independent namespaces, so two branches may legally reuse an internal name (e.g. `mul_out`); `AddTensorWrapper` is idempotent on a name clash and silently mis-wires the second branch onto the first's tensor. The branch-output collision is checked (L298) but branch-internal tensors are not. The tests prefix names with `then_/else_`, masking this. Add branch-unique prefixing (or decline on internal-name intersection) and a regression test with a shared internal name.
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L273` `cond_is_scalar` accepts any all-dims-equal-1 tensor, looser than the ONNX `If` spec (rank-0 scalar). Mostly safe after Select broadcasting; document the intentional relaxation or tighten it.
> 
> **[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/if_op_builder.cc:L32` `MakeIODefFromValueInfo` leaves `quant_param` unset, so implicit inputs / pure-constant branch outputs are treated as non-quantized. This matches the disabled QDQ test limitation; add a comment noting QDQ is not yet supported here.
> 
> **[N-3]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h:L162` The `PushBranchGraphScope` comment says "Used by IfFusion", but this is an op builder (`IfOpBuilder`), not an `IQnnNodeGroup` fusion. Rename to avoid sending readers after a non-existent fusion class.

Addressed all comments.

For M-2 I have added a comment explaining why it is safe.

qti-chuteng (approved):

PR #413, opened 2026-05-19
Revert "Load target backend library for validation when dumping DLC via QnnIr"

qti-yuduo:
Revert "Load target backend library for validation when dumping DLC via QnnIr (#218)"

## Motivation

The new code routes `QnnBackend_validateOpConfig` to a host-side libQnnHtp.so during partitioning. But:

- A `.dlc` is a portable QNN graph artifact. Its consumers (qnn-context-binary-generator, qnn-net-run, downstream tooling) finalize for a specific target with full target arch information — that's where capability checks belong. The DLC-dump path is target-agnostic by contract. It even should be able to execute on GPU/CPU.
- EP not necessarily always get `htp_arch / soc_model`. With no device context, the host HTP shim defaults its op-package arch capability check to v68 and rejects (**false-positive**) kernels that need v73+. `DLC` actually supported for v73 but got rejected due to we validate against default (v68). 

##  Caveat

This revert also removes the unrelated qnn_test_utils.{cc,h} GetGPUSupport() refactor that was bundled into 2366b4de. That improvement should be re-applied as a separate focused commit. @yuhuchua-qti to confirm if we need follow up change here to bring those back. 

##  Future work (not in this PR)

A deliberate fail-fast HTP-validation-during-DLC-dump feature is possible, but would require an explicit opt-in option, a target-arch contract from callers, and QnnDevice_create plumbing against the validator interface so the shim has actual arch context. Separate design conversation.


quic-muchhsu (approved):
Approve to unblock merge.

yath1 (approved):
Approving to restore existing functionality

PR #266, opened 2026-04-20
[QNN EP] Add RandomNormalLike operator support

qti-luharj:
### Summary
This PR adds support for the ONNX `RandomNormalLike` operator in the QNN execution provider.

### Details
- Introduces a new `RandomNormalLikeOpBuilder` for QNN EP.
- Supports `mean`, `scale`, and optional `seed` attributes.
- Registers the new op builder through the QNN op builder factory.
- Adds a QNN EP unit test for RandomNormalLike.

### Backend behavior
- For NPU backends, the operator produces a `UFIXED_POINT_8` intermediate output and inserts a `Dequantize` node to convert to the requested output type.
- For non‑NPU backends, the operator directly produces output in the original data type.

### Files changed
- `onnxruntime/core/providers/qnn/builder/op_builder_factory.{h,cc}`
- `onnxruntime/core/providers/qnn/builder/opbuilder/randomnormallike_op_builder.cc`
- `onnxruntime/test/providers/qnn/randomnormallike_test.cc`

PR #415, opened 2026-05-19
[QNN EP] Fix "Zero tensor size!" for RandomUniformLike / RandomNormalLike

qti-niscmami:
### Description
Register the ONNX `x` input as a `QnnTensorWrapper` (APP_WRITE when it
is a fused-subgraph input) in the `RandomUniformLike` and
`RandomNormalLike` op builders, without adding it to the QNN node's
`input_names`. The QNN op still consumes only the synthesized static
`*_shape` tensor (and optional `*_ort_qnn_ep_seed`).

### Motivation and Context
QNN's `RandomUniformLike` / `RandomNormalLike` ops only take a static
shape tensor. 

The ONNX `x` input is used purely for shape inference at
compile time. The op builders correctly bake `x`'s shape into a static
input, but never registered `x` itself as a tensor wrapper. ORT's
runtime input plumbing iterates the fused EP subgraph's inputs (which
still include `x`) and looks each one up in the wrapper map; with no
wrapper for `x`, the resulting list was empty and `SetupTensors` failed
its `tensor_count == 0` guard.

### Test plan / kill-test fence

  The new tests QnnHTPBackendTests.RandomNormalLike_GraphInputRegistered
  and QnnHTPBackendTests.RandomUniformLike_GraphInputRegistered build a
  solo-node graph whose only consumer of x is the random op. With the
  fix in place, both pass. Verified the tests catch the regression by
  reverting the MakeTensorWrapper block in both op builders, rebuilding
  onnxruntime_provider_test (linux-x86_64, Release, QAIRT 2.46.0.260424),
  and running:

    ./onnxruntime_provider_test \
        --gtest_filter='QnnHTPBackendTests.RandomNormalLike_GraphInputRegistered:QnnHTPBackendTests.RandomUniformLike_GraphInputRegistered'

  Both tests fail with:

    [E:onnxruntime:, qnn_model.cc:365 SetupQnnInputOutput]
      Failed to setup QNN input tensors for graph:
      QNNExecutionProvider_..._1_0. Zero tensor size!

  i.e., they trip the same tensor_count == 0 guard the fix is preventing.
  Restoring the block makes both tests pass.


qti-yuduo:
Nice fix, would be better to add unit test that was failing and now passing.

qti-niscmami:
> Nice fix, would be better to add unit test that was failing and now passing.

Added them.

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/randomuniformlike_op_builder.cc → ProcessInputs`
`RandomNormalLike` already rejects dynamic shapes at the very top of `ProcessInputs()` (pattern [C46], PR #266), but the symmetric `RandomUniformLike` in this PR bakes the input shape into a `QNN_TENSOR_TYPE_STATIC` tensor the same way and is missing the matching guard. With a dynamic / unknown input shape, the current code will silently bake bogus dimensions (negative-as-huge-unsigned) into the QNN graph and fail later with an unrelated message.
```suggestion
  // Explicitly reject dynamic (symbolic/unknown) input shapes.
  // This builder materializes a static shape tensor at compile time, so all
  // dimensions must be statically known.
  if (!input_tensor.shape.has_value()) {
    return MAKE_EP_FAIL(
        "QNN EP RandomUniformLike requires static input dimensions. "
        "Input shape is unknown.");
  }
  for (const auto& dim : *input_tensor.shape) {
    if (dim < 0) {
      return MAKE_EP_FAIL(
          "QNN EP RandomUniformLike requires static input dimensions. "
          "Found symbolic/unknown dimension in input shape.");
    }
  }
```

---

### 🔵 Minor

**[N-1]** `onnxruntime/test/providers/qnn/randomuniformlike_test.cc:25-47` (also `randomnormallike_test.cc:25-47`)
`struct ModelAndBuilder` and `CreateModelInMemory()` are now copied a third time (existing copy in `qnn_basic_test.cc:586`). Promote them to `qnn_test_utils.h/.cc` so future regression tests stop duplicating boilerplate. Can be a follow-up NFC PR if the scope is too large.

**[N-2]** PR description (general)
[G02] kill-test fence not recorded. Please confirm that with the new `MakeTensorWrapper` block reverted in both op builders, `RandomNormalLike_GraphInputRegistered` and `RandomUniformLike_GraphInputRegistered` both fail with `Zero tensor size!`, and add that statement to the PR description. Without it, reviewers cannot tell whether the new tests actually catch the regression.

qti-niscmami:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/randomuniformlike_op_builder.cc → ProcessInputs` `RandomNormalLike` already rejects dynamic shapes at the very top of `ProcessInputs()` (pattern [C46], PR #266), but the symmetric `RandomUniformLike` in this PR bakes the input shape into a `QNN_TENSOR_TYPE_STATIC` tensor the same way and is missing the matching guard. With a dynamic / unknown input shape, the current code will silently bake bogus dimensions (negative-as-huge-unsigned) into the QNN graph and fail later with an unrelated message.
> 
> ```
>   // Explicitly reject dynamic (symbolic/unknown) input shapes.
>   // This builder materializes a static shape tensor at compile time, so all
>   // dimensions must be statically known.
>   if (!input_tensor.shape.has_value()) {
>     return MAKE_EP_FAIL(
>         "QNN EP RandomUniformLike requires static input dimensions. "
>         "Input shape is unknown.");
>   }
>   for (const auto& dim : *input_tensor.shape) {
>     if (dim < 0) {
>       return MAKE_EP_FAIL(
>           "QNN EP RandomUniformLike requires static input dimensions. "
>           "Found symbolic/unknown dimension in input shape.");
>     }
>   }
> ```
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/test/providers/qnn/randomuniformlike_test.cc:25-47` (also `randomnormallike_test.cc:25-47`) `struct ModelAndBuilder` and `CreateModelInMemory()` are now copied a third time (existing copy in `qnn_basic_test.cc:586`). Promote them to `qnn_test_utils.h/.cc` so future regression tests stop duplicating boilerplate. Can be a follow-up NFC PR if the scope is too large.
> 
> **[N-2]** PR description (general) [G02] kill-test fence not recorded. Please confirm that with the new `MakeTensorWrapper` block reverted in both op builders, `RandomNormalLike_GraphInputRegistered` and `RandomUniformLike_GraphInputRegistered` both fail with `Zero tensor size!`, and add that statement to the PR description. Without it, reviewers cannot tell whether the new tests actually catch the regression.

Addressed all comments

qti-yuduo (approved):

qti-chuteng (approved):

PR #417, opened 2026-05-20
Add layernorm decomposition when scale/bais is not aligned with LN axes.

quic-muchhsu:
### Description

Onnx allow scale/bias is not aligned with LN's axes. e.g. [1, 4, 8] ->LN(scale=(4, 8), axes=2), but QNN expects scale has only 1 on none normalized dims. Same for bias.
This PR will decompose such LN to LN -> Mul (scale) + Add (bias) to extend QNN support to align with ONNX models.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc
@@ -516,11 +516,13 @@ Ort::Status LayerNormalizationOpBuilder::BuildDecomposedLayerNorm(QnnModelWrappe
   std::string current = ln_out_name;
 
   if (externalize_scale) {
-    // Owned copy: the externalize_bias=false path moves final_output_info.quant_param into the
-    // Mul output tensor below, so an alias-then-move on .quant_param would leave mul_out_qp
-    // dangling for the requant predicates / GetPerTensorScaleOffset reads.
-    const QnnQuantParamsWrapper mul_out_qp =
-        externalize_bias ? ln_intermediate_qp.Copy() : final_output_info.quant_param.Copy();
+    // Mul produces z*gamma, whose magnitude is bounded by gamma * sqrt(N-1) — well past
+    // ln_intermediate_qp's sqrt(N-1) range whenever |gamma| > 1. final_output_info.quant_param
+    // is sized for z*gamma+beta and trivially covers z*gamma. Owned copy: the
+    // externalize_bias=false branch moves .quant_param into the Mul output tensor below; the
+    // externalize_bias=true branch hands it to the Add output later. Either way, mul_out_qp must
+    // not alias storage that gets moved out before the requant predicates / scale reads run.
+    const QnnQuantParamsWrapper mul_out_qp = final_output_info.quant_param.Copy();
 
     // QNN ELEMENT_WISE_MULTIPLY requires both operands to share a dtype. In QDQ pipelines (e.g.
     // A16W8) the LN output adopts X's dtype while the user scale is still in its own (often
@@ -577,17 +579,17 @@ Ort::Status LayerNormalizationOpBuilder::BuildDecomposedLayerNorm(QnnModelWrappe
       mul_scale_name = requant_scale_name;
     }  // End of requantize scale
 
-    // Mul output: either an internal handoff to the Add (NATIVE, x_dtype, ln_intermediate_qp,
-    // x_shape) or the graph output (final_tensor_type, final dtype/qp/shape). Splitting the two
-    // cases makes the move-from-final_output_info contract obvious — the final-output branch is
-    // the sole writer, so reads of final_output_info.{quant_param,shape} below it are forbidden.
+    // Mul output: either an internal handoff to the Add (NATIVE, x_dtype, mul_out_qp, x_shape)
+    // or the graph output (final_tensor_type, final dtype/qp/shape). Splitting the two cases
+    // makes the move-from-final_output_info contract obvious — the final-output branch is the
+    // sole writer, so reads of final_output_info.{quant_param,shape} below it are forbidden.
     std::string mul_out_name;
     if (externalize_bias) {
       mul_out_name = utils::UniqueNameGenerator().New(node_unit, "_ln_decomposed_mul_out");
       QnnTensorWrapper mul_out_tensor(mul_out_name,
                                       QNN_TENSOR_TYPE_NATIVE,
                                       x_qnn_data_type,
-                                      ln_intermediate_qp.Copy(),
+                                      mul_out_qp.Copy(),
                                       std::vector<uint32_t>(x_shape));
       RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(mul_out_tensor)),
                     "Failed to add decomposed Mul output tensor.");
@@ -613,8 +615,10 @@ Ort::Status LayerNormalizationOpBuilder::BuildDecomposedLayerNorm(QnnModelWrappe
   }
 
   if (externalize_bias) {
-    // Add LHS is the Mul output (when externalize_scale) or the LN output, both encoded with
-    // ln_intermediate_qp.
+    // Bias is the Add RHS; size it to cover beta directly. final_output_info.quant_param is
+    // sized for z*gamma+beta and so covers beta alone with room to spare. ln_intermediate_qp
+    // is too narrow here (its sqrt(N-1) range was sized for z, not beta — a bias whose magnitude
+    // exceeds sqrt(N-1) would silently saturate during requantize).
     std::string bias_name = input_names[2];
     TensorInfo bias_info{};
     RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[2], bias_info));
@@ -630,15 +634,15 @@ Ort::Status LayerNormalizationOpBuilder::BuildDecomposedLayerNorm(QnnModelWrappe
                     "LayerNorm decomposition: externalized bias must be a static initializer.");
       RETURN_IF_NOT(bias_info.quant_param.IsPerTensor(/*include_bw*/ true),
                     "LayerNorm decomposition: externalized bias must be per-tensor quantized.");
-      RETURN_IF_NOT(ln_intermediate_qp.IsPerTensor(/*include_bw*/ true),
-                    "LayerNorm decomposition: requantized bias requires per-tensor intermediate quant params.");
+      RETURN_IF_NOT(final_output_info.quant_param.IsPerTensor(/*include_bw*/ true),
+                    "LayerNorm decomposition: requantized bias requires per-tensor final output quant params.");
 
       float src_scale = 0.0f;
       int32_t src_offset = 0;
       RETURN_IF_ERROR(bias_info.quant_param.GetPerTensorScaleOffset(src_scale, src_offset));
       float dst_scale = 0.0f;
       int32_t dst_offset = 0;
-      RETURN_IF_ERROR(ln_intermediate_qp.GetPerTensorScaleOffset(dst_scale, dst_offset));
+      RETURN_IF_ERROR(final_output_info.quant_param.GetPerTensorScaleOffset(dst_scale, dst_offset));
 
       // Allowlist bias dtypes upfront so a future QDQ pipeline emitting an unknown bias dtype
       // fails loudly here, rather than silently producing an all-zero requantized bias via the
@@ -671,7 +675,7 @@ Ort::Status LayerNormalizationOpBuilder::BuildDecomposedLayerNorm(QnnModelWrappe
       QnnTensorWrapper requant_bias_tensor(requant_bias_name,
                                            QNN_TENSOR_TYPE_STATIC,
                                            x_qnn_data_type,
-                                           ln_intermediate_qp.Copy(),
+                                           final_output_info.quant_param.Copy(),
                                            std::vector<uint32_t>(bias_info.shape),
                                            std::move(requant_bias));
       RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(requant_bias_tensor)),
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc
@@ -2,9 +2,7 @@
 // Licensed under the MIT License.
 
 #include <algorithm>
-#include <cassert>
 #include <cmath>
-#include <cstring>
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
@@ -39,6 +37,7 @@ class LayerNormalizationOpBuilder : public BaseOpBuilder {
   // True iff `shape` (right-aligned to input_rank) has any non-1 dim at an axis < ln_axis.
   // Such dims fall outside QNN/ONNX LayerNorm's normalized axes [ln_axis, input_rank), so the
   // tensor cannot be fed directly to LayerNorm and must be applied via an outer Mul/Add instead.
+  // Precondition: shape.size() <= input_rank (enforced by IsOpSupported).
   static bool HasNonOneDimBeforeAxis(const std::vector<uint32_t>& shape,
                                      size_t input_rank,
                                      size_t ln_axis) {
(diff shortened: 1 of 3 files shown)


qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc → BuildDecomposedLayerNorm()`
The decomposed LN node is built with only `{X, ln_scale_name}` (line ~397) and no bias input. On QNN SDK 2.24–2.27 (the same range guarded by `#if QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 17 && QNN_API_VERSION_MINOR <= 20` in `ProcessInputs`), a LayerNorm without a bias input intermittently fails graph finalize. The decomposition path must replicate the same `AddZeroBiasInput` workaround (using `norm_shape`, on NPU backend with quantized scale + per-tensor X quant params) before emitting the LN node, otherwise every model that hits the externalize-bias branch can intermittently fail on those SDKs.

**[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc:42-53`
`HasNonOneDimBeforeAxis` computes `prefix = input_rank - shape.size()` with both operands `size_t`. If `shape.size() > input_rank` (spec-violating ONNX model, mis-folded scale, etc.), `prefix` underflows and `aligned_axis < ln_axis` is then never true, so the helper silently returns `false` and the bug surfaces only later as an opaque QNN shape-mismatch error. Add an early `if (shape.size() > input_rank) return true;` (or, better, reject in `IsOpSupported()` so the model falls back to CPU EP).

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc:464-480`
The `load_q_value` lambda's `default:` branch silently returns `0` when `bias_info.qnn_data_type` is none of the listed quantized types. Future QDQ pipelines using a different bias dtype would produce an all-zero bias with no error. Either return `Ort::Status` from the lambda or precede the loop with an explicit dtype allowlist check (`RETURN_IF_NOT(IsBiasDtypeSupported(...))`).

**[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc:281`
`std::max(range_abs, 1e-4f)` uses a magic floor without comment. In current code `ln_max_abs_normalized >= 1` so the floor almost never triggers, but if `norm_count` ever degenerates or upstream `GetQuantParams` changes, the choice of `1e-4f` is opaque. Pull it out into a named `constexpr float kMinIntermediateRangeAbs = 1e-4f;` with a one-line rationale.

**[N-3]** `onnxruntime/test/providers/qnn/layernormalization_test.cc`
The 8 new tests all use `axis=-1` and `uint8_t/uint8_t` (QDQ) or fp32 (FP). Suggested additions: a `uint16_t/uint8_t` decomposition test (covers the 16-bit branches of bias requantization and `const_buf` dispatch), and a higher-rank test with mid-axis misalignment (e.g. shape `[2,4,8,16]`, scale `[1,4,1,16]`) to exercise the `prefix` arithmetic. The PR description should also document the kill-test fence result (G02): temporarily disable the new decomposition branch and confirm the new tests fail.

**[N-4]** `onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc:325-353,463-480,487-503`
Three near-identical dtype-switch blocks (fill ones, load int, store requantized int) repeat the same 4-quantized + 2-float dispatch. Consider extracting a small `template <typename T>` helper plus a `DispatchByQnnDtype` macro/visitor so future dtype additions need only one change site instead of three.

qti-yuduo:
---
  P0 — Block on these (silent correctness bugs, verified end-to-end)

  P0-1. quant_one saturation produces a non-identity scale

  - Where: onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc:323–354
  - Exact issue: When externalize_scale=true and the user scale is quantized, the code calls utils::Quantize(1.0, scale, offset, scale_dtype,
  quant_one) and replicates quant_one into the synthesized "ones" tensor. utils::Quantize (qnn_utils.cc:964–973) clamps via Saturate(qmax, qmin,
  …) and returns Status::OK() even after clipping.
  - Why it's a bug — concrete numeric example: Consider a small-amplitude gamma quantized to u8 with range [0, 0.005], so scale = 0.005/255 ≈
  1.96e-5, offset = 0. Then Quantize(1.0) = round(1.0 / 1.96e-5) = 51000 → saturates to 255. Dequantizing gives (255 + 0) * 1.96e-5 = 0.005, not
  1.0 — a 200× silent error. The "ones" tensor isn't ones; the inner LN multiplies by 0.005, the external Mul applies user_scale on top, and the
  final output is silently scaled by ~0.005× the intended value.
  - How to fix: After utils::Quantize, immediately call utils::Dequantize(quant_one) and assert |deq - 1.0| ≤ scale (one-LSB tolerance). If
  outside tolerance, return MAKE_EP_FAIL so the node falls back to CPU. Even simpler and preferred: avoid the synthesized-ones path entirely (see
   S2 below).

  P0-2. 4-bit (include_bw) per-tensor scales read the wrong encoding fields

  - Where: layernormalization_op_builder.cc:273–274, 321–322, 452–455
  - Exact issue: All three sites accept a tensor as "per-tensor" via IsPerTensor(/*include_bw*/ true) — which (per
  qnn_quant_params_wrapper.h:65–69) returns true for both QNN_QUANTIZATION_ENCODING_SCALE_OFFSET (8/16/32-bit) and
  QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET (4-bit). They then read Get().scaleOffsetEncoding.scale / .offset unconditionally.
  - Why it's a bug: qnn_quant_params_wrapper.cc:498–509 (the Init path) shows that when is_int4_type is true (ONNX zp dtype is INT4 or UINT4 —
  :478–479), the data is written to params_.bwScaleOffsetEncoding, not scaleOffsetEncoding. The two are union members in Qnn_QuantizeParams_t.
  Reading the wrong member returns either zero or the bit-pattern of the BW fields, which differs in size and semantics. The downstream
  user_scale_max_abs, the quant_one, and the bias requant all silently use garbage.
  - The codebase's idiomatic pattern is right there: conv_op_builder.cc:347–352 explicitly branches:
  if (act_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) { ... .scaleOffsetEncoding ... }
  else if (... == QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) { ... .bwScaleOffsetEncoding ... }
  - And matmul_op_builder.cc:263 uses IsPerTensor() (the default include_bw=false) before reading scaleOffsetEncoding, which is also safe.
  - How to fix: Either (a) change all three sites to IsPerTensor() (no include_bw) and explicitly reject 4-bit with a clear error, OR (b) extract
   a tiny accessor qp.GetPerTensorScaleOffset() -> {scale, offset} in qnn_quant_params_wrapper.h that branches on quantizationEncoding, and use
  it everywhere. (b) is more reusable.

  P0-3. SDK 2.24–2.27 implicit-bias workaround orphans a tensor in the decomposed-no-user-bias path

  - Where: layernormalization_op_builder.cc:117–134 (the workaround) vs :397 (the decomposed inner LN call)
  - Exact issue: The workaround at :118 runs only when !has_bias_input, on NPU, under QNN API 2.17–2.20. It synthesizes a static int32 zero-bias
  tensor via AddZeroBiasInput (base_op_builder.cc:142–198), which both registers the tensor in model_tensors_map_ (qnn_model_wrapper.cc:120–138)
  and pushes its name onto input_names as slot 2. The decomposed inner LN at :397 constructs the node with only {input_names[0], ln_scale_name} —
   slot 2 is silently ignored.
  - Why it's a bug: Two consequences:
    a. The synthesized bias lives in the graph as an unused static tensor — wasted memory at minimum.
    b. More importantly, the decomposed inner LN node has no bias input — exactly the structural pattern that the workaround was added to dodge
  in QNN SDK 2.24–2.27. Builds against those SDK versions can re-trigger the original validator failure on the decomposed path.
  - Trigger conditions (all must hold for the bug to fire): decomposition active AND !has_bias_input AND IsNpuBackend AND build linked against
  QNN API 2.17–2.20 AND X is per-tensor quantized AND scale is quantized.
  - How to fix: Move the workaround out of ProcessInputs and into the non-decomposition branch of ProcessAttributesAndOutputs. Concretely: detect
   misalignment in ProcessInputs first (or push the workaround into the standard AddToModelBuilder path after the decomposition decision is
  made). The cleanest split: compute externalize_scale/externalize_bias early, then only synthesize the implicit bias when the inner LN node will
   actually consume slot 2.

  ---
  P1 — Should fix before merge

  P1-4. load_q_value silently corrupts bias on unhandled dtypes

  - Where: layernormalization_op_builder.cc:464–479
  - Exact issue: The default: arm of the dtype switch returns 0. The enumerated dtypes are SFIXED_POINT_{8,16,32}, UFIXED_POINT_{8,16}, INT_32.
  Notably missing: UFIXED_POINT_32 (codebase uses it — qnn_utils.cc:872), FLOAT_16/FLOAT_32 (used elsewhere as bias in non-QDQ flows),
  INT_8/INT_16/INT_64, and the 4-bit variants.
  - Why it's a bug: A bias of an unhandled dtype results in every element being treated as Dequantize(src_offset, src_scale, 0.0) = src_offset *
  src_scale, which is generally non-zero garbage. No error is raised. Future dtype additions will silently corrupt; today, callers passing FP
  bias through this branch (which shouldn't reach :444 in practice because bias_info.qnn_data_type == x_qnn_data_type for valid ONNX-FP graphs,
  but defense in depth matters) get silently wrong results.
  - How to fix: Refactor the lambda to return Ort::Status plus an out-param, and use MAKE_EP_FAIL in the default arm. Better — replace the entire
   60-line block by composing existing helpers (see S1 below).

  P1-5. Bias requantize saturation is silent

  - Where: layernormalization_op_builder.cc:486 (utils::Quantize call inside the per-element requant loop)
  - Exact issue: utils::Quantize saturates without warning. For typical dst_scale derived from sqrt(N-1) · user_scale_max_abs / qmax (e.g. N=768,
   user_scale_max=1.0, u8 → dst_scale ≈ 0.217), bias values with |b| > ~27 saturate. Real LN biases are usually small, but when they aren't, the
  quantized bias silently clips and the final output drifts.
  - Why it's a bug: Same class as P0-1 — silent quantization saturation. Output is wrong, no signal to the user.
  - How to fix: Either log a warning when any element saturates, or replace the whole block with composed DequantizePerChannel + QuantizeData
  (S1) and have those helpers grow saturation reporting once.

  P1-6. No HTP rank-4 / non-trivial-N test for the decomposition path

  - Where: onnxruntime/test/providers/qnn/layernormalization_test.cc:228–351
  - Exact issue: Verified: tests LayerNorm_Decomposed_* (:228–274) and LayerNorm_fp_*_misaligned (:324–351) DO trigger HasNonOneDimBeforeAxis (X
  {1,2,3}, scale {1,2,3}, axis=-1: prefix=0, at i=1 aligned_axis=1 < ln_axis=2 and shape[1]=2 ≠ 1 → returns true). So the decomposition IS
  exercised — earlier claim "no test covers the new path" is retracted. However, every test uses N=3 and rank ≤ 3.
  - Why it's a gap: The sqrt(N-1) quant-range derivation, the broadcast-over-non-leading-dims behavior of the externalized Mul, P0-1's saturation
   case (which needs a small-amplitude scale), and per-channel-output handling are all under-exercised at N=3. Real LN N is 768/1024/4096.
  - How to fix: Add at least one rank-4 case, e.g. X {2, 4, 3, 768}, scale {1, 4, 1, 768}, axis=-1, both u8/u8 QDQ and fp16. Add one case with a
  very-low-amplitude scale to catch P0-1 (X {1, 2, 768}, scale {1, 2, 768} with values in [0, 0.005]).

  P1-7. Test typo

  - Where: layernormalization_test.cc:314
  - Exact issue: LayerNorm_fp_no_bais should be LayerNorm_fp_no_bias.
  - How to fix: Rename. Trivial.

  P1-8. Orphan scratch comment in test file

  - Where: layernormalization_test.cc:240
  - Exact issue: A stray line // final_tensor_type / final_output_info.{qnn_data_type, quant_param, shape}). sits between two unrelated tests.
  Looks like an implementation note that escaped into the test file.
  - How to fix: Delete.

  ---
  P2 — Code quality / maintainability

  P2-9. Per-channel-quantized scale gate produces a confusing error message

  - Where: layernormalization_op_builder.cc:319–320
  - Exact issue: When user provides per-channel-quantized scale AND shape is misaligned, BuildDecomposedLayerNorm returns RETURN_IF_NOT(...
  "per-channel quantized scale is not supported."). Verified that IsOpSupported calls AddToModelBuilder(do_op_validation=true)
  (base_op_builder.cc:32), which routes this failure as op-not-supported → CPU fallback. So functionally clean, no runtime crash.
  - Why it matters: The error message reads like a hard global limitation, not a decomposition-path-specific one. Users debugging "why does QNN
  reject my LN" will be misled.
  - How to fix: Reword to "QNN LayerNorm decomposition does not support per-channel-quantized scale combined with shape misalignment; falling
  back to CPU EP." Make it clear this is a narrow path, not a global ban.

  P2-10. Triple GetTensorInfo(inputs[1]) for the same scale tensor

  - Where: :186, :268, :306 (also :193, :438 for bias)
  - Exact issue: Three independent GetTensorInfo fetches for the scale input within one node-build pass, each named differently (scale_info,
  scale_info_for_range, scale_info). Not a correctness bug, but the reader has to verify all three refer to the same input.
  - How to fix: Fetch each input's TensorInfo once at the top of ProcessAttributesAndOutputs and pass const TensorInfo& down. Also collapses 3 of
   the 11 parameters of BuildDecomposedLayerNorm.

  P2-11. Naming inconsistency: default_axis vs ln_axis

  - Where: :83, :161, :164 use default_axis; :44, :231 use ln_axis
  - Exact issue: Two names for the same concept inside one file. default_axis is also a misnomer — ProcessAxisAttribute resolves negative axes
  (base_op_builder.cc:415–419), so by the time it's used the value is the actual axis, not a default.
  - How to fix: Use ln_axis consistently in the new code. (Pre-existing default_axis in untouched code can stay if not in scope.)

  P2-12. Unused <cstring> include

  - Where: layernormalization_op_builder.cc:7
  - Exact issue: No memcpy/memset/strlen/strcpy in the file. Verified.
  - How to fix: Remove. Trivial.

  P2-13. BuildDecomposedLayerNorm takes 11 parameters

  - Where: :55–65 (declaration)
  - Exact issue: 11 positional params including booleans externalize_scale, externalize_bias, do_op_validation, plus a moved-in
  param_tensor_names whose lifecycle precondition (caller pre-built epsilon + axes) is undocumented.
  - How to fix: Inline into ProcessAttributesAndOutputs — there's only one caller, and gemm_op_builder.cc / einsum_op_builder.cc keep their
  decompositions inline. Or group inputs into a DecompositionContext struct.

  ---
  Simplification opportunities (high-confidence)

  ID: S1
  Simplification: Replace the bias requant block (:457–515) with composed utils::DequantizePerChannel (qnn_utils.cc:1096) + utils::QuantizeData
    (:1026). Pass single-element scales/offsets and axis = std::nullopt for the per-tensor case.
  Why it's safe: Both helpers exist, are publicly used, support all the dtypes the PR enumerates AND more (UFIXED_POINT_32 etc.), and propagate
    dtype errors via MAKE_EP_FAIL instead of silently returning 0. Fixes P1-4 and P1-5 simultaneously.
  Lines saved: ~50
  ────────────────────────────────────────
  ID: S2
  Simplification: The "synthesize ones" path is needed only because the inner LN expects a scale tensor. The codebase already calls
    AddZeroBiasInput (base_op_builder.cc:142–198) to synthesize a quantized zero. Add a sibling AddOneScaleInput that synthesizes quantized ones,

    sharing the dtype-dispatch logic. Then the LN-decomposition site shrinks to a single helper call, AND the saturation guard from P0-1 lives in

    one place.
  Why it's safe: The dtype dispatch at :325–354 is identical structurally to the dispatch in AddZeroBiasInput. Concentrating the logic kills the
    duplication and makes P0-1 fixable in one place.
  Lines saved: ~40
  ────────────────────────────────────────
  ID: S3
  Simplification: Inline BuildDecomposedLayerNorm into ProcessAttributesAndOutputs (single caller).
  Why it's safe: Matches the pattern in gemm_op_builder.cc and einsum_op_builder.cc. Removes the 11-param signature (P2-13).
  Lines saved: structural
  ────────────────────────────────────────
  ID: S4
  Simplification: Fetch each input's TensorInfo once at the top of ProcessAttributesAndOutputs; pass const TensorInfo& down.
  Why it's safe: Removes 5 redundant fetches (P2-10), tightens naming.
  Lines saved: ~10

qti-yuduo (commented):
Left comments from Claude after some iterations. Please validate and address if it's valid concerns.

onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc
```diff
@@ -150,6 +351,354 @@ Ort::Status LayerNormalizationOpBuilder::ProcessAttributesAndOutputs(QnnModelWra
...
+
+  const auto& outputs = node_unit.Outputs();
+  const std::string& final_output_name = outputs[0].name;
+  const bool is_graph_output = qnn_model_wrapper.IsGraphOutput(final_output_name);
+  const Qnn_TensorType_t final_tensor_type = is_graph_output ? QNN_TENSOR_TYPE_APP_READ : QNN_TENSOR_TYPE_NATIVE;
+
+  TensorInfo final_output_info{};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(outputs[0], final_output_info));
+
+  // ln_intermediate_qp: LN output. With externalize_scale, LN emits standardized values
```
qti-yuduo, line 375: F1 — ln_intermediate_qp standardized range saturates the Mul output and bias when externalize_scale && externalize_bias. MAJOR
  (latent), HIGH confidence.
  - Where: range built at :357-366 as ±min(sqrt(N-1), 3.0); reused as the Mul output qp (:491-492, the externalize_bias branch) and
  the bias requant dst qp (:602).
  - Why it's wrong: that range is sized for the standardized LN output z (|z| ≤ sqrt(N-1)). But the Mul output is z·γ — exceeds the
  range whenever γ > 1 — and the bias β is an independent addend forced into the same range. Worst case N=3 → ln_abs_max=1.41, uint8
  scale≈0.011; a bias of 5.0 → round(5.0/0.011)=451 clipped to 255 ≈ 1.41, a ~3.5× silent corruption. The :147-151 warning fires
  but the graph is still built with clipped data.
  - Why tests miss it: every decomposition test uses scale∈[0.1,1.0], bias∈[0,1.0], N=3 — constructed to stay in range.
  - Bounded: the bias-only path (externalize_scale=false) correctly uses final_output_info.quant_param (:369), which covers z·γ+β.
  Bug is specific to the scale-externalized branch.
  - Fix: size the Mul-output qp and bias requant dst from a range covering actual operand magnitude — simplest is to reuse
  final_output_info.quant_param for both (it provably covers z·γ+β; QNN gives each tensor its own qp, so independent operand scales
  are fine).

onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc
```diff
@@ -150,6 +351,354 @@ Ort::Status LayerNormalizationOpBuilder::ProcessAttributesAndOutputs(QnnModelWra
...
+                                 x_qnn_data_type,
+                                 ln_intermediate_qp.Copy(),
+                                 std::vector<uint32_t>(x_shape));
+  RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(ln_out_tensor)),
+                "Failed to add decomposed LN intermediate output tensor.");
+
+  std::vector<std::string> ln_inputs = {input_names[0], ln_scale_name};
+
+#if QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 17 && QNN_API_VERSION_MINOR <= 20
+  // Mirror the ProcessInputs workaround: on QNN SDK 2.24-2.27 (API 2.17-2.20), an LN node without
```
qti-yuduo, line 492: Do we really care about this old version?

qti-yuduo (commented):
LGTM overall, left some minor questions.

onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc
quic-muchhsu: From AI hub's standpoint, no we don't.
From EP's point, maybe?, for backward compatibility and consistancy.
@qti-chuteng Any thoughts?

qti-yuduo (approved):

qti-chuteng (approved):

PR #418, opened 2026-05-20
[QNN EP] Add support for Dropout Op

qti-mbadnara:
### Description

Implements the `Dropout` ONNX operator (opsets 7–22) for the QNN Execution Provider across CPU, HTP (FP32, FP16, BF16), and GPU backends.

`Dropout` has no native `QNN Op`, and ONNX Dropout is a pure identity in inference mode (`training_mode = false`), the operator is decomposed into **two independent QNN nodes**:

```
                    ┌──────────────────────────────────────────────────────────────┐
                    │                QNN subgraph for Dropout                      │
                    │                                                              │
                    │   Node 1                                                     │
data (float) ───────►  QNN_OP_TRANSPOSE                                ───────────► output (float)
                    │  perm = [0, 1, ..., N-1]  (identity permutation)             │
                    │                                                              │
                    │   Node 2                                                     │
zeros (STATIC bool) ►  QNN_OP_ELEMENT_WISE_NOT                         ───────────► mask (bool, optional)
all-zeros constant  │  NOT(0) = 1  →  all-ones                                     │
                    │                                                              │
                    └──────────────────────────────────────────────────────────────┘
```

- **Node 1** — `QNN_OP_TRANSPOSE` with an identity permutation passes `data` through unchanged to `output`.
- **Node 2** — `QNN_OP_ELEMENT_WISE_NOT` applied to a static all-zeros `BOOL_8` constant produces `mask` as an all-ones bool tensor. This node is only added when `mask` is a graph output.

The two nodes are completely independent. `Transpose` produces only `output`; `ELEMENT_WISE_NOT` produces only `mask`

---

### Motivation / Context

`Dropout` was previously unsupported in QNN. Without this op the `QNN EP` falls back to the `CPU` for any graph that includes a `Dropout` node, even at inference time where the op does nothing.

ONNX defines `Dropout` as https://onnx.ai/onnx/operators/onnx__Dropout.html

| Opset | Change |
|-------|--------|
| 1–6   | `ratio` is an attribute; no `mask` output |
| 7–9   | `ratio` attribute; optional `mask` output added |
| 10–11 | `seed` attribute added |
| 12–21 | `ratio` and `training_mode` become **inputs**; `seed` attribute kept |
| 22    | `bfloat16` added to `T` type constraints |

This implementation supports all opsets 7–22. At opsets ≥ 12, `ratio` and `training_mode` are inputs; `training_mode` is validated to be a constant `false` (or absent) and then ignored. At opset 22 the `bfloat16` path is exercised by the `HTP BF16` tests.

---

## Files changed

| File | Change |
|------|--------|
| `core/providers/qnn/builder/opbuilder/dropout_op_builder.cc` | `DropoutOpBuilder` implementation |
| `core/providers/qnn/builder/op_builder_factory.h` | Add `CreateDropoutOpBuilder` declaration |
| `core/providers/qnn/builder/op_builder_factory.cc` | Register `"Dropout"` |
| `test/providers/qnn/dropout_test.cc` | Unit tests for all backends and data types |

---

## Implementation notes

**No QDQ variant**: ONNX Dropout has float-only type constraints (`T ∈ {float16, float, double, bfloat16}`), so no QDQ path is needed or registered.

**`training_mode` guard** (`ExplicitOpCheck`): If `inputs[2]` exists and is non-empty, it must be a constant initializer equal to `0`. Dynamic or `true` values cause the node to fall back to the CPU EP.

**Quant param passthrough** (`OverrideOutputQuantParam`): For quantized graphs, output[0]'s scale/offset is copied from input[0] via `SetOutputQParamEqualToInputIfNearlyEqual`, preserving accuracy through the identity.

**Mask output*: Materialised as a static `QNN_DATATYPE_BOOL_8` tensor (`QNN_TENSOR_TYPE_APP_READ` when it is a graph output, `QNN_TENSOR_TYPE_NATIVE` otherwise) filled with `1`. No `QNN op` is added for the mask; it is a constant tensor.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.h
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.h
@@ -262,6 +262,7 @@ class BaseOpBuilder : public IOpBuilder {
   static const std::pair<size_t, size_t> GetInputOutputCountQnnRequired(std::string onnx_op_type) {
     static const std::unordered_map<std::string, std::pair<size_t, size_t>> input_output_count_qnn_required = {
         {"BatchNormalization", {3, 1}},
+        {"Dropout", {1, 1}},
         {"GlobalAveragePool", {0, 1}},
         {"LayerNormalization", {0, 1}},
         {"MaxPool", {0, 1}}};
--- a/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
+++ b/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
@@ -20,6 +20,7 @@ OpBuilderRegistrations::OpBuilderRegistrations() {
   CreateConvOpBuilder("Conv", *this);
   CreateConvOpBuilder("ConvTranspose", *this);
   CreateCumSumOpBuilder("CumSum", *this);
+  CreateDropoutOpBuilder("Dropout", *this);
   CreateEinsumOpBuilder("Einsum", *this);
   CreateExpandOpBuilder("Expand", *this);
   CreateFusedMatMulOpBuilder("FusedMatMul", *this);
--- a/onnxruntime/core/providers/qnn/builder/op_builder_factory.h
+++ b/onnxruntime/core/providers/qnn/builder/op_builder_factory.h
@@ -60,6 +60,7 @@ const IOpBuilder* GetOpBuilder(const std::string& onnx_op_type);
 void CreateArgMaxMinOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
 void CreateBatchNormalizationOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
 void CreateCastOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
+void CreateDropoutOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
 void CreateClipOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
 void CreateConcatOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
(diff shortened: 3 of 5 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.h
```diff
@@ -262,6 +262,7 @@ class BaseOpBuilder : public IOpBuilder {
   static const std::pair<size_t, size_t> GetInputOutputCountQnnRequired(std::string onnx_op_type) {
     static const std::unordered_map<std::string, std::pair<size_t, size_t>> input_output_count_qnn_required = {
         {"BatchNormalization", {3, 1}},
+        {"Dropout", {1, 1}},
```
tirupath-qti, line 265: ORT optimizer typically eliminates Dropout operator in https://github.com/microsoft/onnxruntime/blob/eaf8d943/onnxruntime/core/optimizer/dropout_elimination.cc#L14-L72 
why we need an Opbuilder in QNN EP?

qti-mbadnara: ORT's DropoutElimination optimizer only runs when graph optimizations are enabled. If a user sets ORT_DISABLE_ALL, the Dropout node survives into EP partitioning untouched.

Without a QNN EP op builder, that surviving Dropout node falls back to CPU EP, breaking the graph partition and forcing unnecessary CPU↔QNN memory copies around a node that does literally nothing.

If the optimizer already eliminated Dropout, the builder never gets called; if it didn't, QNN EP can still claim the node cleanly.

qti-ashwshan: Agreed. We should not be adding this dropout op builder just because the user is explicitly disabling all optimizations in ORT run.

qti-mbadnara: Sounds good, will close this PR

PR #326, opened 2026-04-30
[QNN EP] Fix memory leak in OrtTensorTypeAndShapeInfo in ExecuteGraph

qti-ashwshan:
### Description
- Release OrtTensorTypeAndShapeInfo in TensorDataSize lambda
- Set context_created_ after async AddQnnContextHandle
- Clear ep_context_handle_map_ in ReleaseContext

PR #419, opened 2026-05-20
Enable memory leak detection on CI (Linux x86 / ASan)

chunghow-qti:
## Summary

Add a new CI job that builds ORT QNN EP for Linux x86_64 with AddressSanitizer (Debug + ASan) and runs the unit-test suite under it. Any ASan-detectable issue fails CI; ORT-side leaks fail CI; indirect leaks from QAIRT SDK backend libraries are filtered out as known third-party noise.

## What this catches (will fail CI)

Anything ASan / LSan can detect that originates in our code:

- **Heap / stack / global buffer overflow**
- **Use-after-free / use-after-return / use-after-scope**
- **Double-free / invalid-free**
- **`new-delete-type-mismatch` / `alloc-dealloc-mismatch`**
- **Direct memory leaks** (LSan) — ORT or test code

Verified end-to-end (CI fails) by reverting individual fixes from #405.

## What this filters out (will not fail CI)

QAIRT SDK backend libraries do init/runtime allocations that LSan tracks but cannot suppress cleanly:

| Source | Mechanism |
|---|---|
| Direct leaks in `libQnnHtp.so` / `libQnnCpu.so` | Suppressed by `tools/ci_build/lsan_suppressions.txt` (`leak:` patterns match the module path on the stack) |
| Any Indirect leak (regardless of source) | Filtered out by `qcom/scripts/linux/asan_filter_leaks.sh`. Indirect leaks frequently resolve to `<unknown module>` because the QAIRT libraries are stripped of frame-pointer + DWARF info; we do not police them. ORT-side leaks would surface as Direct first. |

## Failure-classification flow

```
test binary exit
    │
    ├─ exit 0                                       → PASS
    ├─ ASan heap error (overflow, UAF, …)           → FAIL
    ├─ gtest assertion failure                      → FAIL
    ├─ "Direct leak" present in output              → FAIL  (ORT-side bug)
    └─ Only "Indirect leak" entries                 → PASS  (not policed)
```

## Changes

The pipeline mirrors the existing **coverage** flow (dedicated workflow, dedicated public task, dedicated test runner script):

| File | Purpose |
|---|---|
| `tools/ci_build/lsan_suppressions.txt` (new) | Suppress direct leaks from `libQnnHtp.so` / `libQnnCpu.so` |
| `qcom/scripts/linux/asan_filter_leaks.sh` (new) | Wrapper that runs the test binary and classifies the failure mode (heap error / gtest fail / Direct leak / Indirect-only) |
| `qcom/scripts/linux/run_asan.sh` (new) | Test-phase entry point. Owns `LD_LIBRARY_PATH`, `ASAN_OPTIONS`, `LSAN_OPTIONS`; invokes `asan_filter_leaks.sh ./onnxruntime_provider_test` and emits gtest JUnit XML. Mirror of `generate_coverage.sh` |
| `qcom/scripts/linux/build.sh` | New `--enable-asan` flag — passes `--enable_address_sanitizer` through to `tools/ci_build/build.py`. Build-time only; test-time concerns live in `run_asan.sh` |
| `qcom/ep_build/tasks/build.py` | New `RunAsanTask` (`BashScriptsWithVenvTask` subclass) — mirror of `GenerateCoverageTask`; calls `run_asan.sh` with `--build-dir` / `--config` |
| `qcom/build_and_test.py` | New `asan_linux_x86_64` `@public_task` — `CompositeTask` of `BuildEpLinuxTask(Debug, --enable-asan)` + `RunAsanTask`. Mirror of `coverage_linux_x86_64` |
| `.github/workflows/qualcomm-internal-asan.yml` (new) | Self-contained workflow that runs `python3 qcom/build_and_test.py asan_linux_x86_64` on a Linux/X64/Build self-hosted runner, gated on PRs touching `onnxruntime/(core\|test)/providers/qnn/`. Mirror of `qualcomm-internal-coverage.yml` |
| `.github/workflows/qualcomm-internal-ci.yml` | New `asan:` job hook on PR events, next to the existing `coverage:` hook |

## Notes

- ASan workflow runs on its own self-hosted runner (`["self-hosted", "X64", "Linux", "Build"]`), not piggybacking on the shared `qualcomm-internal-build-and-test-single-os.yml` workflow. Independent build/test run, isolated `build/linux-x86_64/Debug/`.
- Build artifacts (wheel, test archive) are not produced for the ASan job (`--target-py-version None` skips the wheel build path).
- ONNX model tests (`onnxruntime_plugin_ep_onnx_test`) are skipped under ASan: the binary `dlopen`s ASan-instrumented `.so`s with no ASan in the host process's initial library list, so ASan refuses to initialize.
- Build config is **Debug** for now. Switch to **RelWithDebInfo** (recommended by the sanitizer wiki, faster total CI wall-clock) requires patching `tools/ci_build/build.py` — on Linux the `-fsanitize=address` flag is wired only into the Debug branch. Tracked as a separate follow-up.
- `asan_filter_leaks.sh` and the two `leak:libQnn*.so` suppressions are a stopgap. Once the QAIRT team addresses the leaks in `libQnnHtp.so` / `libQnnCpu.so`, the wrapper can be retired and `run_asan.sh` can run the test binary directly.
- `QnnIRBackendTests` under ASan emits Indirect-only leaks whose stacks resolve to `<unknown module>` and cannot be conclusively attributed to `libQnnIr.so` vs other QAIRT libs loaded at process init. The Indirect-policy in `asan_filter_leaks.sh` makes this benign for CI; root-cause attribution is a follow-up with QAIRT.

### DIFF
--- a/.github/workflows/qualcomm-internal-asan.yml
+++ b/.github/workflows/qualcomm-internal-asan.yml
@@ -5,8 +5,6 @@ name: Qualcomm Internal ASan
 
 permissions:
   contents: read
-  checks: write
-  pull-requests: write
 
 on:
   workflow_call:
@@ -16,12 +14,6 @@ jobs:
   asan-linux-x86_64:
     name: "asan-linux-x86_64"
     runs-on: ["self-hosted", "X64", "Linux", "Build"]
-    env:
-      # Hardcoded Debug because upstream tools/ci_build/build.py only wires
-      # -fsanitize=address into the Linux Debug branch; RelWithDebInfo +
-      # --enable_address_sanitizer would silently produce a non-instrumented
-      # binary. Switching configs requires patching upstream ORT first.
-      ASAN_BUILD_DIR: build/linux-x86_64/Debug
     steps:
       - uses: actions/checkout@v6
         with:
@@ -65,25 +57,3 @@ jobs:
           python3 qcom/build_and_test.py \
             --target-py-version None \
             asan_linux_x86_64
-
-      - name: Publish Test Report
-        uses: mikepenz/action-junit-report@v6
-        # Skipped on fork PRs: the action creates a check run, which the fork-PR GITHUB_TOKEN cannot do.
-        if: ${{ always() && steps.gate.outputs.run == 'true' && github.event.pull_request.head.repo.fork != true }}
-        with:
-          check_name: asan-linux-x86_64-test-results
-          group_reports: false
-          include_passed: true
-          detailed_summary: true
-          verbose_summary: false
-          report_paths: |
-            ${{ env.ASAN_BUILD_DIR }}/*.results.xml
-
-      - name: Upload Test Results
-        uses: actions/upload-artifact@v7
-        if: ${{ always() && steps.gate.outputs.run == 'true' }}
-        with:
-          name: "asan-linux-x86_64-${{ github.sha }}-test-results"
-          retention-days: 7
-          path: |
-            ${{ env.ASAN_BUILD_DIR }}/*.results.xml
--- a/.github/workflows/qualcomm-internal-asan.yml
+++ b/.github/workflows/qualcomm-internal-asan.yml
@@ -1,0 +1,85 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+name: Qualcomm Internal ASan
+
+permissions:
+  contents: read
+  checks: write
+  pull-requests: write
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  asan-linux-x86_64:
+    name: "asan-linux-x86_64"
+    runs-on: ["self-hosted", "X64", "Linux", "Build"]
+    env:
+      ASAN_BUILD_DIR: build/linux-x86_64/Debug
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          # submodules omitted here; initialized below only when ASan is needed.
+
+      - name: Check if ASan is needed
+        id: gate
+        run: |
+          set -euo pipefail
+
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "run=true" >> "${GITHUB_OUTPUT}"
+            exit 0
+          fi
+
+          base_ref="${{ github.base_ref }}"
+          if [ -z "${base_ref}" ]; then
+            echo "::error::github.base_ref is empty (event_name=${{ github.event_name }})"
+            exit 1
+          fi
+
+          if ! changed=$(git diff --name-only "origin/${base_ref}...HEAD"); then
+            echo "::error::git diff against origin/${base_ref}...HEAD failed"
+            exit 1
+          fi
+
+          if echo "${changed}" | grep -qE 'onnxruntime/(core|test)/providers/qnn/'; then
+            echo "run=true" >> "${GITHUB_OUTPUT}"
+          else
+            echo "run=false" >> "${GITHUB_OUTPUT}"
+          fi
+
+      - name: Initialize submodules
+        if: steps.gate.outputs.run == 'true'
+        run: git submodule update --init --recursive
+
+      - name: Run ASan build and unit tests (Linux x86_64, Debug)
+        if: steps.gate.outputs.run == 'true'
+        run: |
+          python3 qcom/build_and_test.py \
+            --target-py-version None \
+            asan_linux_x86_64
+
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v6
+        # Skipped on fork PRs: the action creates a check run, which the fork-PR GITHUB_TOKEN cannot do.
+        if: ${{ always() && steps.gate.outputs.run == 'true' && github.event.pull_request.head.repo.fork != true }}
+        with:
+          check_name: asan-linux-x86_64-test-results
+          group_reports: false
+          include_passed: true
+          detailed_summary: true
+          verbose_summary: false
+          report_paths: |
+            ${{ env.ASAN_BUILD_DIR }}/*.results.xml
+
+      - name: Upload Test Results
+        uses: actions/upload-artifact@v7
+        if: ${{ always() && steps.gate.outputs.run == 'true' }}
+        with:
+          name: "asan-linux-x86_64-${{ github.sha }}-test-results"
+          retention-days: 7
+          path: |
+            ${{ env.ASAN_BUILD_DIR }}/*.results.xml
--- a/.github/workflows/qualcomm-internal-asan.yml
+++ b/.github/workflows/qualcomm-internal-asan.yml
@@ -17,6 +17,10 @@ jobs:
     name: "asan-linux-x86_64"
     runs-on: ["self-hosted", "X64", "Linux", "Build"]
     env:
+      # Hardcoded Debug because upstream tools/ci_build/build.py only wires
+      # -fsanitize=address into the Linux Debug branch; RelWithDebInfo +
+      # --enable_address_sanitizer would silently produce a non-instrumented
+      # binary. Switching configs requires patching upstream ORT first.
       ASAN_BUILD_DIR: build/linux-x86_64/Debug
     steps:
       - uses: actions/checkout@v6
--- a/qcom/build_and_test.py
+++ b/qcom/build_and_test.py
@@ -440,6 +440,7 @@ class TaskLibrary:
                 ],
             )
         )
+
     if is_host_linux() and is_host_x86_64():
 
         @public_task("Build with AddressSanitizer and run unit tests under ASan (Linux x86_64, Debug)")
--- a/qcom/build_and_test.py
+++ b/qcom/build_and_test.py
@@ -40,6 +40,7 @@ from ep_build.tasks.build import (
     GenerateCoverageTask,
     GenerateDiffCoverageTask,
     QdcTestsTask,
+    RunAsanTask,
 )
 from ep_build.tasks.docker import MANYLINUX_2_34_AARCH64_TAG, DockerBuildTask
 from ep_build.tasks.python import (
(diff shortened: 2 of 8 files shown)


qcom/build_and_test.py
```diff
@@ -848,6 +868,24 @@ def test_ort_linux_x86_64(self, plan: Plan) -> str:
                 )
             )
 
+        @task
+        @depends(["build_ort_linux_x86_64_asan"])
+        def test_ort_linux_x86_64_asan(self, plan: Plan) -> str:
```
huaychou: Why do we need to test Degug + Asan? Can we just create another task like GenerateCoverageTask to call asan_filter_leaks.sh?

tools/ci_build/lsan_suppressions.txt
```diff
@@ -0,0 +1,15 @@
+# LeakSanitizer suppressions for QNN EP CI ASan builds.
```
kuanyul-qti, line 4: Copyright is required.

qcom/scripts/linux/build.sh
```diff
@@ -319,6 +328,15 @@ else
   if [ -n "${run_tests}" ]; then
     cd "${build_dir}/${config}/"
 
+    if [ -n "${enable_asan}" ]; then
+      # Make ASan/LSan failures fail CI; suppress known third-party leaks (libQnnHtp.so / libQnnCpu.so).
+      export ASAN_OPTIONS="halt_on_error=1:detect_leaks=1:print_stats=0"
+      export LSAN_OPTIONS="exitcode=1:suppressions=${REPO_ROOT}/tools/ci_build/lsan_suppressions.txt:print_suppressions=0"
```
kuanyul-qti: Just curious, are these two variables used in the ASan tool?

qcom/scripts/linux/asan_filter_leaks.sh
```diff
@@ -0,0 +1,81 @@
...
+# of their code live in anonymous mmap regions. LSan suppression patterns
+# cannot match these stacks. They are, however, indirect children of
+# direct leaks that we already suppress via tools/ci_build/lsan_suppressions.txt
+# (`leak:libQnnHtp.so` / `leak:libQnnCpu.so`), so an Indirect-only result
+# is by construction third-party noise.
+#
+# Usage:
+#   asan_filter_leaks.sh <test_binary> [args...]
+
+set -u
```
kuanyul-qti:
```suggestion
set -u  # NOTE: -e and pipefail are intentionally omitted.
# The core pipeline below must be allowed to "fail" (non-zero binary exit)
# so that PIPESTATUS[0] can be captured and the leak type classified.
# Adding pipefail/errexit would abort the script before the classification
# logic runs, defeating the purpose of this filter.
```

qcom/scripts/linux/build.sh
chunghow-qti: Yes, ASAN includes Adress Sanitizer (for some unfefined behavior) & Leak Sanitizer (for memory leak). Need two variable, one for ASan & another for LSan to control the behavior of two tools.

tools/ci_build/lsan_suppressions.txt
chunghow-qti: Done

qcom/scripts/linux/asan_filter_leaks.sh
chunghow-qti: Done

qcom/build_and_test.py
chunghow-qti: Done, create a new task for ASan, updated in https://github.com/onnxruntime/onnxruntime-qnn/pull/419/commits/ce9219dea1d2e5005cb0af2ff114e58b8810e182.

chunghow-qti: ASan has to be enabled at build time (typically with a Debug build), since it instruments the binary during compilation.
To catch UB and memory leaks, we still need to run tests on the ASan build at runtime.
I’ll also try RelWithDebInfo + ASan to see if we can speed things up a bit while keeping ASan coverage.

chunghow-qti: We considered RelWithDebInfo (-O2 + -g) for faster CI runs and closer-to-production behavior, but kept Debug for now.
  
Reason: tools/ci_build/build.py wires -fsanitize=address into the Debug branch only on Linux; the RelWithDebInfo branch does not add the flag, so --config RelWithDebInfo --enable_address_sanitizer would silently produce a non-instrumented build. Switching would require patching upstream ORT's build.py via cmake/patches/ and upstreaming the fix — out of scope for this CI enablement PR.                                                                                             

Tracking as a follow-up once ASan is stable in CI.
https://github.com/onnxruntime/onnxruntime-qnn/blob/ebebb418d2f272b0441b24445ad65074a37e67cd/tools/ci_build/build.py#L550-L554

chunghow-qti:
End-to-end validated via draft PR #439 (touches a QNN test file to trip the gate filter). ASan workflow ran the full build + test cycle on CI: [job 77815178090](https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/26434828860/job/77815178090) — pass, The draft PR will be closed once this lands.

.github/workflows/qualcomm-internal-asan.yml
```diff
@@ -0,0 +1,85 @@
...
+        if: steps.gate.outputs.run == 'true'
+        run: |
+          python3 qcom/build_and_test.py \
+            --target-py-version None \
+            asan_linux_x86_64
+
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v6
+        # Skipped on fork PRs: the action creates a check run, which the fork-PR GITHUB_TOKEN cannot do.
+        if: ${{ always() && steps.gate.outputs.run == 'true' && github.event.pull_request.head.repo.fork != true }}
```
kuanyul-qti: Is there a reason for setting "always()"?

.github/workflows/qualcomm-internal-asan.yml
```diff
@@ -0,0 +1,85 @@
...
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  asan-linux-x86_64:
+    name: "asan-linux-x86_64"
+    runs-on: ["self-hosted", "X64", "Linux", "Build"]
+    env:
+      ASAN_BUILD_DIR: build/linux-x86_64/Debug
```
kuanyul-qti: Could you add a comment for Debug configuration?

qcom/scripts/linux/run_asan.sh
```diff
@@ -0,0 +1,100 @@
...
+# to initialize.
+#
+# Prerequisites:
+#   - A build compiled with --enable-asan (i.e. -fsanitize=address).
+#     Use: python qcom/build_and_test.py asan_linux_x86_64
+#
+# Usage:
+#   bash run_asan.sh \
+#       --build-dir=/path/to/build/linux-x86_64 \
+#       [--config=Debug|RelWithDebInfo]
```
kuanyul-qti: Have you confirmed that "RelWithDebInfo" config can be used as "Debug" is hard-coded on L34?

chunghow-qti: This is my mistake, it only supports Debug now, since tools/ci_build/build.py in ort upstream only takes -fsanitize=address in Debug build. We just follow this behavior.

Fixed in latest commit.

.github/workflows/qualcomm-internal-asan.yml
chunghow-qti: Done in latest commit, thanks!

.github/workflows/qualcomm-internal-asan.yml
chunghow-qti: **"Publish Test Report"** & **"Upload Test Results"** are removed in latest commit - they're redundant here. 
If there's direct mem leak, the gtest cases themself don't fail (the leaks is reported only at process exit), so the JUnit report would still mark every test as PASSED and end up misleading rather than diagnostic. 

The leak details can be inspected directly via the step's "View raw logs", which is the right surface for diagnosis here.

chunghow-qti:
PR #456 shows validating ASan CI by revering PR #326, which will fail intentionally. Will closed it once ASan landed

tools/ci_build/lsan_suppressions.txt
```diff
@@ -0,0 +1,15 @@
...
+# Format: leak:<substring matched against any frame in the leak stack>
+# See: https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#suppressions
+#
+# Policy: CI fails on any Direct leak; Indirect leaks are filtered out by
+# qcom/scripts/linux/asan_filter_leaks.sh. The suppressions below cover
+# Direct leaks rooted in stripped QAIRT SDK backend libraries which are
+# tracked separately with the QNN team.
+
+leak:libQnnHtp.so
+leak:libQnnCpu.so
```
kuanyul-qti, line 15: Just curious, do we need to add other backend libraries? E.g., libQnnGpu.so, libQnnIr.so, libGenie.so, etc.

kuanyul-qti (commented):
I see the asan job fails in https://github.com/onnxruntime/onnxruntime-qnn/pull/456, could you help check it?

tools/ci_build/lsan_suppressions.txt
chunghow-qti: ASan did not find any memory leaks with other backends so far, so we may not need to suppress them for now. We will continue investigating and report our findings to the CPU and HTP teams.

chunghow-qti: yes, the ci job will fail since it revert the fix for mem leak, which shows our ASan CI works well.

qcom/scripts/linux/run_asan.sh
```diff
@@ -0,0 +1,106 @@
...
+            ;;
+        --config=*)
+            config="${arg#--config=}"
+            ;;
+        -h|--help)
+            cat <<EOF
+Usage: $(basename "${BASH_SOURCE[0]}") --build-dir=<path> [--config=<cfg>]
+
+  --build-dir=<path>    Required. Build root (e.g. build/linux-x86_64).
+  --config=<cfg>        Optional. Build configuration subdirectory. Default: Debug.
```
huaychou, line 54: Just wondering, will we support RelWithDebInfo?

huaychou (approved):
LGTM

kuanyul-qti (approved):

qti-chuteng (approved):

qcom/scripts/linux/run_asan.sh
chunghow-qti: This can be a follow up once ASan is stable. Since, for now, CI. tools/ci_build/build.py wires -fsanitize=address into the Debug branch only on Linux

PR #420, opened 2026-05-20
Avoid private ORT Core header in Test Framework - Phase 1: Removing unnecessary files

qti-hungjuiw:
### Description

- The PR removing unnecessary files in test framework
- Please refer to #222 for further details

### Motivation and Context

- The PR aims to reduce the scope and overhead of #222


onnxruntime/test/unittest_util/model_test_builder.cc
```diff
@@ -4,53 +4,44 @@
 #include "test/unittest_util/graph_transform_test_builder.h"
 
 #include <functional>
+#include <cstddef>
```
minfhong-qti, line 7: Not sure whether included in following PRs. I think this file should be renamed.

onnxruntime/test/util/test_utils.cc
```diff
@@ -3,18 +3,24 @@
...
-#include "core/session/onnxruntime_cxx_api.h"
-#include "core/session/onnxruntime_session_options_config_keys.h"
-#include "core/framework/tensorprotoutils.h"
+#include <algorithm>
+#include <fstream>
+#include <string>
+#include <vector>
+
+#ifdef __clang__
+#pragma clang diagnostic push
```
minfhong-qti, line 12: Just curious. What is this for?

qti-chuteng (approved):

onnxruntime/test/unittest_util/model_test_builder.cc
qti-hungjuiw: Good point! I just renamed it into `model_test_builder.cc/.h`. Please let me know if this looks good.

onnxruntime/test/util/test_utils.cc
qti-hungjuiw: Hi, in `onnx/onnx_pb.h` there is warning about implicitly cast from 64 bits integer to 32 bits. It is treated as error on Linux. The macro allows compiler to skip the check when including this third-party dependency.

minfhong-qti (approved):

PR #424, opened 2026-05-20
[OpBuilder] Support GroupQueryAttention

qti-mattsinc:
### Description
* Support `com.Microsoft.GroupQueryAttention` by mapping onto the new `QNN_OP_GROUP_QUERY_ATTENTION` op.

### Motivation and Context
* Enable GQA nodes in LLMs to run on the GPU for better performance.


onnxruntime/test/providers/qnn/group_query_attention_test.cc
```diff
@@ -0,0 +1,280 @@
...
+
+  std::optional<std::reference_wrapper<TestInputDef<float>>> cos_cache_def = std::ref(cos_def);
+  std::optional<std::reference_wrapper<TestInputDef<float>>> sin_cache_def = std::ref(sin_def);
+
+  std::vector<int64_t> position_ids_data(batch_size * sequence_length);
+  for (int32_t i = 0; i < batch_size * sequence_length; i++) {
+    position_ids_data[i] = (i + 2) % total_seq_len;
+  }
+  static TestInputDef<int64_t> pos_def({batch_size, sequence_length}, true, position_ids_data);
+  std::optional<std::reference_wrapper<TestInputDef<int64_t>>> position_ids_def = std::ref(pos_def);
```
qti-mattsinc: TODO: Get rid of position_ids input in this testcase because the GPU backend doesn't use it. Edit: Done

qti-mattsinc:
The first commit is not strictly needed for the GQA op builder, but having an alternate verifier for the UTs is something I would like to see merged eventually. It could be spun into a separate PR if needed.

onnxruntime/test/providers/qnn/logical_compare_test.cc
```diff
@@ -179,8 +179,15 @@ TEST_F(QnnHTPBackendTests, EqualToCast4D) {
...
-                  ExpectedEPNodeAssignment::All,
-                  1);  // expected nodes in graph
+                  EPVerificationParams{ExpectedEPNodeAssignment::All,
+                                       EPVerificationParams::VerificationType::ElementwiseAbsolute,
+                                       // This test was added in commit 5a4c3b7937c34cad9671e6440a03aa024a98bf19,
+                                       // at which point RunQnnModelTest expected an arg for `num_nodes_in_ep`.
+                                       // This "1" has appeared to have survived changes to RunQnnModelTest and
+                                       // has now erroneously ended up as the fp32_abs_err arg, which means this test
+                                       // is probably not testing for the correct thing anymore.
+                                       // TODO: Fix this test, perhaps by using `EPVerificationParams::graph_verifier`?
```
qti-mattsinc: FYI for anyone familiar with this test

qti-yuduo:
How well the support is for each backend? Do we plan to add tests?

qti-mattsinc:
> How well the support is for each backend? Do we plan to add tests?

I've added some GPU UTs to group_query_attention_test.cc. (I would suggest reviewing the 2 commits individually; I know the first one pollutes the overall diff). Note that they are disabled right now because we have to wait for 2.48 to release and for it to be picked up in packages.yaml for CI. The GPU backend will support this op in 2.48, with support for the packed_qkv format and some restrictions on attributes (e.g. do_rotary required to be enabled, rotary_interleaved not supported, local_window_size not supported fully).

I'm aware that the HTP backend is also intending to support this op, although I'm not sure what their timeline / support matrix will be. @chunghow-qti might be able to comment more about the HTP side.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,259 @@
...
+  RETURN_IF_NOT(node_helper.HasAttr("kv_num_heads"), "required attribute kv_num_heads not provided");
+
+  int32_t local_window_size = node_helper.Get("local_window_size", -1);
+  RETURN_IF(local_window_size != -1 && SafeInt<uint32_t>(local_window_size) < max_sequence_length,
+            "Local attention through local_window_size not supported");
+
+  int32_t qk_output = node_helper.Get("qk_output", 0);
+  RETURN_IF(qk_output != 0, "qk_output != 0 not supported");
+
+  float smooth_softmax = node_helper.Get("smooth_softmax", 1.0f);
```
chunghow-qti: default value of smooth_softmax should be 0
https://github.com/microsoft/onnxruntime/blob/abbc2abc646e1f45e8534cec5991abdcabf9e889/onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h#L89

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,259 @@
...
+  RETURN_IF(smooth_softmax != 1.0f, "smooth_softmax != 1 not supported");
+
```
chunghow-qti, line 103: we need to reject non-zero rotary_interleaved?

```suggestion
int32_t rotary_interleaved = node_helper.Get("rotary_interleaved", 0);
RETURN_IF(rotary_interleaved != 0, "rotary_interleaved != 0 not supported");
```

chunghow-qti:
Just a quick note.  If there's any GPU specific logic or constraint,
I suggest handling it in `IsOpSupport`. That way user gets an explicit reason at partition time, and it'll be easier to extend for CPU & HTP backend support later.
```
bool is_gpu_backend = IsGpuBackend(qnn_model_wrapper.GetQnnBackendType());
```

chunghow-qti:
> > How well the support is for each backend? Do we plan to add tests?
> 
> I've added some GPU UTs to group_query_attention_test.cc. (I would suggest reviewing the 2 commits individually; I know the first one pollutes the overall diff). Note that they are disabled right now because we have to wait for 2.48 to release and for it to be picked up in packages.yaml for CI. The GPU backend will support this op in 2.48, with support for the packed_qkv format and some restrictions on attributes (e.g. do_rotary required to be enabled, rotary_interleaved not supported, local_window_size not supported fully).
> 
> I'm aware that the HTP backend is also intending to support this op, although I'm not sure what their timeline / support matrix will be. @chunghow-qti might be able to comment more about the HTP side.

Based on the information I have, HTP will support this in qairt 2.48.

chunghow-qti:
**[Major] ONNX GQA KV-cache quantization scheme (`k_quant_type` / `v_quant_type` / `kv_cache_bit_width`) is entirely unchecked**

The ONNX `com.microsoft.GroupQueryAttention` spec defines a full op-internal KV-cache quantization scheme; this PR reads none of the attributes:

- `k_quant_type : string` — "NONE" / "PER_TENSOR" / "PER_CHANNEL"
- `v_quant_type : string` — same
- `kv_cache_bit_width : int` — 8 or 4

The L62 rejection

```cpp
RETURN_IF(present_key_tensor_info.quant_param.IsQuantized(), "Quantized KV cache not supported");
```

only catches the **QDQ-wrapper path** (external `DequantizeLinear` / `QuantizeLinear` nodes). ONNX GQA quantization, however, is an **op-internal scheme**: `past_key` / `past_value` are directly `uint8` / `int8` / `float8e4m3fn` with no surrounding DQ/Q, so `IsQuantized()` always returns false here.

Real consequences (combined with L84-86's blanket rejection of `inputs[10..13]`):

- Model sets `k_quant_type="PER_CHANNEL"` and provides `k_scale`/`v_scale` at `inputs[12]`/`inputs[13]` → caught by L84, but the message `"attention_bias, head_sink, k_scale, and v_scale inputs are not supported"` misattributes the cause.
- Model sets `k_quant_type="PER_TENSOR"` but provides no `k_scale` → silently passes → undefined.
- Model sets `kv_cache_bit_width=4` with `uint8` KV cache → silently passes.

When the HTP backend is enabled later, this scheme is essentially mandatory (HTP LLM is the main use case for KV-cache quantization). Adding it would be from scratch since the attributes are not even read today. Near-term: at minimum reject all three attributes explicitly to prevent silent miscompute.

**Suggested fix** (inside `IsOpSupported()`):

```cpp
std::string k_quant_type = node_helper.Get("k_quant_type", std::string("NONE"));
RETURN_IF(k_quant_type != "NONE", "k_quant_type != NONE not supported");
std::string v_quant_type = node_helper.Get("v_quant_type", std::string("NONE"));
RETURN_IF(v_quant_type != "NONE", "v_quant_type != NONE not supported");
RETURN_IF(node_helper.HasAttr("kv_cache_bit_width"),
          "kv_cache_bit_width attribute not supported");
```

qti-mattsinc:
> **[Major] ONNX GQA KV-cache quantization scheme (`k_quant_type` / `v_quant_type` / `kv_cache_bit_width`) is entirely unchecked**
> ...

Good catch, fixed with new commit.



> Just a quick note. If there's any GPU specific logic or constraint, I suggest handling it in `IsOpSupport`. That way user gets an explicit reason at partition time, and it'll be easier to extend for CPU & HTP backend support later.
> 
> ```
> bool is_gpu_backend = IsGpuBackend(qnn_model_wrapper.GetQnnBackendType());
> ```

I can see the benefit of a clearer validation failure reason, but the downside to duplicating backend specific op validation onto the EP side is that, if the backend adds support for a new input/attr/output, then the EP must also be updated to remove the associated check in the op builder. I don't know if there is consensus on this among the current op builders; some of them do have these backend-specific checks but I don't think that's always the case? As it stands right now, the GQA op builder isn't making any GPU-specific checks, so it shouldn't be preventing any CPU/HTP support. `IsOpSupported` is basically figuring out if the onnx GQA can be mapped onto the QNN core GQA; because the QNN op does not support all of the same inputs/attrs/outputs that the onnx op does, this is not always the case.

qti-mattsinc:
Separating this PR into 3 changes:

1. The OpBuilder (this PR)
2. The cosine similarity verifier (#504)
3. The UTs (on the [dev/mattsinc/gqa-tests](https://github.com/onnxruntime/onnxruntime-qnn/tree/dev/mattsinc/gqa-tests) branch but not in a PR yet)

The reason for this is to prioritize getting 1 in without depending on 2 or 3 (esp. since 2 is going to require frequent rebases).

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+                                          std::vector<std::string>&& input_names,
+                                          const Ort::Logger& logger,
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+};
+
+Ort::Status GroupQueryAttentionOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                                        const OrtNodeUnit& node_unit,
+                                                        const Ort::Logger& logger) const {
+  ORT_UNUSED_PARAMETER(logger);
+
```
chunghow-qti, line 45: Could you add IsGpuBackend() check to support the GPU backend for this PR, as the other backends are not ready yet.
Thanks!

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+  if (num_inputs > 4 && inputs[4].Exists()) {
+    TensorInfo past_value_tensor_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[4], past_value_tensor_info));
+    RETURN_IF_NOT(past_value_tensor_info.shape.size() == 4, "Unexpected rank for past_value");
+    RETURN_IF_NOT(past_value_tensor_info.shape[2] == max_sequence_length,
+                  "QNN GroupQueryAttention requires past_value_shape[2] == present_value_shape[2] == max_sequence_length");
+  }
+
+  // At time of writing, the com.micorosoft.GroupQueryAttention op def has 14 inputs and 4 outputs.
```
huaychou:
```suggestion
  // At time of writing, the com.microsoft.GroupQueryAttention op def has 14 inputs and 4 outputs.
```

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
kuanyul-qti, line 1: **[M-1] (Major)** New op builder ships with no unit tests

This PR adds 265 lines of new op-builder code but adds no `group_query_attention_test.cc` under `onnxruntime/test/providers/qnn/`. `IsOpSupported()` contains 11+ reject branches (KV quant_type, local_window_size, qk_output, smooth_softmax, softcap, max input/output count, output_qk, past/present_key shape, required attributes), each of which is a regression point; `ProcessInputs()` additionally re-orders ONNX→QNN inputs and synthesizes NULL tensors — an error in the remapping order silently produces wrong outputs and cannot be caught by static review.

The QNN EP review checklist for new op support requires: every logic change has a corresponding unit test; new ops must include both CPU and HTP backend tests (u8/u16), see [C06]; bug-fix tests must reproduce first, then run a kill-test fence.

**Impact**: Any future QNN SDK uplevel, native-path refactor, or ONNX GQA spec change has no CI signal for this op. A reject branch flipped to accept, or any reorder of the `qnn_idx_to_onnx` remapping table, will go unnoticed.

**Suggested fix:**

Add `onnxruntime/test/providers/qnn/group_query_attention_test.cc` covering: FP32 / FP16 baseline on HTP and CPU backends (small q/kv heads, small seq); negative tests for each reject branch (`k_quant_type="INT8"`, `smooth_softmax=0.5`, `softcap=10.0`, `qk_output=1`) expecting `ExpectedEPNodeAssignment::None`; optional-input paths (past_key/past_value absent, cos_cache/sin_cache absent, position_ids absent); negative test where past_key/present_key shapes disagree on max_sequence_length. Follow the style of `rotary_embedding_test.cc`; because the op lives in the `com.microsoft` domain, declare opset and domain correctly.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
kuanyul-qti: **[N-1] (Minor)** Typo "com.micorosoft" in comment (should be "com.microsoft")

The comment reads `com.micorosoft`; the correct spelling is `com.microsoft`. Although this is only a comment, it makes future `grep`s for the contrib domain miss this line.

```
  // At time of writing, the com.micorosoft.GroupQueryAttention op def has 14 inputs and 4 outputs.
```

**Impact**: Comment-only; no runtime impact, but a codebase search for `com.microsoft` will miss this line.

**Suggested fix:**

```
  // At time of writing, the com.microsoft.GroupQueryAttention op def has 14 inputs and 4 outputs.
```

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+  Qnn_Scalar_t num_heads_scalar = QNN_SCALAR_INIT;
+  num_heads_scalar.dataType = QNN_DATATYPE_UINT_32;
+  num_heads_scalar.uint32Value = SafeInt<uint32_t>(num_heads.value());
+
+  QnnParamWrapper num_heads_param_wrapper(node_unit.Index(),
+                                          node_unit.Name(),
+                                          QNN_OP_GROUP_QUERY_ATTENTION_PARAM_NUM_HEADS,
+                                          num_heads_scalar);
+  param_names.emplace_back(num_heads_param_wrapper.GetParamTensorName());
+  qnn_model_wrapper.AddParamWrapper(std::move(num_heads_param_wrapper));
```
kuanyul-qti: **[N-2] (Minor)** AddParamWrapper bool return value silently discarded at 4 call sites

`QnnModelWrapper::AddParamWrapper(QnnParamWrapper&&)` returns a `bool` (false means the parameter name is empty and is logged as an error). The new `ProcessAttributesAndOutputs` discards the return value at 4 call sites (num_heads / kv_num_heads / do_rotary / scale, lines 180/194/208/223). QNN EP op-builders are inconsistent here — `pool_op_builder.cc`, `lp_pool_op_builder.cc`, and `inverse_op_builder.cc` wrap calls with `RETURN_IF_NOT(...)`; `argmax_argmin`, `clip`, `softmax` do not. New code should adopt the stricter style, aligned with the review checklist principle that status-returning helpers should be wrapped in `RETURN_IF`.

**Impact**: In practice the param name is always `node_unit.Index() + name + param` so the empty-string failure path is unreachable. But if `QnnParamWrapper` adds new failure conditions later (e.g. duplicate-ID conflicts), all four sites silently swallow them.

**Suggested fix:**

Wrap each call as `RETURN_IF_NOT(qnn_model_wrapper.AddParamWrapper(std::move(num_heads_param_wrapper)), "Failed to add num_heads param.");` and apply the same pattern to kv_num_heads / do_rotary / scale.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+
+  std::string k_quant_type = node_helper.Get("k_quant_type", std::string("NONE"));
+  RETURN_IF(k_quant_type != "NONE", "k_quant_type != NONE not supported");
+  std::string v_quant_type = node_helper.Get("v_quant_type", std::string("NONE"));
+  RETURN_IF(v_quant_type != "NONE", "v_quant_type != NONE not supported");
+  RETURN_IF(node_helper.HasAttr("kv_cache_bit_width"),
+            "kv_cache_bit_width attribute not supported");
+
+  int32_t local_window_size = node_helper.Get("local_window_size", -1);
+  RETURN_IF(local_window_size != -1 && SafeInt<uint32_t>(local_window_size) < max_sequence_length,
```
kuanyul-qti, line 89: **[N-3] (Minor)** SafeInt cast on negative local_window_size (other than -1) will throw

The new code `int32_t local_window_size = node_helper.Get("local_window_size", -1);` is followed by `RETURN_IF(local_window_size != -1 && SafeInt<uint32_t>(local_window_size) < max_sequence_length, ...)`. The ONNX GQA spec uses `-1` as the "no local window" sentinel; other negatives (e.g. `-2`) are out-of-spec, but if a model author or model generator writes `local_window_size = -2`, `SafeInt<uint32_t>(-2)` throws `SafeIntException`. An exception in `IsOpSupported` is not gracefully translated to "op unsupported → fallback"; it propagates and fails partitioning.

**Impact**: Spec-compliant models are unaffected. Loose / malformed models cause the partitioner to crash rather than gracefully fall back to CPU.

**Suggested fix:**

Reject any `local_window_size < -1` first (e.g. `RETURN_IF(local_window_size < -1, "local_window_size must be -1 (disabled) or a non-negative value");`), then do the SafeInt / static_cast comparison.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+  QnnParamWrapper kv_num_heads_param_wrapper(node_unit.Index(),
+                                             node_unit.Name(),
+                                             QNN_OP_GROUP_QUERY_ATTENTION_PARAM_KV_NUM_HEADS,
+                                             kv_num_heads_scalar);
+  param_names.emplace_back(kv_num_heads_param_wrapper.GetParamTensorName());
+  qnn_model_wrapper.AddParamWrapper(std::move(kv_num_heads_param_wrapper));
+
+  // do_rotary
+  std::optional<int64_t> do_rotary = node_helper.GetInt64("do_rotary");
+  if (do_rotary.has_value()) {
```
kuanyul-qti: **[N-4] (Minor)** do_rotary / scale silently omitted when ONNX attr absent — relies on QNN-side default

`do_rotary` and `scale` are read via `node_helper.GetInt64` / `GetFloat` (returning `std::optional`) rather than `node_helper.Get("do_rotary", 0)`. When the ONNX model does not set `do_rotary`, the PR omits `QNN_OP_GROUP_QUERY_ATTENTION_PARAM_DO_ROTARY` entirely and lets QNN fall back to its own default.

ONNX `com.microsoft.GroupQueryAttention` defines defaults `do_rotary = 0` (no rotary) and `scale = 1/sqrt(head_size)`. If QNN's defaults differ (e.g. QNN's `do_rotary=1` or `scale=0`), users who omit these attributes will see silently divergent numerics. The PR has no test that proves the two defaults align.

This is the same pattern as `[C73]` (Resize default omitted to let QNN infer), but only safe once QNN's default has been verified against ONNX's.

**Impact**: Real-world ONNX models that omit `do_rotary` / `scale` may produce numerics divergent from the CPU EP, with no decline path triggered.

**Suggested fix:**

Two acceptable fixes: (a) align with the ONNX spec default explicitly: `node_helper.Get("do_rotary", int64_t{0})`, and for `scale` compute `1.0f / sqrtf(head_size)` in the op builder when the attribute is absent — do not rely on the QNN-side default; (b) if QNN's default is verified equal to ONNX's, add a comment that documents the equivalence and include a unit test that exercises the "ONNX omits do_rotary/scale" path against the CPU EP baseline.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+
+Ort::Status GroupQueryAttentionOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                                        const OrtNodeUnit& node_unit,
+                                                        const Ort::Logger& logger) const {
+  ORT_UNUSED_PARAMETER(logger);
+
+  const size_t num_inputs = node_unit.Inputs().size();
+  const auto& inputs = node_unit.Inputs();
+
+  const size_t num_outputs = node_unit.Outputs().size();
```
kuanyul-qti, line 52: **[T-1] (Nit)** Local num_inputs / num_outputs are partially redundant

`IsOpSupported` declares both `num_inputs / inputs` and `num_outputs / outputs`, but later code mixes `outputs.size() > 1` / `outputs[3].Exists()` with `num_outputs`. Pick one style — either always `size()` or always the cached variable.

**Suggested fix:**

(Optional) Drop `num_inputs / num_outputs` and use `inputs.size()` / `outputs.size()` throughout to match the style of other op builders.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+  const auto& outputs = node_unit.Outputs();
+
+  TensorInfo present_key_tensor_info = {};
+  RETURN_IF_NOT(outputs.size() > 1 && outputs[1].Exists(), "Required output tensor present_key not provided");
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(outputs[1], present_key_tensor_info));
+  RETURN_IF_NOT(present_key_tensor_info.shape.size() == 4, "Unexpected rank for present_key");
+  const auto max_sequence_length = present_key_tensor_info.shape[2];
+
+  if (num_inputs > 3 && inputs[3].Exists()) {
```
huaychou: Just wondering, why do we need to check past_key and past_value shape for IsOpSupported?

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+    RETURN_IF_NOT(past_value_tensor_info.shape.size() == 4, "Unexpected rank for past_value");
+    RETURN_IF_NOT(past_value_tensor_info.shape[2] == max_sequence_length,
+                  "QNN GroupQueryAttention requires past_value_shape[2] == present_value_shape[2] == max_sequence_length");
+  }
+
+  // At time of writing, the com.micorosoft.GroupQueryAttention op def has 14 inputs and 4 outputs.
+  const size_t max_num_inputs = 14;
+  const size_t max_num_outputs = 4;
+
+  for (size_t i = 10; i < std::min(num_inputs, max_num_inputs); i++) {
```
huaychou, line 65:
```suggestion
  for (size_t i = 10; i < num_inputs; i++) {
```

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+    RETURN_IF(inputs[i].Exists(), "attention_bias, head_sink, k_scale, and v_scale inputs are not supported");
+  }
+  RETURN_IF(num_inputs > max_num_inputs,
+            ("More than " + std::to_string(max_num_inputs) + " inputs provided, which is unsupported").c_str());
+
+  RETURN_IF(num_outputs > 3 && outputs[3].Exists(), "output_qk output is not supported");
+  RETURN_IF(num_outputs > 4,
```
huaychou:
```suggestion
  RETURN_IF(num_outputs > max_num_outputs ,
```

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+  ORT_UNUSED_PARAMETER(logger);
+
+  OrtNodeAttrHelper node_helper(node_unit);
+
+  std::vector<std::string> param_names;
+
+  // num_heads
+  std::optional<int64_t> num_heads = node_helper.GetInt64("num_heads");
+  RETURN_IF_NOT(num_heads.has_value(), "required attribute num_heads not provided");
+  Qnn_Scalar_t num_heads_scalar = QNN_SCALAR_INIT;
```
huaychou: Can we use AddQnnScalar?

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+                                          node_unit.Name(),
+                                          QNN_OP_GROUP_QUERY_ATTENTION_PARAM_NUM_HEADS,
+                                          num_heads_scalar);
+  param_names.emplace_back(num_heads_param_wrapper.GetParamTensorName());
+  qnn_model_wrapper.AddParamWrapper(std::move(num_heads_param_wrapper));
+
+  // kv_num_heads
+  std::optional<int64_t> kv_num_heads = node_helper.GetInt64("kv_num_heads");
+  RETURN_IF_NOT(kv_num_heads.has_value(), "required attribute kv_num_heads not provided");
+  Qnn_Scalar_t kv_num_heads_scalar = QNN_SCALAR_INIT;
```
huaychou: ditto

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+                                             node_unit.Name(),
+                                             QNN_OP_GROUP_QUERY_ATTENTION_PARAM_KV_NUM_HEADS,
+                                             kv_num_heads_scalar);
+  param_names.emplace_back(kv_num_heads_param_wrapper.GetParamTensorName());
+  qnn_model_wrapper.AddParamWrapper(std::move(kv_num_heads_param_wrapper));
+
+  // do_rotary
+  std::optional<int64_t> do_rotary = node_helper.GetInt64("do_rotary");
+  if (do_rotary.has_value()) {
+    Qnn_Scalar_t do_rotary_scalar = QNN_SCALAR_INIT;
```
huaychou: ditto

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+                                            QNN_OP_GROUP_QUERY_ATTENTION_PARAM_DO_ROTARY,
+                                            do_rotary_scalar);
+    param_names.emplace_back(do_rotary_param_wrapper.GetParamTensorName());
+    qnn_model_wrapper.AddParamWrapper(std::move(do_rotary_param_wrapper));
+  }
+
+  // scale
+  std::optional<float> scale = node_helper.GetFloat("scale");
+  if (scale.has_value()) {
+    Qnn_Scalar_t scale_scalar = QNN_SCALAR_INIT;
```
huaychou: ditto

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+                                        node_unit.Name(),
+                                        QNN_OP_GROUP_QUERY_ATTENTION_PARAM_SCALE,
+                                        scale_scalar);
+    param_names.emplace_back(scale_param_wrapper.GetParamTensorName());
+    qnn_model_wrapper.AddParamWrapper(std::move(scale_param_wrapper));
+  }
+
+  std::vector<std::string> output_names;
+  const auto& outputs = node_unit.Outputs();
+  for (size_t output_idx = 0; output_idx < outputs.size(); ++output_idx) {
```
huaychou:
```suggestion
  for (size_t output_idx = 0; output_idx < outputs.size() && outputs[output_idx].Exists(); ++output_idx) {
```

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: See https://github.com/onnxruntime/onnxruntime-qnn/pull/424#issuecomment-4662683807

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Because I misread the QNN OpDef 😅. Thanks for catching it. Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: The point of the `min` is that, if more inputs get added to the ORT OpDef in the future, and the OpBuilder is called with, say, 16 inputs, then this log message would still be accurate. The log would be:

> More than 14 inputs provided, which is unsupported

Compared to

> attention_bias, head_sink, k_scale, and v_scale inputs are not supported

Which would be inaccurate.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Actually, the default value is -1: https://github.com/microsoft/onnxruntime/blob/abbc2abc646e1f45e8534cec5991abdcabf9e889/onnxruntime/core/graph/contrib_ops/bert_defs.cc#L1242-L1245.

But yes this was wrong originally. Fixed now

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Good catch. Yes we need to reject it. Although I think we could support it by decomposing into unfused QNN RotorayEmbeddings + GQA (because QNN RotaryEmbedding does support the interleaved param, although QNN GQA does not). I've left a comment about it, and it can be left to future work.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Switched to AddQnnScalar

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Changed to mimic `ProcessInputs` (add null tensor if output doesn't exist). Technically, this would be relevant if e.g. present_key does not exist but present_value does (since putting the check in the loop condition would cause it to terminate before reaching present_value). Now, I think that's probably not allowed by either QNN or ORT and the op would be rejected, but it's safest to just handle it correctly

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc → IsOpSupported()`
The 254-line op builder still has no `group_query_attention_test.cc`. `IsOpSupported()` has 10+ reject branches and `ProcessInputs()` re-orders ONNX→QNN inputs via `qnn_idx_to_onnx`; a remapping error produces silently-wrong output. The op is currently `#if`-compiled-out at opset ≤ 2.11, so there is no regression today, but it goes live untested when the guard is dropped at the 2.48 uplevel. At minimum add the GPU-backend fallback negative test (`ExpectedEPNodeAssignment::None`) and per-reject-branch negative tests that run on the current SDK; add FP32/FP16 + input-remapping + optional-input positive tests with the 2.48 enablement.

---

### 🔵 Minor

**[N-3]** `onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc:L86`
`SafeInt<uint32_t>(local_window_size)` throws `SafeIntException` for any negative value other than `-1` (the disabled sentinel). An out-of-spec model with e.g. `local_window_size = -2` would throw inside `IsOpSupported` and fail partitioning instead of falling back to CPU. Reject `local_window_size < -1` first, or compare via a widening `static_cast<int64_t>`.

**[N-5]** `onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc:L159`
`num_heads.value()` (and `kv_num_heads.value()` at L170) is called *before* the `has_value()` check on the next line, making the check dead code — an empty optional throws `std::bad_optional_access` first. Currently shielded by the `HasAttr` guards in `IsOpSupported`, so it is latent, but the order is inverted. Move the `RETURN_IF_NOT(...has_value()...)` above the `value()` call.
```suggestion
  std::optional<int64_t> num_heads = node_helper.GetInt64("num_heads");
  RETURN_IF_NOT(num_heads.has_value(), "required attribute num_heads not provided");
  uint32_t num_heads_u32 = SafeInt<uint32_t>(num_heads.value());
```

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
tirupath-qti: @qti-mattsinc, we can add the suggested decomposition in a follow-up. could you please complete the future work in a follow-up PR if possible.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,254 @@
...
+  //       We could support ONNX GQA w/ rotary_interleaved by decomposing it into QNN RotaryEmbeddings (w/ interleaved)
+  //       + QNN GQA.
+  int32_t rotary_interleaved = node_helper.Get("rotary_interleaved", 0);
+  RETURN_IF(rotary_interleaved != 0, "rotary_interleaved != 0 not supported");
+
+  int32_t smooth_softmax = node_helper.Get("smooth_softmax", -1);
+  RETURN_IF(smooth_softmax != -1, "smooth_softmax != -1 not supported");
+
+  float softcap = node_helper.Get("softcap", 0.0f);
+  RETURN_IF(softcap != 0.0f, "softcap != 0 not supported");
```
tirupath-qti, line 105: We should run QNN validation otherwise we may not see the functional failures until Phase-2 Compile, where there is no CPU fallback (it aborts the functionality).

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Done. Seems like not all op builders are following this pattern (e.g. lstm_op_builder.cc) and should be updated at some point.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: Sure, I'll add the decomposition for interleaved in a later PR.

qti-mattsinc:
> ### 🟠 Major
> 
> **[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc → IsOpSupported()` The 254-line op builder still has no `group_query_attention_test.cc`. `IsOpSupported()` has 10+ reject branches and `ProcessInputs()` re-orders ONNX→QNN inputs via `qnn_idx_to_onnx`; a remapping error produces silently-wrong output. The op is currently `#if`-compiled-out at opset ≤ 2.11, so there is no regression today, but it goes live untested when the guard is dropped at the 2.48 uplevel. At minimum add the GPU-backend fallback negative test (`ExpectedEPNodeAssignment::None`) and per-reject-branch negative tests that run on the current SDK; add FP32/FP16 + input-remapping + optional-input positive tests with the 2.48 enablement.

See https://github.com/onnxruntime/onnxruntime-qnn/pull/424#issuecomment-4662683807

> ### 🔵 Minor
> 
> **[N-3]** `onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc:L86` `SafeInt<uint32_t>(local_window_size)` throws `SafeIntException` for any negative value other than `-1` (the disabled sentinel). An out-of-spec model with e.g. `local_window_size = -2` would throw inside `IsOpSupported` and fail partitioning instead of falling back to CPU. Reject `local_window_size < -1` first, or compare via a widening `static_cast<int64_t>`.

Done

> **[N-5]** `onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc:L159` `num_heads.value()` (and `kv_num_heads.value()` at L170) is called _before_ the `has_value()` check on the next line, making the check dead code — an empty optional throws `std::bad_optional_access` first. Currently shielded by the `HasAttr` guards in `IsOpSupported`, so it is latent, but the order is inverted. Move the `RETURN_IF_NOT(...has_value()...)` above the `value()` call.

Done

huaychou (approved):
LGTM

onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
```diff
@@ -28,6 +28,10 @@ OpBuilderRegistrations::OpBuilderRegistrations() {
   CreateGatherNDOpBuilder("GatherND", *this);
   CreateGemmOpBuilder("Gemm", *this);
   CreateGroupNormalizationOpBuilder("GroupNormalization", *this);
+// GQA not available until opset version 2.12.0 (QAIRT 2.48). TODO: Remove this check once the EP uplevels to 2.48.
+#if !(QNN_OPSET_VERSION_MAJOR < 2 || (QNN_OPSET_VERSION_MAJOR == 2 && QNN_OPSET_VERSION_MINOR <= 11))
```
chunghow-qti: The QNN_OPSET_VERSION_* macros are defined in QnnOpDef.h, which is not (transitively) included in this translation unit before the #if is evaluated. Undefined macros evaluate to 0 in #if, so this gate is always false and the registration silently disappears. This still compiles and links, but at runtime GetOpBuilder("GroupQueryAttention") returns nullptr and the op is reported as "not supported by QNN EP". Better to drop the gate here and keep the version check only in **group_query_attention_op_builder.cc**, where QnnOpDef.h is included (via base_op_builder.h) before the #if:

``` 
# in group_query_attention_op_builder.cc 
#include <cmath>

#include "core/providers/qnn/builder/op_builder_factory.h"
#include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
#include "core/providers/qnn/builder/qnn_utils.h"
#if !(QNN_OPSET_VERSION_MAJOR < 2 || (QNN_OPSET_VERSION_MAJOR == 2 && QNN_OPSET_VERSION_MINOR <= 10))
```

qti-mattsinc: Done

> Better to drop the gate here and keep the version check only in group_query_attention_op_builder.cc, where QnnOpDef.h is included (via base_op_builder.h) before the #if:

We have to ensure the `CreateGroupQueryAttentionOpBuilder` symbol is not referenced when it's not defined. So either a) guard both the definition (group_query_attention_op_builder.cc) and the callsite (op_builder_factory.cc) with the macro or b) make sure it's always defined (but just a no-op if not enabled). Changed to option b.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,265 @@
...
+  // scale
+  RETURN_IF_NOT(num_outputs > 0 && outputs[0].Exists(), "Required output out[0] not provided");
+  TensorInfo output_tensor_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(outputs[0], output_tensor_info));
+  RETURN_IF_NOT(output_tensor_info.shape.size() == 3, "Unexpected rank for output out[0]");
+  const size_t head_size = output_tensor_info.shape[2] / num_heads.value();
+  RETURN_IF(head_size == 0, "head_size can't be zero!");
+
+  const float scale_default = 1.0f / std::sqrtf(static_cast<float>(head_size));
+  const float scale = node_helper.Get("scale", scale_default);
```
chunghow-qti, line 205: https://github.com/microsoft/onnxruntime/blob/abbc2abc646e1f45e8534cec5991abdcabf9e889/onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h#L83
https://github.com/microsoft/onnxruntime/blob/abbc2abc646e1f45e8534cec5991abdcabf9e889/onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h#L950
Seems for ORT CPU EP, scale == 0 is treated as a sentinel for "use the default" (1/sqrt(head_size)).

```suggestion
  const float scale_default = 1.0f / std::sqrtf(static_cast<float>(head_size));
  float scale = node_helper.Get("scale", scale_default);
  if (scale == 0.0f) {
    scale = scale_default;
  }
```

chunghow-qti (approved):
LGTM

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: That's interesting. I don't see that mentioned in the ORT spec for the com.microsoft.GroupQueryAttention op (though maybe it's just assumed since scale=0 wouldn't make sense otherwise). That said, this *is* actually mentioned in the QNN OpDef for GQA:

> __scale__
> Scalar multiplier applied to QK^T before softmax. When set to 0.0 or omitted, defaults to 1 / sqrt(head_size).
> 
> Mandatory: false
> 
> Data type: QNN_DATATYPE_FLOAT_32
> 
> Shape: scalar
> 
> Default: 0

So, we'll get the same behavior as ORT CPU EP as things are currently, by passing 0.0 along to the QNN op.

qti-mattsinc:
Thank you @huaychou and @chunghow-qti for reviewing!

@qti-chuteng, we need sign-off from a code-owner to merge. Could you have a final look?

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,298 @@
...
+  const uint32_t kv_num_heads_u32 = SafeInt<uint32_t>(kv_num_heads.value());
+  RETURN_IF_ERROR(AddQnnScalar(qnn_model_wrapper,
+                               node_unit.Index(),
+                               node_unit.Name(),
+                               kv_num_heads_u32,
+                               QNN_OP_GROUP_QUERY_ATTENTION_PARAM_KV_NUM_HEADS,
+                               param_names));
+
+  // do_rotary
+  const int64_t do_rotary = node_helper.Get("do_rotary", 0ll);
```
huaychou, line 187:
```suggestion
  const int64_t do_rotary = node_helper.Get("do_rotary", static_cast<int64_t>(0));
```

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -0,0 +1,298 @@
...
+
+  // scale
+  RETURN_IF_NOT(num_outputs > 0 && outputs[0].Exists(), "Required output out[0] not provided");
+  TensorInfo output_tensor_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(outputs[0], output_tensor_info));
+  RETURN_IF_NOT(output_tensor_info.shape.size() == 3, "Unexpected rank for output out[0]");
+  const size_t head_size = output_tensor_info.shape[2] / num_heads.value();
+  RETURN_IF(head_size == 0, "head_size can't be zero!");
+
+  const float scale_default = 1.0f / std::sqrtf(static_cast<float>(head_size));
```
huaychou, line 204:
```suggestion
  const float scale_default = 1.0f / std::sqrt(static_cast<float>(head_size));
```

huaychou:
Should be fixed in https://github.com/onnxruntime/onnxruntime-qnn/pull/579

PR #427, opened 2026-05-21
[QNN EP] Add support to build ARM64 (ARM64x) PDB Build

qti-mbadnara:
This PR adds support to build PDB builds for ARM64 (ARM64x) needed for WinML workflows


huaychou:
The change looks good to me. We need to check:
1. Trigger workflow and success
2. Check the pdb is for arm64x QNN EP

kuanyul-qti (approved):
Please also check the name of the arm64x zip file.

qti-chuteng (approved):

PR #428, opened 2026-05-21
[QNN EP] Add support to publish Archive to GitHub

qti-mbadnara:
## Description

Adds `github` as a valid `index_server_to` destination for ZIP and TGZ artifacts, enabling automated GitHub Release creation as part of the existing uplevel workflow. When selected, the script creates an annotated git tag, opens a draft release, and attaches the artifacts.

## Summary

- Extends `index_server_to` to accept `github` as a valid destination for `zip` and `tgz` artifact formats
- When `index_server_to=github`, the script automatically creates an annotated git tag `v<version_to>`, opens a **draft** GitHub Release, and attaches the artifacts
- Existing tags and release assets are handled gracefully on re-runs (tag reused, assets overwritten via `--clobber`)

### `qcom/scripts/upleveling/qnn_ep_uplevel.py`
- `ArtifactUpleveler.__init__`: skip Artifactory URL lookup when `index_server_to=github`
- `ZipUpleveler._upload_to_github()`: new method — creates git tag, creates draft GitHub Release, uploads artifacts
- `ZipUpleveler.upload_artifacts()`: dispatches to `_upload_to_github()` when destination is `github`; `TgzUpleveler` inherits this for free
- `parse_arguments`: updated `--index_server_to` help text to document the new `github` option

### `.github/workflows/qualcomm-internal-upload-artifactory.yml`
- `contents: write` permission added (required to push tags and create releases)
- `index_server_to` description updated to document the `github` option
- `GH_TOKEN`: added to the Publish step env in both zip and tgz jobs

## How to use

| Input | Value |
|---|---|
| `artifact_format` | `zip`, `tgz`, or `all` |
| `version_from` | `2.2.0rc1` |
| `version_to` | `2.2.0` |
| `index_server_from` | `project` |
| `index_server_to` | `github` |

The script downloads from Artifactory, renames artifacts from `2.2.0rc1` → `2.2.0`, creates git tag `v2.2.0`, opens a draft release, and attaches the files. The release must be manually published on GitHub after review.


qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -1069,8 +1077,43 @@ def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir:
...
+            subprocess.run(["git", "tag", "-a", tag, "-m", f"Release {tag}"], check=True)
+            subprocess.run(["git", "push", "origin", tag], check=True)
+            logging.info(f"Created and pushed git tag {tag}")
+        else:
+            logging.info(f"Git tag {tag} already exists, reusing it")
+        # Create the GitHub Release if it does not exist yet (tag is already pinned above).
+        subprocess.run(["gh", "release", "create", tag, "--title", tag, "--notes", "", "--draft"], check=False)
+
+        # Attach assets; --clobber replaces any existing asset with the same name (safe for re-runs).
+        subprocess.run(["gh", "release", "upload", tag, "--clobber", *files], check=True)
```
kuanyul-qti, line 1142: It appears this process will generate a draft release with a version tag and artifacts. Afterward, does the user need to manually update the release notes on the GitHub release page?

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -1069,8 +1077,43 @@ def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir:
...
+            == 0
+        )
+        if not tag_exists:
+            subprocess.run(["git", "tag", "-a", tag, "-m", f"Release {tag}"], check=True)
+            subprocess.run(["git", "push", "origin", tag], check=True)
+            logging.info(f"Created and pushed git tag {tag}")
+        else:
+            logging.info(f"Git tag {tag} already exists, reusing it")
+        # Create the GitHub Release if it does not exist yet (tag is already pinned above).
+        subprocess.run(["gh", "release", "create", tag, "--title", tag, "--notes", "", "--draft"], check=False)
```
kuanyul-qti: If a release tag is generated, what happens with this line?
For example, we've triggered the GitHub release flow for zip archive, and then we trigger the GitHub release flow for Tgz archive.

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Yes that is the goal!

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: When the tgz job runs after the zip job has already created the draft release:

- Verify if tag exists -> Yes, created by zip job → Skip tag creation
- gh release create → exits non-zero with release already exists
- gh release upload ... --clobber → succeeds, attaches tgz files to the existing draft release
- Net result: the same draft release ends up with both zip and tgz assets. 

That's the intended design

But when they run in parallel, it will endup in a race condition. 

If both check the git tag at the same instant, both will create the local tag, and both attempt git push origin <tag> — one push succeeds, the other is rejected.

I have made the tag-push step race-tolerant in `qnn_ep_uplevel.py` as part of the latest commit

- git push runs with check=False.
- If push fails, we git fetch origin tag <tag> and re-verify with git rev-parse.
- If the tag is now present on the remote (the other job pushed it first), we log "pushed concurrently by another job, reusing it" and continue.
- If the tag still isn't there, we raise with the captured stderr.
- So zip and tgz can run in parallel safely; both gets published to the same draft release

.github/workflows/qualcomm-internal-upload-artifactory.yml
```diff
@@ -185,6 +186,7 @@ jobs:
           ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
           ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }}
           JFROG_API_KEY: ${{ secrets.JFROG_API_KEY }}
+          GH_TOKEN: ${{ secrets.github.token }}
```
kuanyul-qti: Should it be `GH_TOKEN: ${{ github.token }}`?

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -145,11 +145,19 @@ def __init__(self, args: argparse.Namespace):
         self.url_from = self.config_manager.get_repository_url(
             args.index_server_from, args.product_name, args.version_from
         )
-        self.url_to = self.config_manager.get_repository_url(args.index_server_to, args.product_name, args.version_to)
+
+        # GitHub Releases has no Artifactory URL; url_to is unused in that path.
+        if args.index_server_to == "github":
+            self.url_to = ""
+            self.url_to_display = "GitHub Releases"
```
kuanyul-qti, line 152: Since `url_to` is not used in GitHub release path, is the change necessary?

qcom/scripts/upleveling/qnn_ep_uplevel.py
kuanyul-qti: `gh release create` is called with `check=False` and no return-code inspection. While the intent is to tolerate "release already exists" on re-runs, a genuine auth failure (e.g., empty-token issue) or repo-not-found error also passes silently, and the subsequent `gh release upload` will fail with a confusing error message pointing to upload rather than the root cause.
Check `stderr` for `"already exists"` with `capture_output=True` to distinguish the two cases and raise a `RuntimeError` on unexpected failures.

.github/workflows/qualcomm-internal-upload-artifactory.yml
```diff
@@ -5,7 +5,7 @@ name: Publish ONNX Runtime Execution Provider wheel, nuget, archive and maven
 
 permissions:
   checks: write
-  contents: read
+  contents: write
```
kuanyul-qti: Minor: `contents: write` is granted at the workflow level, applying to all jobs including `publish-wheel-workflow` and `publish-nuget-workflow` that do not need it. Follow least-privilege: restore the top-level to `contents: read` and add job-level `permissions: contents: write` only to `publish-zip-workflow` and `publish-tgz-workflow`.

.github/workflows/qualcomm-internal-upload-artifactory.yml
```diff
@@ -219,6 +225,10 @@ jobs:
     name: "Publish TGZ artifacts to Artifactory"
```
kuanyul-qti:
```suggestion
    name: "Publish TGZ artifacts to Artifactory or GitHub"
```

kuanyul-qti (approved):

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -1069,8 +1077,77 @@ def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir:
...
+        if tag_exists:
+            logging.info(f"Git tag {tag} already exists, reusing it")
+        else:
+            subprocess.run(["git", "tag", "-a", tag, "-m", f"Release {tag}"], check=True)
+            push_result = subprocess.run(["git", "push", "origin", tag], check=False, capture_output=True)
+            if push_result.returncode == 0:
+                logging.info(f"Created and pushed git tag {tag}")
+            else:
+                # Push lost a race with a concurrent job — pull the remote tag and reuse it.
+                subprocess.run(["git", "fetch", "origin", "tag", tag], check=False, capture_output=True)
```
huaychou, line 1110: Just wondering. why don't we fetch first to prevent this situation?

tirupath-qti (approved):
LGTM

PR #429, opened 2026-05-22
[QNN EP] Conv BQ Phase 1: support Conv with block-quantized weight on HTP

huaychou:
### Description

- Enables block-quantized weight Conv on QNN HTP via the BW_FLOAT_BLOCK kernel (BQ path)
- Add int2 support in core and test framework
- Add corresponding unit tests

### Motivation and Context

- Support Block Quantization in QNN

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -357,6 +357,23 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
       const int64_t IC = static_cast<int64_t>(input_info.shape[1]);
       const int64_t nb = bq_scale_shape[1];  // num_blocks_per_oc (validated: IC % nb == 0)
       const int64_t block_size = IC / nb;
+      const uint32_t bitwidth = GetBQBitwidth(inputs[1].type);
+
+      // For unsigned types (UINT2/UINT4/UINT8), shift weight data to the signed domain.
+      // QNN BW_FLOAT_BLOCK only supports SFIXED_POINT_8 (signed); unsigned data must be converted.
+      // TransformUnsignedToSignedFixedPoint XORs the MSB of each packed element:
+      //   - UINT8 (bitwidth=8, mask=0x80): 1 byte/element — directly converts [0,255] → [-128,127].
+      //   - UINT4 (bitwidth=4, mask=0x88): after TransposeFromNchwToHwcn unpack, each byte holds
+      //     a UINT4 value in its lower nibble (upper nibble=0). XOR with 0x88 flips bits 3 and 7;
+      //     QNN reads only the lower nibble for bitwidth=4, so the result is equivalent to
+      //     flipping bit 3, which shifts [0,15] → signed-nibble-interpreted [-8,7] correctly.
+      const bool is_unsigned_weight = (inputs[1].type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT2 ||
+                                       inputs[1].type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4 ||
+                                       inputs[1].type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8);
+      if (is_unsigned_weight) {
+        RETURN_IF_ERROR(utils::TransformUnsignedToSignedFixedPoint(unpacked_tensor,
+                                                                   static_cast<int64_t>(bitwidth)));
+      }
 
       // QNN BW_FLOAT_BLOCK blockSize for HWCN weight: {1, 1, block_size, 1}.
       // Each block spans block_size consecutive IC elements; H and W dimensions are not blocked.
@@ -369,9 +386,11 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
                     "QNN EP: BQ Conv scale size mismatch");
       // QNN BW_FLOAT_BLOCK expects scales in [OC, nb] order — same as ONNX, no reordering needed.
 
-      // Float offsets in [OC, nb] order, derived from zero_points: offset[i] = -zp[i].
-      // BW_FLOAT_BLOCK offsets are unconstrained floats; defaults to 0.0f when no zero_points.
-      std::vector<float> offsets_qnn(static_cast<size_t>(OC * nb), 0.0f);
+      // Float offsets in [OC, nb] order.
+      // Signed:   offset = -onnx_zp
+      // Unsigned: offset = (1 << (bits-1)) - onnx_zp   (compensates for unsigned→signed shift above)
+      const float unsigned_bias = is_unsigned_weight ? static_cast<float>(1u << (bitwidth - 1)) : 0.0f;
+      std::vector<float> offsets_qnn(static_cast<size_t>(OC * nb), unsigned_bias);
       if (inputs[1].quant_param->zero_point != nullptr) {
         std::vector<int32_t> zp_values;
         ONNXTensorElementDataType zp_onnx_type = ONNX_TENSOR_ELEMENT_DATA_TYPE_UNDEFINED;
@@ -380,24 +399,18 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
         RETURN_IF_NOT(static_cast<int64_t>(zp_values.size()) == OC * nb,
                       "QNN EP: BQ Conv zero_point size must match OC * num_blocks");
         for (size_t idx = 0; idx < static_cast<size_t>(OC * nb); ++idx) {
-          offsets_qnn[idx] = static_cast<float>(-zp_values[idx]);
+          offsets_qnn[idx] = unsigned_bias - static_cast<float>(zp_values[idx]);
         }
       }
 
       QnnQuantParamsWrapper bq_quant_params(gsl::span<const float>(scales_qnn),
                                             gsl::span<const float>(offsets_qnn),
-                                            GetBQBitwidth(inputs[1].type),
+                                            bitwidth,
                                             gsl::span<const uint32_t>(block_size_arr));
 
-      // Unpacked weight: each element occupies 1 byte regardless of bitwidth.
-      // Use signed or unsigned INT8 storage matching the weight element type.
-      const bool is_signed_weight = (inputs[1].type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT2 ||
-                                     inputs[1].type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4 ||
-                                     inputs[1].type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8);
-      const Qnn_DataType_t weight_qnn_dtype = is_signed_weight ? QNN_DATATYPE_SFIXED_POINT_8
-                                                               : QNN_DATATYPE_UFIXED_POINT_8;
+      // Always use SFIXED_POINT_8: unsigned types are pre-converted by TransformUnsignedToSignedFixedPoint.
       QnnTensorWrapper bq_weight_wrapper(input1_name, tensor_type,
-                                         weight_qnn_dtype,
+                                         QNN_DATATYPE_SFIXED_POINT_8,
                                          std::move(bq_quant_params),
                                          std::move(hwcn_shape),
                                          std::move(unpacked_tensor));
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -12,6 +12,7 @@
 namespace onnxruntime {
 namespace qnn {
 
+namespace {
 // HTP BQ Conv: supported bitwidths and their block_size divisor constraints.
 // block_size must be a multiple of the corresponding value (same as MatMulNBits HTP constraints).
 const std::unordered_map<uint32_t, int64_t> kHtpConvBQBitsAndBlockSizeMultipliers{
@@ -33,6 +34,7 @@ static uint32_t GetBQBitwidth(ONNXTensorElementDataType onnx_type) {
       return 0;
   }
 }
+}  // namespace
 
 // ONNX convolution types supported by this builder.
 // We translate node_unit.OpType() into this enum to avoid repeated string comparisons.
@@ -137,11 +139,9 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
   // Validate quantization axis for per-channel quantized weights.
   bool is_npu_backend = IsNpuBackend(qnn_model_wrapper.GetQnnBackendType());
   if (is_npu_backend) {
-    // Detect block-quantized weight. Two supported scale layouts:
-    //   (a) Standard (ONNX opset 21): scale rank == weight rank, e.g. [OC, IC/bs, H, W] for
-    //       axis=1; scale_shape[1] == IC/block_size < weight_shape[1].
-    //   (b) Legacy rank-2: [OC, num_blocks]; scale_shape[1] == IC/block_size.
-    // In both cases scale_shape[1] == num_blocks_per_oc for 1x1 or IC-axis blocking.
+    // Detect block-quantized weight. Per ONNX opset 21, the scale rank equals the weight
+    // rank: for a Conv2D weight [OC, IC, H, W] blocked on axis=1 the scale is
+    // [OC, IC/block_size, H, W], so scale_shape[1] == num_blocks_per_oc < weight_shape[1].
     if (inputs[1].quant_param.has_value() && inputs[1].quant_param->scale != nullptr) {
       const auto weight_scale_shape = utils::GetInitializerShape(inputs[1].quant_param->scale,
                                                                  qnn_model_wrapper.GetOrtApi());
@@ -150,11 +150,10 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
           weight_shape.size() != 4) {
         // BQ only supported for Conv2D (rank-4 weight); skip for Conv1D (rank-3) and Conv3D (rank-5).
       } else {
-        const bool is_rank2_scale = (weight_scale_shape.size() == 2);
-        const bool is_rankN_scale = (weight_scale_shape.size() == weight_shape.size() &&
+        const bool is_block_quant = (weight_scale_shape.size() == weight_shape.size() &&
                                      weight_scale_shape.size() > 1 &&
                                      weight_scale_shape[1] < weight_shape[1]);
-        if (is_rank2_scale || is_rankN_scale) {
+        if (is_block_quant) {
           const int64_t num_blocks_per_oc = weight_scale_shape[1];
           RETURN_IF(num_blocks_per_oc <= 0, "QNN EP: BQ num_blocks must be positive");
           const int64_t ic = static_cast<int64_t>(weight_shape[1]);
@@ -177,10 +176,10 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
 
           // Return success; full QNN validation is done in the NHWC IsOpSupported path.
           return Ort::Status();
-        }  // end is_rank2_scale || is_rankN_scale
+        }  // end is_block_quant
       }  // end else (weight shape obtainable)
     }  // end quant_param check
-
+    // checking for per-channel quantization
     const auto& input_1 = inputs[1];  // weight
     bool is_per_axis_quant = false;
     int64_t quant_axis = 0;
@@ -250,11 +249,9 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
   //
   RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[0], logger, input_names));
 
-  // Detect block-quantized weight. Accepted scale layouts:
-  //   Standard (ONNX opset 21): scale rank == weight rank, scale_shape[1] < weight_shape[1].
-  //     Weight is always NCHW [OC,IC,H,W] in both NCHW and NHWC domains; IC at index 1.
-  //   Legacy rank-2: [OC, num_blocks].
-  // scale_shape[1] == num_blocks_per_oc in all supported cases.
+  // Detect block-quantized weight. Per ONNX opset 21, the scale rank equals the weight rank
+  // with scale_shape[1] < weight_shape[1] (the blocked IC axis). Weight is always NCHW
+  // [OC,IC,H,W] in both NCHW and NHWC domains; IC at index 1, so scale_shape[1] == num_blocks_per_oc.
   bool is_bq_weight = false;
   std::vector<int64_t> bq_scale_shape;
   if (IsNpuBackend(qnn_model_wrapper.GetQnnBackendType()) &&
@@ -266,11 +263,8 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
         bq_scale_shape.size() > 1 &&
         bq_weight_shape.size() == 4) {  // BQ only for Conv2D (rank-4 weight); not Conv3D (rank-5)
       // Weight is NCHW [OC,IC,H,W]: IC is always at index 1.
-      const bool rank2 = (bq_scale_shape.size() == 2);
-      const bool rankN = (bq_scale_shape.size() == bq_weight_shape.size() &&
-                          bq_weight_shape.size() > 1 &&
-                          bq_scale_shape[1] < static_cast<int64_t>(bq_weight_shape[1]));
-      is_bq_weight = rank2 || rankN;
+      is_bq_weight = (bq_scale_shape.size() == bq_weight_shape.size() &&
+                      bq_scale_shape[1] < static_cast<int64_t>(bq_weight_shape[1]));
     }
   }
 
@@ -289,11 +283,10 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
     // Activation handling: BQ kernel (BW_FLOAT_BLOCK) requires FP16, so INT16 → FP16 via Dequantize.
     //                      LPBQ kernel (BLOCKWISE_EXPANSION) accepts INT16 as-is.
     if (!is_lpbq) {
-      const std::string act_name = input_names.back();  // copy before pop_back invalidates ref
+      const std::string act_name = input_names[0];  // activation (input 0), pushed by ProcessInput above
       const auto& act_wrapper = qnn_model_wrapper.GetQnnTensorWrapper(act_name);
       const Qnn_DataType_t act_dtype = act_wrapper.GetTensorDataType();
       if (act_dtype == QNN_DATATYPE_SFIXED_POINT_16 || act_dtype == QNN_DATATYPE_UFIXED_POINT_16) {
-        input_names.pop_back();
         const std::string fp16_act_name = utils::UniqueNameGenerator().New(act_name, "_fp16");
         std::vector<uint32_t> act_shape = act_wrapper.GetTensorDims();
         QnnTensorWrapper fp16_act_wrapper(fp16_act_name, QNN_TENSOR_TYPE_NATIVE,
@@ -306,11 +299,11 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
                           QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_DEQUANTIZE,
                           {act_name}, {fp16_act_name}, {}, do_op_validation),
                       "Failed to add INT16→FP16 Dequantize node for BQ Conv activation.");
-        input_names.push_back(fp16_act_name);
+        input_names[0] = fp16_act_name;
       }
     }
 
-    // Common: transpose weight data from OIHW→HWCN (or CNHW→HWCN for ConvTranspose).
+    // Common: transpose weight data from OIHW→HWIO (or IOHW→HWIO for ConvTranspose).
     // TransposeFromNchwToHwcn unpacks INT4 to INT8 internally (1 byte per element).
     const bool is_3d = (input_info.shape.size() == 5);
     std::vector<uint8_t> unpacked_tensor;
@@ -541,8 +534,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
         RETURN_IF_NOT(quant_param_wrapper.IsPerTensor(),
                       "Conv's INT16 weight inputs only support INT16 per-tensor quantization");
 
-        // Pop Conv weight. Insert Convert op after Weight
-        input_names.pop_back();
+        // Insert Convert op after Weight, replacing the weight name (last element) in place.
         std::string convert_output_name = utils::UniqueNameGenerator().New(weight_input_name, "_convert");
 
         RETURN_IF_ERROR(utils::InsertConvertOp(qnn_model_wrapper,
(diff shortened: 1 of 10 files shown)


onnxruntime/core/providers/qnn/ort_api.h
```diff
@@ -33,6 +33,7 @@
 #include "onnxruntime_session_options_config_keys.h"
 
 #include "core/providers/qnn/common/int4.h"
+#include "core/providers/qnn/common/int2.h"
```
minfhong-qti, line 35: I think `int2` should be ahead of `int4`.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -375,6 +375,34 @@ Ort::Status UnpackInt4ToInt8(size_t num_int4_elems, std::vector<uint8_t>& data_b
...
+    std::vector<uint8_t> packed_int2_bytes = std::move(data_bytes);
+    data_bytes = std::vector<uint8_t>(num_int2_elems);
+
+    auto dst = gsl::make_span(reinterpret_cast<int8_t*>(data_bytes.data()), data_bytes.size());
+    auto src = gsl::make_span(reinterpret_cast<const Int2x4*>(packed_int2_bytes.data()), packed_int2_bytes.size());
+    RETURN_IF_NOT(Int2x4::Unpack(dst, src), "Failed to unpack Tensor<Int2x4> for QNN");
+
+    // Mask off top 6 bits (keep lower 2 bits), consistent with the INT4 masking workaround for QNN.
+    for (size_t i = 0; i < dst.size(); i++) {
+      dst[i] &= 0x03;  // e.g., -1 (0b11111111) becomes 3 (0b00000011)
```
minfhong-qti, line 392: Just curious why mask here but not for unsigned branch? Although HTP backend will automatically ignore top 6 bits by themselves.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -861,12 +865,14 @@ bool OnnxDataTypeToQnnDataType(const ONNXTensorElementDataType onnx_data_type,
   };
 
   const std::unordered_map<ONNXTensorElementDataType, Qnn_DataType_t> onnx_to_qnn_data_type_quantized = {
+      {ONNX_TENSOR_ELEMENT_DATA_TYPE_INT2, QNN_DATATYPE_SFIXED_POINT_8},
```
minfhong-qti, line 868: I'm not sure whether we should map 2-bit to 8-bit here as this mapping in fact is HTP's constraint while QNN already has 2-bit type.

onnxruntime/test/util/include/int2.h
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
minfhong-qti, line 1: Should use QCOM copyrights?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -115,6 +137,50 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
...
+    // Detect block-quantized weight. Two supported scale layouts:
+    //   (a) Standard (ONNX opset 21): scale rank == weight rank, e.g. [OC, IC/bs, H, W] for
+    //       axis=1; scale_shape[1] == IC/block_size < weight_shape[1].
+    //   (b) Legacy rank-2: [OC, num_blocks]; scale_shape[1] == IC/block_size.
+    // In both cases scale_shape[1] == num_blocks_per_oc for 1x1 or IC-axis blocking.
+    if (inputs[1].quant_param.has_value() && inputs[1].quant_param->scale != nullptr) {
+      const auto weight_scale_shape = utils::GetInitializerShape(inputs[1].quant_param->scale,
+                                                                 qnn_model_wrapper.GetOrtApi());
+      std::vector<uint32_t> weight_shape;
+      if (!qnn_model_wrapper.GetOnnxShape(inputs[1].shape, weight_shape) ||
```
minfhong-qti, line 149: Not sure whether worth return not supported if `GetOnnxShape` returns `false` which means non-static shape. Would it be handled somewhere later?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +250,562 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+      RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[2], bias_info));
+      RETURN_IF(!bias_info.is_initializer, "QNN EP: BQ Conv bias must be a constant initializer");
+
+      // Read the INT32 quantized bias and its scale, then dequantize to FP16.
+      // QNN BW_FLOAT_BLOCK Conv2d requires FP16 bias matching the computation precision.
+      std::vector<uint8_t> raw_bias_bytes;
+      RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(bias_info.initializer_tensor,
+                                                              raw_bias_bytes));
+      std::vector<float> bias_scale_vals;
+      RETURN_IF_ERROR(qnn_model_wrapper.UnpackScales(inputs[2].quant_param->scale, bias_scale_vals));
```
kuanyul-qti: The new BQ bias branch dereferences `inputs[2].quant_param->scale` without checking `has_value()`. `OrtNodeUnitIODef::quant_param` is a `std::optional<QuantParam>`; using `->` when the optional is empty is undefined behavior. The non-BQ branch correctly gates on `bias_info.quant_param.IsQuantized()` (a wrapper, always safe). Add the following before dequant:

```suggestion
      RETURN_IF_NOT(inputs[2].quant_param.has_value() &&
                    inputs[2].quant_param->scale != nullptr,
                    "QNN EP: BQ Conv bias must be quantized (INT32 with scale)");
      RETURN_IF_ERROR(qnn_model_wrapper.UnpackScales(inputs[2].quant_param->scale, bias_scale_vals));
```

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +250,562 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+      const float bias_scale = bias_scale_vals.empty() ? 1.0f : bias_scale_vals[0];
```
kuanyul-qti: Do we support per-channel bias?

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -2884,6 +2884,365 @@ TEST_F(QnnHTPBackendTests, ConvU8U8S32_LargeInput_Dilations_Pads) {
...
+                  ExpectedEPNodeAssignment::All,
+                  /*fp32_abs_err=*/1e-2f);
+}
+
+// INT2, block_size=16: ORT rejects tensor(int2) as a DequantizeLinear input type at model
+// load time (ONNX type check), so session initialization throws. Use EXPECT_THROW to
+// actively guard against silent regressions. When ONNX spec and ORT model validation
+// accept tensor(int2), replace with a normal RunQnnModelTest (verify_outputs=false until
+// the CPU DQ kernel also supports int2).
+TEST_F(QnnHTPBackendTests, ConvBQ_U16Int2_1x1_BlockSize16) {
```
kuanyul-qti: The two INT2/UINT2 tests use `EXPECT_THROW`, but the throw originates in ORT's model loader (rejecting `tensor(int2)` as a `DequantizeLinear` input), which fires *before* the QNN EP `GetCapability`/`Compile` path. Removing every line of the new INT2/UINT2 op-builder code (`UnpackInt2ToInt8`, `UnpackZeroPoints` INT2 case, `is_unsigned_weight` UINT2 branch, `GetBQBitwidth` 2-bit case) would not change these tests' outcome. The kill-test fence is invalid. Could keep it but add unit tests that exercise `UnpackInt2ToInt8` and the `INT2`/`UINT2` cases of `UnpackZeroPoints` directly without going through ORT model load; or delete the two `EXPECT_THROW` tests.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -12,6 +12,28 @@
...
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4:
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4:
+      return 4;
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8:
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8:
+      return 8;
+    default:
+      return 0;
+  }
+}
```
kuanyul-qti, line 36: Per-backend constraint tables and helpers should live in an anonymous `namespace { ... }` for internal linkage and ODR safety. Wrap `kHtpConvBQBitsAndBlockSizeMultipliers` and `GetBQBitwidth` in an unnamed namespace.

```suggestion
namespace {

// HTP BQ Conv: supported bitwidths and their block_size divisor constraints.
// block_size must be a multiple of the corresponding value (same as MatMulNBits HTP constraints).
const std::unordered_map<uint32_t, int64_t> kHtpConvBQBitsAndBlockSizeMultipliers{
    {2, 16}, {4, 8}, {8, 4}};

// Returns BQ weight bitwidth (2/4/8) from an ONNX element data type, or 0 if unsupported.
uint32_t GetBQBitwidth(ONNXTensorElementDataType onnx_type) {
  switch (onnx_type) {
    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT2:
    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT2:
      return 2;
    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4:
    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4:
      return 4;
    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8:
    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8:
      return 8;
    default:
      return 0;
  }
}

}  // namespace
```

onnxruntime/test/util/include/int2.h
qti-hungjuiw, line 1: Not sure if this should be Qualcomm copyright

minfhong-qti (commented):
Will continue further review later.

onnxruntime/test/util/include/int2.h
huaychou: No, it is copy from MSFT

huaychou: ditto

onnxruntime/test/providers/qnn/conv_test.cc
huaychou: Actually, 2-bit ConvBQ is not supported by HTP backend until 2.47. I think we may need to disable these UTs and remove the support of 2-bit for now.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
huaychou: I think QNN doesn't take 2-bit type directly now.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
huaychou: Copying from UnpackInt4ToInt8. I think this is due to unpack unsigned int wouldn't result in signed bit.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: What is ODR safety? Could you elaborate that?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: I think it is handled in ProcessAttributesAndOutputs

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Bias is in FP16. We just need to dequantize it.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: BQ bias no need to be quantized.

onnxruntime/core/providers/qnn/ort_api.h
huaychou: Will fix

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -115,6 +137,50 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
   // Validate quantization axis for per-channel quantized weights.
   bool is_npu_backend = IsNpuBackend(qnn_model_wrapper.GetQnnBackendType());
   if (is_npu_backend) {
+    // Detect block-quantized weight. Two supported scale layouts:
+    //   (a) Standard (ONNX opset 21): scale rank == weight rank, e.g. [OC, IC/bs, H, W] for
+    //       axis=1; scale_shape[1] == IC/block_size < weight_shape[1].
+    //   (b) Legacy rank-2: [OC, num_blocks]; scale_shape[1] == IC/block_size.
```
minfhong-qti: Pre-opset 21 QuantizeLinear does not support block quant. Where does it come from?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -115,6 +137,50 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
...
+      const auto weight_scale_shape = utils::GetInitializerShape(inputs[1].quant_param->scale,
+                                                                 qnn_model_wrapper.GetOrtApi());
+      std::vector<uint32_t> weight_shape;
+      if (!qnn_model_wrapper.GetOnnxShape(inputs[1].shape, weight_shape) ||
+          weight_shape.size() != 4) {
+        // BQ only supported for Conv2D (rank-4 weight); skip for Conv1D (rank-3) and Conv3D (rank-5).
+      } else {
+        const bool is_rank2_scale = (weight_scale_shape.size() == 2);
+        const bool is_rankN_scale = (weight_scale_shape.size() == weight_shape.size() &&
+                                     weight_scale_shape.size() > 1 &&
```
minfhong-qti, line 154: Since `weight_shape` is guaranteed to be rank 4 and `weight_scale_shape` has identical shape with `weight_shape`, I think this check is redundant.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -115,6 +137,50 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
...
+    // In both cases scale_shape[1] == num_blocks_per_oc for 1x1 or IC-axis blocking.
+    if (inputs[1].quant_param.has_value() && inputs[1].quant_param->scale != nullptr) {
+      const auto weight_scale_shape = utils::GetInitializerShape(inputs[1].quant_param->scale,
+                                                                 qnn_model_wrapper.GetOrtApi());
+      std::vector<uint32_t> weight_shape;
+      if (!qnn_model_wrapper.GetOnnxShape(inputs[1].shape, weight_shape) ||
+          weight_shape.size() != 4) {
+        // BQ only supported for Conv2D (rank-4 weight); skip for Conv1D (rank-3) and Conv3D (rank-5).
+      } else {
+        const bool is_rank2_scale = (weight_scale_shape.size() == 2);
```
minfhong-qti: I guess it is checking whether is BQ, but is the check here really sufficient?
BTW, I would prefer explicitly named as `is_rank2_bq_scale`.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -115,6 +137,50 @@ Ort::Status ConvOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
...
+                     " must be a multiple of " + std::to_string(bq_it->second) +
+                     " for " + std::to_string(bitwidth) + "-bit weight")
+                        .c_str());
+
+          // Return success; full QNN validation is done in the NHWC IsOpSupported path.
+          return Ort::Status();
+        }  // end is_rank2_scale || is_rankN_scale
+      }  // end else (weight shape obtainable)
+    }  // end quant_param check
+
```
minfhong-qti: Nit. Add a comment describing below is checking for per-channel quant.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +250,562 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+                                                qnn_model_wrapper.GetOrtApi());
+    std::vector<uint32_t> bq_weight_shape;
+    if (qnn_model_wrapper.GetOnnxShape(inputs[1].shape, bq_weight_shape) &&
+        bq_scale_shape.size() > 1 &&
+        bq_weight_shape.size() == 4) {  // BQ only for Conv2D (rank-4 weight); not Conv3D (rank-5)
+      // Weight is NCHW [OC,IC,H,W]: IC is always at index 1.
+      const bool rank2 = (bq_scale_shape.size() == 2);
+      const bool rankN = (bq_scale_shape.size() == bq_weight_shape.size() &&
+                          bq_weight_shape.size() > 1 &&
+                          bq_scale_shape[1] < static_cast<int64_t>(bq_weight_shape[1]));
```
minfhong-qti: ditto

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +250,562 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+      is_bq_weight = rank2 || rankN;
+    }
+  }
+
+  if (is_bq_weight) {
+    // Determine BQ vs LPBQ by reading weight quant params via GetTensorInfo.
+    // When PR307 is active, Init() auto-converts BQ scales → LPBQ (BLOCKWISE_EXPANSION).
+    // Before PR307: Init() cannot handle block_size, so IsLPBQ() is false → BQ path.
```
minfhong-qti, line 274: Is here considering another unmerged PR? It seems a bit weird. Probably we should sync beforehand and reach some consensus.
BTW, I guess all LPBQ branches are not tested, and thus the discussion may include who is responsible of adding testcases for Conv's LPBQ.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +250,562 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+                                          QNN_DATATYPE_FLOAT_16, QnnQuantParamsWrapper(),
+                                          std::vector<uint32_t>(act_shape));
+        RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(fp16_act_wrapper)),
+                      "Failed to add FP16 activation tensor for BQ Conv.");
+        RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(
+                          utils::UniqueNameGenerator().New(act_name, "_int16_dequantize"),
+                          QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_DEQUANTIZE,
+                          {act_name}, {fp16_act_name}, {}, do_op_validation),
+                      "Failed to add INT16→FP16 Dequantize node for BQ Conv activation.");
+        input_names.push_back(fp16_act_name);
```
minfhong-qti: I personally would directly modify `input_names[0]` instead of pop and push.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +250,562 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+    }
 
-    std::string actual_name = input_info.is_initializer ? input1_name : utils::UniqueNameGenerator().New(input1_name, "_transpose");
-    input_names.push_back(actual_name);
-
-    std::vector<uint32_t> actual_shape;
-    actual_shape.resize(input_info.shape.size());
-
-    // Change shape to HWCN, it could be initializer or normal input
+    // Common: transpose weight data from OIHW→HWCN (or CNHW→HWCN for ConvTranspose).
```
minfhong-qti: Should the symbol unify to `OI` or `CN`? Is a bit confusing now.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
kuanyul-qti: **ODR** = **One Definition Rule** ([cppreference](https://en.cppreference.com/w/cpp/language/definition)). It says any class type, inline function, or template visible across multiple translation units (TUs) must be **defined identically in every TU** that sees it. Violations are *ill-formed, no diagnostic required* — the linker is allowed to silently pick one definition and discard the rest, leading to:

- Memory corruption when the "winning" definition has a different layout than the caller's TU expected
- Heisenbugs that only reproduce in Release / `-O2` / one specific build flavor
- Debug vs Release behavior divergence that is hard to bisect

**Why it matters here:**

`kHtpConvBQBitsAndBlockSizeMultipliers` and `GetBQBitwidth` are currently at file scope without an anonymous namespace wrapper. Concretely:

1. If another `.cc` ever defines a symbol with the same name (e.g. a different op builder also wants `GetBQBitwidth`), the linker has to resolve it. With external linkage you get either a multiple-definition error or — worse — silent dedup to one definition.
2. If the table's contents are ever conditionally compiled and the gating macro isn't truly project-global, two TUs end up with structurally different "same" tables → ODR violation → undefined behavior.
3. Wrapping in `namespace { ... }` gives the symbol internal linkage, makes it invisible to other TUs entirely, and removes any possibility for a downstream TU to even *see* a divergent definition. This is the convention used elsewhere in the codebase — e.g. `qnn_node_group.cc`'s `kFusionAttempts`, and the LPBQ constraint table from PR #288 (`[C63]`).

It's not just style — once the table and helper live only in one `.cc` inside an unnamed namespace, ODR violations on them become impossible by construction.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```suggestion
                                                   const bool unpack_to_8_bit) const {
```
I think this should be renamed.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
minfhong-qti: Should we check `inputs[2].quant_param.has_value()` ahead instead of here?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: According to https://onnx.ai/onnx/operators/onnx__DequantizeLinear.html, there is no `block_size` before opset 21.

minfhong-qti: I mean is there rank-2 block quant?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Currently, `IsLPBQ()` always return false. We may need PR307 to enable LPBQ path.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

huaychou: Will add

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

onnxruntime/test/providers/qnn/conv_test.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Removed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Removed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Removed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +249,564 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+    const bool is_lpbq = input_info.quant_param.IsLPBQ();
+
+    // Activation handling: BQ kernel (BW_FLOAT_BLOCK) requires FP16, so INT16 → FP16 via Dequantize.
+    //                      LPBQ kernel (BLOCKWISE_EXPANSION) accepts INT16 as-is.
+    if (!is_lpbq) {
+      const std::string act_name = input_names[0];  // activation (input 0), pushed by ProcessInput above
+      const auto& act_wrapper = qnn_model_wrapper.GetQnnTensorWrapper(act_name);
+      const Qnn_DataType_t act_dtype = act_wrapper.GetTensorDataType();
+      if (act_dtype == QNN_DATATYPE_SFIXED_POINT_16 || act_dtype == QNN_DATATYPE_UFIXED_POINT_16) {
+        const std::string fp16_act_name = utils::UniqueNameGenerator().New(act_name, "_fp16");
```
tirupath-qti: Since the DQ nodes are part of Conv node unit, we can extract the tensor names from DQ node and use them instead of creating new tensor names so that the tensor name mapping from ONNX graph will be retained.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -184,353 +249,564 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+        std::vector<uint32_t> act_shape = act_wrapper.GetTensorDims();
+        QnnTensorWrapper fp16_act_wrapper(fp16_act_name, QNN_TENSOR_TYPE_NATIVE,
+                                          QNN_DATATYPE_FLOAT_16, QnnQuantParamsWrapper(),
+                                          std::vector<uint32_t>(act_shape));
+        RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(fp16_act_wrapper)),
+                      "Failed to add FP16 activation tensor for BQ Conv.");
+        RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(
+                          utils::UniqueNameGenerator().New(act_name, "_int16_dequantize"),
```
tirupath-qti, line 302: same as comment on Line 290. extract the original tensor names from DQ node.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -375,6 +375,34 @@ Ort::Status UnpackInt4ToInt8(size_t num_int4_elems, std::vector<uint8_t>& data_b
   return Ort::Status();
 }
 
+// Re-writes a buffer of packed 2-bit elements to a buffer of unpacked 8-bit elements.
+// QNN requires that 2-bit weights are unpacked to 8-bit (stored in lower 2 bits).
```
tirupath-qti, line 379: Unpacking 2-bit weights into 8-bit values increases memory bandwidth requirements, thereby diminishing the benefits of using 2-bit weights.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -760,6 +770,7 @@ Ort::Status QnnModelWrapper::UnpackZeroPoints(const OrtValueInfo* zp_tensor,
       transform_zero_points(ReinterpretAsSpan<const int8_t>(gsl::make_span(initializer_bytes)));
       break;
     case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4:  // UINT4 zero-points are unpacked as 8-bit values for QNN
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT2:  // UINT2 zero-points are unpacked as 8-bit values for QNN
```
tirupath-qti, line 773: unpacking as 8-bit values diminish the benefits of 2-bit weights. the unpacking should be done at runtime instead of constant time.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Done

onnxruntime/core/providers/qnn/builder/qnn_utils.h
huaychou: The 8-bit weights are packed back into 2-bit when preparing in backend, so it won't increase the memory bandwidth.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Done

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
huaychou: ditto

tirupath-qti (approved):
LGTM

minfhong-qti (approved):

PR #430, opened 2026-05-22
Avoid private ORT Core header in Test Framework - Phase 3: Handle device/model compatibility testcases

qti-hungjuiw:
### Description

- `test/providers/qnn/*`: Updates QNN-specific tests
    - (`gru_test.cc`, `lstm_test.cc`, `qnn_device_compatibility_test.cc`, `qnn_ep_context_test.cc`) to use the public API surface only.
- `cmake/onnxruntime_unittests.cmake`: Drops the remaining private header search path
- `core/providers/qnn/qnn_provider_factory.cc`: Enhances `ValidateCompiledModelCompatibilityInfo` to construct a temporary QnnEp when the factory has not yet created a persistent EP, rather than returning an early error.
- Extends `qcom/linters/check_private_ort_headers.py` and `.lintrunner.toml` to cover `onnxruntime_unittests.cmake` and `onnxruntime_prebuilt.cmake`.
    - Adds per-target approved include sets for `onnxruntime_providers_qnn` and `onnxruntime_provider_test`.

### Motivation and Context

- The PR aims to reduce the scope and overhead of https://github.com/onnxruntime/onnxruntime-qnn/pull/222

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -354,7 +354,7 @@ OrtStatus* ORT_API_CALL QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl
   // EP has not been created yet. Create a temporary QNN EP for validation.
   const auto provider_prefix = GetProviderOptionPrefix(factory->ep_name_);
 
-  // Determine backend type from the provided devices (prefer NPU, then GPU).
+  // Determine backend type from the provided devices (Only supports NPU currently).
   std::string backend_type;
   for (size_t i = 0; i < num_devices; ++i) {
     auto device_type = factory->ort_api.HardwareDevice_Type(devices[i]);
@@ -363,9 +363,12 @@ OrtStatus* ORT_API_CALL QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl
       break;
     }
   }
-  RETURN_IF(backend_type.empty(),
-            "Currently QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl only supports OrtHardwareDeviceType_NPU, but "
-            "no OrtHardwareDeviceType_NPU is found in the `devices` argument.");
+  if (backend_type.empty()) {
+    *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
+    return factory->ort_api.CreateStatus(ORT_EP_FAIL,
+                                         "Currently QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl only supports OrtHardwareDeviceType_NPU, but "
+                                         "no OrtHardwareDeviceType_NPU is found in the `devices` argument.");
+  }
 
   using SessionOptionsUniquePtr = std::unique_ptr<OrtSessionOptions, std::function<void(OrtSessionOptions*)>>;
   OrtSessionOptions* temp_session_options = nullptr;
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -364,17 +364,23 @@ OrtStatus* ORT_API_CALL QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl
     }
   }
   RETURN_IF(backend_type.empty(),
-    "Currently QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl only supports OrtHardwareDeviceType_NPU, but "
-    "no OrtHardwareDeviceType_NPU is found in the `devices` argument.");
+            "Currently QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl only supports OrtHardwareDeviceType_NPU, but "
+            "no OrtHardwareDeviceType_NPU is found in the `devices` argument.");
 
   using SessionOptionsUniquePtr = std::unique_ptr<OrtSessionOptions, std::function<void(OrtSessionOptions*)>>;
   OrtSessionOptions* temp_session_options = nullptr;
-  RETURN_IF_NOT_NULL(factory->ort_api.CreateSessionOptions(&temp_session_options));
+  if (OrtStatus* _status = factory->ort_api.CreateSessionOptions(&temp_session_options)) {
+    *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
+    return _status;
+  }
   SessionOptionsUniquePtr session_options(temp_session_options, factory->ort_api.ReleaseSessionOptions);
 
-  RETURN_IF_NOT_NULL(factory->ort_api.AddSessionConfigEntry(session_options.get(),
-                                                            (provider_prefix + "backend_type").c_str(),
-                                                            backend_type.c_str()));
+  if (OrtStatus* _status = factory->ort_api.AddSessionConfigEntry(session_options.get(),
+                                                                  (provider_prefix + "backend_type").c_str(),
+                                                                  backend_type.c_str())) {
+    *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
+    return _status;
+  }
 
   if (!OrtLoggingManager::HasDefaultLogger()) {
     *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -384,7 +384,7 @@ OrtStatus* ORT_API_CALL QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl
 
   if (!OrtLoggingManager::HasDefaultLogger()) {
     *model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE;
-    return factory->ort_api.CreateStatus(ORT_FAIL, "Default logger is not available for model compatibility check.");
+    return factory->ort_api.CreateStatus(ORT_EP_FAIL, "Default logger is not available for model compatibility check.");
   }
   const OrtLogger* logger = OrtLoggingManager::GetDefaultLoggerPtr();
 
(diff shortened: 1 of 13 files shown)


qti-hungjuiw:
Hi @huaychou, @minfhong-qti would you mind reviewing device compatibility and model compatibility part?

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -344,17 +344,65 @@ OrtStatus* ORT_API_CALL QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl
...
-    // Currently we require EP must first be created as QNN backend is mandatory for validating the compatibility.
-    // Possibly consider creating a fake EP for validation only if necessary.
+  if (factory->qnn_ep_ != nullptr) {
+    return factory->qnn_ep_->ValidateCompiledModelCompatibilityInfo(devices,
+                                                                    num_devices,
+                                                                    compatibility_info,
+                                                                    model_compatibility);
+  }
+
+  // EP has not been created yet. Create a temporary QNN EP for validation.
```
huaychou, line 354: I don't think we need to create a temporary QNN EP here.

onnxruntime/test/providers/qnn/qnn_device_compatibility_test.cc
```diff
@@ -35,41 +34,57 @@ class QnnDeviceCompatibilityTests : public ::testing::Test {
...
+    return ep_api_->CreateHardwareDevice(
+        type,
+        vendor_id,
+        0 /* device_id */,
+        "QCOM" /* vendor_name */,
+        nullptr /* metadata */,
+        &mock_hw_device);
   }
 
+  const uint32_t qualcomm_vendor_id{'Q' | ('C' << 8) | ('O' << 16) | ('M' << 24)};
```
huaychou, line 62: Just wondering, why do we need to define it here?

onnxruntime/test/providers/qnn/qnn_device_compatibility_test.cc
```diff
@@ -86,26 +101,21 @@ TEST_F(QnnDeviceCompatibilityTests, CPUDeviceIsCompatible) {
...
-  RegisterQnnEpLibrary(registered_ep_device, so, onnxruntime::kQnnExecutionProvider, options);
+  RegisterQnnEpLibrary(registered_ep_device, so, kQnnExecutionProvider, options);
 
   ASSERT_NE(registered_ep_device, nullptr);
-  ASSERT_NE(registered_ep_device->ep_factory, nullptr);
-
-  // Get the Qualcomm vendor ID from the factory
-  uint32_t qualcomm_vendor_id = registered_ep_device->ep_factory->GetVendorId(registered_ep_device->ep_factory);
 
-  // Create a mock NPU device with Qualcomm vendor ID
```
huaychou, line 102: Could we preserve the comment?

onnxruntime/test/providers/qnn/qnn_device_compatibility_test.cc
```diff
@@ -122,26 +132,21 @@ TEST_F(QnnDeviceCompatibilityTests, NPUDeviceWithQualcommVendorIsCompatible) {
...
-  RegisterQnnEpLibrary(registered_ep_device, so, onnxruntime::kQnnExecutionProvider, options);
+  RegisterQnnEpLibrary(registered_ep_device, so, kQnnExecutionProvider, options);
 
   ASSERT_NE(registered_ep_device, nullptr);
-  ASSERT_NE(registered_ep_device->ep_factory, nullptr);
 
-  // Get the Qualcomm vendor ID from the factory
-  uint32_t qualcomm_vendor_id = registered_ep_device->ep_factory->GetVendorId(registered_ep_device->ep_factory);
-
-  // Create a mock GPU device with Qualcomm vendor ID
```
huaychou, line 138: ditto

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc → ValidateCompiledModelCompatibilityInfoImpl`
The new device-loop only updates `backend_type` for NPU/GPU and explicitly skips CPU; on a CPU-only or empty `devices` array `backend_type` stays empty and silently falls back to `"htp"`, which then constructs an HTP-backed temporary EP to validate a model that may have been compiled for CPU. Either include CPU in the preference order (NPU > GPU > CPU) or return `ORT_INVALID_ARGUMENT` when `num_devices == 0` / no match.

**[M-2]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:L378–L381`
Both early `RETURN_IF_NOT_NULL(... CreateSessionOptions ...)` and `RETURN_IF_NOT_NULL(... AddSessionConfigEntry ...)` propagate a non-null status without writing `*model_compatibility`. Every other failure path in this function sets `*model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE` first; align these two sites with that pattern (and apply the same to `GetHardwareDeviceIncompatibilityDetailsImpl`).

---

### 🔵 Minor

**[N-1]** `onnxruntime/test/providers/qnn/qnn_ep_context_test.cc:L1082, L1676`
`buffer.resize(narrow<size_t>(file.tellg()))` was changed to `static_cast<size_t>(file.tellg())`. `tellg()` is signed and returns `-1` on failure; the prior `narrow<>` would throw, the new `static_cast` silently produces `SIZE_MAX` and OOMs. Switch to `std::filesystem::file_size(path)` (unsigned, naturally fits `size_t`) or restore the checked narrow.

**[N-2]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:L387, L396, L399`
The three new failure paths use `ORT_FAIL`; the original code used `ORT_EP_FAIL`. `ORT_EP_FAIL` is more specific and lets callers branch on EP-specific failures vs generic ORT failures — please switch back.

onnxruntime/test/providers/qnn/qnn_device_compatibility_test.cc
qti-hungjuiw: Fixed, thanks!

onnxruntime/test/providers/qnn/qnn_device_compatibility_test.cc
qti-hungjuiw: Fixed, thanks!

qti-chuteng (approved):

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-hungjuiw: Per offline discussion with @huaychou and @minfhong-qti, we create a temporary QNN EP here since we do not expect the users of the `QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl` API do CreateEP themselves. Also, we error out if there is no `OrtHardwareDeviceType_NPU` passed in.

onnxruntime/test/providers/qnn/qnn_device_compatibility_test.cc
qti-hungjuiw: Previously we used `registered_ep_device->ep_factory->GetVendorId` to acquire `qualcomm_vendor_id`. This requires the private header `core/session/abi_devices.h` and need to be modified.
Given that the `qualcomm_vendor_id` is used to create mock `OrtHardwareDevice` for testing, I think it is ok to do the hardcode. Please let me know if this cause issues.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -344,17 +344,65 @@ OrtStatus* ORT_API_CALL QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl
...
+  const auto provider_prefix = GetProviderOptionPrefix(factory->ep_name_);
+
+  // Determine backend type from the provided devices (prefer NPU, then GPU).
```
minfhong-qti: Update the comment.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -344,17 +344,65 @@ OrtStatus* ORT_API_CALL QnnEpFactory::ValidateCompiledModelCompatibilityInfoImpl
...
+  std::string backend_type;
+  for (size_t i = 0; i < num_devices; ++i) {
+    auto device_type = factory->ort_api.HardwareDevice_Type(devices[i]);
+    if (device_type == OrtHardwareDeviceType_NPU) {
+      backend_type = "htp";
+      break;
+    }
+  }
+  RETURN_IF(backend_type.empty(),
```
minfhong-qti: I don't think `RETURN_IF` is appropriate here as the function requires `OrtStatue*` instead of `Ort::Status`.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
minfhong-qti: I think this case can be removed as the original intention is not applicable now.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-hungjuiw: Thanks for pointing this out. Fixed!

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
qti-hungjuiw: Thanks for pointing this out. Removed!

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-hungjuiw: Thanks for pointing this out. Fixed!

qti-hungjuiw:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc → ValidateCompiledModelCompatibilityInfoImpl` The new device-loop only updates `backend_type` for NPU/GPU and explicitly skips CPU; on a CPU-only or empty `devices` array `backend_type` stays empty and silently falls back to `"htp"`, which then constructs an HTP-backed temporary EP to validate a model that may have been compiled for CPU. Either include CPU in the preference order (NPU > GPU > CPU) or return `ORT_INVALID_ARGUMENT` when `num_devices == 0` / no match.
> 
> **[M-2]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:L378–L381` Both early `RETURN_IF_NOT_NULL(... CreateSessionOptions ...)` and `RETURN_IF_NOT_NULL(... AddSessionConfigEntry ...)` propagate a non-null status without writing `*model_compatibility`. Every other failure path in this function sets `*model_compatibility = OrtCompiledModelCompatibility_EP_NOT_APPLICABLE` first; align these two sites with that pattern (and apply the same to `GetHardwareDeviceIncompatibilityDetailsImpl`).
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/test/providers/qnn/qnn_ep_context_test.cc:L1082, L1676` `buffer.resize(narrow<size_t>(file.tellg()))` was changed to `static_cast<size_t>(file.tellg())`. `tellg()` is signed and returns `-1` on failure; the prior `narrow<>` would throw, the new `static_cast` silently produces `SIZE_MAX` and OOMs. Switch to `std::filesystem::file_size(path)` (unsigned, naturally fits `size_t`) or restore the checked narrow.
> 
> **[N-2]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:L387, L396, L399` The three new failure paths use `ORT_FAIL`; the original code used `ORT_EP_FAIL`. `ORT_EP_FAIL` is more specific and lets callers branch on EP-specific failures vs generic ORT failures — please switch back.

All the comments are fixed in the latest commit.

minfhong-qti (approved):

PR #432, opened 2026-05-22
[QNN EP] Enable static MSVC runtime for x86_64

qti-mbadnara:
This PR adds support to enable static MSVC runtime for Windows x86_64 builds


qti-ashwshan (approved):
LGTM

PR #433, opened 2026-05-23
[QNN EP] Pass --nightly_build as explicit CLI argument to build.py

qti-mbadnara:
`tools/ci_build/build.py` previously read `NIGHTLY_BUILD` via env var while all other build flags are passed as explicit CLI arguments, requiring separate translation layers for Linux and Windows. This PR adds `--nightly_build` as a CLI argument and has both shell scripts pass it explicitly when `ORT_NIGHTLY_BUILD=1`. The redundant NIGHTLY_BUILD translation layers are removed.

Building on top review comment from #308


huaychou (approved):
LGTM :shipit:

qti-chuteng (approved):

kuanyul-qti (approved):

PR #444, opened 2026-05-27
[QNN EP] Add support to build Linux x86_64 wheels on Ubuntu 22.04 Docker

qti-mbadnara:
### Motivation

The existing `Linux x86_64` build runs natively on self-hosted runners, producing `linux_x86_64` wheels tied to whatever distro happens to be on the runner. This PR adds a parallel Docker-based pipeline that builds inside an `Ubuntu 22.04` container, producing a **PyPI-shippable `manylinux_2_35_x86_64` wheel** across a multi-Python matrix (3.11–3.14), to
match what we already ship for `aarch64_manylinux_2_34`. The arch identifier `x86_64_ubuntu_22_04` is plumbed end-to-end so all outputs land in their own build directory and never collide with the native `x86_64` path (which we intend to deprecate in a follow-up).

---

#### New Docker image (`qcom/scripts/linux/ubuntu_22_04/Dockerfile`)
`Ubuntu 22.04` from the internal mirror `docker-registry.qualcomm.com/library/ubuntu:22.04`
(Docker Hub is unreachable from the runners). Includes:
- Build tools: `build-essential`, `cmake`, `ninja-build`, `git`, `wget`, `curl`, `pkg-config`, `ca-certificates`.
- `libc++1` + `libc++abi1` — QAIRT backend `.so`s are clang/libc++-linked and `dlopen` `libc++.so.1` at runtime.
- `patchelf` — required by `auditwheel`.
- `Python 3.10–3.14` (`-dev` + `-venv`) via deadsnakes; 3.10 is needed to bootstrap the in-container build venv.
- `useradd -m` for a writable `$HOME` (pip/uv caches).
- `ENV AUDITWHEEL_PLAT=manylinux_2_35_x86_64` — the signal `setup.py` reads to enable manylinux packaging + `auditwheel repair`.

#### Shippable manylinux_2_35 wheel
- `setup.py`: added `manylinux_2_35_x86_64` to `manylinux_tags`, so the build bundles QAIRT libs and runs `auditwheel repair` → `…-manylinux_2_35_x86_64.whl` (pip-installable on glibc ≥ 2.35).
- `tools/ci_build/github/linux/python/requirements.txt`: added `auditwheel` (manylinux images ship it; Ubuntu doesn't).

#### Docker task infrastructure (`qcom/ep_build/tasks/docker.py`)
Added `UBUNTU_22_04_X86_64_TAG`. Reused existing `platform` / `inner_task` / `docker_tag`
params (default `linux/aarch64`, overridden to `linux/amd64`).

#### New build/test tasks (`qcom/build_and_test.py`, `qcom/ep_build/typing.py`)
`x86_64_ubuntu_22_04` added to `TargetArchLinuxT` and plumbed through so the build dir is
`build/linux-x86_64_ubuntu_22_04/` and the archive is `onnxruntime-tests-linux-x86_64_ubuntu_22_04.tar.bz2`:
- `docker_build_ubuntu_22_04_x86_64`, `build_ort_linux_x86_64_ubuntu_22_04`, `_build_ort_linux_x86_64_ubuntu_22_04`.
- `archive_ort_linux_x86_64_ubuntu_22_04` (depends on `create_venv`), `extract_ort_linux_x86_64_ubuntu_22_04`.
- `test_ort_linux_x86_64_ubuntu_22_04`: runs C++ unit + ONNX model tests natively on the host;
  passes `target_py_version=None` so it uses the activated host venv (no in-container `venv-3.12` required).

#### Supporting changes
- `qcom/ep_build/tasks/python.py`: added `x86_64_ubuntu_22_04 → x86_64` mapping and an arch dispatch in `__find_wheel` to locate the `manylinux_*_x86_64.whl`.
- `qcom/scripts/all/archive_tests.py`: added `linux-x86_64_ubuntu_22_04` to the `--target-platform` choices.

#### CI workflows
- **`qualcomm-internal-build-and-test.yml`**: added `build-and-test-linux-x86_64-ubuntu-22-04` (build + native test, wheel/test-archive/TGZ upload).
- **`qualcomm-internal-build-artifacts.yml`**: added the `linux_target_py_vsns` input (default `['3.11','3.12','3.13','3.14']`, shared by aarch64 manylinux_2_34 and x86_64 Ubuntu 22.04) and the `build-linux-x86_64-ubuntu-22-04` matrix job.
- **`qualcomm-internal-build-and-test-single-os.yml`**: `target_arch` description now lists `x86_64_ubuntu_22_04`.
- **`qualcomm-internal-upload-artifactory-from-github.yml`**: download/consolidate/publish for the x86_64 Ubuntu 22.04 wheels (3.11–3.14) and TGZ; wheel count check 12 → **16**, TGZ 1 → **2**.
- **`qualcomm-internal-release.yml`**: download steps for x86_64 Ubuntu 22.04 wheels (3.11–3.14), test archive, and TGZ; wheel count 14 → **18**, `.tar.bz2` 4 → **5**, TGZ 1 → **2**.


qcom/ep_build/tasks/build.py
```diff
@@ -359,4 +363,6 @@ def ort_build_env_vars(old_env: Mapping[str, str] | None = None) -> dict[str, st
     env = os.environ.copy() if old_env is None else dict(old_env)
     if env.get("ORT_NIGHTLY_BUILD", "0") == "1":
         env["Build_SourceVersion"] = git_head_sha()
+    elif env.get("ORT_NIGHTLY_BUILD", "0") == "0":
+        env["NIGHTLY_BUILD"] = "0"
```
kuanyul-qti: Why adding this back?

kuanyul-qti (commented):
1. Should we align the linux x86_64 wheel version with linux aarch64, for example, using manylinux_2_34_x86_64? Or is linux aarch64 expected to update to manylinux_2_35 as well?
2. The version gap with ORT, which uses manylinux_2_27, is increasing. Is this expected? Could you provide more context on this?

qcom/ep_build/tasks/build.py
qti-mbadnara: Will fix this, thanks for catching. This got added in the rebase!

qti-mbadnara:
> 1. Should we align the linux x86_64 wheel version with linux aarch64, for example, using manylinux_2_34_x86_64? Or is linux aarch64 expected to update to manylinux_2_35 as well?
> 2. The version gap with ORT, which uses manylinux_2_27, is increasing. Is this expected? Could you provide more context on this?

The Linux x86_64 wheels must use manylinux_2_35 because the bundled QAIRT libraries require GLIBC ≥ 2.35, which aligns with Ubuntu 22.04/24.04. Due to this dependency, aligning with older baselines (like manylinux_2_27 used by ORT) or even manylinux_2_34 is not feasible. The GLIBC of QAIRT work with manylinux_2_34 on ARM64 Linux, therefore no change is needed there.

As a result, the divergence from ORT’s manylinux version is expected and driven by these runtime requirements rather than a packaging choice.

kuanyul-qti:
### 🟠 Major


**[M-1]** `README.md:L89-L93`
This PR adds 4 Linux x86_64 manylinux wheels and 1 Linux x86_64 .tgz to the release flow (wheel count 14→18, tgz count 1→2 in both `qualcomm-internal-release.yml` and `qualcomm-internal-upload-artifactory-from-github.yml`), but the per-version availability table in `README.md` still only mentions ARM64. Any release-flow PR that adds a per-platform artifact must update the README availability table in the same PR — otherwise downstream users have no signal for which release first ships the new x86_64 manylinux wheel. The new manylinux wheel is also a *different wheel spec* from the existing non-manylinux py3.10/3.12 Linux x86_64 wheel, so the distinction deserves an explicit per-version note.

```suggestion
### Linux Wheels and .tgz Files

- **X.Y.Z+** (x86_64): Linux x86_64 manylinux_2_35 Wheels (py3.11–3.14) and .tgz files available
- **2.1.1+**: Linux ARM64 Wheels and .tgz files available
- **2.1.0**: Linux ARM64 preview wheels available
- **2.0.0**: No Linux ARM64 Wheels or .tgz files
```


**[M-2]** `.github/workflows/qualcomm-internal-build-and-test.yml:L100-L101` (NEW)
The new `build-and-test-linux-x86_64-ubuntu-22-04` job sets `build_archive: true` + `upload_archive: true`. But `qualcomm-internal-build-and-test.yml` is the PR/push CI workflow (called by `qualcomm-internal-ci.yml` on `pull_request`/`push` with `do_all: true`), not the release flow — and it's the *only* job in that workflow setting these flags (both default `false` in `*-single-os.yml`; every sibling job leaves them off). `upload_archive: true` triggers the "Upload Archive to Artifactory" step, so every internal-branch PR/push now builds and uploads a release `.tgz` to Artifactory. Per [P03], artifact-upload flags should be true only in the release workflow. (The release path is already covered by `build-linux-x86_64-ubuntu-22-04` in `qualcomm-internal-build-artifacts.yml`, which gates these flags on the first matrix entry.)

(i.e. drop the `build_archive: true` / `upload_archive: true` lines. If PR-time validation that the tgz packs is desired, keep `build_archive: true` but still remove `upload_archive: true`.)

---

### 🔵 Minor


**[N-1]** `.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml:L23`
The input description still references a non-existent target: `"JSON list of Python versions to build for Linux targets (aarch64 manylinux_2_34 and x86_64 manylinux_2_34)"`. The matching description in `qualcomm-internal-build-artifacts.yml:42` was correctly updated to `"x86_64 Ubuntu 22.04"`. This is the text shown on the GitHub Actions "Run workflow" UI; please bring the two in sync.

```suggestion
        description: "JSON list of Python versions to build for Linux targets (aarch64 manylinux_2_34 and x86_64 Ubuntu 22.04)"
```


**[N-2]** `qcom/scripts/linux/ubuntu_22_04/Dockerfile`
The new Dockerfile installs 5 Python versions (py3.10–3.14) from deadsnakes PPA, each with `-dev` and `-venv`, adding ~300 MB of fixed image cost regardless of which py version a given matrix entry actually uses. This works, but compared to the symmetric `manylinux_2_34/Dockerfile` (which uses `quay.io/pypa/manylinux_2_34_aarch64` as base and inherits multi-version CPython for free) it has poorer layer-cache hit rate and requires editing this Dockerfile every time a new Python version is added (e.g. py3.15). Consider parameterizing the version list as a `PYTHON_VERSIONS` build-arg fed from `linux_target_py_vsns`. Not blocking.

qti-mbadnara:
> **[M-1]** `README.md:L89-L93` This PR adds 4 Linux x86_64 manylinux wheels and 1 Linux x86_64 .tgz to the release flow (wheel count 14→18, tgz count 1→2 in both `qualcomm-internal-release.yml` and `qualcomm-internal-upload-artifactory-from-github.yml`), but the per-version availability table in `README.md` still only mentions ARM64. Any release-flow PR that adds a per-platform artifact must update the README availability table in the same PR — otherwise downstream users have no signal for which release first ships the new x86_64 manylinux wheel. The new manylinux wheel is also a _different wheel spec_ from the existing non-manylinux py3.10/3.12 Linux x86_64 wheel, so the distinction deserves an explicit per-version note.
> 
> ```
> ### Linux Wheels and .tgz Files
> 
> - **X.Y.Z+** (x86_64): Linux x86_64 manylinux_2_35 Wheels (py3.11–3.14) and .tgz files available
> - **2.1.1+**: Linux ARM64 Wheels and .tgz files available
> - **2.1.0**: Linux ARM64 preview wheels available
> - **2.0.0**: No Linux ARM64 Wheels or .tgz files
> ```

M-1 will be fixed when the release notes for 2.3.0 are added

qti-mbadnara:
> **[M-2]** `.github/workflows/qualcomm-internal-build-and-test.yml:L100-L101` (NEW) The new `build-and-test-linux-x86_64-ubuntu-22-04` job sets `build_archive: true` + `upload_archive: true`. But `qualcomm-internal-build-and-test.yml` is the PR/push CI workflow (called by `qualcomm-internal-ci.yml` on `pull_request`/`push` with `do_all: true`), not the release flow — and it's the _only_ job in that workflow setting these flags (both default `false` in `*-single-os.yml`; every sibling job leaves them off). `upload_archive: true` triggers the "Upload Archive to Artifactory" step, so every internal-branch PR/push now builds and uploads a release `.tgz` to Artifactory. Per [P03], artifact-upload flags should be true only in the release workflow. (The release path is already covered by `build-linux-x86_64-ubuntu-22-04` in `qualcomm-internal-build-artifacts.yml`, which gates these flags on the first matrix entry.)
> 
> (i.e. drop the `build_archive: true` / `upload_archive: true` lines. If PR-time validation that the tgz packs is desired, keep `build_archive: true` but still remove `upload_archive: true`.)



> **[M-2]** `.github/workflows/qualcomm-internal-build-and-test.yml:L100-L101` (NEW) The new `build-and-test-linux-x86_64-ubuntu-22-04` job sets `build_archive: true` + `upload_archive: true`. But `qualcomm-internal-build-and-test.yml` is the PR/push CI workflow (called by `qualcomm-internal-ci.yml` on `pull_request`/`push` with `do_all: true`), not the release flow — and it's the _only_ job in that workflow setting these flags (both default `false` in `*-single-os.yml`; every sibling job leaves them off). `upload_archive: true` triggers the "Upload Archive to Artifactory" step, so every internal-branch PR/push now builds and uploads a release `.tgz` to Artifactory. Per [P03], artifact-upload flags should be true only in the release workflow. (The release path is already covered by `build-linux-x86_64-ubuntu-22-04` in `qualcomm-internal-build-artifacts.yml`, which gates these flags on the first matrix entry.)
> 
> (i.e. drop the `build_archive: true` / `upload_archive: true` lines. If PR-time validation that the tgz packs is desired, keep `build_archive: true` but still remove `upload_archive: true`.)
> 
> ### 🔵 Minor
> **[N-1]** `.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml:L23` The input description still references a non-existent target: `"JSON list of Python versions to build for Linux targets (aarch64 manylinux_2_34 and x86_64 manylinux_2_34)"`. The matching description in `qualcomm-internal-build-artifacts.yml:42` was correctly updated to `"x86_64 Ubuntu 22.04"`. This is the text shown on the GitHub Actions "Run workflow" UI; please bring the two in sync.
> 
> ```
>         description: "JSON list of Python versions to build for Linux targets (aarch64 manylinux_2_34 and x86_64 Ubuntu 22.04)"
> ```

Fixed with latest commit

qcom/scripts/linux/ubuntu_22_04/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM docker-registry.qualcomm.com/library/ubuntu:22.04
```
huaychou, line 1: Is it available for external developer?

qcom/build_and_test.py
```diff
@@ -353,6 +383,23 @@ def archive_ort_linux_x86_64(self, plan: Plan) -> str:
                 )
             )
 
+    @task
```
huaychou: The indent is not correct

qcom/scripts/linux/ubuntu_22_04/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+RUN groupadd -g $BUILD_GID $BUILD_USER && useradd -m -u $BUILD_UID -g $BUILD_USER $BUILD_USER
+
+ARG ORT_NIGHTLY_BUILD=0
+ENV ORT_NIGHTLY_BUILD=$ORT_NIGHTLY_BUILD
+
+ARG ORT_VERSION_SUFFIX=
+ENV ORT_VERSION_SUFFIX=$ORT_VERSION_SUFFIX
+
+ENV AUDITWHEEL_PLAT=manylinux_2_35_x86_64
```
huaychou, line 40: Why do we need this? I though auditwheel repair wheel accoring to the libs instead of AUDITWHEEL_PLAT

qcom/scripts/linux/ubuntu_22_04/Dockerfile
qti-mbadnara: This mirror cannot be accessed by external users. But the core image is a base Ubuntu 22.04 image. Users can use that instead. This will be mentioned in the release notes

qcom/scripts/linux/ubuntu_22_04/Dockerfile
qti-mbadnara: `setup.py` reads it to decide whether to run `auditwheel` repair at all. Without it on plain Ubuntu, the wheel is produced as `linux_x86_64` with no repair or lib-bundling.

On plain `Ubuntu 22.04` there is no AUDITWHEEL_PLAT set anywhere (unlike the pypa/manylinux base images which bake it in). Without it, `auditwheel's` built-in policy caps at `manylinux_2_34` and would reject our libs since `QAIRT` requires `glibc 2.35`. Setting it in the `Dockerfile` is the Ubuntu equivalent of what the `manylinux` base image does for `aarch64`

qcom/build_and_test.py
qti-mbadnara: Fixed in latest commit

huaychou (approved):
LGTM

tirupath-qti (approved):

PR #445, opened 2026-05-27
[QNN EP] Support for Release Testing Workflow

qti-mbadnara:
## Description

Adds end-to-end release validation infrastructure for `onnxruntime-qnn` artifacts. This includes a GitHub Actions workflow, a Python orchestrator (`release_testing.py`), platform-specific bootstrap scripts, and test/verification scripts for all supported artifact formats (wheel, zip, tgz, nuget).

---

## Motivation & Context

Before this PR, there was no automated validation gate between publishing artifacts to Artifactory and declaring a release good. Build CI only verified that artifacts could be built — not that the published packages were correctly signed, versioned, and functional on real hardware. This PR closes that gap by adding the first end-to-end release validation pipeline.

The workflow is invoked manually (`workflow_dispatch`) or called from a parent release orchestration workflow (`workflow_call`). There is no push trigger — it is a deliberate, gated action.

### Design decisions

**Python orchestrator (`release_testing.py`) as the single entry point** — All test orchestration lives in Python, not in the workflow YAML. The workflow is a thin wrapper that calls `run_release_testing.ps1` (Windows) or `run_release_testing.sh` (Linux), which bootstraps a venv, installs dependencies, and invokes `release_testing.py`. This means the same tests can run locally on a dev machine, on a host device, or on any CI system — not just GitHub Actions.

**HTTP-based Artifactory downloads (no JFrog CLI)** — Artifacts are downloaded via `requests` using HTTP basic auth against Artifactory storage URLs (configured in `config.ini`). This eliminates the JFrog CLI dependency and makes the scripts portable. Authentication uses `ARTIFACTORY_USERNAME` / `ARTIFACTORY_PASSWORD` environment variables, matching the existing `qnn_ep_uplevel.py` pattern.

**Consolidated metadata verification** — Three separate metadata verification scripts (wheel, zip, nuget) were consolidated into a single `verify_windows_metadata.ps1` with an `-ArtifactType` parameter. Shared logic (Authenticode signature check, FileVersion normalization to `major.minor.patch`) is factored into helper functions.

**Exit code propagation for external executables** — All PowerShell scripts check `$LASTEXITCODE` after every `python`, `pip`, and native executable call. `$ErrorActionPreference = 'Stop'` does not propagate non-zero exit codes from external executables in PowerShell.

**Per-job targeted downloads** — Each job downloads only the artifacts it needs. The metadata job downloads all Windows wheels; install/test jobs download only the single wheel for their platform and Python version. This avoids redundant multi-GB downloads across parallel matrix jobs.

**`test_package_version` is separate from `artifact_version`** — The test package archive (`onnxruntime-qnn-<ver>-test_package.zip`) is built and published on a different cadence than the release artifacts. Using the same version string would silently download the wrong package; callers must provide both explicitly.

**ARM64 host detection on Windows** — `is_host_arm64()` checks `PROCESSOR_ARCHITEW6432` before `platform.machine()`. On a WoS device running an x64-emulated Python interpreter, `platform.machine()` returns `AMD64`; `PROCESSOR_ARCHITEW6432` correctly identifies the true host architecture.

---

## Architecture

```
Workflow YAML
  └── run_release_testing.ps1 / .sh  (bootstrap: venv + deps)
        └── release_testing.py        (orchestrator: download + dispatch)
              ├── verify_windows_metadata.ps1  (Authenticode + version)
              ├── install_and_test_wheel.ps1 / .sh  (venv + install + sample)
              ├── verify_backward_compatibility.ps1  (downgrade ORT + sample)
              ├── smoke_test_zip.ps1  (extract + perf_test)
              ├── smoke_test_tgz.sh   (extract + perf_test)
              └── test_nuget_package.ps1  (dotnet build + inference)
```

The orchestrator supports three modes of operation:

| Flag | Behavior |
|---|---|
| `--only-count-check` | Query Artifactory and verify all expected artifacts are present (16 wheels, 1 NuGet, 3 zips, 2 tgz). Runs before all other jobs. |
| `--only-metadata` | Run metadata verification only (Authenticode + FileVersion). |
| `--only-backward-compat` | Run backward compatibility test only (wheel, Windows ARM64). |
| *(default)* | Run install/smoke tests only (no metadata, no backward compat). |

---

## Workflow Inputs

| Input | Required | Description |
|---|---|---|
| `artifact_format` | Yes | One of `wheel`, `nuget`, `zip`, `tgz`, `all`. |
| `artifact_version` | Yes | Version of the artifact to test. Locates `onnxruntime-qnn/<version>/` in Artifactory. |
| `test_package_version` | No (required for zip/tgz) | Version of `test_packages.zip` in `aisw-zip-testproj-generic-virtual`. May differ from `artifact_version`. |

---

## Jobs

All jobs depend on `verify-artifact-counts`, which runs first and fails fast if any expected artifacts are absent on Artifactory.

### Artifact count job (always runs)

| Job | Runner | What it does |
|---|---|---|
| `verify-artifact-counts` | Linux x86_64 | Queries Artifactory and verifies: 16 wheels (4 Python versions x 4 arches), 1 NuGet package, 3 Windows zips (excluding `-pdb.zip` and `-win-arm64ec.zip`), 2 Linux tgz archives. Fails with a detailed list of missing artifacts before any testing begins. |

### Wheel jobs (`artifact_format == wheel`)

| Job | Runner | What it does |
|---|---|---|
| `verify-windows-wheel-metadata` | Windows Build | Downloads all Windows wheels. Runs `verify_windows_metadata.ps1 -ArtifactType wheel`: checks Authenticode signature (signed by `QUALCOMM INCORPORATED`) and `FileVersion` for each DLL. AMD64 wheels carry two DLLs (`libs/amd64/` + `libs/arm64ec/`); both must pass for the wheel-level verdict. Version comparison normalizes to `major.minor.patch` to handle pre-release suffixes and four-part FileVersions. |
| `test-wheel-on-windows` | Windows ARM64 + x86_64 (8 matrix jobs) | Downloads only the wheel for this platform and Python version. Creates a venv via `py` launcher, installs wheel, asserts `onnxruntime_qnn.__version__ == artifact_version`, runs `qcom/samples/test_wheel.py`. Checks `$LASTEXITCODE` after every `python`/`pip` call. Cleans up venv in `finally` block. |
| `test-wheel-on-linux` | Linux ARM64 + x86_64 (8 matrix jobs) | Same as Windows, running natively on the self-hosted runner. Uses `actions/setup-python` to ensure the correct Python version is available. |
| `verify-backward-compatibility` | Windows ARM64 | Downloads the `cp312 win_arm64` wheel. Installs it, downgrades `onnxruntime` to `1.24.2`, then re-runs the sample to confirm the QNN EP plugin still loads against an older ORT runtime. |

### Zip jobs (`artifact_format == zip`)

| Job | Runner | What it does |
|---|---|---|
| `verify-windows-zip-metadata` | Windows Build | Downloads all 3 Windows zips (`win-arm64`, `win-arm64x`, `win-x64`). Runs `verify_windows_metadata.ps1 -ArtifactType zip`: extracts each zip, checks `onnxruntime_providers_qnn.dll` for Authenticode signature and `FileVersion`. |
| `smoke-test-zip-on-windows` | Windows ARM64 + x86_64 (2 matrix jobs) | Downloads the specific zip + `test_package.zip`. Extracts release zip, copies `onnxruntime_perf_test.exe`, `onnxruntime.dll`, and `onnxruntime_providers_shared.dll` from test_package. Runs 1 inference via plugin EP against `cmake/external/onnx/.../test_abs/model.onnx`. Platform auto-detected by the orchestrator. |

### TGZ jobs (`artifact_format == tgz`)

| Job | Runner | What it does |
|---|---|---|
| `smoke-test-tgz-on-linux` | Linux ARM64 + x86_64 (2 matrix jobs) | Downloads the specific tgz + `test_package.zip`. Extracts release tgz, copies `onnxruntime_perf_test`, `libonnxruntime.so`, and `libonnxruntime_providers_shared.so` from test_package. Creates `libonnxruntime.so.1` symlink, sets `LD_LIBRARY_PATH` and `ADSP_LIBRARY_PATH`. Runs 1 inference. Linux ARM64 uses `libQnnHtp.so`; x86_64 uses `libQnnCpu.so`. |

### NuGet jobs (`artifact_format == nuget`)

| Job | Runner | What it does |
|---|---|---|
| `verify-windows-nuget-metadata` | Windows Build | Downloads the `.nupkg`. Extracts it, checks `<version>` in `.nuspec` matches `artifact_version`. Checks Authenticode + `FileVersion` for all three DLLs: managed (`lib/netstandard2.0/`), native win-arm64, native win-x64. Managed DLL `FileVersion` in `x.y.z.0` format is normalized to `x.y.z` before comparison. |
| `test-nuget-on-windows` | Windows ARM64 + x86_64 | Downloads the `.nupkg`. Creates a temp `net8.0` console project, adds the local `.nupkg` as a NuGet source, installs `System.Numerics.Tensors`, `Microsoft.ML.OnnxRuntime` (pinned to the version from `cmake/deps.txt`), and `Qualcomm.ML.OnnxRuntime.QNN`. Builds for the host RID, writes a C# smoke test, runs QNN EP inference. |

---

## Files Changed

| File | Purpose |
|---|---|
| `.github/workflows/qualcomm-internal-release-testing.yml` | Workflow — 10 jobs covering all 4 artifact formats |
| `qcom/scripts/release_testing/release_testing.py` | Python orchestrator — CLI entry point for all release testing |
| `qcom/scripts/release_testing/run_release_testing.ps1` | Bootstrap script (Windows) — creates venv, installs deps, runs orchestrator |
| `qcom/scripts/release_testing/run_release_testing.sh` | Bootstrap script (Linux) — same as above |
| `qcom/scripts/release_testing/verify_windows_metadata.ps1` | Consolidated metadata verification for wheel, zip, and nuget artifacts |
| `qcom/scripts/release_testing/install_and_test_wheel.ps1` | Create venv, install wheel, version check, run sample (Windows) |
| `qcom/scripts/release_testing/install_and_test_wheel.sh` | Same for Linux; detects Python via `python{ver}` or manylinux `/opt/python/` path |
| `qcom/scripts/release_testing/verify_backward_compatibility.ps1` | Install wheel, downgrade ORT to 1.24.2, re-run sample |
| `qcom/scripts/release_testing/smoke_test_zip.ps1` | Extract zip + test_package, copy ORT DLLs, run perf_test (Windows) |
| `qcom/scripts/release_testing/smoke_test_tgz.sh` | Extract tgz + test_package, copy ORT libs, set env vars, run perf_test (Linux) |
| `qcom/scripts/release_testing/test_nuget_package.ps1` | dotnet restore + build per RID, write C# smoke test, run QNN EP inference |
| `qcom/scripts/upleveling/config.ini` | Added storage URL sections for Artifactory download |

### DIFF
--- a/.github/workflows/qualcomm-internal-release-testing.yml
+++ b/.github/workflows/qualcomm-internal-release-testing.yml
@@ -1,0 +1,361 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+name: Perform ONNX Runtime Execution Provider wheel, nuget, archive Release Testing
+
+permissions:
+  checks: write
+  contents: read
+  id-token: write
+
+on:
+  workflow_call:
+    inputs: &common_inputs
+      artifact_format:
+        description: "The format of artifacts to test. Choose one of [wheel, nuget, zip, tgz, all]."
+        type: string
+        required: true
+      artifact_version:
+        description: |
+          Source version of artifact.
+          For python wheel, the format should be <version>.<version>.<version><.><suffix>.
+            And the suffix should be <letters><numbers>.
+          For nuget package, the format should be <version>.<version>.<version>-<suffix>.
+            Please note that underscore (_) is not allowed in the version format.
+        type: string
+        required: true
+      test_package_version:
+        description: |
+          Version folder of the test package archive.
+          Required when artifact_format is zip, tgz, or all.
+        type: string
+        default: "2.5.0rc3"
+  workflow_dispatch:
+    inputs: *common_inputs
+
+env:
+  ARTIFACT_VERSION:     ${{ inputs.artifact_version }}
+  TEST_PACKAGE_VERSION: ${{ inputs.test_package_version }}
+  ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
+  ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_UPLOADER_PASSWORD }}
+
+jobs:
+
+  # ===========================================================================
+  # ARTIFACT COUNT CHECK (runs first — gates all other jobs)
+  # ===========================================================================
+
+  verify-artifact-counts:
+    name: "Verify Artifact Counts"
+    runs-on: [Linux, self-hosted, X64]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Verify Artifact Counts
+        shell: bash
+        run: |
+          bash "${GITHUB_WORKSPACE}/qcom/scripts/release_testing/run_release_testing.sh" \
+            --artifact-type all \
+            --artifact-version "${ARTIFACT_VERSION}" \
+            --source-directory "${GITHUB_WORKSPACE}/release_testing" \
+            --only-count-check
+
+      - name: Summary
+        if: always()
+        shell: bash
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  # ===========================================================================
+  # WHEEL JOBS
+  # ===========================================================================
+
+  verify-windows-wheel-metadata:
+    name: "Verify Windows Wheel Metadata"
+    runs-on: [Windows, self-hosted, Build]
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'wheel' || inputs.artifact_format == 'all'
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Verify Windows Wheel Metadata
+        shell: powershell
+        run: |
+          & "${{ github.workspace }}/qcom/scripts/release_testing/run_release_testing.ps1" `
+            --artifact-type wheel `
+            --artifact-version "$env:ARTIFACT_VERSION" `
+            --source-directory "${{ github.workspace }}/release_testing" `
+            --download-from-artifactory `
+            --only-metadata
+
+      - name: Summary
+        if: always()
+        shell: powershell
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  test-wheel-on-windows:
+    name: "Test Wheel - ${{ matrix.platform.name }} - Py${{ matrix.python-version }}"
+    runs-on: ${{ matrix.platform.os }}
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'wheel' || inputs.artifact_format == 'all'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: "Windows ARM64"
+            os: [Windows, self-hosted, ARM64]
+          - name: "Windows x86_64"
+            os: [Windows, self-hosted, X64]
+        python-version: ['3.11', '3.12', '3.13', '3.14']
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Test Wheel
+        shell: powershell
+        run: |
+          & "${{ github.workspace }}/qcom/scripts/release_testing/run_release_testing.ps1" `
+            --artifact-type wheel `
+            --artifact-version "$env:ARTIFACT_VERSION" `
+            --source-directory "${{ github.workspace }}/release_testing" `
+            --download-from-artifactory `
+            --python-versions "${{ matrix.python-version }}"
+
+      - name: Summary
+        if: always()
+        shell: powershell
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  test-wheel-on-linux:
+    name: "Test Wheel - ${{ matrix.platform.name }} - Py${{ matrix.python-version }}"
+    runs-on: ${{ matrix.platform.os }}
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'wheel' || inputs.artifact_format == 'all'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: "Linux ARM64"
+            os: [Linux, self-hosted, ARM64]
+          - name: "Linux x86_64"
+            os: [Linux, self-hosted, X64]
+        python-version: ['3.11', '3.12', '3.13', '3.14']
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v6
+        with:
+          python-version: '${{ matrix.python-version }}'
+
+      - name: Test Wheel
+        shell: bash
+        run: |
+          bash "${GITHUB_WORKSPACE}/qcom/scripts/release_testing/run_release_testing.sh" \
+            --artifact-type wheel \
+            --artifact-version "${ARTIFACT_VERSION}" \
+            --source-directory "${GITHUB_WORKSPACE}/release_testing" \
+            --download-from-artifactory \
+            --python-versions "${{ matrix.python-version }}"
+
+      - name: Summary
+        if: always()
+        shell: bash
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  verify-backward-compatibility:
+    name: "Verify Backward Compatibility (Windows ARM64, Python 3.12)"
+    runs-on: [Windows, self-hosted, ARM64]
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'wheel' || inputs.artifact_format == 'all'
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Backward Compatibility Test
+        shell: powershell
+        run: |
+          & "${{ github.workspace }}/qcom/scripts/release_testing/run_release_testing.ps1" `
+            --artifact-type wheel `
+            --artifact-version "$env:ARTIFACT_VERSION" `
+            --source-directory "${{ github.workspace }}/release_testing" `
+            --download-from-artifactory `
+            --only-backward-compat
+
+      - name: Summary
+        if: always()
+        shell: powershell
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  # ===========================================================================
+  # ZIP JOBS
+  # ===========================================================================
+
+  verify-windows-zip-metadata:
+    name: "Verify Windows Zip Metadata"
+    runs-on: [Windows, self-hosted, Build]
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'zip' || inputs.artifact_format == 'all'
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Verify Windows Zip Metadata
+        shell: powershell
+        run: |
+          & "${{ github.workspace }}/qcom/scripts/release_testing/run_release_testing.ps1" `
+            --artifact-type zip `
+            --artifact-version "$env:ARTIFACT_VERSION" `
+            --source-directory "${{ github.workspace }}/release_testing" `
+            --test-package-version "$env:TEST_PACKAGE_VERSION" `
+            --download-from-artifactory `
+            --only-metadata
+
+      - name: Summary
+        if: always()
+        shell: powershell
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  smoke-test-zip-on-windows:
+    name: "Smoke Test Zip - ${{ matrix.platform.name }}"
+    runs-on: ${{ matrix.platform.os }}
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'zip' || inputs.artifact_format == 'all'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: "Windows ARM64"
+            os: [Windows, self-hosted, ARM64]
+          - name: "Windows x86_64"
+            os: [Windows, self-hosted, X64]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Smoke Test Zip
+        shell: powershell
+        run: |
+          & "${{ github.workspace }}/qcom/scripts/release_testing/run_release_testing.ps1" `
+            --artifact-type zip `
+            --artifact-version "$env:ARTIFACT_VERSION" `
+            --source-directory "${{ github.workspace }}/release_testing" `
+            --test-package-version "$env:TEST_PACKAGE_VERSION" `
+            --download-from-artifactory
+
+      - name: Summary
+        if: always()
+        shell: powershell
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  # ===========================================================================
+  # TGZ JOBS
+  # ===========================================================================
+
+  smoke-test-tgz-on-linux:
+    name: "Smoke Test TGZ - ${{ matrix.platform.name }}"
+    runs-on: ${{ matrix.platform.os }}
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'tgz' || inputs.artifact_format == 'all'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: "Linux ARM64"
+            os: [Linux, self-hosted, ARM64]
+          - name: "Linux x86_64"
+            os: [Linux, self-hosted, X64]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Smoke Test TGZ
+        shell: bash
+        run: |
+          bash "${GITHUB_WORKSPACE}/qcom/scripts/release_testing/run_release_testing.sh" \
+            --artifact-type tgz \
+            --artifact-version "${ARTIFACT_VERSION}" \
+            --source-directory "${GITHUB_WORKSPACE}/release_testing" \
+            --test-package-version "${TEST_PACKAGE_VERSION}" \
+            --download-from-artifactory
+
+      - name: Summary
+        if: always()
+        shell: bash
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  # ===========================================================================
+  # NUGET JOBS
+  # ===========================================================================
+
+  verify-windows-nuget-metadata:
+    name: "Verify Windows NuGet Metadata"
+    runs-on: [Windows, self-hosted, Build]
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'nuget' || inputs.artifact_format == 'all'
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Verify NuGet Metadata
+        shell: powershell
+        run: |
+          & "${{ github.workspace }}/qcom/scripts/release_testing/run_release_testing.ps1" `
+            --artifact-type nuget `
+            --artifact-version "$env:ARTIFACT_VERSION" `
+            --source-directory "${{ github.workspace }}/release_testing" `
+            --download-from-artifactory `
+            --only-metadata
+
+      - name: Summary
+        if: always()
+        shell: powershell
+        run: |
+          echo "Job result: ${{ job.status }}"
+
+  test-nuget-on-windows:
+    name: "Test NuGet - ${{ matrix.platform.name }}"
+    runs-on: ${{ matrix.platform.os }}
+    needs: verify-artifact-counts
+    if: inputs.artifact_format == 'nuget' || inputs.artifact_format == 'all'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: "Windows ARM64"
+            os: [Windows, self-hosted, ARM64]
+          - name: "Windows x86_64"
+            os: [Windows, self-hosted, X64]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Test NuGet Package
+        shell: powershell
+        run: |
+          & "${{ github.workspace }}/qcom/scripts/release_testing/run_release_testing.ps1" `
+            --artifact-type nuget `
+            --artifact-version "$env:ARTIFACT_VERSION" `
+            --source-directory "${{ github.workspace }}/release_testing" `
+            --download-from-artifactory
+
+      - name: Summary
+        if: always()
+        shell: powershell
+        run: |
+          echo "Job result: ${{ job.status }}"
--- a/.github/workflows/qualcomm-internal-release-testing.yml
+++ b/.github/workflows/qualcomm-internal-release-testing.yml
@@ -114,6 +114,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          submodules: recursive
 
       - name: Test Wheel
         shell: powershell
@@ -149,6 +151,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          submodules: recursive
 
       - name: Setup Python ${{ matrix.python-version }}
         uses: actions/setup-python@v6
@@ -179,6 +183,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          submodules: recursive
 
       - name: Backward Compatibility Test
         shell: powershell
@@ -243,6 +249,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          submodules: recursive
 
       - name: Smoke Test Zip
         shell: powershell
@@ -281,6 +289,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          submodules: recursive
 
       - name: Smoke Test TGZ
         shell: bash
@@ -344,6 +354,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          submodules: recursive
 
       - name: Test NuGet Package
         shell: powershell
--- a/qcom/scripts/release_testing/install_and_test_wheel.ps1
+++ b/qcom/scripts/release_testing/install_and_test_wheel.ps1
@@ -1,0 +1,91 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+param(
+    [Parameter(Mandatory=$true)]
+    [string]$PythonVersion,
+    [Parameter(Mandatory=$true)]
+    [string]$WheelArch,
+    [Parameter(Mandatory=$true)]
+    [string]$WheelDirectory,
+    [Parameter(Mandatory=$true)]
+    [string]$ExpectedVersion,
+    [Parameter(Mandatory=$true)]
+    [string]$SamplePath
+)
+
+$ErrorActionPreference = 'Stop'
+
+# Force UTF-8 for native-process I/O so ONNX Runtime's wide-char log output
+# isn't misrendered with spaces between every character.
+[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
+$OutputEncoding           = [System.Text.Encoding]::UTF8
+
+# Suppress ANSI color escape codes ([0;93m...[m) in ORT's log output —
+# they render as literal text in the GitHub Actions log viewer.
+$env:NO_COLOR = "1"
+
+$pyNoDot = $PythonVersion.Replace(".", "")
+$envName = "py${pyNoDot}_release_testing_env"
+
+# Find the wheel that matches both the Python version and the platform
+$wheel = Get-ChildItem -Path $WheelDirectory `
+  -Filter "*cp${pyNoDot}-cp${pyNoDot}-${WheelArch}.whl" |
+  Select-Object -First 1
+
+if (-not $wheel) {
+    Write-Host "No wheel found matching cp${pyNoDot}-${WheelArch} in $WheelDirectory" -ForegroundColor Red
+    exit 1
+}
+Write-Host "Found wheel: $($wheel.FullName)" -ForegroundColor Cyan
+
+# Create venv named py{XYZ}_release_testing_env using the py launcher
+# to pick the requested Python version (no actions/setup-python needed).
+$pyTag = if ($WheelArch -eq "win_arm64") { "$PythonVersion-arm64" } else { "$PythonVersion" }
+if (Test-Path $envName) { Remove-Item -Path $envName -Recurse -Force }
+py -$pyTag -m venv $envName
+if ($LASTEXITCODE -ne 0) {
+    Write-Host "Failed to create venv (exit $LASTEXITCODE)" -ForegroundColor Red
+    exit $LASTEXITCODE
+}
+
+try {
+    # Activate venv (dot-source so the PATH update persists in this script's scope)
+    . "$envName/Scripts/Activate.ps1"
+
+    # Upgrade pip and install the local wheel
+    python -m pip install --upgrade pip
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "pip upgrade FAILED (exit $LASTEXITCODE)" -ForegroundColor Red
+        exit $LASTEXITCODE
+    }
+
+    python -m pip install $wheel.FullName
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "Wheel install FAILED (exit $LASTEXITCODE)" -ForegroundColor Red
+        exit $LASTEXITCODE
+    }
+
+    # Verify the package reports the expected version
+    $reported = (python -c "import onnxruntime_qnn as qnn_ep; print(qnn_ep.__version__)").Trim()
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "Version query FAILED (exit $LASTEXITCODE)" -ForegroundColor Red
+        exit $LASTEXITCODE
+    }
+    Write-Host "onnxruntime_qnn.__version__ = $reported"
+    if ($reported -ne $ExpectedVersion) {
+        Write-Host "Version mismatch: expected '$ExpectedVersion', got '$reported'" -ForegroundColor Red
+        exit 1
+    }
+    Write-Host "Version check PASS" -ForegroundColor Green
+
+    # Run the sample test
+    python $SamplePath
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "Sample test FAILED (exit $LASTEXITCODE)" -ForegroundColor Red
+        exit $LASTEXITCODE
+    }
+    Write-Host "Sample test PASS" -ForegroundColor Green
+} finally {
+    Remove-Item -Path $envName -Recurse -Force -ErrorAction SilentlyContinue
+}
--- a/qcom/scripts/release_testing/release_testing.py
+++ b/qcom/scripts/release_testing/release_testing.py
@@ -1,0 +1,719 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+"""
+Release testing orchestrator for onnxruntime-qnn artifacts.
+
+Runs validation tests for wheel, zip, tgz, and nuget artifacts either from
+a local directory or by downloading from Artifactory first. Designed to be
+invocable both from CI (GitHub Actions) and from a local dev/host machine.
+
+Usage:
+    # Local mode — artifacts already downloaded
+    python qcom/scripts/release_testing/release_testing.py --artifact-type wheel --artifact-version 2.6.0 \\
+        --source-directory ./release_testing
+
+    # Download mode — fetch from Artifactory first
+    python qcom/scripts/release_testing/release_testing.py --artifact-type wheel --artifact-version 2.6.0 \\
+        --source-directory ./release_testing --download-from-artifactory
+"""
+
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+import platform
+import re
+import subprocess
+import sys
+from configparser import ConfigParser
+from pathlib import Path
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+SCRIPT_DIR = Path(__file__).resolve().parent
+REPO_ROOT = SCRIPT_DIR.parent.parent.parent
+RELEASE_SCRIPTS_DIR = SCRIPT_DIR
+
+# Upleveling infrastructure paths (reused for certs and config)
+UPLEVELING_DIR = SCRIPT_DIR.parent / "upleveling"
+INI_FILE = UPLEVELING_DIR / "config.ini"
+ARTIFACTORY_CERTS_FILE = UPLEVELING_DIR / "certs" / "artifactory-ca.pem"
+
+# Hardcoded defaults
+PRODUCT_NAME = "onnxruntime-qnn"
+SAMPLE_PATH = REPO_ROOT / "qcom" / "samples" / "test_wheel.py"
+MODEL_PATH = (
+    REPO_ROOT
+    / "cmake"
+    / "external"
+    / "onnx"
+    / "onnx"
+    / "backend"
+    / "test"
+    / "data"
+    / "node"
+    / "test_abs"
+    / "model.onnx"
+)
+BACKWARD_COMPAT_ORT_VERSION = "1.24.2"
+
+# Storage index servers per artifact type (config.ini sections with direct storage URLs)
+ARTIFACT_STORAGE_INDEXES = {
+    "wheel": "artifactory-pypi-public-storage",
+    "nuget": "artifactory-nuget-public-storage",
+    "zip": "artifactory-zip-public-storage",
+    "tgz": "artifactory-zip-public-storage",
+    "test_package": "artifactory-zip-testproj-storage",
+}
+
+ARTIFACT_SUFFIXES = {
+    "wheel": ".whl",
+    "nuget": ".nupkg",
+    "zip": ".zip",
+    "tgz": ".tgz",
+}
+
+log = logging.getLogger("release_test")
+
+
+# =============================================================================
+# Platform detection
+# =============================================================================
+
+
+def is_host_windows() -> bool:
+    return platform.system() == "Windows"
+
+
+def is_host_linux() -> bool:
+    return platform.system() == "Linux"
+
+
+def is_host_arm64() -> bool:
+    # On Windows ARM64 with x64-emulated Python, platform.machine()
+    # returns "AMD64"; check PROCESSOR_ARCHITEW6432 for the true arch.
+    arch_w6432 = os.environ.get("PROCESSOR_ARCHITEW6432", "").upper()
+    if arch_w6432 == "ARM64":
+        return True
+    return platform.machine().lower() in ("aarch64", "arm64")
+
+
+def is_host_x86_64() -> bool:
+    return platform.machine().lower() in ("x86_64", "amd64")
+
+
+# =============================================================================
+# Artifactory download (follows qnn_ep_uplevel.py pattern)
+# =============================================================================
+
+
+def _get_artifactory_credentials() -> tuple[str, str]:
+    user = os.environ.get("ARTIFACTORY_USERNAME", "")
+    password = os.environ.get("ARTIFACTORY_PASSWORD", "")
+    missing = []
+    if not user:
+        missing.append("ARTIFACTORY_USERNAME")
+    if not password:
+        missing.append("ARTIFACTORY_PASSWORD")
+    if missing:
+        log.error(f"Missing env var(s) required for downloads: {', '.join(missing)}")
+        sys.exit(1)
+    return user, password
+
+
+def _get_repository_url(artifact_type: str, version: str) -> str:
+    config = ConfigParser()
+    config.read([str(INI_FILE)])
+    index = ARTIFACT_STORAGE_INDEXES[artifact_type]
+    base_url = config.get(index, "repository")
+    return f"{base_url}/{PRODUCT_NAME}/{version}"
+
+
+def download_artifacts(
+    artifact_type: str,
+    version: str,
+    suffix: str,
+    dest_dir: Path,
+    filename_filter: str | None = None,
+) -> list[str]:
+    """Download artifacts from Artifactory using HTTP (no JFrog CLI needed)."""
+    url = _get_repository_url(artifact_type, version)
+    user, password = _get_artifactory_credentials()
+    auth = HTTPBasicAuth(user, password)
+    verify = str(ARTIFACTORY_CERTS_FILE) if ARTIFACTORY_CERTS_FILE.exists() else True
+
+    log.info(f"Fetching artifact list from: {url}")
+    response = requests.get(url, auth=auth, verify=verify)
+    if response.status_code != 200:
+        log.error(f"Failed to fetch artifact list from {url} (HTTP {response.status_code})")
+        sys.exit(1)
+
+    # Parse Artifactory HTML directory listing (same regex as qnn_ep_uplevel.py)
+    artifact_list = [
+        m.group(1)
+        for m in (
+            re.search(r'href=["\']([^"\']*' + re.escape(suffix) + r')["\']', line)
+            for line in response.text.splitlines()
+            if suffix in line
+        )
+        if m
+    ]
+
+    if filename_filter:
+        artifact_list = [f for f in artifact_list if re.search(filename_filter, f)]
+
+    if not artifact_list:
+        log.error(f"No artifacts with suffix '{suffix}' found at {url}")
+        sys.exit(1)
+
+    dest_dir.mkdir(parents=True, exist_ok=True)
+
+    for artifact_file in artifact_list:
+        artifact_url = f"{url}/{artifact_file}"
+        download_path = dest_dir / artifact_file
+        log.info(f"Downloading {artifact_file}")
+        r = requests.get(artifact_url, auth=auth, verify=verify)
+        if r.status_code != 200:
+            log.error(f"Failed to download {artifact_file} (HTTP {r.status_code})")
+            sys.exit(1)
+        download_path.write_bytes(r.content)
+        log.info(f"Saved {download_path}")
+
+    return artifact_list
+
+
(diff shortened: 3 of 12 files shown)


qcom/scripts/release_testing/test_nuget_package.ps1
```diff
@@ -0,0 +1,160 @@
...
+            Select-Object -First 1
+
+        if (-not $dll) {
+            Write-Host "  DLL FAIL: onnxruntime_providers_qnn.dll not found in $outputDir" -ForegroundColor Red
+            $buildFailures++
+        } else {
+            Write-Host "  BUILD PASS: onnxruntime_providers_qnn.dll found at $($dll.FullName)" -ForegroundColor Green
+        }
+    }
+
```
kuanyul-qti, line 180: Do we need to run a real inference?

kuanyul-qti (commented):
Do we need to include Android AAR test for Maven release?

qcom/scripts/release_testing/test_nuget_package.ps1
qti-mbadnara: Done!

qti-mbadnara:
> Do we need to include Android AAR test for Maven release?

Can be taken up as a follow-up

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,520 @@
...
+permissions:
+  checks: write
+  contents: read
+  id-token: write
+
+on:
+  workflow_call:
+    inputs: &common_inputs
+      artifact_format:
+        description: "The format of artifact. Choose one of [wheel, nuget, zip, tgz, all]."
```
qti-kromero: nit: "The format of artifacts **to test**"

qti-kromero: generally do we expect the use case for only testing a subset of the artifacts to persist? wondering if the effort of maintaining that is worth the potential benefit or if it should just be all or nothing

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,520 @@
...
+        type: string
+        required: true
+      version_from:
```
qti-kromero: I think source_artifact_version might be a better name for this or artifact_version

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,520 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+name: Perform ONNX Runtime Execution Provider wheel, nuget, archive Release Testing
```
qti-kromero, line 4: I would generally kind of prefer if this were integrated into the build_and_test.py infrastructure or a similar script which does the orchestration and then the GitHub Actions workflow file is just a wrapper over that

the benefit of this primarily being that we can run the workflow more easily outside of GitHub Actions in case we needed to either on a host device or if we wanted to migrate automation frameworks at some point

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,520 @@
...
+  # WHEEL JOBS
+  # ===========================================================================
+
+  verify-windows-wheel-metadata:
+    name: "Verify Windows Wheel Metadata"
+    runs-on: [Windows, self-hosted, Build]
+    if: inputs.artifact_format == 'wheel' || inputs.artifact_format == 'all'
+
+    steps:
+      - uses: actions/checkout@v6
```
qti-kromero, line 81: do these jobs need to checkout? I don't think they need access to the source code as-is. if we change to rely on a script in the source tree then they might

qcom/scripts/release_testing/verify_windows_nuget_metadata.ps1
```diff
@@ -0,0 +1,150 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
```
qti-kromero: I feel like these files could have one entrypoint which shares some of the common bits of them - 

i.e. verify_windows_artifact.ps1 which takes in a --wheel, --zip, or --nuget flag and can differentiate the validation according to that

some of the common validation/enumeration of errors and printing of a summary appears superficially common though. they also basically have the same arguments as well with path to artifact + version

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,520 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
```
qti-kromero, line 2: kind of another general note but why not integrate this testing into the building of the release artifacts themselves? then instead of downloading something that may or may not be valid, verifying it, and then the thing that may or may not be valid is still up in our artifactory, we could have certainty that the thing we are uploading to begin with is valid and passes smoke testing

qcom/scripts/release_testing/smoke_test_zip.ps1
```diff
@@ -0,0 +1,99 @@
...
+# --- Find onnxruntime_perf_test.exe in the test binaries ---
+$perfTest = Get-ChildItem -Path $testBinDir.FullName -Recurse -Filter "onnxruntime_perf_test.exe" |
+    Select-Object -First 1
+
+if (-not $perfTest) {
+    Write-Host "ERROR: onnxruntime_perf_test.exe not found in $($testBinDir.FullName)" -ForegroundColor Red
+    exit 1
+}
+
+# --- Copy onnxruntime_perf_test.exe to the extracted release folder ---
```
qti-kromero, line 63: something feels a little code smelly between smoke_test_tgz.sh and smoke_test_zip.ps1 with all the downloading/manipulation of the binaries on disk. seems like there are a few potential optimizations but kind of unsure if we want to consider any of them

1 - we could have onnxruntime_perf_test distributed as part of the zip/tgz release as a helper binary
2 - onnxruntime.dll/onnxruntime_providers_shared.dll do not technically need to come from the test package, they could come from any onnxruntime version and could also fulfill backwards compat test case

I think the main thing I'm trying to avoid is to have to download both the test package and the artifact under test to only get some bits of the test package. seems kind of like a waste

qcom/scripts/release_testing/install_and_test_wheel.ps1
```diff
@@ -0,0 +1,64 @@
...
+$reported = (python -c "import onnxruntime_qnn as qnn_ep; print(qnn_ep.__version__)").Trim()
+Write-Host "onnxruntime_qnn.__version__ = $reported"
+if ($reported -ne $ExpectedVersion) {
+    Write-Host "Version mismatch: expected '$ExpectedVersion', got '$reported'" -ForegroundColor Red
+    exit 1
+}
+Write-Host "Version check PASS" -ForegroundColor Green
+
+# Run the sample test
+python $SamplePath
```
huaychou: **[M-1 Major]** `$ErrorActionPreference = 'Stop'` only throws on PowerShell Cmdlet errors — it does **not** propagate non-zero exit codes from external executables such as `python.exe`. All five `python` invocations in this script (lines 51, 52, 55, 64, and the venv activation) are unchecked. If `python $SamplePath` fails at runtime (e.g., QNN EP fails to load), the PowerShell script exits with code 0 and the CI job falsely reports success, defeating the entire purpose of the release test.

**Fix**: add `$LASTEXITCODE` checks after every `python` (and `pip`) call, e.g.:
```powershell
python $SamplePath
if ($LASTEXITCODE -ne 0) {
    Write-Host "Sample test FAILED with exit code $LASTEXITCODE" -ForegroundColor Red
    exit $LASTEXITCODE
}
```
Same pattern needed for the `pip install` calls at lines 51–52.

qcom/scripts/release_testing/install_and_test_wheel.ps1
```diff
@@ -0,0 +1,64 @@
...
+if (-not $wheel) {
+    Write-Host "No wheel found matching cp${pyNoDot}-${WheelArch} in $WheelDirectory" -ForegroundColor Red
+    exit 1
+}
+Write-Host "Found wheel: $($wheel.FullName)" -ForegroundColor Cyan
+
+# Create venv named py{XYZ}_release_testing_env using the py launcher
+# to pick the requested Python version (no actions/setup-python needed).
+$pyTag = if ($WheelArch -eq "win_arm64") { "$PythonVersion-arm64" } else { "$PythonVersion" }
+py -$pyTag -m venv $envName
```
huaychou, line 46: **[N-1 Minor]** The venv `$envName` is created in `${{ github.workspace }}` and never cleaned up on script exit. When the CI job is re-run (GitHub Actions re-run), the directory already exists and `python -m venv` may reuse a stale or mis-versioned environment, causing intermittent failures that are hard to diagnose.

**Fix**: remove the old venv before creating it and clean up in a `finally` block:
```powershell
if (Test-Path $envName) { Remove-Item -Path $envName -Recurse -Force }
py -$pyTag -m venv $envName
# ... test logic ...
finally { Remove-Item -Path $envName -Recurse -Force -ErrorAction SilentlyContinue }
```

qcom/scripts/release_testing/verify_backward_compatibility.ps1
```diff
@@ -0,0 +1,60 @@
...
+
+# Install onnxruntime-qnn from the local wheel (pulls onnxruntime as a dependency)
+python -m pip install $wheel.FullName
+
+# Replace the bundled onnxruntime with the older version to verify backward compatibility
+python -m pip uninstall -y onnxruntime
+python -m pip install "onnxruntime==$OnnxruntimeVersion"
+
+# Run the sample test against the older onnxruntime
+python $SamplePath
```
huaychou: **[M-2 Major]** Same root cause as M-1 in `install_and_test_wheel.ps1`: `python -m pip install $wheel.FullName` (line 53), `python -m pip uninstall -y onnxruntime` (line 56), `python -m pip install "onnxruntime==$OnnxruntimeVersion"` (line 57), and `python $SamplePath` (line 60) are all unchecked for `$LASTEXITCODE`. The entire point of this script is to confirm that a downgraded `onnxruntime` still works with the new QNN EP wheel — if `python $SamplePath` fails, CI reports success and the compatibility regression is invisible.

**Fix**: add `$LASTEXITCODE` checks after every `python` call, especially the final one:
```powershell
python $SamplePath
if ($LASTEXITCODE -ne 0) {
    Write-Host "Backward compat test FAILED (exit $LASTEXITCODE)" -ForegroundColor Red
    exit $LASTEXITCODE
}
Write-Host "Backward compat test PASSED" -ForegroundColor Green
```

qcom/scripts/release_testing/verify_windows_wheel_metadata.ps1
```diff
@@ -0,0 +1,183 @@
...
+    } elseif ($signature.SignerCertificate.Subject -like "*QUALCOMM INCORPORATED*") {
+        Write-Host "  [$Label] CERTIFICATE PASS" -ForegroundColor Green
+        $result.CertPass = $true
+    } else {
+        Write-Host "  [$Label] CERTIFICATE FAIL: Not signed by QUALCOMM INCORPORATED (Subject: $($signature.SignerCertificate.Subject))" -ForegroundColor Red
+    }
+
+    $fileVersion = [System.Diagnostics.FileVersionInfo]::GetVersionInfo($Dll.FullName).FileVersion
+
+    if ($fileVersion -eq $ExpectedVersion) {
```
huaychou: **[M-3 Major]** `$fileVersion -eq $ExpectedVersion` is a direct string comparison between `FileVersionInfo.FileVersion` (which returns a four-part version like `"2.5.0.0"`) and `$ExpectedVersion` (which for a pre-release wheel is `"2.5.0.rc3"`). These will never be equal, so every RC wheel will report `VERSION FAIL` even when the version is correct. Compare `verify_windows_nuget_metadata.ps1`, which normalises `FileVersion` to `major.minor.patch` with a comment explaining why.

**Fix**: normalise both sides to `major.minor.patch` before comparing:
```powershell
# Normalise to major.minor.patch to tolerate pre-release suffix (e.g. 2.5.0.rc3 -> 2.5.0)
$parts = $fileVersion.Split('.')
$normFile = "$($parts[0]).$($parts[1]).$($parts[2])"
$expParts = $ExpectedVersion.Split('.')
$normExp  = "$($expParts[0]).$($expParts[1]).$($expParts[2])"
if ($normFile -eq $normExp) { ... }
```

qcom/scripts/release_testing/test_nuget_package.ps1
```diff
@@ -0,0 +1,190 @@
...
+    # Step 4: Install packages
+    # =============================================================================
+
+    Write-Host ""
+    Write-Host "Step 4: Installing packages..." -ForegroundColor Cyan
+
+    & dotnet add package System.Numerics.Tensors --version 9.0.0
+    Write-Host "  Installed System.Numerics.Tensors 9.0.0" -ForegroundColor Green
+
+    & dotnet add package Microsoft.ML.OnnxRuntime
```
huaychou: Do we need to specific version of onnxruntime?

qcom/scripts/release_testing/test_nuget_package.ps1
```diff
@@ -0,0 +1,190 @@
...
+    Write-Host "Step 3 PASSED" -ForegroundColor Green
+
+    # =============================================================================
+    # Step 4: Install packages
+    # =============================================================================
+
+    Write-Host ""
+    Write-Host "Step 4: Installing packages..." -ForegroundColor Cyan
+
+    & dotnet add package System.Numerics.Tensors --version 9.0.0
```
huaychou: **[T-1 Nit]** `System.Numerics.Tensors --version 9.0.0` is hardcoded with no comment explaining the choice. If this version is yanked from nuget.org or becomes incompatible with the .NET 8.0 target, the test breaks silently.

**Fix**: add an inline comment:
```powershell
& dotnet add package System.Numerics.Tensors --version 9.0.0  # minimum version supporting .NET 8 Tensor<T>
```

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,520 @@
...
+    runs-on: ${{ matrix.platform.os }}
+    if: inputs.artifact_format == 'nuget' || inputs.artifact_format == 'all'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - name: "Windows ARM64"
+            os: [Windows, self-hosted, ARM64]
+            runtime_identifiers: "win-arm64"
```
huaychou: **[N-3 Minor]** The `test-nuget-on-windows` matrix includes `win-arm64` and `win-x64` but omits `win-arm64ec`. `verify_windows_wheel_metadata.ps1` explicitly validates that the AMD64 wheel contains both `libs/amd64/` and `libs/arm64ec/` DLLs, confirming ARM64EC is a supported deployment target. Any breakage in `win-arm64ec` dotnet restore/build goes undetected.

**Fix**: add a `win-arm64ec` entry (can run on an x64 machine since ARM64EC is a cross-build target):
```yaml
- name: "Windows ARM64EC"
  os: [Windows, self-hosted, X64]
  runtime_identifiers: "win-arm64ec"
  backend_dll: QnnCpu.dll  # ARM64EC cannot use HTP on an x64 host
```

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,520 @@
...
+          For nuget package, the format should be <version>.<version>.<version>-<suffix>.
+            Please note that underscore (_) is not allowed in the version format.
+        type: string
+        required: true
+      test_package_version:
+        description: |
+          Version folder of the test package archive.
+          Required when artifact_format is zip, tgz, or all.
+        type: string
+        default: "2.5.0rc3"
```
huaychou, line 32: Just wondering, can it be retrieved by flag or branch name?

qcom/scripts/release_testing/smoke_test_tgz.sh
```diff
@@ -0,0 +1,92 @@
...
+echo ""
+echo "Extracting release tgz: $release_tgz_name"
+mkdir -p "$release_extract_dir"
+tar -xf "$release_tgz" -C "$release_extract_dir"
+
+# --- Extract test_package.zip ---
+test_pkgs_dir="$(dirname "$test_package_zip")/test_package"
+echo "Extracting test_package.zip"
+mkdir -p "$test_pkgs_dir"
+python3 -c "import zipfile; zipfile.ZipFile('${test_package_zip}').extractall('${test_pkgs_dir}')"
```
huaychou: **[T-2 Nit]** Paths are embedded directly in a quoted Python one-liner: `python3 -c "import zipfile; zipfile.ZipFile('${test_package_zip}').extractall('${test_pkgs_dir}')"`. If either path contains a single quote, the command breaks. `unzip` is available on all Linux runners and is cleaner.

**Fix** (preferred):
```bash
unzip -q "${test_package_zip}" -d "${test_pkgs_dir}"
```
Or pass paths as arguments to avoid string interpolation:
```bash
python3 -c "import zipfile,sys; zipfile.ZipFile(sys.argv[1]).extractall(sys.argv[2])" "${test_package_zip}" "${test_pkgs_dir}"
```

qcom/scripts/release_testing/verify_windows_zip_metadata.ps1
```diff
@@ -0,0 +1,107 @@
...
+
+# Verify the directory exists
+if (-not (Test-Path $ArchiveDirectory -PathType Container)) {
+    Write-Host ""
+    Write-Error "Directory not found: $ArchiveDirectory"
+    exit 1
+}
+
+# All Windows zip release variants to verify
+$zipArchs = @("win-arm64", "win-arm64x", "win-x64")
```
huaychou: **[T-3 Nit]** `$zipArchs = @("win-arm64", "win-arm64x", "win-x64")` is a hardcoded list. When a new ZIP variant is added in the future (e.g., `win-x64-avx2`), if this script is not updated alongside it, verification silently passes while the new variant's issues go undetected.

**Fix**: add an inline comment:
```powershell
$zipArchs = @("win-arm64", "win-arm64x", "win-x64")  # TODO: update when new ZIP release variants are added
```

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: Can be taken as a follow up but can be left as a user input for now

qcom/scripts/release_testing/test_nuget_package.ps1
qti-mbadnara: Done! We can take ort_core version from cmake

qcom/scripts/release_testing/verify_windows_wheel_metadata.ps1
qti-mbadnara: Done

qcom/scripts/release_testing/verify_backward_compatibility.ps1
qti-mbadnara: Done

qcom/scripts/release_testing/install_and_test_wheel.ps1
qti-mbadnara: Done

qcom/scripts/release_testing/install_and_test_wheel.ps1
qti-mbadnara: Done

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: 99% of the time we will be testing every artifact but we might need the option to test one artifact at a time for a staggered release.

qti-mbadnara: Done

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: The jobs reference qcom/scripts/release_testing/*.ps1 scripts, so might be needed to checkout

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: Done!

qcom/scripts/release_testing/verify_windows_nuget_metadata.ps1
qti-mbadnara: Done!

qcom/scripts/release_testing/smoke_test_zip.ps1
qti-mbadnara: 1) I think the executable is not shipped as part of the external suite
2) onnxruntime.dll/onnxruntime_providers_shared.dll do not technically need to come from the test package, they could come from any onnxruntime version and could also fulfill backwards compat test case - True but the goal here is to ensure the onnxruntime.dll we are currently build EP with is compatible. 

We can take this up in a follow up PR if necessary

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: Done!

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: The post-publish validation is intentional, it tests the actual artifact that users will download, not the build output. This catches issues that only surface after packaging and publishing

qti-mbadnara:
Sample Workflow: https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/33561296177

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,274 @@
...
+name: Perform ONNX Runtime Execution Provider wheel, nuget, archive Release Testing
+
+permissions:
+  checks: write
+  contents: read
+  id-token: write
+
+on:
+  pull_request:
+    branches: [main]
```
kuanyul-qti: Why is this needed if release testing is supposed to be triggered by the user?

qcom/scripts/upleveling/config.ini
```diff
@@ -54,3 +54,19 @@ repository: https://artifactory-qdc-global.qualcomm.com/artifactory/aisw-maven-v
...
+repository: https://artifactory-qdc-global.qualcomm.com/artifactory/aisw-public-nuget-virtual
+cert: certs/artifactory-ca.pem
+
+[artifactory-zip-public-storage]
+repository: https://artifactory-qdc-global.qualcomm.com/artifactory/aisw-zip-public-generic-virtual
+cert: certs/artifactory-ca.pem
+
+[artifactory-zip-testproj-storage]
+repository: https://artifactory-qdc-global.qualcomm.com/artifactory/aisw-zip-testproj-generic-virtual
+cert: certs/artifactory-ca.pem
```
kuanyul-qti: What’s the reason for creating separate indices here instead of using the existing ones, such as `artifactory-nuget-public`?

qcom/scripts/release_testing/release_testing.py
```diff
@@ -0,0 +1,589 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
kuanyul-qti, line 1: How about placing this file under `qcom/scripts/release_testing`?

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,274 @@
...
+jobs:
+
+  # ===========================================================================
+  # WHEEL JOBS
+  # ===========================================================================
+
+  verify-windows-wheel-metadata:
+    name: "Verify Windows Wheel Metadata"
+    runs-on: [Windows, self-hosted, Build]
+    if: inputs.artifact_format == 'wheel' || inputs.artifact_format == 'all' || inputs.artifact_format == ''
```
kuanyul-qti: If `inputs.artifact_format` is required, can `inputs.artifact_format == ''` still exist?

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,274 @@
...
+          Version folder of the test package archive.
+          Required when artifact_format is zip, tgz, or all.
+        type: string
+        default: "2.5.0rc3"
+  workflow_dispatch:
+    inputs: *common_inputs
+
+env:
+  ARTIFACT_VERSION:     ${{ inputs.artifact_version || '2.5.0' }}
+  TEST_PACKAGE_VERSION: ${{ inputs.test_package_version || '2.5.0rc3' }}
```
kuanyul-qti: I don't think these two need to be set as environment variables. They can be obtained by `input.*`. Currently, the `release_testing.py` script takes them as arguments instead of reading them from env.

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: This was for local validation, will remove this in the final squash commit

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: This is for local testing, will remove in the final squash commit

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: This is for local testing, will remove in the final squash commit

qcom/scripts/release_testing/release_testing.py
qti-mbadnara: Sure will do

qcom/scripts/upleveling/config.ini
qti-mbadnara: The existing URLs are only usable for uploads, hitting 404 when I try downloading artifacts with the same URL

qcom/scripts/release_testing/release_testing.py
```diff
@@ -0,0 +1,589 @@
...
+def is_host_windows() -> bool:
+    return platform.system() == "Windows"
+
+
+def is_host_linux() -> bool:
+    return platform.system() == "Linux"
+
+
+def is_host_arm64() -> bool:
+    return platform.machine().lower() in ("aarch64", "arm64")
```
huaychou, line 101: **[Major] M-4** On a Windows ARM64 host running an x64-emulated Python interpreter (e.g., the system-default Python on a WoS device), `platform.machine()` returns `"AMD64"`, not `"ARM64"`. So `is_host_arm64()` returns `False`, and `test_wheel` (line 268) selects the `win_amd64` wheel instead of `win_arm64` — ARM64 validation is silently skipped.

**Suggested fix:** Check `PROCESSOR_ARCHITEW6432` (set to `"ARM64"` when an x64-emulated process runs on an ARM64 host):
```python
def is_host_arm64() -> bool:
    # On Windows ARM64 with x64-emulated Python, platform.machine()
    # returns "AMD64"; check PROCESSOR_ARCHITEW6432 for the true arch.
    arch_w6432 = os.environ.get("PROCESSOR_ARCHITEW6432", "").upper()
    if arch_w6432 == "ARM64":
        return True
    return platform.machine().lower() in ("aarch64", "arm64")
```

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,274 @@
...
+  # ===========================================================================
+  # WHEEL JOBS
+  # ===========================================================================
+
+  verify-windows-wheel-metadata:
+    name: "Verify Windows Wheel Metadata"
+    runs-on: [Windows, self-hosted, Build]
+    if: inputs.artifact_format == 'wheel' || inputs.artifact_format == 'all' || inputs.artifact_format == ''
+
+    steps:
```
huaychou, line 80: **[Minor] N-5** None of the 9 jobs includes a final step guarded by `if: always()` that prints a test summary. When a step fails, the GitHub Actions UI only shows "step failed" — full diagnostics require hunting through multi-runner logs.

**Suggested fix:** Add to each job:
```yaml
- name: Summary
  if: always()
  shell: bash
  run: echo "Job result: ${{ job.status }}"
```

qcom/scripts/release_testing/run_release_testing.sh
```diff
@@ -0,0 +1,33 @@
...
+# Clean stale venv
+rm -rf "${VENV_NAME}"
+
+# Create and activate venv
+python3 -m venv "${VENV_NAME}"
+source "${VENV_NAME}/bin/activate"
+
+# Install dependencies
+pip install uv
+uv pip install -r "${REPO_ROOT}/qcom/requirements.txt"
```
huaychou: Do we need `--system-certs` as `run_release_testing.ps1`?

qcom/scripts/release_testing/run_release_testing.ps1
```diff
@@ -0,0 +1,32 @@
...
+$ErrorActionPreference = 'Stop'
+
+$RepoRoot = git rev-parse --show-toplevel
+$VenvName = "release_testing_venv"
+
+# Clean stale venv
+if (Test-Path $VenvName) { Remove-Item -Path $VenvName -Recurse -Force }
+
+# Create and activate venv
+python -m venv $VenvName
```
huaychou, line 21: Just wondering, will this python be amd64 or arm64?

qti-mbadnara: This will be ARM64

qcom/scripts/release_testing/run_release_testing.sh
qti-mbadnara: Done!

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: Done!

qcom/scripts/release_testing/release_testing.py
qti-mbadnara: Done!

qcom/scripts/release_testing/run_release_testing.ps1
```diff
@@ -0,0 +1,36 @@
...
+$VenvName = "release_testing_venv"
+
+# Clean stale venv
+if (Test-Path $VenvName) { Remove-Item -Path $VenvName -Recurse -Force }
+
+# Create and activate venv.
+# This venv only runs the Python orchestrator (release_testing.py); its
+# architecture does not affect which wheels are tested.  Wheel testing
+# creates its own arch-specific venv via install_and_test_wheel.ps1 using
+# the py launcher (e.g. py -3.12-arm64).
```
huaychou: **[N-1] (Minor)** run_release_testing.ps1: venv creation and dependency install steps lack $LASTEXITCODE checks

Problem Code

```
# no $LASTEXITCODE check after venv creation
python -m venv $VenvName
. "$VenvName/Scripts/Activate.ps1"

pip install uv
uv pip install --system-certs -r "$RepoRoot/qcom/requirements.txt"

```

`$ErrorActionPreference = 'Stop'` only throws on Cmdlet errors — it does not propagate non-zero exit codes from external processes such as `python`, `pip`, or `uv`. If venv creation fails (e.g., Python not found), the script continues and attempts to dot-source a non-existent `Activate.ps1`, causing confusing downstream errors.

**Impact**: When the Python environment is broken or a corporate proxy blocks pip, CI fails with a misleading error rather than clearly reporting the venv or dependency install failure.

**Suggested fix:**

Add $LASTEXITCODE checks after each external process call:

```powershell
python -m venv $VenvName
if ($LASTEXITCODE -ne 0) { Write-Host "Failed to create venv" -ForegroundColor Red; exit $LASTEXITCODE }

. "$VenvName/Scripts/Activate.ps1"

pip install uv
if ($LASTEXITCODE -ne 0) { Write-Host "pip install uv FAILED" -ForegroundColor Red; exit $LASTEXITCODE }

uv pip install --system-certs -r "$RepoRoot/qcom/requirements.txt"
if ($LASTEXITCODE -ne 0) { Write-Host "uv pip install FAILED" -ForegroundColor Red; exit $LASTEXITCODE }
```

qcom/scripts/release_testing/verify_windows_metadata.ps1
```diff
@@ -0,0 +1,337 @@
...
+    [string]$ExpectedVersion
+)
+
+$ErrorActionPreference = 'Stop'
+
+# Force UTF-8 so any native tool output renders correctly in the GHA log viewer.
+[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
+$OutputEncoding           = [System.Text.Encoding]::UTF8
+
+# =============================================================================
```
huaychou: **[N-2] (Minor)** verify_windows_metadata.ps1: Normalize-Version cannot handle pre-release suffixes embedded in the numeric part

Problem Code

```
function Normalize-Version([string]$Version) {
    $parts = $Version.Split('.')
    return "$($parts[0]).$($parts[1]).$($parts[2])"
}

```

This function correctly normalizes 4-part DLL FileVersions (e.g., `"2.5.0.0"` → `"2.5.0"`). However, when `ExpectedVersion` uses the no-dot pre-release format (e.g., `"2.5.0rc3"` — the format used in the YAML default), `Split('.')` produces `["2", "5", "0rc3"]`, so `Normalize-Version("2.5.0rc3") = "2.5.0rc3"`, which does not match `"2.5.0"` (from FileVersion `"2.5.0.0"`).

**Impact**: Running metadata verification with the no-dot RC version string (e.g., `"2.5.0rc3"`) causes false FAIL on the version check.

**Suggested fix:**

Use a regex to extract only the major.minor.patch prefix:

```powershell
function Normalize-Version([string]$Version) {
    # Extract major.minor.patch, discarding 4th segment or embedded alpha suffix
    if ($Version -match '^(\d+\.\d+\.\d+)') {
        return $Matches[1]
    }
    return $Version
}
```

This handles both "2.5.0.0" (4-part) and "2.5.0rc3" (embedded suffix).

qcom/scripts/release_testing/test_nuget_package.ps1
```diff
@@ -0,0 +1,200 @@
...
+    Write-Host "  Added nuget.org source" -ForegroundColor Green
+
+    Write-Host ""
+    Write-Host "Step 3 PASSED" -ForegroundColor Green
+
+    # =============================================================================
+    # Step 4: Install packages
+    # =============================================================================
+
+    Write-Host ""
```
huaychou: **[N-3] (Minor)** test_nuget_package.ps1: dotnet external commands lack $LASTEXITCODE checks

Problem Code

```
# Step 3: no $LASTEXITCODE check
& dotnet nuget add source $NuGetDirectory --name QnnLocalSource
& dotnet nuget add source "https://api.nuget.org/v3/index.json" --name nuget

# Step 4: dotnet add package also unchecked
& dotnet add package System.Numerics.Tensors --version 9.0.0
& dotnet add package Microsoft.ML.OnnxRuntime --version $ortVersion

```

`$ErrorActionPreference = 'Stop'` only throws on Cmdlet errors — `& dotnet ...` is an external process and its non-zero exit code is not automatically propagated. If the nuget source command fails due to a name conflict or a package cannot be resolved, the script continues and eventually fails with an unrelated error.

**Impact**: NuGet source setup or package install failures do not immediately abort the script, making debugging difficult.

**Suggested fix:**

Add an exit code check after each & dotnet ... call:

```powershell
& dotnet nuget add source $NuGetDirectory --name QnnLocalSource
if ($LASTEXITCODE -ne 0) { Write-Host "dotnet nuget add source (local) FAILED" -ForegroundColor Red; exit $LASTEXITCODE }
```

Apply the same pattern to all dotnet add package calls.

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,361 @@
...
+          For nuget package, the format should be <version>.<version>.<version>-<suffix>.
+            Please note that underscore (_) is not allowed in the version format.
+        type: string
+        required: true
+      test_package_version:
+        description: |
+          Version folder of the test package archive.
+          Required when artifact_format is zip, tgz, or all.
+        type: string
+        default: "2.5.0rc3"
```
huaychou: **[T-1] (Nit)** test_package_version default hardcoded to a specific RC version — becomes stale after release

The `test_package_version` default is hardcoded to `"2.5.0rc3"` and will need manual updates for each release.

**Suggested fix:**

Change the default to an empty string and document that the caller must supply it, or derive the default automatically from artifact_version in the script.

qcom/scripts/release_testing/run_release_testing.ps1
```diff
@@ -0,0 +1,36 @@
...
+python -m venv $VenvName
+. "$VenvName/Scripts/Activate.ps1"
+
```
huaychou: **[T-2] (Nit)** uv installed without version pin — upstream updates may silently change behavior

`pip install uv` (both PowerShell and shell variants) has no version pin, violating the CI tool pinning convention ([P29]).

**Suggested fix:**

Pin to a specific or bounded version, e.g., pip install "uv>=0.4,<1", with a comment explaining the ceiling.

qcom/scripts/release_testing/test_nuget_package.ps1
```diff
@@ -0,0 +1,200 @@
...
+    # Parse the ort_core tag from cmake/deps.txt so the pin stays in sync automatically.
+    $depsFile = Join-Path $PSScriptRoot "../../../cmake/deps.txt"
+    $ortLine  = Get-Content $depsFile | Where-Object { $_ -match '^ort_core;' }
+    if ($ortLine -match '/tags/v([^/]+)\.zip') {
+        $ortVersion = $Matches[1]
+    } else {
+        Write-Host "  ERROR: Could not parse ort_core version from cmake/deps.txt" -ForegroundColor Red
+        exit 1
+    }
+    & dotnet add package Microsoft.ML.OnnxRuntime --version $ortVersion
```
huaychou: **[T-3] (Nit)** System.Numerics.Tensors version hardcoded to 9.0.0

`dotnet add package System.Numerics.Tensors --version 9.0.0` has a hardcoded version that requires manual updates when the .NET version changes.

**Suggested fix:**

Derive the version from cmake/deps.txt or another config file, or use a floating range (e.g., --version "[9.0,10.0)") with an explanatory comment.

qcom/scripts/release_testing/verify_windows_metadata.ps1
```diff
@@ -0,0 +1,337 @@
...
+            try {
+                Expand-Artifact $wheel.FullName $extractDir
+
+                $allDlls    = Get-ChildItem -Path $extractDir -Recurse -Filter "onnxruntime_providers_qnn.dll"
+                $amd64Dll   = $allDlls | Where-Object { $_.FullName -match "[\\/]libs[\\/]amd64[\\/]"   } | Select-Object -First 1
+                $arm64ecDll = $allDlls | Where-Object { $_.FullName -match "[\\/]libs[\\/]arm64ec[\\/]" } | Select-Object -First 1
+
+                # libs/amd64/onnxruntime_providers_qnn.dll
+                $amd64Path = if ($amd64Dll) { $amd64Dll.FullName } else { "" }
+                $r = Test-QnnDll -DllPath $amd64Path -Label "amd64  " -ExpectedVersion $ExpectedVersion
```
huaychou: **[T-4] (Nit)** verify_windows_metadata.ps1: $zipArchs list has no maintenance comment about updating when variants change

`$zipArchs = @("win-arm64", "win-arm64x", "win-x64")` hardcodes the supported zip architectures with no comment reminding maintainers to update it when arch variants change.

**Suggested fix:**

Add a comment:
```powershell
# Note: update this list when zip package architectures change
$zipArchs = @("win-arm64", "win-arm64x", "win-x64")
```

qcom/scripts/release_testing/smoke_test_tgz.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# --- Locate the test binaries folder in test_package ---
+test_bin_dir="${test_pkgs_dir}/${test_bin_arch}"
```
huaychou: **[T-5] (Nit)** smoke_test_tgz.sh: inline Python for zip extraction — unzip would be simpler

`python3 -c "import zipfile; ..."` uses inline Python for zip extraction in a shell script. On the target Linux environment, `unzip` is typically pre-installed and simpler.

**Suggested fix:**

Use unzip instead:
```bash
unzip -q "$test_package_zip" -d "$test_pkgs_dir"
```

qcom/scripts/release_testing/release_testing.py
```diff
@@ -0,0 +1,716 @@
...
+        default="3.11,3.12,3.13,3.14",
+        help="Comma-separated Python versions for wheel testing (default: 3.11,3.12,3.13,3.14)",
+    )
+    parser.add_argument(
+        "--only-count-check",
+        action="store_true",
+        help="Only verify that all expected artifacts are present on Artifactory.",
+    )
+    parser.add_argument(
+        "--only-backward-compat",
```
huaychou: **[T-6] (Nit)** release_testing.py: logging calls use f-strings instead of lazy % formatting

Multiple `log.info(f"...")` calls use f-strings, violating the Python logging convention of lazy `%` formatting ([P06]). F-strings evaluate the string even when the log level is not active.

**Suggested fix:**

Use % formatting:
```python
log.info("Platform: %s %s", platform.system(), platform.machine())
log.info("Artifact type: %s", args.artifact_type)
```

qcom/scripts/release_testing/test_nuget_package.ps1
```diff
@@ -0,0 +1,200 @@
...
+    Write-Host "Step 3 PASSED" -ForegroundColor Green
+
+    # =============================================================================
+    # Step 4: Install packages
+    # =============================================================================
+
+    Write-Host ""
+    Write-Host "Step 4: Installing packages..." -ForegroundColor Cyan
+
+    & dotnet add package System.Numerics.Tensors --version 9.0.0
```
huaychou, line 88: Just wondering, why do we need to specify the version of `System.Numerics.Tensors`

.github/workflows/qualcomm-internal-release-testing.yml
huaychou, line 32: Should it be 2.5.0.rc3?

qti-mbadnara: For zips it is 2.5.0rc3

qcom/scripts/release_testing/test_nuget_package.ps1
qti-mbadnara: That is the base environment used to test, ideally, we can use any version, but this is used for replication and resolve any issues which might potentially arise

.github/workflows/qualcomm-internal-release-testing.yml
```diff
@@ -0,0 +1,361 @@
...
+  # ===========================================================================
+  # ARTIFACT COUNT CHECK (runs first — gates all other jobs)
+  # ===========================================================================
+
+  verify-artifact-counts:
+    name: "Verify Artifact Counts"
+    runs-on: [Linux, self-hosted, X64]
+
+    steps:
+      - uses: actions/checkout@v6
```
huaychou, line 53: Do we need with submodules to fetch models under MODEL_PATH?
```suggestion
      - uses: actions/checkout@v6
        with:
          submodules: recursive      
```

qcom/scripts/upleveling/config.ini
```diff
@@ -54,3 +54,19 @@ repository: https://artifactory-qdc-global.qualcomm.com/artifactory/aisw-maven-v
 
 [maven-central]
 repository: https://central.sonatype.com/api/v1/publisher/upload
+
+[artifactory-pypi-public-storage]
+repository: https://artifactory-qdc-global.qualcomm.com/artifactory/aisw-public-pypi-virtual
+cert: certs/artifactory-ca.pem
```
huaychou: I think the cert is defined in `ARTIFACTORY_CERTS_FILE` and the cert is not used.

qcom/scripts/release_testing/verify_windows_metadata.ps1
```diff
@@ -0,0 +1,337 @@
...
+[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
+$OutputEncoding           = [System.Text.Encoding]::UTF8
+
+# =============================================================================
+# Shared helpers
+# =============================================================================
+
+function Normalize-Version([string]$Version) {
+    # Reduce to major.minor.patch — handles four-part FileVersion ("2.5.0.0"),
+    # pre-release suffixes ("2.5.0.rc3"), and plain three-part ("2.5.0").
```
huaychou: If the version is `2.5.0rc`, can Normalize-Version still work?

qcom/scripts/release_testing/test_nuget_package.ps1
huaychou: Acknowledge

qcom/scripts/upleveling/config.ini
qti-mbadnara: Done!

qcom/scripts/release_testing/verify_windows_metadata.ps1
qti-mbadnara: Fixed

.github/workflows/qualcomm-internal-release-testing.yml
qti-mbadnara: Done!

huaychou (approved):
LGTM. Is it possible to trigger the workflow before merging this PR, or it need to be merged first?

tirupath-qti (approved):

PR #448, opened 2026-05-27
[QNN EP] Set ORT Core Version to >=1.24.2

qti-mbadnara:
This PR sets the floor of `onnxruntime` in `requirements.txt` to `>= 1.24.2`. This prevents pinning `ORT Core` to `1.24.4`


qti-yuduo (approved):

tirupath-qti (commented):
pinning a specific ORT version is not acceptable behavior. This should go into 2.1.1, 2.2.0, 2.3.0 as well.

PR #458, opened 2026-05-28
Add ut for qnn backend manager

yuhuchua-qti:
## Summary

Component-level unit tests for `qnn_backend_manager.cc`, added to the `unit/` tier. Two suites, both compiled only under the coverage build (`QNN_EP_INTERNAL_SYMBOL_ACCESS`):

- **`QnnUnit_BackendManagerTest`** (12 tests) — stub-based, load no QNN shared library. Cover `QnnSerializerConfig` construction/configuration, `SetupBackend` error paths (invalid `.so` path), pre-setup early-return paths in `ResetQnnLogLevel` / `GetContextBinaryBuffer`, and `ParseLoraConfig` file-I/O error handling.
- **`QnnUnit_BackendManagerHtpTest`** (24 tests) — load a real backend via the HTP emulator (`libQnnHtp.so`) and IR backend (`libQnnIr.so`); each skips via `GTEST_SKIP()` when the SDK is unavailable. Cover all 8 `ContextPriority` values incl. the `UNDEFINED` error path, device configs (`soc_model`, `htp_arch`), profiling (`BASIC`/`DETAILED`, ETW merge, `SetProfilingLevelETW`), `ResetQnnLogLevel` after setup, context binary buffer (new `GetContextBinaryBuffer` signature), `LoadCachedQnnContextFromBuffer` with invalid buffer, IR backend loaded directly (SERIALIZER type), and Saver / Ir serializer with HTP as validator.

The two-suite split is required: GTest forbids mixing `TEST()` and `TEST_F()` under one suite name, and the two groups differ semantically — the stub group must run on every host (plain CI regression protection), while the HTP group skips gracefully when the SDK is absent.

## Coverage

Measured on Linux x86_64 with the HTP emulator (`qnn_backend_manager.cc`):

| Filter | Lines | Functions |
|---|---|---|
| This PR's tests only (`*QnnUnit_BackendManager*`) | 679/1311 (51.8%) | 72/104 (69.2%) |
| Full suite (all `*Qnn*`, incl. op-level) | **965/1311 (73.6%)** | **88/104 (84.6%)** |

~73.6% is the practical ceiling on x86_64. The remaining ~346 uncovered lines / 16 functions fall into three buckets:

- **Device-only**: VTCM weight sharing, DMA context-binary loading, RPC shared memory, HTP power config, on-chip platform info — require real HTP hardware.
- **Mock-required**: QNN log/profile error switches — would need either a testing `.so` or a production-source test seam, both intentionally avoided under current project policy.
- **Architecture-limited**: DSP backend (no `libQnnDsp.so` in test env), Windows-only DX12 shared-memory paths.

## Notable design decisions

- **Unit tier only — no `integration/` files.** These tests construct `QnnBackendManager` directly (via EP-internal headers) and never create an ORT session, so they are component tests, not pipeline integration tests. The `unit/` tier already sanctions real-`.so` op-validation tests and already named `qnn_backend_manager.cc` as a target, so they belong here. (An earlier revision placed the 24 real-backend tests under `integration/`; they were moved to keep `integration/` strictly public-API + session.)
- **All `libQnnCpu.so`-based tests removed.** The QNN CPU backend is no longer supported in the QNN EP; tests that exercised general behaviour via CPU were rewritten to use HTP.
- **No production source changes.** Test files only — `onnxruntime/core/providers/qnn/` is untouched.
- **`StubApiEnv` helper in `qnn_unit_test_utils.h`.** Bundles zero-init `OrtApi` / `OrtEpApi` / `OrtModelEditorApi` stubs + null logger + `ApiPtrs` view, so a `QnnBackendManager` can be constructed without boilerplate.

## Test plan

- [x] `onnxruntime_provider_test --gtest_filter="QnnUnit_BackendManagerTest.*"` — 12/12 pass
- [x] `onnxruntime_provider_test --gtest_filter="QnnUnit_BackendManagerHtpTest.*"` — 24/24 pass (skip when HTP unavailable)
- [x] Rebased onto latest `main` (post-#575) — rebuilt, still green
- [x] `python qcom/build_and_test.py lint` — clean


qti-chuteng:
### 🟪 Architecture Fit

**[D-2]** `🟠 Concern` `[Arch]` — `integration/qnn_backend_manager_test.cc` (vs `integration/README.md` + `ort_api_test.cc`)
The integration tests **hard-fail** when a library can't load — the header says "missing libraries are treated as CI environment misconfiguration, not 'feature not available'", and each test uses `ASSERT_TRUE(status.IsOK())` to FAIL. This is the opposite of the established practice under the same convention: `integration/README.md` states tests "skip automatically via `GTEST_SKIP()` when the required backend is unavailable", and the sibling `ort_api_test.cc` uses `GTEST_SKIP()` for a missing backend/EP (even where its message says "CI configuration error", it still skips). Two files in the same folder giving opposite behavior for "backend missing" muddies CI signal semantics.
*Better design:* Switch to `GTEST_SKIP()` mirroring `ort_api_test.cc` when a library/backend is unavailable. If fail-fast is genuinely intended, first codify the new policy in `integration/README.md` and update `ort_api_test.cc` to match, so the convention stays singular — and call out the change in the PR description.

### 🔵 Completeness

**[D-1]** `🔴 Blocker` `[Completeness]` — `integration/` + `unit/qnn_backend_manager_test.cc` (vs current-main header)
The PR head (`a4d9f63`) has diverged from main — main carries 11 commits the PR branch lacks, and PR #482 (Flexible Context Binary) changed `GetContextBinaryBuffer`'s signature. PR-branch header: `std::unique_ptr<unsigned char[]> GetContextBinaryBuffer(uint64_t& written_buffer_size)`. Current-main header: `Ort::Status GetContextBinaryBuffer(bool is_multi_soc_buffer, unsigned char** context_buffer, uint64_t& buffer_size)`. The tests call the old form — `auto buffer = manager->GetContextBinaryBuffer(written_size); EXPECT_NE(buffer, nullptr)` (integration `GetContextBinaryBuffer_HTP_AfterSetup_ReturnsValidBuffer`) and `EXPECT_EQ(manager->GetContextBinaryBuffer(written_size), nullptr)` (unit `GetContextBinaryBuffer_BeforeSetup_ReturnsNull`). The tests pass only against the PR's own stale base; merging requires a rebase, after which these two fail to compile and break the provider-test target link. The claimed 74.1% coverage was also measured against the old code.
*Better design:* Rebase onto current main and rewrite the two `GetContextBinaryBuffer` tests for the new signature (pass `bool is_multi_soc_buffer` and the `unsigned char**` out-param; assert on the returned `Ort::Status` and the out buffer). Re-run on a coverage build and refresh the coverage numbers. Confirm remaining calls stay compatible (`LoadCachedQnnContextFromBuffer` gained a defaulted `is_multi_soc_buffer`; `SetupBackend` gained two defaulted params).

**[D-3]** `🔵 Improvement` `[Completeness]` — `integration/qnn_backend_manager_test.cc` vs `integration/ort_api_test.cc`
The new integration file's guard is `#if !defined(ORT_MINIMAL_BUILD) && defined(__linux__) && QNN_EP_INTERNAL_SYMBOL_ACCESS`; the existing `ort_api_test.cc` has only `#if !defined(ORT_MINIMAL_BUILD) && defined(__linux__)`. The difference is itself **correct** — this file includes the EP-internal header `qnn_backend_manager.h` and constructs `QnnBackendManager` directly, so it must require `QNN_EP_INTERNAL_SYMBOL_ACCESS`, whereas `ort_api_test.cc` goes through the public API only. But two files in the same folder now guard differently, and this is effectively a new sub-class of integration test — easy for a future contributor to pick the wrong guard.
*Better design:* Add a row to the guard section of `integration/README.md` ("integration tests that directly construct EP-internal objects must add the `QNN_EP_INTERNAL_SYMBOL_ACCESS` guard"), or a one-line header note spelling out the difference from `ort_api_test.cc`.

qti-chuteng (approved):

PR #459, opened 2026-05-28
Handle SSR in AOT embed_mode=0 flow

qti-hungjuiw:
### Description

- `QnnModel::RecoverFromSSR`: When `graphExecute` returns `QNN_COMMON_ERROR_SYSTEM_COMMUNICATION` (error 1007), the model recovers by:
    - Releasing the stale context (contextFree)
    - Reading the `.bin` file from disk
    - Calling `contextCreateFromBinary` with proper context configs (priority + spill fill)
    - Calling `graphRetrieve` to get a fresh graph handle
    - Updating `graph_info_` in-place (tensor metadata preserved via `GraphInfo::ResetHandles`)
    - Retrying `graphExecute`
- `ORT_ENGINE_ERROR`: In **JIT** or **embed_mode=1** flows (no .bin file to reload from), the SSR error is propagated as `ORT_ENGINE_ERROR` so applications can programmatically detect NPU crashes and recreate the session.
- `QnnMockSSR.dll`: Test mock that returns `QNN_COMMON_ERROR_SYSTEM_COMMUNICATION` on the first         graphExecute call (no real PD reset — portable across all devices).
- 5 test cases covering: single EPContext, multi-partition, weight-sharing, JIT (ORT_ENGINE_ERROR), embed_mode=1 (ORT_ENGINE_ERROR).

### Motivation and Context

- On Qualcomm SoCs, the NPU (HTP) can undergo a Subsystem Restart (SSR) — a PD (Process Domain) crash and recovery that invalidates all existing QNN context and graph handles. Without SSR handling, any `Session.Run()` after an SSR event would fail permanently, requiring the application to fully recreate the session.
- For production AI workloads running continuously on Windows-on-ARM devices (e.g., always-on inference pipelines), this is unacceptable. The AOT `embed_mode=0` flow stores the compiled context binary as an external `.bin` file, which enables transparent recovery: on SSR detection, QNN EP reloads the binary, recreates the context, retrieves the graph, and retries — all invisible to the application.
- For flows where recovery is impossible (JIT compilation or embedded binaries), the new
  `ORT_ENGINE_ERROR` code gives applications a clear signal to distinguish NPU hardware failures from logic errors (`ORT_EP_FAIL`), enabling proper error handling and session recreation.


qti-chuteng:
### 🔴 Critical

**[C-1]** `onnxruntime/core/providers/qnn/builder/qnn_model.cc → RecoverFromSSR()`
The check-release-create sequence is not atomic. In weight-sharing scenarios, multiple `QnnModel` instances sharing the same `Qnn_ContextHandle_t` can each hold their own `graph_exec_mutex_` and call `RecoverFromSSR` concurrently. Both may see `HasContextHandle(old_context) == true` and both call `ReleaseSpecificContextHandle`, causing a double-free. Alternatively, the second model may call `GetQnnContext(0)` while the first is still between release and re-register, getting a null or stale context. SSR is a global NPU event that affects all graphs simultaneously, so concurrent recovery in weight-sharing is not a theoretical race — it is the expected runtime behavior.

```suggestion
// Add a recovery_mutex_ to QnnBackendManager and hold it across the entire
// check → release → create → register sequence, or expose an atomic
// RecoverContext(old_ctx, filepath, ...) method from QnnBackendManager.
```

---

### 🟠 Major

**[M-1]** `onnxruntime/test/providers/qnn/ssr/qnn_mock_ssr.cc — anonymous namespace`
Static initializers call `real_QnnInterface_getProviders` via a function pointer obtained from `GetProcAddress`. If `QnnHtp.dll` is absent or the symbol is missing, the function pointer is null and calling it is undefined behavior — crashing the entire test binary at DLL load time with a hard-to-diagnose error instead of a skipped test.
Add null checks for `qnn_htp.get()`, `addr`, and `real_providerList` after the static init block, and guard both `QnnGraph_execute` and `QnnInterface_getProviders` against a null `real_providerList`.

**[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_model.h → SetContextRecoveryInfo()`
`node_name_` is move-assigned in `SetContextRecoveryInfo` but `RecoverFromSSR` always uses `graph_info_->Name()` and never reads `node_name_`. This dead state member is confusing: it implies `node_name` and `graph_info_->Name()` might differ in some cases, but there is no comment explaining the relationship. Remove `node_name_` and the corresponding parameter from `SetContextRecoveryInfo`, or document why it is stored if it may be needed later.

---

### 🔵 Minor

**[N-1]** `onnxruntime/test/providers/qnn/ssr/qnn_mock_ssr.cc:L1` + `ssr_test.cc:L1`
Both new files carry the Microsoft copyright header. Per team convention [C10], newly added files must use the Qualcomm copyright:
```suggestion
// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
// SPDX-License-Identifier: MIT
```

**[N-2]** `cmake/onnxruntime_unittests.cmake`
The PR branch has a rebase conflict against `origin/main` in this file (commit `d8ede12b5a` cannot be applied). CI merge checks will fail until the conflict is resolved. Please rebase onto the latest main and force-push.

qti-hungjuiw:
> ### 🔴 Critical
> **[C-1]** `onnxruntime/core/providers/qnn/builder/qnn_model.cc → RecoverFromSSR()` The check-release-create sequence is not atomic. In weight-sharing scenarios, multiple `QnnModel` instances sharing the same `Qnn_ContextHandle_t` can each hold their own `graph_exec_mutex_` and call `RecoverFromSSR` concurrently. Both may see `HasContextHandle(old_context) == true` and both call `ReleaseSpecificContextHandle`, causing a double-free. Alternatively, the second model may call `GetQnnContext(0)` while the first is still between release and re-register, getting a null or stale context. SSR is a global NPU event that affects all graphs simultaneously, so concurrent recovery in weight-sharing is not a theoretical race — it is the expected runtime behavior.
> 
> ```
> // Add a recovery_mutex_ to QnnBackendManager and hold it across the entire
> // check → release → create → register sequence, or expose an atomic
> // RecoverContext(old_ctx, filepath, ...) method from QnnBackendManager.
> ```
> 
> ### 🟠 Major
> **[M-1]** `onnxruntime/test/providers/qnn/ssr/qnn_mock_ssr.cc — anonymous namespace` Static initializers call `real_QnnInterface_getProviders` via a function pointer obtained from `GetProcAddress`. If `QnnHtp.dll` is absent or the symbol is missing, the function pointer is null and calling it is undefined behavior — crashing the entire test binary at DLL load time with a hard-to-diagnose error instead of a skipped test. Add null checks for `qnn_htp.get()`, `addr`, and `real_providerList` after the static init block, and guard both `QnnGraph_execute` and `QnnInterface_getProviders` against a null `real_providerList`.
> 
> **[M-2]** `onnxruntime/core/providers/qnn/builder/qnn_model.h → SetContextRecoveryInfo()` `node_name_` is move-assigned in `SetContextRecoveryInfo` but `RecoverFromSSR` always uses `graph_info_->Name()` and never reads `node_name_`. This dead state member is confusing: it implies `node_name` and `graph_info_->Name()` might differ in some cases, but there is no comment explaining the relationship. Remove `node_name_` and the corresponding parameter from `SetContextRecoveryInfo`, or document why it is stored if it may be needed later.
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/test/providers/qnn/ssr/qnn_mock_ssr.cc:L1` + `ssr_test.cc:L1` Both new files carry the Microsoft copyright header. Per team convention [C10], newly added files must use the Qualcomm copyright:
> 
> ```
> ```
> 
> **[N-2]** `cmake/onnxruntime_unittests.cmake` The PR branch has a rebase conflict against `origin/main` in this file (commit `d8ede12b5a` cannot be applied). CI merge checks will fail until the conflict is resolved. Please rebase onto the latest main and force-push.

[C-1] (context_recovery_mutex_)
Fixed. Added context_recovery_mutex_ to QnnBackendManager and hold it across the entire check → release → create → register sequence in RecoverFromSSR(). All models sharing the same backend manager now contend on this mutex: exactly one model performs the actual context recreation while the others fall through to the GetQnnContext(0) reuse path.

[M-1] (null checks in qnn_mock_ssr.cc)                                                                                                                                                                                                                                                                                      Fixed. Moved the static initialization into a StaticInit struct constructor with null checks on qnn_htp.get() and addr. If QnnHtp.dll is absent or the symbol is missing, real_providerList stays null and the DLL loads without crashing. Added null guard on real_providerList in QnnGraph_execute (returns QNN_COMMON_ERROR_GENERAL) and in QnnInterface_getProviders (returns a zeroed-out mock interface with backendId = 0).                                                                                                                                                                                                      

[M-2] (dead node_name_ member)                                                                                                                                                                                                                                                                                              Removed. node_name_ was stored but never read — RecoverFromSSR uses graph_info_->Name() exclusively. Removed the member and the corresponding parameter from SetContextRecoveryInfo(). Updated both call sites in qnn_backend_manager.cc.

[N-1] (copyright headers)
Fixed. Both qnn_mock_ssr.cc and ssr_test.cc now use the Qualcomm copyright header.

[N-2] (rebase conflict)
Resolved. Rebased onto latest main and force-pushed.

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/qnn_model.cc:L442`
`RecoverFromSSR()` clears/rebuilds `qnn_input_infos_`/`qnn_output_infos_` and rewrites the graph handle via `ResetHandles()` + `SetupQnnInputOutput()`, but it is invoked from the part of `ExecuteGraph` that does NOT hold `graph_exec_mutex_` (proactive branch ~L660, retry branch ~L715). `context_recovery_mutex_` only serializes cross-model context create/free, not per-model `qnn_input_infos_` access. This introduces a concurrent writer where previously there were only concurrent readers, racing with the documented multi-threaded `Run()` support on a single session.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -464,144 +564,174 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
+          mutable_data, qnn_output_info.tensor_byte_size,
+          graph_info_->GraphContext(),
+          qnn_outputs.back()));
+    }
 
-    // Acquire mutex before calling QNN APIs to support calling session.Run() from multiple threads.
-    std::lock_guard<std::mutex> lock(graph_exec_mutex_);
+    Qnn_ErrorHandle_t execute_status = QNN_GRAPH_NO_ERROR;
+
+    // If a sibling model sharing the same context already recovered from SSR
```
minfhong-qti: "Sibling model" here is a bit confusing at first glance. I guess it means other `QnnModel` (i.e., other graph) in the same context, right? I think these info must be explicitly commented for better understanding.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -9,6 +9,7 @@
 #include <thread>
 
 #include "QnnOpDef.h"
+#include "HTP/QnnHtpContext.h"
```
minfhong-qti, line 11: Nit. Should be ahead `QnnOpDef.h`.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -464,144 +564,174 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
-    ORT_CXX_RETURN_ON_API_FAIL(api_ptrs_.ort_api.KernelContext_GetInput(context, qnn_input_info.ort_index, &ort_input_tensor));
-    auto ort_tensor_size = TensorDataSize(ort_input_tensor);
-    ORT_CXX_LOG(logger,
-                ORT_LOGGING_LEVEL_VERBOSE,
-                ("Qnn tensor size: " + std::to_string(qnn_input_info.tensor_byte_size) +
-                 " Ort tensor size: " + std::to_string(ort_tensor_size))
-                    .c_str());
-    RETURN_IF_NOT(qnn_input_info.tensor_byte_size == ort_tensor_size,
-                  "ORT Tensor data size does not match QNN tensor data size.");
+    std::vector<Qnn_Tensor_t> qnn_inputs;
```
minfhong-qti: I guess we can wrap all original binding and execution logic in another function and reuse it instead of exploiting the for loop which looks a bit weird.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1820,12 +1820,21 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
     // the graph name from the context binary may not match the EPContext node name
     auto qnn_model = std::make_unique<qnn::QnnModel>(this, api_ptrs_);
     RETURN_IF_ERROR(qnn_model->DeserializeGraphInfoFromBinaryInfo(graphs_info[0], context));
+    // Seed recovery info for embed_mode=0 so ExecuteGraph can reload after SSR.
```
minfhong-qti: Can we iterate `qnn_models` and set info outside this if branch? Then no need to write in two places.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1820,12 +1820,21 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
+    if (!context_bin_filepath.empty()) {
+      qnn_model->SetContextRecoveryInfo(context_bin_filepath, max_spill_fill_size);
+    }
     qnn_models.emplace(node_name, std::move(qnn_model));
   } else {
     for (uint32_t i = 0; i < graph_count; ++i) {
       auto qnn_model = std::make_unique<qnn::QnnModel>(this, api_ptrs_);
       RETURN_IF_ERROR(qnn_model->DeserializeGraphInfoFromBinaryInfo(graphs_info[i], context));
-      qnn_models.emplace(graphs_info[i].graphInfoV1.graphName, std::move(qnn_model));
+      std::string graph_name(graphs_info[i].graphInfoV1.graphName);
```
minfhong-qti: This change seems irrelevant.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -464,144 +564,174 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
 
-    // Acquire mutex before calling QNN APIs to support calling session.Run() from multiple threads.
-    std::lock_guard<std::mutex> lock(graph_exec_mutex_);
+    Qnn_ErrorHandle_t execute_status = QNN_GRAPH_NO_ERROR;
+
+    // If a sibling model sharing the same context already recovered from SSR
+    // (freeing the old context), our handles are stale.  Proactively recover
+    // before calling graphExecute with dangling handles.
+    if (!context_bin_filepath_.empty() &&
+        !qnn_backend_manager_->HasContextHandle(graph_info_->GraphContext())) {
```
minfhong-qti: Should context recovery lock be acquired before checking context existence?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -429,6 +430,105 @@ static Ort::Status BindQnnTensorMemoryToOrtValueMemory(const OrtApi& ort_api,
...
+      RETURN_IF(!cache_file || !cache_file.good(),
+                ("SSR recovery: failed to open context file: " + context_bin_filepath_).c_str());
+      cache_file.seekg(0, cache_file.end);
+      const size_t buffer_size = static_cast<size_t>(cache_file.tellg());
+      RETURN_IF(buffer_size == 0, ("SSR recovery: context binary file is empty: " + context_bin_filepath_).c_str());
+      cache_file.seekg(0, cache_file.beg);
+      auto buffer = std::make_unique<char[]>(buffer_size);
+      cache_file.read(buffer.get(), static_cast<std::streamsize>(buffer_size));
+      RETURN_IF(!cache_file, ("SSR recovery: failed to read context binary: " + context_bin_filepath_).c_str());
+      cache_file.close();
```
minfhong-qti: I believe this part is copied from `onnx_ctx_model_helper.cc`, and I'm wondering whether `QnnBackendManager::ReadContextBinIfValid` is any different.
I prefer to unify how context binary is read from disk into one place as @tzuhsuanwei 's current work on encryption.

minfhong-qti: In addition, can the entire logic of reading and creating context leverages `QnnBackendManager::LoadCachedQnnContextFromBuffer`?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Done. Thanks!

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Done. Thanks!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: Done. Thanks!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: Done. Thanks for pointing this out!

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Done. Thanks for pointing this out!

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Thanks for catching this. The latest implementation now leverages `QnnBackendManager::ReadContextBinIfValid`.

As for `QnnBackendManager::LoadCachedQnnContextFromBuffer`, it will create new `qnn_model` instead of just updating the existing ones. Using it might require temporary `qnn_models` map and additional copy. Let me know if this is really required.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Done. Please let me know if this meets expectation.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -176,19 +173,43 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+  // Returns true if the context was freed (poisoned). Used to detect use-after-free.
+  bool IsContextPoisoned(Qnn_ContextHandle_t context_handle) const {
+    return poisoned_contexts_.find(context_handle) != poisoned_contexts_.end();
+  }
+
+  // Returns the mutex that serializes SSR context recovery across models sharing this backend.
+  std::mutex& GetContextRecoveryMutex() { return context_recovery_mutex_; }
+
+  // Returns the context priority setting for this backend.
+  ContextPriority GetContextPriority() const { return context_priority_; }
```
minfhong-qti: Can this be provided during `SetContextRecoveryInfo`?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -693,6 +686,16 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   // HtpSharedMemoryAllocator allocation cleanup callback.
   std::unordered_map<Qnn_ContextHandle_t, std::shared_ptr<QnnContextHandleRecord>> context_map_;
 
+  // Tracks freed context handles to detect use-after-free (see IsContextPoisoned).
+  std::unordered_set<Qnn_ContextHandle_t> poisoned_contexts_;
```
minfhong-qti: Is this ever set somewhere?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -588,56 +672,118 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
+  {
+    auto ctx = graph_info_->GraphContext();
+    if (qnn_backend_manager_->IsContextPoisoned(ctx)) {
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_ERROR,
+                  ("SSR POISON DETECTED: context handle " +
+                   std::to_string(reinterpret_cast<uintptr_t>(ctx)) +
+                   " for graph " + graph_info_->Name() +
+                   " was freed by sibling recovery. Proactive recovery will follow.")
+                      .c_str());
+    }
```
minfhong-qti: Same question like above in `qnn_backend_manager.h`.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -588,56 +672,118 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
+      RETURN_IF_ERROR(BindAndExecuteGraph(context, logger, execute_status));
+      if (QNN_COMMON_ERROR_SYSTEM_COMMUNICATION == execute_status) {
+        return Ort::Status("NPU crashed again after SSR recovery.", ORT_ENGINE_ERROR);
+      }
+      if (QNN_GRAPH_NO_ERROR == execute_status) {
+        ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING, "SSR recovery succeeded.");
+      }
+    } else {
+      std::ostringstream oss;
+      oss << "NPU crashed. SSR detected. Caused QNN graph execute error. Error code: " << execute_status;
```
minfhong-qti: Should the message here describe there is no SSR handle and thus skip it?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -588,56 +672,118 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_ERROR, oss.str().c_str());
+      return Ort::Status(oss.str().c_str(), ORT_ENGINE_ERROR);
+    }
   }
 
   if (QNN_GRAPH_NO_ERROR != execute_status) {
```
minfhong-qti:
```suggestion
} else if (QNN_GRAPH_NO_ERROR != execute_status) {
```
Nit?

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
```diff
@@ -0,0 +1,674 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Not sure. Since `ssr` folder is created to place mock files, how about moving this file into it as well?
In addition, probably rename to `qnn_ssr_test.cc`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
qti-hungjuiw: Added, thanks!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
qti-hungjuiw: Removed. Thanks!

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Removed. Thanks for pointing this out!

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Good point! I've described that the empty `context_bin` path means JIT or `embed_mode=1`.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Done. Thanks!

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
qti-hungjuiw: Done. Thanks!

minfhong-qti (approved):

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -470,35 +470,118 @@ static Ort::Status BindQnnTensorMemoryToOrtValueMemory(const OrtApi& ort_api,
...
+
+      const auto& qnn_interface = qnn_backend_manager_->GetQnnInterface();
+
+      // Build context configs: priority + spill fill buffer.
+      QnnContext_Config_t priority_config = QNN_CONTEXT_CONFIG_INIT;
+      priority_config.option = QNN_CONTEXT_CONFIG_OPTION_PRIORITY;
+      priority_config.priority = QNN_PRIORITY_NORMAL;
+
+      if (context_priority_ == ContextPriority::LOW)
+        priority_config.priority = QNN_PRIORITY_LOW;
```
qti-chuteng: Now we have 8 context_priority

qti-hungjuiw: Fixed. Thanks!

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -470,35 +471,110 @@ static Ort::Status BindQnnTensorMemoryToOrtValueMemory(const OrtApi& ort_api,
...
 
+Ort::Status QnnModel::BindAndExecuteGraph(OrtKernelContext* context,
+                                          const Ort::Logger& logger,
+                                          Qnn_ErrorHandle_t& execute_status) {
   using namespace qnn::utils;
   auto TensorDataSize = [&ort_api = api_ptrs_.ort_api](auto ort_tensor) -> size_t {
     OrtTensorTypeAndShapeInfo* tensor_type_and_shape = nullptr;
     OrtStatusPtr tensor_status = ort_api.GetTensorTypeAndShape(ort_tensor, &tensor_type_and_shape);
     if (tensor_status != nullptr) {
-      return 0;  // Return 0 on error, will be handled by caller
```
tirupath-qti, line 489: seems the caller handles the error by reading execute_status. But we are not updating execute_status here.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -470,35 +471,110 @@ static Ort::Status BindQnnTensorMemoryToOrtValueMemory(const OrtApi& ort_api,
...
+      return 0;
     }
     size_t length;
     tensor_status = ort_api.GetTensorShapeElementCount(tensor_type_and_shape, &length);
     if (tensor_status != nullptr) {
       ort_api.ReleaseTensorTypeAndShapeInfo(tensor_type_and_shape);
-      return 0;  // Return 0 on error, will be handled by caller
```
tirupath-qti, line 495: same as line 489

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -589,59 +665,110 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
+  // First attempt: bind tensors and execute.
+  Qnn_ErrorHandle_t execute_status = QNN_GRAPH_NO_ERROR;
+  RETURN_IF_ERROR(BindAndExecuteGraph(context, logger, execute_status));
+
+  if (QNN_COMMON_ERROR_SYSTEM_COMMUNICATION == execute_status) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_ERROR,
+                "NPU crashed. SSR detected during QNN graph execute.");
+    if (!context_bin_filepath_.empty()) {
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING, "Attempting SSR recovery.");
+      RETURN_IF_ERROR(RecoverFromSSR(logger));
```
tirupath-qti, line 751: If EP SSR recovery is failed, it should be fallback to application SSR recovery. We should return ORT_DEVICE_RESET error.

onnxruntime/core/providers/qnn/builder/qnn_model.h
```diff
@@ -180,6 +201,11 @@ class QnnModel {
   // Mutex acquired during graph execution to support multi-threaded inference of a single session.
   std::mutex graph_exec_mutex_;
   const ApiPtrs api_ptrs_;
+
+  // SSR recovery state (embed_mode=0 only). An empty filepath disables recovery.
+  std::string context_bin_filepath_;
+  int64_t max_spill_fill_size_ = 0;
```
tirupath-qti, line 207: do we need to store is_multi_soc_buffer?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -470,35 +471,110 @@ static Ort::Status BindQnnTensorMemoryToOrtValueMemory(const OrtApi& ort_api,
...
+    std::lock_guard<std::mutex> recovery_lock(qnn_backend_manager_->GetContextRecoveryMutex());
+
+    if (qnn_backend_manager_->HasContextHandle(old_context)) {
+      // We are the first model to recover from this SSR event.
+      // Free the old (shared) context and create a new one from the binary.
+      qnn_backend_manager_->ReleaseSpecificContextHandle(old_context);
+
+      // Use the unified file I/O helper instead of duplicating the read logic.
+      std::vector<char> buffer;
+      RETURN_IF_ERROR(qnn_backend_manager_->ReadContextBinIfValid(context_bin_filepath_, buffer));
```
tirupath-qti, line 500: We should continue to filemap the ctxbin even after the SSR.

onnxruntime/core/providers/qnn/builder/qnn_model.h
qti-hungjuiw: No worry. Currently, regardless of whether `is_multi_soc_buffer` is true or false, `contextCreateFromBinary` can handle the buffer correctly. Therefore, SSR can recover the context in the same way without performing an additional check on `is_multi_soc_buffer`.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Sure, as discussed offline, the `ORT_DEVICE_RESET` will be accessible after **ORT 1.28**. Will do this in the separate PR: https://github.com/onnxruntime/onnxruntime-qnn/pull/700

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Sure, thank for pointing this out! Will do this on the separating PR:
https://github.com/onnxruntime/onnxruntime-qnn/pull/677

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Hi, the `TensorDataSize` does not invoke QNN API, so it will never encounter SSR and thus do not need to consider specific `QNN_COMMON_ERROR_SYSTEM_COMMUNICATION`.

Please let me know if I missed anything.

PR #464, opened 2026-05-28
[QNN EP] Make graph partitioner aware of multi-NodeUnit fusions

qti-ashwshan:
Make the QNN EP graph partitioner schedule on IQnnNodeGroup (multi-NodeUnit fusions) as the atomic unit, instead of on individual OrtNodeUnits.

  Problem

  The partitioner's BFS understands OrtNodeUnit atomicity but is unaware of IQnnNodeGroup. When an unsupported op lies on the activation path of a fusion's target, BFS can place the members of one fusion into different partitions. At Compile time, QnnModel::ComposeGraph re-runs
  GetQnnNodeGroups on each partition subgraph — with members missing, the pattern fails to match and the fusion silently drops. Depending on the fusion this either degrades to a slower per-op path or aborts compile entirely.

  Worked example — LPBQ MatMul + GroupQueryAttention

  The LPBQ MatMul fusion is built from 3 ONNX nodes grouped into 2 NodeUnits:

  - Scale_DQL NodeUnit — a DequantizeLinear whose inputs are two initializers (per-block uint scale + per-channel float scale).
  - MatMul NodeUnit — the weight DequantizeLinear, the activation DequantizeLinear, the MatMul, and its output QuantizeLinear.

  The fusion's MatMul sits downstream of an unsupported GroupQueryAttention on the activation path. On mainline, BFS does this:

  1. Scale_DQL has in_degree == 0 (initializer-fed) → admitted into the first HTP partition on the opening BFS wave.
  2. BFS walks the activation path forward → reaches GroupQueryAttention → unsupported → defers it and closes the partition.
  3. GroupQueryAttention is exiled to its own CPU partition.
  4. A new HTP partition opens after GroupQueryAttention. The LPBQ MatMul (with its weight DQ and activation DQ absorbed into the same NodeUnit) lands here.
  5. The fusion is now split across two partitions — Scale_DQL in the first, MatMul in the second.

  The session aborts on the first post-GroupQueryAttention LPBQ MatMul, even though the model contains many more.

  - Any multi-NodeUnit fusion is exposed to the same hazard the moment an unsupported op lands upstream of its target.
  - Fusions with a usable per-op fallback (Gelu, LayerNorm with constant γ, ChannelShuffle, …) silently degrade to slower paths — the output is correct, but the fused QNN op never gets built.
  - Fusions whose encoding can only be consumed by the fused builder (LPBQ MatMul/Gemm, where the weight is per-channel block-quantized) cause outright Compile failures.
  - The partitioner is the right place to fix it. It's the only layer that decides partition boundaries, and fixing it generalizes to every existing and future multi-NodeUnit fusion automatically.

  ---
  Approach

  - QnnNodeGroupInfo describes each IQnnNodeGroup: members, target NodeUnit, supported flag, external_in_degree. Every OrtNodeUnit ends up in exactly one group — real fusion or 1-member QnnNodeUnitWrapper — so BFS becomes branch-free.
  - ComputeGroupExternalInDegree counts only edges that cross the group boundary. Intra-group edges and initializer edges don't count.
  - Group-level BFS in CreateSupportedPartitionNodeGroups: when a group is admitted, all member OrtNodes are emitted atomically into the current partition.
  - Cycle detection + demotion in GetSupportedNodes: in the rare case where a multi-member group can't be scheduled atomically (would require a cyclic dependency through an unsupported op), it's demoted to 1-member groups and each member is re-checked via the standalone op
  builder. Supported members run independently on QNN; unsupported ones fall back to CPU.

  What the fix does for the LPBQ example

  - Scale_DQL is no longer scheduled standalone — it's a member of the LPBQ group, not its own BFS unit.
  - The LPBQ group's external_in_degree > 0 because MatMul depends on the activation path through GroupQueryAttention.
  - The group cannot be admitted until its activation predecessors are emitted → it lands in the post-GroupQueryAttention HTP partition, atomically.
  - Scale_DQL, the weight DQ, and MatMul all enter the same partition together → fusion re-detection in ComposeGraph finds every member → ✅ the LPBQ MatMul builder emits the fused QNN op.


minfhong-qti:
Do you have an example that would fail in current partition? Since this change extremely complicates the partition logic, I would prefer to seek for alternative solutions first before modifying it. Thanks!

qti-ashwshan:
> Do you have an example that would fail in current partition? Since this change extremely complicates the partition logic, I would prefer to seek for alternative solutions first before modifying it. Thanks!

This will help - https://github.com/microsoft/onnxruntime/issues/26325

qti-chuteng:
---

### 🟪 Architecture Fit

**[D-2]** `🟠 Concern` `[Arch]` — `qnn_execution_provider.cc` (demotion loop)
After the dry-run topo sort, the code demotes **every** multi-member group with residual `dry_in_degree > 0`. But a group has residual in-degree not only when it's *in* a cycle, but also when it's merely **transitively downstream** of one (its cyclic predecessors never reach zero, so it's never reached). Those downstream fusions are perfectly schedulable once the offending group is demoted — yet they get demoted in the same pass, before the cycle is broken, each losing its fused QNN op. Correctness-safe (members are rechecked standalone), but it directly undercuts the reason this PR exists: one non-convex fusion can knock out an arbitrary number of innocent downstream fusions.
*Better design:* Demote only groups actually in a cycle (SCC detection), or demote one blocked group at a time and re-run the dry run before demoting more. Cycles are rare, so the per-iteration re-check cost is negligible and it preserves downstream fusions.

**[D-3]** `🔵 Improvement` `[Arch]` — `ort_api.cc:436` / `qnn_ep_utils.cc:1996`
`GetOutputNodes` enumerates consumers only from `target_node_`/clip through the unit's own Q nodes — never from a DQ node's output. But `ComputeGroupExternalInDegree` walks the inputs of *all* member nodes including DQ. If a DQ node's output is ever consumed *outside* its own NodeUnit, the input side counts an edge the output side can never decrement → the downstream group's in-degree never reaches 0 → the terminal `ORT_CXX_API_THROW(..., ORT_EP_FAIL)` aborts the **entire** session compile. This mirrors the old node-level BFS (not a regression, and "cannot arise in standard QDQ selection"), but the blast radius is worse than a graceful fallback and this PR adds new routes to the throw (group cycles). The new `if (d_in > 0)` decrement guard is a good addition, but it only masks over-decrement; the dangerous direction (in-degree seeded too high) still throws.
*Better design:* Consider downgrading the terminal mismatch from a throw to a logged error + graceful bail (return best-effort partitions), so a future accounting bug degrades to slower execution rather than a dead session. The error message itself is good — keep it.

### 🟩 Maintainability

**[D-4]** `🔵 Improvement` `[Maint]` — `qnn_execution_provider.cc` (`check_standalone` lambda, ~L1432)
`check_standalone` reaches directly into `qnn::GetOpBuilder(...)->IsOpSupported(...)`, re-implementing what `QnnNodeUnitWrapper::IsSupported` (`qnn_node_group.cc:49-59`) already does. Verified equivalent *today* (the wrapper is a thin `Ort::Status` wrapper with no extra validation), but if the wrapper ever gains validation or QDQ-specific handling the two paths silently diverge — a demoted member's verdict would no longer match a regular 1-member group's. This is the C1.1 "same decision in ≥2 places with no single owner" smell.
*Better design:* On demotion, construct a `QnnNodeUnitWrapper` and call `IsSupported`, or extract a shared helper, so "is this single NodeUnit supported" has exactly one definition.

### 🟩 Simplicity

**[D-5]** `🔵 Improvement` `[Simplicity]` — `qnn_execution_provider.cc` (demotion member loop, ~L1553)
`supported_nodes.erase(std::remove(...))` inside the per-member demotion loop rescans the whole `supported_nodes` vector on each removal. Demotion is rare so the impact is minor, but on large models once a group is demoted it becomes a full scan per member node.
*Better design:* Collect the nodes to remove into an `unordered_set` during demotion and do a single `remove_if` pass over `supported_nodes` at the end.

### 🔵 Completeness

**[D-1]** `🔴 Blocker` `[Completeness]` — `qnn_execution_provider.cc` (`GetSupportedNodes` / `GetMultiSocSupportedNodes` / `GetCapabilityImpl`)
This PR branched from a merge-base (`5153bd48`) that **predates** main's multi-SoC feature (PR #482, `ce2752c8`). The PR base has 0 occurrences of `GetMultiSocSupportedNodes`; current main has 2. After rebasing onto main:
1. **Compile break** — `GetMultiSocSupportedNodes` (main `qnn_execution_provider.cc:1587`) calls `GetSupportedNodes` with the **old 5-arg signature**, but this PR changes it to require 7 args (`groups`, `node_unit_to_group_id`). That call site won't compile.
2. **Silent functional regression** — main's `GetCapabilityImpl` is now an `if/else` (`:2039-2047`): the `enable_multi_soc_ep_context_` branch calls `GetMultiSocSupportedNodes`, which never populates `node_groups`. This PR only threads groups through the single (pre-multi-soc) call site. If the multi-soc branch passes an empty `groups` to `CreateSupportedPartitionNodeGroups`, then `live_group_count == 0`, the BFS queue is empty, and it returns **zero partitions** — a multi-SoC EP-context session offloads nothing to QNN, with no error.

Also folded in here: the diff adds **no test** exercising a fusion split by an unsupported op on the activation path (the LPBQ MatMul scenario the PR describes) — the regression this PR fixes is left uncovered.
*Better design:* Rebase onto current main and explicitly wire the group / `node_unit_to_group_id` plumbing through `GetMultiSocSupportedNodes` (including how demotion interacts with the per-SoC set-intersection). Add a multi-SoC test and a fusion-split regression test.

qti-ashwshan:
> ### 🟪 Architecture Fit
> **[D-2]** `🟠 Concern` `[Arch]` — `qnn_execution_provider.cc` (demotion loop) After the dry-run topo sort, the code demotes **every** multi-member group with residual `dry_in_degree > 0`. But a group has residual in-degree not only when it's _in_ a cycle, but also when it's merely **transitively downstream** of one (its cyclic predecessors never reach zero, so it's never reached). Those downstream fusions are perfectly schedulable once the offending group is demoted — yet they get demoted in the same pass, before the cycle is broken, each losing its fused QNN op. Correctness-safe (members are rechecked standalone), but it directly undercuts the reason this PR exists: one non-convex fusion can knock out an arbitrary number of innocent downstream fusions. _Better design:_ Demote only groups actually in a cycle (SCC detection), or demote one blocked group at a time and re-run the dry run before demoting more. Cycles are rare, so the per-iteration re-check cost is negligible and it preserves downstream fusions.
> 
> **[D-3]** `🔵 Improvement` `[Arch]` — `ort_api.cc:436` / `qnn_ep_utils.cc:1996` `GetOutputNodes` enumerates consumers only from `target_node_`/clip through the unit's own Q nodes — never from a DQ node's output. But `ComputeGroupExternalInDegree` walks the inputs of _all_ member nodes including DQ. If a DQ node's output is ever consumed _outside_ its own NodeUnit, the input side counts an edge the output side can never decrement → the downstream group's in-degree never reaches 0 → the terminal `ORT_CXX_API_THROW(..., ORT_EP_FAIL)` aborts the **entire** session compile. This mirrors the old node-level BFS (not a regression, and "cannot arise in standard QDQ selection"), but the blast radius is worse than a graceful fallback and this PR adds new routes to the throw (group cycles). The new `if (d_in > 0)` decrement guard is a good addition, but it only masks over-decrement; the dangerous direction (in-degree seeded too high) still throws. _Better design:_ Consider downgrading the terminal mismatch from a throw to a logged error + graceful bail (return best-effort partitions), so a future accounting bug degrades to slower execution rather than a dead session. The error message itself is good — keep it.
> 
> ### 🟩 Maintainability
> **[D-4]** `🔵 Improvement` `[Maint]` — `qnn_execution_provider.cc` (`check_standalone` lambda, ~L1432) `check_standalone` reaches directly into `qnn::GetOpBuilder(...)->IsOpSupported(...)`, re-implementing what `QnnNodeUnitWrapper::IsSupported` (`qnn_node_group.cc:49-59`) already does. Verified equivalent _today_ (the wrapper is a thin `Ort::Status` wrapper with no extra validation), but if the wrapper ever gains validation or QDQ-specific handling the two paths silently diverge — a demoted member's verdict would no longer match a regular 1-member group's. This is the C1.1 "same decision in ≥2 places with no single owner" smell. _Better design:_ On demotion, construct a `QnnNodeUnitWrapper` and call `IsSupported`, or extract a shared helper, so "is this single NodeUnit supported" has exactly one definition.
> 
> ### 🟩 Simplicity
> **[D-5]** `🔵 Improvement` `[Simplicity]` — `qnn_execution_provider.cc` (demotion member loop, ~L1553) `supported_nodes.erase(std::remove(...))` inside the per-member demotion loop rescans the whole `supported_nodes` vector on each removal. Demotion is rare so the impact is minor, but on large models once a group is demoted it becomes a full scan per member node. _Better design:_ Collect the nodes to remove into an `unordered_set` during demotion and do a single `remove_if` pass over `supported_nodes` at the end.
> 
> ### 🔵 Completeness
> **[D-1]** `🔴 Blocker` `[Completeness]` — `qnn_execution_provider.cc` (`GetSupportedNodes` / `GetMultiSocSupportedNodes` / `GetCapabilityImpl`) This PR branched from a merge-base (`5153bd48`) that **predates** main's multi-SoC feature (PR #482, `ce2752c8`). The PR base has 0 occurrences of `GetMultiSocSupportedNodes`; current main has 2. After rebasing onto main:
> 
> 1. **Compile break** — `GetMultiSocSupportedNodes` (main `qnn_execution_provider.cc:1587`) calls `GetSupportedNodes` with the **old 5-arg signature**, but this PR changes it to require 7 args (`groups`, `node_unit_to_group_id`). That call site won't compile.
> 2. **Silent functional regression** — main's `GetCapabilityImpl` is now an `if/else` (`:2039-2047`): the `enable_multi_soc_ep_context_` branch calls `GetMultiSocSupportedNodes`, which never populates `node_groups`. This PR only threads groups through the single (pre-multi-soc) call site. If the multi-soc branch passes an empty `groups` to `CreateSupportedPartitionNodeGroups`, then `live_group_count == 0`, the BFS queue is empty, and it returns **zero partitions** — a multi-SoC EP-context session offloads nothing to QNN, with no error.
> 
> Also folded in here: the diff adds **no test** exercising a fusion split by an unsupported op on the activation path (the LPBQ MatMul scenario the PR describes) — the regression this PR fixes is left uncovered. _Better design:_ Rebase onto current main and explicitly wire the group / `node_unit_to_group_id` plumbing through `GetMultiSocSupportedNodes` (including how demotion interacts with the per-SoC set-intersection). Add a multi-SoC test and a fusion-split regression test.

Would it be possible to share in line comments for the same? Would be easier for review. Thanks

johnpaultaken:
@qti-ashwshan   How does this change affect the GPU backend ? Anything we need to be looking closely at ?

qti-ashwshan:
> @qti-ashwshan How does this change affect the GPU backend ? Anything we need to be looking closely at ?

Hi John, yes, it could affect the GPU backend as well since the change involves around the QNN node group fusions and a lot of these fusions work for both GPU and NPU backend so any multi-partition model where these fusions span a boundary should now ideally get better partitions and ensures that members stay in the same split for the fusions to occur. (As of now LPBQ is the only case we have seen which is NPU specific)

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1599,14 +1623,193 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+    const size_t original_count = groups.size();
+    for (size_t i = 0; i < original_count; ++i) {
+      utils::QnnNodeGroupInfo& g = groups[i];
+      if (g.is_defunct) {
+        continue;
+      }
+      if (g.member_node_units.size() <= 1) {
+        continue;
+      }
+      if (dry_in_degree[g.group_id] == 0) {
```
qti-yuduo, line 1944: The dry run is plain Kahn, so a group has residual `dry_in_degree > 0` if it is in a cycle **or anywhere downstream of one** — and this filter is exactly `dry_in_degree != 0 && member_node_units.size() > 1`. One localized cycle therefore dissolves every multi-NodeUnit fusion in its entire descendant cone.

Demoted members are erased from `supported_nodes` and only restored if `check_standalone` passes, and that is `GetOpBuilder(nu.OpType())`, which is nullptr for ops that only exist inside a fusion. `Erf` for instance is not registered in `op_builder_factory.cc`, so demoting a `GeluFusion` anywhere downstream of a cycle pushes the `Erf` node to CPU and splits the partition around it. "Over-demotion is safe" holds for loop termination, not for coverage.

*Fix:* compute SCCs (Tarjan) over the group graph and demote only groups inside a non-trivial SCC. Also worth gating on support: the comment above describes `member -> unsupported non-member -> member`, but the dry run never reads `is_supported`, so a *supported* intervening NodeUnit — harmless, it lands in the same partition — is treated as a cycle.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1569,13 +1572,34 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
   for (const std::unique_ptr<qnn::IQnnNodeGroup>& qnn_node_group : qnn_node_groups) {
     Ort::Status support_status = qnn_node_group->IsSupported(qnn_model_wrapper, logger_);
     const bool supported = support_status.IsOK();
 
     LogNodeSupport(logger_, *qnn_node_group, support_status);
+
+    utils::QnnNodeGroupInfo info;
+    info.group_id = groups.size();
+    info.target_node_unit = qnn_node_group->GetTargetNodeUnit();
+    for (const OrtNodeUnit* nu : qnn_node_group->GetNodeUnits()) {
```
qti-yuduo, line 1789: `member_set` dedupes here, `member_node_units` does not. `UDOQDQFusion` builds `all_nodes_` from `std::map<size_t, const OrtNodeUnit*> input_node_units` keyed by *input index* (`udo_fusion.cc:336`), so one DQ NodeUnit feeding two inputs of the UDO node is pushed twice. Two independent consequences:

- `ComputeGroupExternalInDegree` (`qnn_ep_utils.cc:1919`) walks `member_node_units`, so the duplicated NodeUnit's external input edge is counted twice while the producer's `GetOutputNodes()` yields one edge -> one decrement. In-degree sticks at 1 -> false cycle -> demotion -> `GetOpBuilder(<udo op type>)` is nullptr -> the UDO node is dropped from `supported_nodes`. Demotion also creates *two* 1-member groups for the same NodeUnit while `node_unit_to_group_id` keeps only the last, leaving a live-but-unreachable group -> `num_groups_processed != live_group_count` -> `ORT_CXX_API_THROW` at `qnn_ep_utils.cc:2089`.
- When the duplicated NodeUnit's inputs are initializers (constant UDO weight, no producer) the in-degree is not inflated, the group schedules normally, and `qnn_ep_utils.cc:2036` emits that NodeUnit's nodes **twice** into the partition's node list.

Mainline is unaffected: it computed in-degree per NodeUnit via `GetInputEdgesCount`, which matched `GetOutputNodes`' per-input-slot multiplicity.

*Fix:* push to `member_node_units` only when `member_set.insert(nu).second`. (Fixing `UDOQDQFusion::all_nodes_` alone leaves the invariant unenforced for future fusions.)

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1942,92 +1995,98 @@ std::vector<std::vector<const OrtNode*>> CreateSupportedPartitionNodeGroups(
...
+      ++num_groups_processed;
+      continue;
+    }
 
-    if (!is_node_supported && Contains(supported_group_border, node)) {
-      // An unsupported node on the border will be processed after the current partition node group.
-      nodes_to_process_with_next_group.push_back(node);
+    // A group is supported if IQnnNodeGroup::IsSupported accepted it AND the target OrtNode is not already claimed
+    // by another EP AND our own supported_nodes set contains the target (belt-and-suspenders check).
+    const OrtNode* target_ortnode = &g.target_node_unit->GetNode();
```
qti-yuduo: Mainline evaluated `Node_GetEpName` on each NodeUnit's target node as it was popped. This reads it only from the group's target NodeUnit, while the emission loop below adds every member NodeUnit's nodes unconditionally. In a multi-EP session where an earlier EP claimed a non-target member of a fusion, QNN EP will now claim that node as well.

*Fix:* run the `ep_name` (and `supported_nodes_set`) check over all `member_node_units` and treat the group as unsupported if any member is owned by another EP.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1599,14 +1623,193 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
+                  ("[QNN EP] Demoting un-fittable fusion group (target: " +
+                   g.target_node_unit->OpType() + " '" + g.target_node_unit->Name() +
+                   "'). Members will be rechecked individually.")
+                      .c_str());
+
+      // Remove this group's member OrtNodes from supported_nodes; they will be re-added below if still supported.
+      for (const OrtNodeUnit* member_nu : g.member_node_units) {
+        for (const OrtNode* n : member_nu->GetAllNodesInGroup()) {
+          supported_nodes.erase(std::remove(supported_nodes.begin(), supported_nodes.end(), n),
```
qti-yuduo: `supported_nodes` has one entry per ONNX node, so each demoted member node rescans and shifts the whole vector — O(N) per node inside a double loop. Only on the demotion path, but that path is the whole downstream cone flagged above. (The common path is net faster in this PR — mainline's per-node `std::find` is now a hash set.)

*Fix:* accumulate the nodes to drop in an `unordered_set` and do one `remove_if` after the demotion loop.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1599,14 +1623,193 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+        new_g.target_node_unit = member_nu;
+        new_g.member_node_units.push_back(member_nu);
+        new_g.member_set.insert(member_nu);
+        new_g.is_supported = check_standalone(*member_nu);
+        new_g.is_defunct = false;
+        new_g.external_in_degree = 0;  // Recomputed next iteration.
+
+        node_unit_to_group_id[member_nu] = new_g.group_id;
+
+        if (new_g.is_supported) {
```
qti-yuduo, line 1987: This never updates `unsupported_nodes`. Since the demotion filter ignores `is_supported`, unsupported multi-member groups are demoted too: members that now pass `check_standalone` are added to `supported_nodes` but stay listed as unsupported, and members of a previously-supported group that fail are removed from `supported_nodes` but never added. With `enable_framework_op_trace_` the trace is wrong for exactly the nodes this path touches. Diagnostics only.

*Fix:* erase newly-supported members from `unsupported_nodes` here, and append newly-unsupported ones with the `check_standalone` status message.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1873,6 +1875,56 @@ std::vector<OrtNodeGroup> OrtSelectorManager::GetOrtQDQSelections(const OrtGraph
...
 
+size_t ComputeGroupExternalInDegree(
+    const QnnNodeGroupInfo& group,
+    const std::unordered_map<const OrtNode*, const OrtNodeUnit*>& node_unit_map,
+    const OrtApi& ort_api) {
+  size_t external = 0;
+
+  auto count_external_producers = [&](const OrtNode* node) {
+    size_t num_inputs = 0;
+    if (OrtStatus* s = ort_api.Node_GetNumInputs(node, &num_inputs); s != nullptr) {
```
minfhong-qti, line 1959: Use CXX API and no need of these handlings. I believe `OrtApi` should never be used if CXX API adopted.
Please help revise those existing C API usage as well, thanks!

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1890,49 +1942,50 @@ std::vector<std::vector<const OrtNode*>> CreateSupportedPartitionNodeGroups(
...
+  for (const QnnNodeGroupInfo& g : groups) {
+    if (g.is_defunct) {
       continue;
     }
-
-    size_t degree = node_unit->GetInputEdgesCount(ort_api);
-    in_degree.insert({node_unit->Index(), degree});
-    if (degree == 0) {
-      nodes_to_process.push_back(node);
+    if (group_in_degree[g.group_id] == 0) {
```
minfhong-qti:
```suggestion
    if (!g.is_defunct && group_in_degree[g.group_id] == 0) {
```

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1890,49 +1942,50 @@ std::vector<std::vector<const OrtNode*>> CreateSupportedPartitionNodeGroups(
...
+      queue_current.push_back(g.group_id);
     }
   }
 
   std::vector<const OrtNode*> supported_group{};
-  // The partition node group's border is the aggregate of its nodes' output nodes.
-  InlinedHashSet<const OrtNode*> supported_group_border{};
+  // Group ids of unprocessed downstream groups reachable from the currently in-progress partition. Mirrors the
+  // `supported_group_border` concept in the old BFS but tracks groups instead of OrtNodes.
+  std::unordered_set<size_t> supported_group_border{};
```
minfhong-qti, line 2058: I think no need to mention previous BFS as it could be confusing due to previous code no longer visible.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1942,92 +1995,98 @@ std::vector<std::vector<const OrtNode*>> CreateSupportedPartitionNodeGroups(
...
+    queue_current.pop_front();
 
-    // A node that is already assigned to an EP other than current EP is unsupported.
-    const char* node_ep_name;
-    ORT_CONTINUE_ON_ERROR(ort_api.Node_GetEpName(node, &node_ep_name), ort_api);
-    const bool is_node_supported = ((std::string(node_ep_name).empty() || node_ep_name == ep_type) &&
-                                    std::find(supported_nodes.cbegin(), supported_nodes.cend(), node) != supported_nodes.cend());
+    const QnnNodeGroupInfo& g = groups[gid];
+    if (g.is_defunct) {
+      ++num_groups_processed;
```
minfhong-qti, line 2082: `live_group_count` does not include defunct groups. Wouldn't `num_groups_processed` never be equal to `live_group_count`?

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1890,49 +1942,50 @@ std::vector<std::vector<const OrtNode*>> CreateSupportedPartitionNodeGroups(
...
-  // #inputs from unprocessed nodes (in-degree) per node.
-  std::unordered_map<size_t, size_t> in_degree{};
-  // Nodes that are ready to process.
-  std::deque<const OrtNode*> nodes_to_process{};
-  // Nodes that will be processed when considering the next partition node group.
-  std::deque<const OrtNode*> nodes_to_process_with_next_group{};
+  // Per-group in-degree; indexed by group_id.
+  std::vector<size_t> group_in_degree(groups.size(), 0);
+  size_t live_group_count = 0;
+  for (const QnnNodeGroupInfo& g : groups) {
```
minfhong-qti:
```suggestion
  for (const QnnNodeGroupInfo& group : groups) {
```
Since it will be widely used, prefer avoiding single-char name. And as well for all the others.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1942,92 +1995,98 @@ std::vector<std::vector<const OrtNode*>> CreateSupportedPartitionNodeGroups(
...
-        if (output_reshape_node) {
-          supported_group.push_back(output_reshape_node);
-          supported_group_border.erase(output_reshape_node);
+    // Propagate to downstream groups. For each distinct edge from any member OrtNode to a NodeUnit outside the
+    // group, decrement the downstream group's in-degree. Edge-count semantics match ComputeGroupExternalInDegree:
+    // multiple edges from the group to the same downstream group produce multiple decrements.
+    for (const OrtNodeUnit* member_nu : g.member_node_units) {
+      for (const OrtNode* output_node : member_nu->GetOutputNodes(ort_api)) {
+        auto it_nu = node_unit_map.find(output_node);
+        if (it_nu == node_unit_map.cend()) {
```
minfhong-qti, line 2128: Is it ever expected that a node not in `node_unit_map`? Would simply skipping it here cause any issue later?
Same for find group id below.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1942,92 +1995,98 @@ std::vector<std::vector<const OrtNode*>> CreateSupportedPartitionNodeGroups(
...
-      auto& downstream_node_in_degree = in_degree[downstream_node_unit->Index()];
-      --downstream_node_in_degree;
+        if (is_group_supported) {
+          supported_group_border.insert(downstream_gid);
+        }
 
-      if (downstream_node_in_degree == 0) {
-        nodes_to_process.push_back(downstream_node);
+        auto& d_in = group_in_degree[downstream_gid];
+        if (d_in > 0) {
```
minfhong-qti, line 2146: Should it check `d_in` is not expected to <=0?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1599,14 +1623,193 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+      if (g.is_defunct) {
+        continue;
+      }
+      g.external_in_degree = utils::ComputeGroupExternalInDegree(g, node_unit_map, ort_api);
+    }
+  };
+
+  // Iterative cycle-detection + demotion loop.
+  //
+  // A multi-member fusion is "un-fittable" if there's a path member -> unsupported_non_member -> member, which
```
minfhong-qti, line 1857: Is there existing node group that is possible to generate this pattern?
I think it should be node group's responsibility to check whether the node(s) that will be not fused in the path must as well be supported; otherwise, this node group is in fact useless.

PR #477, opened 2026-06-03
[QNN EP] Gemm BQ Phase 1: support Gemm with block-quantized weight on HTP

huaychou:
### Description

- QNN HTP FullyConnected accepts BW_FLOAT_BLOCK on a 2-D weight [N, K] with block_sizes = {1, block_size} (K is the contraction axis, axis 1). Activation stays 2-D [M, K]; no reshape needed.
- Supports INT4/UINT4/INT8/UINT8 weights with transB=0 and transB=1.

### Motivation and Context

- Adds BW_FLOAT_BLOCK (block-quantized) weight support to the QNN EP Gemm op builder, translating BQ Gemm to QNN FullyConnected. Requires QAIRT >= 2.47.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
@@ -281,22 +281,24 @@ Ort::Status GemmOpBuilder::ProcessInputsForBQGemm(QnnModelWrapper& qnn_model_wra
     const std::string act_name = input_names[0];
     const auto& act_wrapper = qnn_model_wrapper.GetQnnTensorWrapper(act_name);
     const Qnn_DataType_t act_dtype = act_wrapper.GetTensorDataType();
-    if (act_dtype == QNN_DATATYPE_SFIXED_POINT_16 || act_dtype == QNN_DATATYPE_UFIXED_POINT_16) {
-      // Reuse the original DequantizeLinear node's output name for the FP16 tensor.
-      const std::string fp16_name = Ort::ConstNode(&node_unit.GetNode()).GetInputs()[0].GetName();
-      const std::vector<uint32_t> act_shape_2d = act_wrapper.GetTensorDims();
-      QnnTensorWrapper fp16_wrapper(fp16_name, QNN_TENSOR_TYPE_NATIVE,
-                                    QNN_DATATYPE_FLOAT_16, QnnQuantParamsWrapper(),
-                                    std::vector<uint32_t>(act_shape_2d));
-      RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(fp16_wrapper)),
-                    "Failed to add FP16 activation tensor for BQ Gemm.");
-      RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(
-                        utils::UniqueNameGenerator().New(act_name, "_int16_dequantize"),
-                        QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_DEQUANTIZE,
-                        {act_name}, {fp16_name}, {}, do_op_validation),
-                    "Failed to add INT16→FP16 Dequantize node for BQ Gemm activation.");
-      input_names[0] = fp16_name;
-    }
+    // BW_FLOAT_BLOCK FC requires FP16 activation. The only activation dtype reaching this
+    // path through the QDQ selector is INT16 (SFIXED or UFIXED), so anything else is unexpected.
+    RETURN_IF_NOT(act_dtype == QNN_DATATYPE_SFIXED_POINT_16 || act_dtype == QNN_DATATYPE_UFIXED_POINT_16,
+                  "QNN EP: BQ Gemm activation must be INT16-quantized for the BW_FLOAT_BLOCK kernel");
+    // Reuse the original DequantizeLinear node's output name for the FP16 tensor.
+    const std::string fp16_name = Ort::ConstNode(&node_unit.GetNode()).GetInputs()[0].GetName();
+    const std::vector<uint32_t> act_shape_2d = act_wrapper.GetTensorDims();
+    QnnTensorWrapper fp16_wrapper(fp16_name, QNN_TENSOR_TYPE_NATIVE,
+                                  QNN_DATATYPE_FLOAT_16, QnnQuantParamsWrapper(),
+                                  std::vector<uint32_t>(act_shape_2d));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(fp16_wrapper)),
+                  "Failed to add FP16 activation tensor for BQ Gemm.");
+    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(
+                      utils::UniqueNameGenerator().New(act_name, "_int16_dequantize"),
+                      QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_DEQUANTIZE,
+                      {act_name}, {fp16_name}, {}, do_op_validation),
+                  "Failed to add INT16→FP16 Dequantize node for BQ Gemm activation.");
+    input_names[0] = fp16_name;
   }
 
   //
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
@@ -1,13 +1,78 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#include <unordered_map>
+
+#include <gsl/gsl>
+
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
 #include "core/providers/qnn/builder/qnn_model_wrapper.h"
 #include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/providers/qnn/ort_api.h"
 
 namespace onnxruntime {
 namespace qnn {
+
+namespace {
+
+// HTP BQ Gemm/FC: supported weight bitwidths and their block_size divisor constraints.
+const std::unordered_map<uint32_t, int64_t> kHtpGemmBQBitsAndBlockSizeMultipliers{
+    {2, 16}, {4, 8}, {8, 4}};
+
+// Returns BQ weight bitwidth (2/4/8) from an ONNX element data type, or 0 if unsupported.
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
```diff
@@ -153,17 +237,284 @@ Ort::Status GemmOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
...
+                                         output_info.quant_param.Copy(), std::vector<uint32_t>(output_shape));
+      RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(int16_out_wrapper)),
+                    "Failed to add INT16 BQ Gemm output tensor.");
+      RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(
+                        utils::UniqueNameGenerator().New(org_output_name, "_fp16_quantize"),
+                        QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_QUANTIZE,
+                        {fc_fp16_out}, {org_output_name}, {}, do_op_validation),
+                    "Failed to add FP16→INT16 Quantize node for BQ Gemm output.");
+    }
+    // Unquantized (float) output: the FC node already outputs to org_output_name.
```
minfhong-qti: Not sure whether really needs to be fixed since there is a comment explicitly here but FYI by Agnet.

The float (unquantized) graph-output path is broken and untested. `out_tensor_type` (`APP_READ` for graph outputs) is computed at L480 but the FC FP16 output wrapper at L489 always uses `QNN_TENSOR_TYPE_NATIVE`. On the float path `fc_fp16_out == org_output_name`, so a float graph output is created as NATIVE and won't be readable — yet the final commit is titled "fix float output issue". The Conv BQ reference (conv_op_builder.cc L1344-1345, L1380) always builds the graph output as `is_graph_output ? APP_READ : NATIVE`. All 7 tests use quantized output (`AddQDQNodePairWithOutputAsGraphOutput`), so this branch has no coverage.

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
```diff
@@ -69,18 +142,29 @@ Ort::Status GemmOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
...
     RETURN_IF_ERROR(ExplictOpCheck(node_unit));
   }
+
+  OrtNodeAttrHelper node_helper(node_unit);
+  const int64_t trans_b = node_helper.Get("transB", static_cast<int64_t>(0));
+  const float beta = node_helper.Get("beta", 1.0f);
+  const auto& inputs = node_unit.Inputs();
+
+  // Block-quantized weight: translate to QNN FullyConnected with BW_FLOAT_BLOCK weight.
+  {
```
minfhong-qti: The bracket seems redundant.

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
```diff
@@ -69,18 +142,29 @@ Ort::Status GemmOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
...
+    int64_t num_blocks = 0, block_size = 0;
```
minfhong-qti: Why declare this as output parameter of `IsBQGemmWeight` but never used later?

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
huaychou: The test is not available since the QDQ selector of Gemm always has quantized output. We may add this kind of use case in the future.
The issue of tensor type exists, will fix this issue.

huaychou: After discussion, we will support float output in another PR, so assert quantized output now.

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
huaychou: Removed

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
huaychou: Removed

qti-chuteng:
### **[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc:L105`

**BQ path does not validate `transA` — silently produces wrong results when transA=1**

`ProcessInputsForBQGemm` never reads or handles the `transA` attribute. When `transA=1`, activation A has logical shape `[K, M]` but the BQ path treats it as `[M, K]` and feeds it directly to QNN FullyConnected, causing wrong output values silently. The non-BQ path handles transA via `input_trans_flag` and a Transpose node; the BQ path is completely missing this logic. `ExplictOpCheck` also does not reject the `transA=1` BQ case.

Suggested fix — add at the top of `ProcessInputsForBQGemm`:
```cpp
OrtNodeAttrHelper node_helper(node_unit);
RETURN_IF_NOT(node_helper.Get("transA", static_cast<int64_t>(0)) == 0,
              "QNN EP: BQ Gemm does not support transA=1");
```
Also add a negative test asserting `ExpectedEPNodeAssignment::None` for transA=1 BQ Gemm.

---

### **[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc:L436`

**BQ bias dequantization ignores zero_point — wrong values when INT32 bias zp ≠ 0**

The BQ bias dequantization loop computes `fp32 = scale * raw_int32`, but the correct ONNX DequantizeLinear formula is `fp32 = scale * (raw - zero_point)`. `UnpackZeroPoints` is called for the weight but never for the bias. The current test case always sets `bias_zp = 0`, so this bug is never triggered. If a PTQ tool produces a bias with non-zero zero_point (valid per ONNX spec), the BQ Gemm will produce shifted output values silently.

Suggested fix — unpack bias zero_point before the dequantization loop:
```cpp
std::vector<int32_t> bias_zp_values;
if (inputs[2].quant_param.has_value() && inputs[2].quant_param->zero_point != nullptr) {
  ONNXTensorElementDataType zp_type = ONNX_TENSOR_ELEMENT_DATA_TYPE_UNDEFINED;
  RETURN_IF_ERROR(qnn_model_wrapper.UnpackZeroPoints(
      inputs[2].quant_param->zero_point, bias_zp_values, zp_type));
}
// in loop:
const int32_t zp = bias_zp_values.empty() ? 0
    : (bias_zp_values.size() == 1 ? bias_zp_values[0] : bias_zp_values[i]);
const Ort::Float16_t fp16(static_cast<float>(i32_ptr[i] - zp) * scale);
```

---

### **[N-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc:L20`

**`kHtpGemmBQBitsAndBlockSizeMultipliers` uses `std::unordered_map` instead of `InlinedHashMap`**

QNN EP convention ([B3]) requires small maps to use `InlinedHashMap<K, V>` (from the ORT public API) to avoid heap allocation overhead. Replace with:
```cpp
const InlinedHashMap<uint32_t, int64_t> kHtpGemmBQBitsAndBlockSizeMultipliers{
    {2, 16}, {4, 8}, {8, 4}};
```
Also remove the now-unnecessary `#include <unordered_map>`.

---

### **[N-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc:L54`

**`IsBQGemmWeight` does not confirm weight is an initializer before `GetInitializerShape`**

`GetInitializerShape` is called on `weight.quant_param->scale` without first verifying the weight is a static initializer. While the `scale_shape.size() != 2` guard may catch non-initializer cases implicitly, adding an explicit `if (!weight.is_initializer) return false;` guard (symmetric with the check in `ProcessInputsForBQGemm`) makes the intent explicit.

huaychou:
> ### **[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc:L105`
> **BQ path does not validate `transA` — silently produces wrong results when transA=1**
> 
> `ProcessInputsForBQGemm` never reads or handles the `transA` attribute. When `transA=1`, activation A has logical shape `[K, M]` but the BQ path treats it as `[M, K]` and feeds it directly to QNN FullyConnected, causing wrong output values silently. The non-BQ path handles transA via `input_trans_flag` and a Transpose node; the BQ path is completely missing this logic. `ExplictOpCheck` also does not reject the `transA=1` BQ case.
> 
> Suggested fix — add at the top of `ProcessInputsForBQGemm`:
> 
> ```c++
> OrtNodeAttrHelper node_helper(node_unit);
> RETURN_IF_NOT(node_helper.Get("transA", static_cast<int64_t>(0)) == 0,
>               "QNN EP: BQ Gemm does not support transA=1");
> ```
> 
> Also add a negative test asserting `ExpectedEPNodeAssignment::None` for transA=1 BQ Gemm.
> 
> ### **[M-2]** `onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc:L436`
> **BQ bias dequantization ignores zero_point — wrong values when INT32 bias zp ≠ 0**
> 
> The BQ bias dequantization loop computes `fp32 = scale * raw_int32`, but the correct ONNX DequantizeLinear formula is `fp32 = scale * (raw - zero_point)`. `UnpackZeroPoints` is called for the weight but never for the bias. The current test case always sets `bias_zp = 0`, so this bug is never triggered. If a PTQ tool produces a bias with non-zero zero_point (valid per ONNX spec), the BQ Gemm will produce shifted output values silently.
> 
> Suggested fix — unpack bias zero_point before the dequantization loop:
> 
> ```c++
> std::vector<int32_t> bias_zp_values;
> if (inputs[2].quant_param.has_value() && inputs[2].quant_param->zero_point != nullptr) {
>   ONNXTensorElementDataType zp_type = ONNX_TENSOR_ELEMENT_DATA_TYPE_UNDEFINED;
>   RETURN_IF_ERROR(qnn_model_wrapper.UnpackZeroPoints(
>       inputs[2].quant_param->zero_point, bias_zp_values, zp_type));
> }
> // in loop:
> const int32_t zp = bias_zp_values.empty() ? 0
>     : (bias_zp_values.size() == 1 ? bias_zp_values[0] : bias_zp_values[i]);
> const Ort::Float16_t fp16(static_cast<float>(i32_ptr[i] - zp) * scale);
> ```
> 

Resolved

minfhong-qti (approved):

qti-chuteng (approved):

PR #478, opened 2026-06-03
Add explicit relu if encoding zp > 0

yath1:
### Description
This PR adds an explicit relu op if zero point > 0 in output encodings of Conv + Relu 

### Motivation and Context

tldr: fixes YOLOv6 accuracy issue. Previous: 28.9 mAP. After this fix: 42.626 mAP (simulated accuracy is 42.6 as well).

Explanation: 
If we have a model like this Q/DQ (s1, zp1) -> Conv -> Relu -> Q/DQ (s2, zp2). The output of Relu would have a min=0 (or zp2 == 0). EP currently fuses Relu into Conv since the output encodings would have a min of zero which essentially simulates Relu layer.  

But in YoloV6, output of Relu is fed into a concat op as below
image

HTP has a requirement that the concat op's inputs and output encoding match to run on target. So, AIMET updates the encodings at the output of Relu to be equal to other tensors surrounding the concat op. In this scenario, because of this requirement of concat, tensor 2 has a zp > 0. Since EP is fusing Relu into Conv layer, the effective graph does not actually include a Relu layer which causes accuracy drops on-device. 

This PR fixes this by observing the output encodings, and add an explicit Relu layer if zp > 0.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_ep_utils.cc
+++ b/onnxruntime/core/providers/qnn/qnn_ep_utils.cc
@@ -6,6 +6,7 @@
 #include <iostream>
 #include <string>
 
+#include "core/providers/qnn/builder/qnn_utils.h"
 #include "core/providers/qnn/common/inlined_containers.h"
 #include "core/providers/qnn/common/qnn_graph_utils.h"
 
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -1377,17 +1377,69 @@ Ort::Status ConvOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
...
-                                                    output_node_type,
-                                                    std::move(input_names),
-                                                    {output_name},
-                                                    std::move(param_tensor_names),
-                                                    do_op_validation),
-                    "Failed to add node.");
+      // Check if Relu is fused into this Conv and the output encoding allows negatives.
+      // If so, emit a separate Relu (ElementWiseNeuron) after Conv because QNN HTP
+      // does not implicitly apply Relu clamping when the output encoding has offset < 0.
+      const OrtNode* relu_node = node_unit.GetRedundantClipNode();
```
qti-yuduo: This method can potentially return `Clip` -- not just `ReLU`. If we unconditionally eimit plain `ReLU`, are we risking forcing lower bound to 0? Perhaps we should reject finate-bound Clips in the selector so they route to standard Clip op builder?

qti-yuduo: E.g., reject min!=0 / finite-max Clips in qnn_ep_utils.cc ~1253-1286

qti-yuduo: Ideally add a unit test for this, verify it would fail initially.

qti-yuduo (commented):
Nice triage and fix! Left one question to double check.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
yath1: Thanks for the comment. I have resolved this, and have added a generic solution to handle both Relu and Clip. Clip requires further changes since its max value can also have the same is what this PR is solving. 

As a side effect of this change, this would also get applied for other combinations like Gemm + Relu, Add + Relu etc.

qti-yuduo (approved):
LGTM! Also validated this fix also improving YoloV6: on-device mAP (26.7 → 35.7).

PR #481, opened 2026-06-04
Allow float bias with Conv QDQ node group

qti-ashimaj:
### Description
Allow bias to be float when selecting QDQ node group for Conv Op

### Motivation and Context
If the bias is float (non-quantized) while the weights and activations are quantized, the Q and DQ nodes aren't getting properly partitioned into 1 group along with the conv op. As a result, graph finalization error occurs. This change will allow the node group to be formed even when bias is float.

The bias quantization is handled in https://github.com/onnxruntime/onnxruntime-qnn/pull/483

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -177,6 +177,40 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
   return Ort::Status();
 }
 
+// Quantize a float bias using bias_scale = activation_scale * weight_scale.
+static Ort::Status ProcessFloatBias(QnnModelWrapper& qnn_model_wrapper,
+                                    const Ort::Logger& logger,
+                                    const OrtNodeUnitIODef& bias_def,
+                                    const TensorInfo& bias_info,
+                                    gsl::span<const float> weights_scales,
+                                    float activation_scale,
+                                    std::vector<std::string>& input_names) {
+  ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
+              ("Quantizing float bias " + bias_def.name + " using activation_scale * weight_scale[c]").c_str());
+  std::vector<uint8_t> original_bias_data;
+  RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(bias_info.initializer_tensor, original_bias_data));
+  const size_t num_channels = bias_info.shape[0];
+  RETURN_IF_NOT(original_bias_data.size() == num_channels * sizeof(float),
+                "Unexpected bias data size for float bias quantization");
+  std::vector<uint8_t> new_bias_data;
+  std::vector<float> new_scales;
+  std::vector<int32_t> new_offsets;
+  RETURN_IF_ERROR(utils::QuantizeFloatBiasTensor(
+      gsl::make_span<const float>(reinterpret_cast<const float*>(original_bias_data.data()), num_channels),
+      weights_scales, activation_scale, new_bias_data, new_scales, new_offsets));
+  int32_t bias_quant_axis = 0;
+  QnnQuantParamsWrapper quant_params = (weights_scales.size() == 1)
+                                           ? QnnQuantParamsWrapper::PerTensor(new_scales[0], 0)
+                                           : QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, bias_quant_axis);
+  QnnTensorWrapper bias_wrapper(bias_def.name, QNN_TENSOR_TYPE_STATIC, QNN_DATATYPE_SFIXED_POINT_32,
+                                std::move(quant_params), std::vector<uint32_t>(bias_info.shape),
+                                std::move(new_bias_data));
+  RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(bias_wrapper)),
+                "Failed to add bias tensor.");
+  input_names.push_back(bias_def.name);
+  return Ort::Status();
+}
+
 Ort::Status ConvOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
                                          const OrtNodeUnit& node_unit,
                                          const Ort::Logger& logger,
@@ -740,6 +774,62 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
               }
             }
           }
+        }
+      } else if (bias_info.is_initializer && !bias_info.quant_param.IsQuantized()) {
+        // Get activation and weight quantization parameters
+        TensorInfo input0_info = {};
+        TensorInfo input1_info = {};
+        RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input0_info));
+        RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[1], input1_info));
+
+        if (input0_info.quant_param.IsQuantized() && input1_info.quant_param.IsQuantized()) {
+          // Get activation scale (must be per-tensor for Conv)
+          float activation_scale = 1.0f;
+          const auto& act_quant_params = input0_info.quant_param.Get();
+          if (act_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+            activation_scale = act_quant_params.scaleOffsetEncoding.scale;
+          } else if (act_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) {
+            activation_scale = act_quant_params.bwScaleOffsetEncoding.scale;
+          }
+
+          // Get weight scales (per-tensor or per-channel)
+          std::vector<float> weights_scales;
+
+          if (input1_info.quant_param.IsPerTensor()) {
+            // Handle per-tensor quantization (encodings 0 and 2)
+            const auto& weight_quant_params = input1_info.quant_param.Get();
+
+            if (weight_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+              weights_scales.push_back(weight_quant_params.scaleOffsetEncoding.scale);
+            } else if (weight_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) {
+              weights_scales.push_back(weight_quant_params.bwScaleOffsetEncoding.scale);
+            }
+          } else {
+            // Handle per-channel quantization (encodings 1 and 3)
+            const auto& weight_quant_params = input1_info.quant_param.Get();
+
+            if (weight_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_AXIS_SCALE_OFFSET) {
+              if (weight_quant_params.axisScaleOffsetEncoding.scaleOffset != nullptr &&
+                  weight_quant_params.axisScaleOffsetEncoding.numScaleOffsets > 0) {
+                for (size_t i = 0; i < weight_quant_params.axisScaleOffsetEncoding.numScaleOffsets; ++i) {
+                  weights_scales.push_back(weight_quant_params.axisScaleOffsetEncoding.scaleOffset[i].scale);
+                }
+              }
+            } else if (weight_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_AXIS_SCALE_OFFSET) {
+              if (weight_quant_params.bwAxisScaleOffsetEncoding.scales != nullptr &&
+                  weight_quant_params.bwAxisScaleOffsetEncoding.numElements > 0) {
+                for (size_t i = 0; i < weight_quant_params.bwAxisScaleOffsetEncoding.numElements; ++i) {
+                  weights_scales.push_back(weight_quant_params.bwAxisScaleOffsetEncoding.scales[i]);
+                }
+              }
+            }
+          }
+
+          // Safety check to prevent crashes
+          RETURN_IF_NOT(!weights_scales.empty(), "No weight scales found for quantized weights");
+
+          // Quantize a float bias using bias_scale = activation_scale * weight_scale
+          ProcessFloatBias(qnn_model_wrapper, logger, bias_input, bias_info, weights_scales, activation_scale, input_names);
         }
       }
 
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -200,7 +200,7 @@ static Ort::Status ProcessFloatBias(QnnModelWrapper& qnn_model_wrapper,
       weights_scales, activation_scale, new_bias_data, new_scales, new_offsets));
   int32_t bias_quant_axis = 0;
   QnnQuantParamsWrapper quant_params = (weights_scales.size() == 1)
-                                           ? QnnQuantParamsWrapper::PerTensor(new_scales[0], 0)
+                                           ? QnnQuantParamsWrapper::PerTensor(new_scales[0], new_offsets[0])
                                            : QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, bias_quant_axis);
   QnnTensorWrapper bias_wrapper(bias_def.name, QNN_TENSOR_TYPE_STATIC, QNN_DATATYPE_SFIXED_POINT_32,
                                 std::move(quant_params), std::vector<uint32_t>(bias_info.shape),
@@ -591,6 +591,8 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
 
         if (input0_info.quant_param.IsQuantized() && input1_info.quant_param.IsQuantized()) {
           // Get activation scale (must be per-tensor for Conv)
+          RETURN_IF_NOT(input0_info.quant_param.IsPerTensor(/*include_bw*/ true),
+                        "Activation must be per-tensor quantized for Conv 2D");
           float activation_scale = 1.0f;
           const auto& act_quant_params = input0_info.quant_param.Get();
           if (act_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
@@ -632,7 +634,6 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
             }
           }
 
-          // Safety check to prevent crashes
           RETURN_IF_NOT(!weights_scales.empty(), "No weight scales found for quantized weights");
 
           // Check bias quantization type
@@ -784,6 +785,8 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
 
         if (input0_info.quant_param.IsQuantized() && input1_info.quant_param.IsQuantized()) {
           // Get activation scale (must be per-tensor for Conv)
+          RETURN_IF_NOT(input0_info.quant_param.IsPerTensor(/*include_bw*/ true),
+                        "Activation must be per-tensor quantized for Conv 2D");
           float activation_scale = 1.0f;
           const auto& act_quant_params = input0_info.quant_param.Get();
           if (act_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
@@ -825,11 +828,10 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
             }
           }
 
-          // Safety check to prevent crashes
           RETURN_IF_NOT(!weights_scales.empty(), "No weight scales found for quantized weights");
 
           // Quantize a float bias using bias_scale = activation_scale * weight_scale
-          ProcessFloatBias(qnn_model_wrapper, logger, bias_input, bias_info, weights_scales, activation_scale, input_names);
+          RETURN_IF_ERROR(ProcessFloatBias(qnn_model_wrapper, logger, bias_input, bias_info, weights_scales, activation_scale, input_names));
         }
       }
 
--- a/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
@@ -1741,6 +1741,35 @@ bool CheckBiasScaleMatch(float bias_scale, float weights_scale, float activation
   return std::abs(bias_scale - expected_scale) <= tolerance;
 }
 
+Ort::Status QuantizeFloatBiasTensor(gsl::span<const float> float_bias_data,
+                                    gsl::span<const float> weights_scales,
+                                    float activation_scale,
+                                    /*out*/ std::vector<uint8_t>& quantized_bias_bytes,
+                                    /*out*/ std::vector<float>& bias_scales,
+                                    /*out*/ std::vector<int32_t>& bias_offsets) {
+  RETURN_IF_NOT(!float_bias_data.empty(), "Float bias data must not be empty");
+  RETURN_IF_NOT(!weights_scales.empty(), "Weight scales must not be empty");
+  const size_t num_channels = float_bias_data.size();
+  bias_scales.resize(num_channels);
+  bias_offsets.assign(num_channels, 0);
+  // Compute bias_scale = activation_scale * weight_scale and quantize to int32.
+  // If weights_scales has a single element (per-tensor weight), all channels share the same scale.
+  std::vector<int32_t> quantized_bias(num_channels, 0);
+  for (size_t c = 0; c < num_channels; ++c) {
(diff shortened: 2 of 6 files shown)


qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/qnn_ep_utils.cc → OrtConvNodeGroupSelector::Check()`

The semantic change to allow `dq_nodes.size() == 2` (float-bias Conv) is a functional bug fix with a clear regression risk, but the PR diff contains no unit test. Per team convention, a bug-fix PR must include a reproducing test (CI red), a post-fix verification (CI green), and a kill-test fence — temporarily reverting the fix to confirm the test fails. Without a test, a future accidental revert to the default `num_dq_inputs = -1` would go undetected.

Please add two test cases in `onnxruntime/test/providers/qnn/`:
1. **Float bias**: Conv + DQ(input) + DQ(weight), bias as a float initializer (no DQ node) → assert `Check()` returns true
2. **INT32 bias regression**: Conv + DQ(input) + DQ(weight) + DQ(bias, INT32) → confirm existing behavior is unchanged

Record the kill-test result in the PR description's Validation section.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/qnn_ep_utils.cc:L853`

After passing `static_cast<int>(dq_nodes.size())` as `num_dq_inputs`, the "reject if DQ count mismatches" guard inside `CheckQDQNodes` becomes a no-op (the passed value always equals the actual value). If `Check()` is ever called with `dq_nodes.size() < 2`, `CheckQDQNodes` can no longer intercept it, and the subsequent `dq_nodes[0]` at L858 / `dq_nodes[1]` at L859 would be out-of-bounds.

```suggestion
  if (!CheckQDQNodes(graph, ort_api, node, redundant_clip_node, dq_nodes, q_nodes, static_cast<int>(dq_nodes.size()))) {
    return false;
  }

  if (dq_nodes.size() < 2) {
    return false;  // Conv requires at least DQ nodes for input and weight
  }
```

qti-ashimaj:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/qnn_ep_utils.cc → OrtConvNodeGroupSelector::Check()`
> 
> The semantic change to allow `dq_nodes.size() == 2` (float-bias Conv) is a functional bug fix with a clear regression risk, but the PR diff contains no unit test. Per team convention, a bug-fix PR must include a reproducing test (CI red), a post-fix verification (CI green), and a kill-test fence — temporarily reverting the fix to confirm the test fails. Without a test, a future accidental revert to the default `num_dq_inputs = -1` would go undetected.
> 
> Please add two test cases in `onnxruntime/test/providers/qnn/`:
> 
> 1. **Float bias**: Conv + DQ(input) + DQ(weight), bias as a float initializer (no DQ node) → assert `Check()` returns true
> 2. **INT32 bias regression**: Conv + DQ(input) + DQ(weight) + DQ(bias, INT32) → confirm existing behavior is unchanged
> 
> Record the kill-test result in the PR description's Validation section.
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/qnn_ep_utils.cc:L853`
> 
> After passing `static_cast<int>(dq_nodes.size())` as `num_dq_inputs`, the "reject if DQ count mismatches" guard inside `CheckQDQNodes` becomes a no-op (the passed value always equals the actual value). If `Check()` is ever called with `dq_nodes.size() < 2`, `CheckQDQNodes` can no longer intercept it, and the subsequent `dq_nodes[0]` at L858 / `dq_nodes[1]` at L859 would be out-of-bounds.
> 
> ```
>   if (!CheckQDQNodes(graph, ort_api, node, redundant_clip_node, dq_nodes, q_nodes, static_cast<int>(dq_nodes.size()))) {
>     return false;
>   }
> 
>   if (dq_nodes.size() < 2) {
>     return false;  // Conv requires at least DQ nodes for input and weight
>   }
> ```

**[M-1]** TCs have been added in conv lpbq branch
**[N-1]** Agree, added the check

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -850,10 +850,14 @@ bool OrtConvNodeGroupSelector::Check(const OrtGraph* graph, const OrtApi& ort_ap
...
                                      const std::vector<const OrtNode*>& q_nodes) const {
-  if (!CheckQDQNodes(graph, ort_api, node, redundant_clip_node, dq_nodes, q_nodes)) {
+  if (!CheckQDQNodes(graph, ort_api, node, redundant_clip_node, dq_nodes, q_nodes, static_cast<int>(dq_nodes.size()))) {
     return false;
   }
 
+  if (dq_nodes.size() < 2) {
+    return false;  // Conv requires at least DQ nodes for input and weight
+  }
+
```
yath1: dq nodes are positional - Conv can have Input, Output, Weight and Bias Q/DQs. This case would as well pass if only weight and bias have DQ nodes but not input and weight like it says in the PR. Its unfortunate we are relying on the count of DQ nodes in these selectors rather than making it concrete (Not introduced by your PR though). Do you want to take a stab at making it concrete?

qti-ashimaj: Ok, this can be taken up as a refactoring task in another PR. For now, can we merge with this change as this PR is a blocker for lpbq feature enablement.

qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🟠 Concern` `[Arch]` — `qnn_ep_utils.cc:853`
Passing `CheckQDQNodes(..., static_cast<int>(dq_nodes.size()))` makes the internal guard `num_dq_inputs != dq_nodes.size()` (qnn_ep_utils.cc:548) always false — the "every Conv input must be quantized" gate is switched off entirely, not narrowed to "bias may lack a DQ". That gate was load-bearing: `dq_nodes` is built in `GetOrtQDQSelection` (qnn_ep_utils.cc:1355-1373) by pushing only DQ-producing inputs in order and discarding the input index, so the positional contract `dq_nodes[0]=data / [1]=weight / [2]=bias` holds only when every input is quantized. The new `dq_nodes.size() < 2` guard prevents the OOB read but not the mis-mapping. Concrete leak: a Conv with float data + quantized weight (uint16) + quantized bias (int32) yields `dq_nodes = [weight_dq, bias_dq]`, size 2, passes `<2`; `dt_input` then reads the weight (uint16) and `dt_weight` reads the bias (int32), `dt_input==dt_output` passes, the `==3` bias branch is skipped, and with `allow_16bit_` true it returns true — the old code rejected this cleanly. Float-weight is symmetric.
*Better design:* Follow the sibling `OrtClipNodeGroupSelector`: after relaxing the count, restore a positional check — confirm the producers of node input[0] and input[1] are DQ nodes in `dq_nodes` (the only input allowed to lack a DQ is the bias at index 2), or have `GetOrtQDQSelection` preserve each DQ's input slot so type lookups use the slot (data=0/weight=1/bias=2). Add a one-line comment on why this diverges from ORT-core `ConvNodeGroupSelector` (which does not relax this check).

### 🟩 Simplicity

**[D-3]** `🔵 Improvement` `[Simplicity]` — `qnn_ep_utils.cc:857`
The mechanism (reusing Clip/Pad's variable-count idiom) is fine, but the form diverges. `OrtClipNodeGroupSelector` (:693-698) and Pad (:1120-1125) do an explicit two-sided range check *before* calling `CheckQDQNodes` (Clip uses `[1,3]`). This PR inverts it: call first, then a floor-only `< 2` with no ceiling. Not a live bug (Conv can't have 4 quantized inputs today), but the guard doesn't state the legal range `[2,3]`, and the next reader will wonder if the missing upper bound is intentional.
*Better design:* `const size_t n = dq_nodes.size(); if (n < 2 || n > 3) return false;` then a single `CheckQDQNodes(..., static_cast<int>(n))` call. Same line count, states the `[2,3]` range outright, restores the dropped upper bound, aligns with the Clip idiom.

### 🔵 Completeness

**[D-2]** `🟠 Concern` `[Completeness]` — `conv_test.cc`
This PR changes a QDQ node-group selector's behavior (the description explicitly fixes a "float bias -> graph finalization error") with no test at all. The only float-bias QDQ Conv helper in the tree, `BuildQDQConvWithFusedActivationTestCase` (conv_test.cc:1971), was introduced by PR #478 (`71f4859d4c`) for the Relu/Clip path — it is not this PR's test; the existing `ConvU8U8S32_*` tests are all int32-bias. Reverting this PR's `static_cast` argument to `-1` leaves the suite green, so coverage credit is zero.
*Better design:* Add a positive test (QDQ Conv with quantized activation+weight and a float bias forms one group assigned to QNN — assert the group structure, not just a node count) plus a negative test (float data / float weight → `ExpectedEPNodeAssignment::None`), on CPU and HTP. The negative test would catch D-1.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
qti-ashimaj: Updated the conv selector check, refactoring of full file including all node selectors can be deferred to another PR

qti-ashimaj:
> ### 🟪 Architecture Fit
> **[D-1]** `🟠 Concern` `[Arch]` — `qnn_ep_utils.cc:853` Passing `CheckQDQNodes(..., static_cast<int>(dq_nodes.size()))` makes the internal guard `num_dq_inputs != dq_nodes.size()` (qnn_ep_utils.cc:548) always false — the "every Conv input must be quantized" gate is switched off entirely, not narrowed to "bias may lack a DQ". That gate was load-bearing: `dq_nodes` is built in `GetOrtQDQSelection` (qnn_ep_utils.cc:1355-1373) by pushing only DQ-producing inputs in order and discarding the input index, so the positional contract `dq_nodes[0]=data / [1]=weight / [2]=bias` holds only when every input is quantized. The new `dq_nodes.size() < 2` guard prevents the OOB read but not the mis-mapping. Concrete leak: a Conv with float data + quantized weight (uint16) + quantized bias (int32) yields `dq_nodes = [weight_dq, bias_dq]`, size 2, passes `<2`; `dt_input` then reads the weight (uint16) and `dt_weight` reads the bias (int32), `dt_input==dt_output` passes, the `==3` bias branch is skipped, and with `allow_16bit_` true it returns true — the old code rejected this cleanly. Float-weight is symmetric. _Better design:_ Follow the sibling `OrtClipNodeGroupSelector`: after relaxing the count, restore a positional check — confirm the producers of node input[0] and input[1] are DQ nodes in `dq_nodes` (the only input allowed to lack a DQ is the bias at index 2), or have `GetOrtQDQSelection` preserve each DQ's input slot so type lookups use the slot (data=0/weight=1/bias=2). Add a one-line comment on why this diverges from ORT-core `ConvNodeGroupSelector` (which does not relax this check).
> 
> ### 🟩 Simplicity
> **[D-3]** `🔵 Improvement` `[Simplicity]` — `qnn_ep_utils.cc:857` The mechanism (reusing Clip/Pad's variable-count idiom) is fine, but the form diverges. `OrtClipNodeGroupSelector` (:693-698) and Pad (:1120-1125) do an explicit two-sided range check _before_ calling `CheckQDQNodes` (Clip uses `[1,3]`). This PR inverts it: call first, then a floor-only `< 2` with no ceiling. Not a live bug (Conv can't have 4 quantized inputs today), but the guard doesn't state the legal range `[2,3]`, and the next reader will wonder if the missing upper bound is intentional. _Better design:_ `const size_t n = dq_nodes.size(); if (n < 2 || n > 3) return false;` then a single `CheckQDQNodes(..., static_cast<int>(n))` call. Same line count, states the `[2,3]` range outright, restores the dropped upper bound, aligns with the Clip idiom.
> 
> ### 🔵 Completeness
> **[D-2]** `🟠 Concern` `[Completeness]` — `conv_test.cc` This PR changes a QDQ node-group selector's behavior (the description explicitly fixes a "float bias -> graph finalization error") with no test at all. The only float-bias QDQ Conv helper in the tree, `BuildQDQConvWithFusedActivationTestCase` (conv_test.cc:1971), was introduced by PR #478 (`71f4859d4c`) for the Relu/Clip path — it is not this PR's test; the existing `ConvU8U8S32_*` tests are all int32-bias. Reverting this PR's `static_cast` argument to `-1` leaves the suite green, so coverage credit is zero. _Better design:_ Add a positive test (QDQ Conv with quantized activation+weight and a float bias forms one group assigned to QNN — assert the group structure, not just a node count) plus a negative test (float data / float weight → `ExpectedEPNodeAssignment::None`), on CPU and HTP. The negative test would catch D-1.

**[D-1]**  Done
**[D-3]**  Done
**[D-2]**  The TCs are added in the conv lpbq PR

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -741,6 +775,62 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+
+        if (input0_info.quant_param.IsQuantized() && input1_info.quant_param.IsQuantized()) {
+          // Get activation scale (must be per-tensor for Conv)
+          float activation_scale = 1.0f;
+          const auto& act_quant_params = input0_info.quant_param.Get();
+          if (act_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+            activation_scale = act_quant_params.scaleOffsetEncoding.scale;
+          } else if (act_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) {
+            activation_scale = act_quant_params.bwScaleOffsetEncoding.scale;
+          }
```
huaychou, line 796: Could we add an else case or assertion to ensure that act_quant_params.quantizationEncoding are QNN_QUANTIZATION_ENCODING_SCALE_OFFSET or QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET?
Otherwise, the activation_scale will be 1.0f if act_quant_params.quantizationEncoding is invalid.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -741,6 +775,62 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+
+          // Get weight scales (per-tensor or per-channel)
+          std::vector<float> weights_scales;
+
+          if (input1_info.quant_param.IsPerTensor()) {
+            // Handle per-tensor quantization (encodings 0 and 2)
```
huaychou, line 802: What does encodings 0 and 2 mean?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -741,6 +775,62 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+            const auto& weight_quant_params = input1_info.quant_param.Get();
+
+            if (weight_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+              weights_scales.push_back(weight_quant_params.scaleOffsetEncoding.scale);
+            } else if (weight_quant_params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) {
+              weights_scales.push_back(weight_quant_params.bwScaleOffsetEncoding.scale);
+            }
+          } else {
+            // Handle per-channel quantization (encodings 1 and 3)
```
huaychou, line 811: What does encodings 1 and 3 mean? Do we need to consider block quant case?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -741,6 +775,62 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+              if (weight_quant_params.bwAxisScaleOffsetEncoding.scales != nullptr &&
+                  weight_quant_params.bwAxisScaleOffsetEncoding.numElements > 0) {
+                for (size_t i = 0; i < weight_quant_params.bwAxisScaleOffsetEncoding.numElements; ++i) {
+                  weights_scales.push_back(weight_quant_params.bwAxisScaleOffsetEncoding.scales[i]);
+                }
+              }
+            }
+          }
+
+          // Safety check to prevent crashes
```
huaychou: Can be removed

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -741,6 +775,62 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+          RETURN_IF_NOT(!weights_scales.empty(), "No weight scales found for quantized weights");
+
+          // Quantize a float bias using bias_scale = activation_scale * weight_scale
+          ProcessFloatBias(qnn_model_wrapper, logger, bias_input, bias_info, weights_scales, activation_scale, input_names);
```
huaychou:
```suggestion
          RETURN_IF_ERROR(ProcessFloatBias(qnn_model_wrapper, logger, bias_input, bias_info, weights_scales, activation_scale, input_names));
```

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1741,6 +1741,35 @@ bool CheckBiasScaleMatch(float bias_scale, float weights_scale, float activation
...
+                                    /*out*/ std::vector<uint8_t>& quantized_bias_bytes,
+                                    /*out*/ std::vector<float>& bias_scales,
+                                    /*out*/ std::vector<int32_t>& bias_offsets) {
+  RETURN_IF_NOT(!float_bias_data.empty(), "Float bias data must not be empty");
+  RETURN_IF_NOT(!weights_scales.empty(), "Weight scales must not be empty");
+  const size_t num_channels = float_bias_data.size();
+  bias_scales.resize(num_channels);
+  bias_offsets.assign(num_channels, 0);
+  // Compute bias_scale = activation_scale * weight_scale and quantize to int32.
+  // If weights_scales has a single element (per-tensor weight), all channels share the same scale.
```
huaychou, line 1756: Could we check the size of weights_scales first, then resize the bias_scales and bias_offsets?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -177,6 +177,40 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
...
+                "Unexpected bias data size for float bias quantization");
+  std::vector<uint8_t> new_bias_data;
+  std::vector<float> new_scales;
+  std::vector<int32_t> new_offsets;
+  RETURN_IF_ERROR(utils::QuantizeFloatBiasTensor(
+      gsl::make_span<const float>(reinterpret_cast<const float*>(original_bias_data.data()), num_channels),
+      weights_scales, activation_scale, new_bias_data, new_scales, new_offsets));
+  int32_t bias_quant_axis = 0;
+  QnnQuantParamsWrapper quant_params = (weights_scales.size() == 1)
+                                           ? QnnQuantParamsWrapper::PerTensor(new_scales[0], 0)
```
huaychou:
```suggestion
                                           ? QnnQuantParamsWrapper::PerTensor(new_scales[0], new_offsets[0])
```

onnxruntime/test/providers/qnn/unit/qnn_ep_utils_test.cc
```diff
@@ -607,6 +644,73 @@ TEST(QnnUnit_EpUtilsTest, Conv_Rejects3DqWithNonInt32Bias) {
...
+
+  FakeValueInfo q_out{"z", ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8, {}};
+  FakeNode q{"q", "QuantizeLinear", "", 13, {}, {&q_out}};
+
+  // Install producer stub: conv_in1 → dq_wt, conv_in2 → dq_bias.
+  // float_input has no entry → producer = nullptr → positional check fails at slot 0.
+  ConvProducerGuard guard({{conv_in1.AsValueInfo(), dq_wt.AsNode()},
+                           {conv_in2.AsValueInfo(), dq_bias.AsNode()}});
+  ctx.api.ValueInfo_GetValueProducer = &FakeConvProducerStub;
+  // OrtGlobalApiOverride not needed: producer for inputs[0] is nullptr so
```
huaychou, line 705: @yuhuchua-qti, could you help comment if OrtGlobalApiOverride is required or not?

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -367,9 +368,15 @@ static GetTestQDQModelFn<ActivationQType> BuildQDQConvTestCase(
 
     // bias ->
     if (!bias_def.GetShape().empty()) {
-      // Bias requirement taken from python quantization tool: onnx_quantizer.py::quantize_bias_static()
-      const float bias_scale = input_qparams.scale * weights_qparams.scale;
-      conv_input_names.push_back(MakeTestQDQBiasInput(builder, "bias", bias_def, bias_scale, use_contrib_qdq));
+      if (use_float_bias) {
+        QNN_ASSERT(bias_def.IsInitializer() && bias_def.IsRawData());
```
huaychou:
```suggestion
        ASSERT_TRUE(bias_def.IsInitializer() && bias_def.IsRawData()) << "Float bias must be an initializer with raw data";
```

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Added an assert check that activations are per tensor quantized

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: per tensor encodings in Qnn_QuantizationEncoding_t enum items in QAIRT are 0 and 2

image

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: These are the per channel encodings.
Since block quant works on float activations, the conditions required for requantizing bias won't meet

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Done

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Done

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
qti-ashimaj: bias_scales and bias_offsets size depends on bias quantization being per tensor or per channel. The size of weight_scales is not relevant here.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Done

huaychou (approved):
LGTM

onnxruntime/test/providers/qnn/unit/qnn_ep_utils_test.cc
yuhuchua-qti: Confirmed — OrtGlobalApiOverride is genuinely not needed here.

This test intentionally omits slot 0 from ConvProducerGuard, so FakeConvProducerStub returns `producer = nullptr` for inputs[0]. 
At qnn_ep_utils.cc:882, the `producer == nullptr `short-circuit fires before the `Ort::ConstNode(producer).GetOperatorType()` call on the right of the ||, so the C++ wrapper (which would otherwise dispatch through the process-global `Ort::detail::Global::Api()`) is never reached on the fake pointer. 
Selector returns false via the null-producer path — no global api routing required.

The sibling `Conv_AcceptsFloatBias` right above is the mirror case: both slot 0 and slot 1 have real fake DQ producers, so the wrapper is invoked and the override is required. 
The two tests together cover both branches of the `||`.

quic-calvnguy (approved):

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1741,6 +1741,35 @@ bool CheckBiasScaleMatch(float bias_scale, float weights_scale, float activation
   return std::abs(bias_scale - expected_scale) <= tolerance;
 }
 
+Ort::Status QuantizeFloatBiasTensor(gsl::span<const float> float_bias_data,
+                                    gsl::span<const float> weights_scales,
+                                    float activation_scale,
+                                    /*out*/ std::vector<uint8_t>& quantized_bias_bytes,
+                                    /*out*/ std::vector<float>& bias_scales,
+                                    /*out*/ std::vector<int32_t>& bias_offsets) {
+  RETURN_IF_NOT(!float_bias_data.empty(), "Float bias data must not be empty");
```
qti-kromero: nit: I feel like the double negative here is a bit confusing, I think this could just be 

RETURN_IF(float_bias_data.empty(), ...)
...

qti-kromero (approved):

onnxruntime/core/providers/qnn/builder/qnn_utils.cc

PR #483, opened 2026-06-04
[QNN EP] Support LPBQ encodings for Conv Op

qti-ashimaj:
### Description
This PR adds support for the LPBQ (BLOCKWISE_EXPANSION) encoding path in the QNN EP Conv operator builder.

Route block-quantized weights to the HTP LPBQ encodings (`QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION`) when supported enabling the HTP backend to use its native low-precision block quantization kernel.
+ conv_op_builder: Selects the encoding based on the weight and activation datatype and other necessary conditions for LPBQ. Also adds the support for quantization of bias when float bias is present, but weight and activations are quantized.
+ conv_test: Adds lpbq tests for conv op.

### Motivation and Context
Previously, the QNN EP Conv builder only supported the BW_FLOAT_BLOCK path (FP16 activation + block-quantized INT4 weight) alongside per-tensor and per-channel quantization. This PR adds the LPBQ path that is triggered when:

1. The activation is INT16-quantized (uint16 or int16)
2. The weight is 4-bit signed and symmetric and has ONNX block quantization (`block_size` attribute on DQ node)
3. The provider option `enable_block_quant_weight_optimization=1` is set

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -235,61 +235,6 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
   return Ort::Status();
 }
 
-// Quantize a float bias using bias_scale = activation_scale * weight_scale.
-static Ort::Status ProcessFloatBias(QnnModelWrapper& qnn_model_wrapper,
-                                    const Ort::Logger& logger,
-                                    const OrtNodeUnitIODef& bias_def,
-                                    const TensorInfo& bias_info,
-                                    gsl::span<const float> weights_scales,
-                                    float activation_scale,
-                                    std::vector<std::string>& input_names) {
-  ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
-              ("Quantizing float bias " + bias_def.name + " using activation_scale * weight_scale[c]").c_str());
-  std::vector<uint8_t> original_bias_data;
-  RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(bias_info.initializer_tensor, original_bias_data));
-  const size_t num_channels = bias_info.shape[0];
-  RETURN_IF_NOT(original_bias_data.size() == num_channels * sizeof(float),
-                "Unexpected bias data size for float bias quantization");
-  std::vector<uint8_t> new_bias_data;
-  std::vector<float> new_scales;
-  std::vector<int32_t> new_offsets;
-  RETURN_IF_ERROR(utils::QuantizeFloatBiasTensor(
-      gsl::make_span<const float>(reinterpret_cast<const float*>(original_bias_data.data()), num_channels),
-      weights_scales, activation_scale, new_bias_data, new_scales, new_offsets));
-  int32_t bias_quant_axis = 0;
-  QnnQuantParamsWrapper quant_params = (weights_scales.size() == 1)
-                                           ? QnnQuantParamsWrapper::PerTensor(new_scales[0], new_offsets[0])
-                                           : QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, bias_quant_axis);
-  QnnTensorWrapper bias_wrapper(bias_def.name, QNN_TENSOR_TYPE_STATIC, QNN_DATATYPE_SFIXED_POINT_32,
-                                std::move(quant_params), std::vector<uint32_t>(bias_info.shape),
-                                std::move(new_bias_data));
-  RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(bias_wrapper)),
-                "Failed to add bias tensor.");
-  input_names.push_back(bias_def.name);
-  return Ort::Status();
-}
-
-Ort::Status ConvOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
-                                         const OrtNodeUnit& node_unit,
-                                         const Ort::Logger& logger,
-                                         std::vector<std::string>& input_names,
-                                         bool do_op_validation) const {
-  const auto& inputs = node_unit.Inputs();
-  assert(inputs.size() >= 2);
-
-  std::vector<uint32_t> input0_shape;
-  RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[0].shape, input0_shape),
-                "QNN EP: Cannot get shape for first input");
-
-  if (input0_shape.size() == 3) {
-    return ProcessConv1DInputs(qnn_model_wrapper, node_unit, logger, input_names, do_op_validation);
-  } else if (input0_shape.size() == 4 || input0_shape.size() == 5) {
-    return ProcessConv2D3DInputs(qnn_model_wrapper, node_unit, logger, input_names, do_op_validation);
-  }
-
-  return MAKE_EP_FAIL("QNN Conv only supports 3D(rank 5), 2D (rank 4) or 1D (rank 3) inputs.");
-}
-
 // Dequantize INT32 bias to FP16 for BW_FLOAT_BLOCK Conv.
 static Ort::Status ProcessBqFp16Bias(QnnModelWrapper& qnn_model_wrapper,
                                      const OrtNodeUnitIODef& bias_def,
@@ -373,6 +318,40 @@ static Ort::Status ProcessRequantizeBias(QnnModelWrapper& qnn_model_wrapper,
   return Ort::Status();
 }
 
+// Quantize a float bias using bias_scale = activation_scale * weight_scale.
+static Ort::Status ProcessFloatBias(QnnModelWrapper& qnn_model_wrapper,
+                                    const Ort::Logger& logger,
+                                    const OrtNodeUnitIODef& bias_def,
+                                    const TensorInfo& bias_info,
+                                    gsl::span<const float> weights_scales,
+                                    float activation_scale,
+                                    std::vector<std::string>& input_names) {
+  ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
+              ("Quantizing float bias " + bias_def.name + " using activation_scale * weight_scale[c]").c_str());
+  std::vector<uint8_t> original_bias_data;
+  RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(bias_info.initializer_tensor, original_bias_data));
+  const size_t num_channels = bias_info.shape[0];
+  RETURN_IF_NOT(original_bias_data.size() == num_channels * sizeof(float),
+                "Unexpected bias data size for float bias quantization");
+  std::vector<uint8_t> new_bias_data;
+  std::vector<float> new_scales;
+  std::vector<int32_t> new_offsets;
+  RETURN_IF_ERROR(utils::QuantizeFloatBiasTensor(
+      gsl::make_span<const float>(reinterpret_cast<const float*>(original_bias_data.data()), num_channels),
+      weights_scales, activation_scale, new_bias_data, new_scales, new_offsets));
+  int32_t bias_quant_axis = 0;
+  QnnQuantParamsWrapper quant_params = (weights_scales.size() == 1)
+                                           ? QnnQuantParamsWrapper::PerTensor(new_scales[0], new_offsets[0])
+                                           : QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, bias_quant_axis);
+  QnnTensorWrapper bias_wrapper(bias_def.name, QNN_TENSOR_TYPE_STATIC, QNN_DATATYPE_SFIXED_POINT_32,
+                                std::move(quant_params), std::vector<uint32_t>(bias_info.shape),
+                                std::move(new_bias_data));
+  RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(bias_wrapper)),
+                "Failed to add bias tensor.");
+  input_names.push_back(bias_def.name);
+  return Ort::Status();
+}
+
 Ort::Status ConvOpBuilder::ProcessConvBias(QnnModelWrapper& qnn_model_wrapper,
                                            const Ort::Logger& logger,
                                            const std::vector<OrtNodeUnitIODef>& inputs,
@@ -419,6 +398,27 @@ Ort::Status ConvOpBuilder::ProcessConvBias(QnnModelWrapper& qnn_model_wrapper,
 
   // Process bias normally: non-initializer, or activation/weight not quantized, or scales already match.
   return ProcessInput(qnn_model_wrapper, bias_def, logger, input_names);
+}
+
+Ort::Status ConvOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
+                                         const OrtNodeUnit& node_unit,
+                                         const Ort::Logger& logger,
+                                         std::vector<std::string>& input_names,
+                                         bool do_op_validation) const {
+  const auto& inputs = node_unit.Inputs();
+  assert(inputs.size() >= 2);
+
+  std::vector<uint32_t> input0_shape;
+  RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[0].shape, input0_shape),
+                "QNN EP: Cannot get shape for first input");
+
+  if (input0_shape.size() == 3) {
+    return ProcessConv1DInputs(qnn_model_wrapper, node_unit, logger, input_names, do_op_validation);
+  } else if (input0_shape.size() == 4 || input0_shape.size() == 5) {
+    return ProcessConv2D3DInputs(qnn_model_wrapper, node_unit, logger, input_names, do_op_validation);
+  }
+
+  return MAKE_EP_FAIL("QNN Conv only supports 3D(rank 5), 2D (rank 4) or 1D (rank 3) inputs.");
 }
 
 Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrapper,
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -13,6 +13,12 @@
 namespace onnxruntime {
 namespace qnn {
 
+namespace {
+inline bool IsQuant16bit(Qnn_DataType_t qnn_data_type) {
+  return qnn_data_type == QNN_DATATYPE_UFIXED_POINT_16 || qnn_data_type == QNN_DATATYPE_SFIXED_POINT_16;
+}
+}  // namespace
+
 // ONNX convolution types supported by this builder.
 // We translate node_unit.OpType() into this enum to avoid repeated string comparisons.
 enum class OnnxConvType {
@@ -248,6 +254,10 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
   // Input 0
   //
   RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[0], logger, input_names));
+  const std::string act_name = input_names[0];  // activation (input 0)
+  const auto& act_wrapper = qnn_model_wrapper.GetQnnTensorWrapper(act_name);
+  const Qnn_DataType_t act_dtype = act_wrapper.GetTensorDataType();
+  const bool is_act_16bit = act_dtype == IsQuant16bit(act_dtype);
 
   // Detect block-quantized weight. Per ONNX opset 21, the scale rank equals the weight rank
   // with scale_shape[1] < weight_shape[1] (the blocked IC axis). Weight is always NCHW
@@ -272,14 +282,16 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
   TensorInfo input_info = {};
   RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[1], input_info));
 
-  if (is_bq_weight && !input_info.quant_param.IsLPBQ()) {
+  const bool is_lpbq_weight = IsNpuBackend(qnn_model_wrapper.GetQnnBackendType()) &&
+                              is_act_16bit &&
+                              input_info.is_initializer &&
+                              input_info.quant_param.IsLPBQ();
+
+  if (is_bq_weight && !is_lpbq_weight) {
     RETURN_IF(!input_info.is_initializer, "QNN EP: BQ Conv weight must be a constant initializer");
 
     // Activation handling: BQ kernel (BW_FLOAT_BLOCK) requires FP16, so INT16 → FP16 via Dequantize.
-    const std::string act_name = input_names[0];  // activation (input 0), pushed by ProcessInput above
-    const auto& act_wrapper = qnn_model_wrapper.GetQnnTensorWrapper(act_name);
-    const Qnn_DataType_t act_dtype = act_wrapper.GetTensorDataType();
-    if (act_dtype == QNN_DATATYPE_SFIXED_POINT_16 || act_dtype == QNN_DATATYPE_UFIXED_POINT_16) {
+    if (is_act_16bit) {
       // Reuse the original DequantizeLinear output name for the FP16 tensor so the QNN graph
       // stays aligned with the ONNX graph naming.
       const std::string fp16_act_name = Ort::ConstNode(&node_unit.GetNode()).GetInputs()[0].GetName();
@@ -341,10 +353,10 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
     RETURN_IF_ERROR(bq::ComputeBQOffsets(qnn_model_wrapper, inputs[1].quant_param->zero_point,
                                          is_unsigned_weight, bitwidth, OC * nb, offsets_qnn));
 
-      QnnQuantParamsWrapper bq_quant_params = QnnQuantParamsWrapper::BwFloatBlock(gsl::span<const float>(scales_qnn),
-                                                                                  gsl::span<const float>(offsets_qnn),
-                                                                                  bitwidth,
-                                                                                  gsl::span<const uint32_t>(block_size_arr));
+    QnnQuantParamsWrapper bq_quant_params = QnnQuantParamsWrapper::BwFloatBlock(gsl::span<const float>(scales_qnn),
+                                                                                gsl::span<const float>(offsets_qnn),
+                                                                                bitwidth,
+                                                                                gsl::span<const uint32_t>(block_size_arr));
 
     // Always use SFIXED_POINT_8: unsigned types are pre-converted by TransformUnsignedToSignedFixedPoint.
     QnnTensorWrapper bq_weight_wrapper(input1_name, tensor_type,
@@ -375,7 +387,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
     }
 
     bool is_3d = (input_info.shape.size() == 5);
-    RETURN_IF(is_3d && input_info.quant_param.IsLPBQ(), "LPBQ is only supported for Conv2d (rank-4 weights)");
+    RETURN_IF(is_3d && is_lpbq_weight, "LPBQ is only supported for Conv2d (rank-4 weights)");
 
     std::vector<uint8_t> unpacked_tensor;
     if (input_info.is_initializer) {
@@ -389,7 +401,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
       }
 
       // Transpose quantization parameter's axis if this is using per-channel or LPBQ quantization.
-      if (input_info.quant_param.IsPerChannel() || input_info.quant_param.IsLPBQ()) {
+      if (input_info.quant_param.IsPerChannel() || is_lpbq_weight) {
         std::vector<size_t> perm;
         if (is_3d) {
           perm = conv_type == OnnxConvType::kConv ? nchw2hwcn_perm_3d : cnhw2hwcn_perm_3d;
@@ -402,7 +414,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
       }
     } else {
       // Add transpose node above weight input.
-      RETURN_IF(input_info.quant_param.IsPerChannel() || input_info.quant_param.IsLPBQ(),
+      RETURN_IF(input_info.quant_param.IsPerChannel() || is_lpbq_weight,
                 "Non-constant Conv inputs only support per-tensor quantization");
       bool is_graph_input = qnn_model_wrapper.IsGraphInput(input1_name);
       ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, ("Add HWCN Transpose node after input: " + input1_name).c_str());
@@ -498,7 +510,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
   if (has_bias_input) {
     // For BQ Conv: QNN BW_FLOAT_BLOCK Conv2d requires FP16 bias matching the computation precision.
     // The bias in the QDQ NodeUnit is an INT32 quantized initializer — dequantize it to FP16.
-    if (is_bq_weight && !input_info.quant_param.IsLPBQ()) {
+    if (is_bq_weight && !is_lpbq_weight) {
       TensorInfo bias_info = {};
       RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[2], bias_info));
       RETURN_IF(!bias_info.is_initializer, "QNN EP: BQ Conv bias must be a constant initializer");
@@ -570,16 +582,16 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
             activation_scale = act_quant_params.bwScaleOffsetEncoding.scale;
           }
 
+          int32_t bias_quant_axis = 0;
           std::vector<float> weights_scales;
-          RETURN_IF_ERROR(utils::GetWeightQuantScales(input1_info.quant_param, bias_info.shape[0], weights_scales));
+          RETURN_IF_ERROR(utils::GetWeightQuantScales(input1_info.quant_param, weights_scales));
           RETURN_IF_NOT(!weights_scales.empty(), "No weight scales found for bias quantization");
 
           if (bias_info.quant_param.IsQuantized()) {
             // Bias is already quantized: check if scales match, requantize if needed.
             std::vector<float> current_scales;
             std::vector<int32_t> current_offsets;
-            int32_t quant_axis = 0;
-            RETURN_IF_ERROR(utils::GetBiasQuantScalesAndOffsets(bias_info.quant_param, current_scales, current_offsets, quant_axis));
+            RETURN_IF_ERROR(utils::GetBiasQuantScalesAndOffsets(bias_info.quant_param, current_scales, current_offsets, bias_quant_axis));
 
             const size_t num_channels = current_scales.size();
             bool needs_requantization = false;
@@ -600,7 +612,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
               std::vector<uint8_t> requantized_bias_data;
               std::vector<float> new_scales;
               std::vector<int32_t> new_offsets;
-              const std::optional<int64_t> axis_opt = (num_channels > 1) ? std::optional<int64_t>(quant_axis) : std::nullopt;
+              const std::optional<int64_t> axis_opt = (num_channels > 1) ? std::optional<int64_t>(bias_quant_axis) : std::nullopt;
               RETURN_IF_ERROR(utils::RequantizeBiasTensor(
                   original_bias_data, bias_info.shape, current_scales, current_offsets,
                   weights_scales, activation_scale, bias_info.qnn_data_type,
@@ -610,7 +622,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
               if (new_scales.size() == 1) {
                 new_quant_params = QnnQuantParamsWrapper::PerTensor(new_scales[0], new_offsets[0]);
               } else {
-                new_quant_params = QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, quant_axis, false);
+                new_quant_params = QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, bias_quant_axis);
               }
 
               std::string bias_name = bias_input.name;
@@ -647,7 +659,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
             if (weights_scales.size() == 1) {
               new_quant_params = QnnQuantParamsWrapper::PerTensor(new_scales[0], 0);
             } else {
-              new_quant_params = QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, /*axis=*/0, /*is_int4=*/false);
+              new_quant_params = QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, bias_quant_axis);
             }
 
             std::string bias_name = bias_input.name;
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -17,6 +17,47 @@ namespace {
 inline bool IsQuant16bit(Qnn_DataType_t qnn_data_type) {
   return qnn_data_type == QNN_DATATYPE_UFIXED_POINT_16 || qnn_data_type == QNN_DATATYPE_SFIXED_POINT_16;
 }
+
+// Extracts the per-tensor activation scale from a QNN quant params wrapper.
+// Returns 1.0f for unrecognized encodings (treated as unquantized).
+inline float GetActivationScale(const QnnQuantParamsWrapper& quant_params) {
+  const auto& qp = quant_params.Get();
+  if (qp.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+    return qp.scaleOffsetEncoding.scale;
+  }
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -563,7 +528,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
   if (has_bias_input) {
     // For BQ Conv: QNN BW_FLOAT_BLOCK Conv2d requires FP16 bias matching the computation precision.
     // The bias in the QDQ NodeUnit is an INT32 quantized initializer — dequantize it to FP16.
-    if (is_bq_weight) {
+    if (is_bq_weight && !input_info.quant_param.IsLPBQ()) {
```
huaychou: **[M-1] (Major)** Reads moved-from input_info.quant_param (use-after-move); bias-branch LPBQ check is correct only by POD coincidence

In the non-BQ weight branch (which is also the branch LPBQ weights take), line 477 `std::move`s `input_info.quant_param` into the `QnnTensorWrapper`. The bias section then reads the same object at line 531: `if (is_bq_weight && !input_info.quant_param.IsLPBQ())`. For an LPBQ Conv with bias (`is_bq_weight==true` and `IsLPBQ()==true`), control enters the else branch and hits the move at 477, after which line 531 calls `IsLPBQ()` on the moved-from object.

`QnnQuantParamsWrapper` uses a defaulted move (header lines 26-27). Its member `Qnn_QuantizeParams_t params_` is a POD union; a defaulted move byte-copies it and leaves the source unchanged, so `IsLPBQ()`, which reads only the scalar `encodingDefinition`/`quantizationEncoding` fields, happens to return the correct value today. However, the same moved-from `params_` still holds a `blockwiseExpansion` raw pointer that now aliases a buffer owned by the destination object.

**Impact:** This is an access to a moved-from object whose correctness rests purely on the implementation detail that a POD move does not clear the source. If anyone changes the move to reset `params_` (e.g. set the moved-from `encodingDefinition` to UNDEFINED), an LPBQ Conv with bias would mis-detect `IsLPBQ()==false`, wrongly take the BQ INT32→FP16 bias dequant path, and produce silently wrong results. Fragile and regression-prone.

**Suggested fix:**

Cache the predicate into a local before the move at line 477 and use it everywhere afterwards instead of touching input_info.quant_param again. e.g. right after GetTensorInfo (~line 273) add const bool weight_is_lpbq = input_info.quant_param.IsLPBQ(); and replace both input_info.quant_param.IsLPBQ() reads at lines 275 and 531 with weight_is_lpbq.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1630,6 +1630,121 @@ bool CheckBiasScaleMatch(float bias_scale, float weights_scale, float activation
...
+                  "Invalid BW_AXIS_SCALE_OFFSET weight quant params");
+    for (size_t i = 0; i < qp.bwAxisScaleOffsetEncoding.numElements; ++i) {
+      weights_scales.push_back(qp.bwAxisScaleOffsetEncoding.scales[i]);
+    }
+  } else if (qp.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION) {
+    RETURN_IF_NOT(qp.blockwiseExpansion != nullptr &&
+                      qp.blockwiseExpansion->scaleOffsets != nullptr,
+                  "Invalid BLOCKWISE_EXPANSION weight quant params");
+    for (size_t c = 0; c < num_output_channels; ++c) {
+      weights_scales.push_back(qp.blockwiseExpansion->scaleOffsets[c].scale);
```
huaychou, line 1774: **[M-2] (Major)** GetWeightQuantScales LPBQ branch reads scaleOffsets bounded only by caller-supplied num_output_channels, with no self-consistent length guard (potential OOB read)

The BLOCKWISE_EXPANSION (LPBQ) branch reads the array with `for (size_t c = 0; c < num_output_channels; ++c) weights_scales.push_back(qp.blockwiseExpansion->scaleOffsets[c].scale);`, where `num_output_channels` is supplied by the caller (conv_op_builder passes `bias_info.shape[0]`). It only checks `scaleOffsets != nullptr`; it never verifies the array actually has >= num_output_channels elements.

Contrast the sibling branches: AXIS_SCALE_OFFSET loops to `numScaleOffsets` and BW_AXIS_SCALE_OFFSET loops to `numElements` — both bound by the encoding's own count. Only the LPBQ branch trusts an externally-supplied channel count.

**Impact:** If the weight's LPBQ scaleOffsets element count does not match bias_shape[0] (e.g. a rewritten graph, or a bias whose OC does not line up with the weight), this is a heap out-of-bounds read that may crash or read a garbage scale and silently produce wrong bias quantization.

**Suggested fix:**

Bound the loop by the encoding's own element count and assert it equals num_output_channels first — read the relevant count field on qp.blockwiseExpansion (per QnnTypes), RETURN_IF_NOT(count == num_output_channels, ...) then loop — so this branch is self-consistent like the others.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1630,6 +1630,121 @@ bool CheckBiasScaleMatch(float bias_scale, float weights_scale, float activation
...
+  bias_scales.resize(num_channels);
+  bias_offsets.assign(num_channels, 0);
+
+  // Compute bias_scale = activation_scale * weight_scale and quantize to int32.
+  // If weights_scales has a single element (per-tensor weight), all channels share the same scale.
+  std::vector<int32_t> quantized_bias(num_channels, 0);
+  for (size_t c = 0; c < num_channels; ++c) {
+    const float weight_scale = (c < weights_scales.size()) ? weights_scales[c] : weights_scales[0];
+    bias_scales[c] = activation_scale * weight_scale;
+    if (bias_scales[c] > 0.0f) {
```
huaychou: **[N-1] (Minor)** QuantizeFloatBiasTensor silently quantizes a channel to 0 when bias_scale <= 0, with no error or log

`if (bias_scales[c] > 0.0f) { quantized_bias[c] = round(float_bias_data[c] / bias_scales[c]); }` — when `bias_scales[c] <= 0` (i.e. activation_scale or weight_scale is 0/negative) the channel's quantized value stays 0. This avoids a divide-by-zero but silently swallows degenerate input.

**Impact:** If an upstream degenerate encoding yields scale=0, the bias is silently zeroed rather than reported, making it harder to spot during later debugging.

**Suggested fix:**

For a non-positive bias_scale, surface it via RETURN_IF_NOT / MAKE_EP_FAIL (or at minimum ORT_CXX_LOG a WARNING for that channel) instead of silently zeroing.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -630,187 +598,103 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+            RETURN_IF_ERROR(utils::QuantizeFloatBiasTensor(
+                gsl::span<const float>(bias_float_data, num_channels),
+                weights_scales, activation_scale,
+                quantized_bias_data, new_scales, new_offsets));
+
+            QnnQuantParamsWrapper new_quant_params;
+            if (weights_scales.size() == 1) {
+              new_quant_params = QnnQuantParamsWrapper(new_scales[0], 0);
+            } else {
+              new_quant_params = QnnQuantParamsWrapper(new_scales, new_offsets, /*axis=*/0, /*is_int4=*/false);
```
huaychou: **[T-1] (Nit)** Inconsistent per-channel axis source between the already-quantized requantize path and the float-bias quantize path

The already-quantized requantize branch builds the per-channel `QnnQuantParamsWrapper` with the `quant_axis` returned by `GetBiasQuantScalesAndOffsets` (line 641), while the float-bias branch hardcodes axis 0 (line 678). For a bias of shape [OC], axis 0 is correct in both, so there is no functional difference today; but using different sources across the two paths can mislead future maintainers.

**Suggested fix:**

Either add a comment in the float-bias branch noting 'bias is 1-D [OC], so the quant axis is always 0', or unify how both paths obtain the axis, to reduce cognitive load.

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -638,6 +638,169 @@ static void RunHTPConvOpPerChannelTest(const std::string& conv_op_type, const Te
   TestQDQModelAccuracy(f32_fn, qdq_fn, provider_options, opset, expected_ep_assignment, tolerance);
 }
 
+template <typename ActivationQType, typename WeightQType>
+static GetTestQDQModelFn<ActivationQType> BuildQDQBlockQuantConv2DTestCase(
```
huaychou: Could you leverage BuildBQConvTestCase to build test cases?

huaychou:
The title and the description are quite different from the change. Could you help update them?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Added a const bool flag to decide if weight is of type lpbq

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
qti-ashimaj: The qnn struct Qnn_BlockwiseExpansion_t doesn't have any element count member, hence the number of channels were used previously. Added a getter in qnn_quant_params_wrapper to get the count

onnxruntime/core/providers/qnn/builder/qnn_utils.cc

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Declared an axis variable at the beginning

onnxruntime/test/providers/qnn/conv_test.cc
qti-ashimaj: The current BuildBQConvTestCase only support uint16 activations. Also, they do not have any provision for providing bias as float instead of quantized. Lmk if it's fine to add these functionalities without breaking BQ TCs, I can use the same builder for BQ and LPBQ TCs. @huaychou

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1741,6 +1741,120 @@ bool CheckBiasScaleMatch(float bias_scale, float weights_scale, float activation
...
+  bias_offsets.assign(num_channels, 0);
+
+  // Compute bias_scale = activation_scale * weight_scale and quantize to int32.
+  // If weights_scales has a single element (per-tensor weight), all channels share the same scale.
+  std::vector<int32_t> quantized_bias(num_channels, 0);
+  for (size_t c = 0; c < num_channels; ++c) {
+    const float weight_scale = (c < weights_scales.size()) ? weights_scales[c] : weights_scales[0];
+    bias_scales[c] = activation_scale * weight_scale;
+    RETURN_IF_NOT(bias_scales[c] > 0.0f, "Bias scale value is non-positive");
+    quantized_bias[c] = static_cast<int32_t>(std::round(float_bias_data[c] / bias_scales[c]));
```
huaychou: **[M-1] (Major)** QuantizeFloatBiasTensor missing INT32 range guard — potential undefined behavior

Problem Code

```
quantized_bias[c] = static_cast<int32_t>(std::round(float_bias_data[c] / bias_scales[c]));
```

When `float_bias_data[c] / bias_scales[c]` exceeds the `INT32` range (±2,147,483,647), the `static_cast<int32_t>` is undefined behavior per C++ standard §7.3.10.

**Impact:** If the bias value is large and the quantization scale is very small, the out-of-range cast typically truncates to 0 or INT_MIN on x86, but the behavior is compiler/platform-dependent and produces no error.

**Suggested fix:**

Per [B4] convention, use `SafeInt<int32_t>`; or add an explicit range clamp before the cast:
`const double val = std::round(static_cast<double>(float_bias_data[c]) / bias_scales[c]);
quantized_bias[c] = static_cast<int32_t>(std::clamp(val, static_cast<double>(INT32_MIN), static_cast<double>(INT32_MAX)));`

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -565,187 +545,103 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+            ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
+                        ("Quantizing float bias " + bias_input.name + " using activation_scale * weight_scale[c]").c_str());
+
+            std::vector<uint8_t> original_bias_data;
+            RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(bias_info.initializer_tensor, original_bias_data));
+
+            const size_t num_channels = bias_info.shape[0];
+            RETURN_IF_NOT(original_bias_data.size() == num_channels * sizeof(float),
+                          "Unexpected bias data size for float bias quantization");
+            const float* bias_float_data = reinterpret_cast<const float*>(original_bias_data.data());
```
huaychou:
```suggestion
            auto bias_float_data = gsl::make_span<const float>(reinterpret_cast<const float*>(original_bias_data.data()), num_channels;
```

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -3416,6 +3579,23 @@ TEST_F(QnnHTPBackendTests, ConvBQ_U16UInt8_1x1_BlockSize4) {
 
 #endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
 
+#if defined(__linux__)
```
huaychou: **[N-2] (Minor)** New test uses #if defined(__linux__) instead of the three-platform guard consistent with existing LPBQ tests

Test `Conv2DOp_QDQ_BlockQuant` uses `#if defined(__linux__)`, inconsistent with the three-platform guard `#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)` used by all other LPBQ/BQ Conv tests in the same file (e.g. `ConvBQ_U16UInt8_1x1_BlockSize4` at line 3416).

The comment says "only Linux for now", meaning x86-64 Linux via QNN simulator; however, `__linux__` also matches ARM64 Linux (`__aarch64__`), which supports HTP hardware tests too.

**Impact:** Readability issue — future maintainers may not understand the intended platform restriction, and when the BQ→LPBQ path is later enabled on Windows ARM64 (`_M_ARM64`), the guard update may be overlooked.

**Suggested fix:**

Add a clear comment to the `#if` guard explaining why it is Linux-only (x86-64 QNN Simulator), or if ARM64 Linux can also run this test, switch to the canonical three-platform form:
`#if defined(__linux__) // BQ->LPBQ conversion currently only enabled on Linux (x86-64 via QNN Simulator); extend to _M_ARM64 when ready`

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -234,32 +243,27 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
-        const std::string fp16_act_name = Ort::ConstNode(&node_unit.GetNode()).GetInputs()[0].GetName();
-        RETURN_IF_ERROR(bq::AddInt16ToFp16DequantForActivation(qnn_model_wrapper, act_name,
-                                                               fp16_act_name, do_op_validation, "Conv"));
-        input_names[0] = fp16_act_name;
-      }
+    if (IsQuant16bit(act_dtype)) {
+      // Reuse the original DequantizeLinear output name for the FP16 tensor so the QNN graph
+      // stays aligned with the ONNX graph naming.
+      const std::string fp16_act_name = Ort::ConstNode(&node_unit.GetNode()).GetInputs()[0].GetName();
+      std::vector<uint32_t> act_shape = act_wrapper.GetTensorDims();
```
huaychou: **[T-1] (Nit)** act_shape declared but never used — dead variable leftover from refactoring

```
std::vector<uint32_t> act_shape = act_wrapper.GetTensorDims();
```

This variable is declared but never used — dead code left over from refactoring.

**Suggested fix:**

Remove the `act_shape` line.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -707,7 +708,7 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
     const int64_t qnn_axis = 1 - axis;
 
     *this = QnnQuantParamsWrapper::LowPowerBlockwise(per_channel_scales, per_block_int_scales, lpbq_offsets,
-                                                     qnn_axis, /*block_scale_bitwidth=*/is_int4_type ? 4 : 8);
+                                                     qnn_axis, /*block_scale_bitwidth=*/4);
```
huaychou: **[T-2] (Nit)** block_scale_bitwidth hardcoded to 4 with no explanatory comment

```
*this = QnnQuantParamsWrapper::LowPowerBlockwise(..., /*block_scale_bitwidth=*/4);
```

The original code was `is_int4_type ? 4 : 8`; this PR changes it to a hard-coded `4` without a comment explaining why. While `4` is correct (because this code path is only reachable when `io_def.type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4`), future maintainers may not be aware of this precondition.

**Suggested fix:**

Add a comment stating the precondition, e.g.: `/*block_scale_bitwidth=*/4  // LPBQ conversion only supports INT4; guarded by the check at the start of this block`

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -565,187 +545,103 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
...
+            const size_t num_channels = bias_info.shape[0];
+            RETURN_IF_NOT(original_bias_data.size() == num_channels * sizeof(float),
+                          "Unexpected bias data size for float bias quantization");
+            const float* bias_float_data = reinterpret_cast<const float*>(original_bias_data.data());
+
+            std::vector<uint8_t> quantized_bias_data;
+            std::vector<float> new_scales;
+            std::vector<int32_t> new_offsets;
+            RETURN_IF_ERROR(utils::QuantizeFloatBiasTensor(
+                gsl::span<const float>(bias_float_data, num_channels),
```
huaychou:
```suggestion
                bias_float_data,
```

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
huaychou: Acknowledge

onnxruntime/test/providers/qnn/conv_test.cc
huaychou: Sure. It's fine to update the utility function with more options to support different use cases.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Done

onnxruntime/test/providers/qnn/conv_test.cc
qti-ashimaj: LPBQ support is now added to windows, hence no longer necessary to add the guard

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Done

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: Added comment

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Updated

onnxruntime/test/providers/qnn/conv_test.cc
qti-ashimaj: Using BuildBQConvTestCase for LPBQ TCs as well now

qti-ashimaj:
> The title and the description are quite different from the change. Could you help update them?

Done

qti-chuteng:
### 🟪 Architecture Fit

**[D-3]** `🔵 Improvement` `[Arch]` — `qnn_quant_params_wrapper.cc` (Init) + `conv_op_builder.cc:250` ⟦C1.2⟧
Whether the LPBQ path actually engages depends on conditions spread across two files with no single owner: `QnnQuantParamsWrapper::Init` silently `return`s the float-BQ fallback on three conditions (`enable_block_quant_weight_optimization` off, non-int4, asymmetric zero-points), while `conv_op_builder` gates `is_lpbq_weight` on three more (NPU backend, 16-bit activation, initializer). A maintainer debugging "why did my model take the float BQ path instead of LPBQ" has to correlate VERBOSE logs across both files to reconstruct the decision. This is an undocumented implicit precondition split across a module boundary.
*Better design:* emit one VERBOSE log at the conv-builder decision point stating the chosen path and the deciding condition (e.g. `is_lpbq_weight=0: act_dtype not 16-bit`), so the path selection is diagnosable from a single line.

### 🟩 Maintainability

**[D-2]** `🟠 Concern` `[Maint]` — `conv_op_builder.cc:521-642` (ProcessConv2D3DInputs) ⟦C1.1⟧
`ProcessConv2D3DInputs` is ~430 lines; the bias section alone is ~120 lines nested 5–6 levels deep (`has_bias → is_initializer → both-quantized → is_quantized/float → needs_requant → per-channel loop`). This PR adds to that depth rather than extracting. The function now carries three distinct bias strategies (BQ FP16 dequant, requantize-if-mismatch, quantize-float) inline. The utils extraction elsewhere in this same PR proves the author knows the pattern — the bias block is the natural next candidate.
*Better design:* extract `ProcessConvBias(qnn_model_wrapper, inputs, is_lpbq_weight, input_names, ...)` mirroring the weight/activation helpers already pulled into `qnn_utils`. This makes the two-file control flow far easier to follow and shrinks the function back toward a readable size.

**[D-4]** `🔵 Improvement` `[Maint]` — `conv_op_builder.cc:250` ⟦C1.3⟧
`is_lpbq_weight = IsNpuBackend && IsQuant16bit(act) && is_initializer && IsLPBQ()`. The name reads as a property of the weight, but it actually means "take the LPBQ code path" and depends on backend + activation dtype + weight. A reader hitting `is_bq_weight && !is_lpbq_weight` has to reverse-engineer that an 8-bit-activation LPBQ-quant weight deliberately falls back to the BW_FLOAT_BLOCK path.
*Better design:* rename to `use_lpbq_path` and add a one-line comment at the definition enumerating the four conditions and why each is required.

### 🟩 Simplicity

**[D-1]** `🟠 Concern` `[Simplicity]` — `conv_op_builder.cc:599-634`, `qnn_utils.cc` (QuantizeFloatBiasTensor) ⟦C8.4⟧ (minimal-change discipline)
`QuantizeFloatBiasTensor` and the "bias is float → quantize with `activation_scale · weight_scale`" branch are **not** LPBQ-specific — they fire for any quantized-activation + quantized-weight Conv that has a float bias initializer. Bundling a general new capability into a PR scoped to "LPBQ encodings for Conv" inflates the diff's stated boundary and couples two independently-testable, independently-revertable features into one unit: if either the float-bias path or the LPBQ path later needs reverting, they can't be separated, and blame/bisect get noisier.
*Better design:* ideally split into two PRs (LPBQ weight path; general float-bias quantization). If kept together, state it explicitly in the title/description and close the coverage gap called out in D-5.

### 🔵 Completeness

**[D-5]** `🔵 Improvement` `[Completeness]` — `conv_test.cc` (new LPBQ tests) ⟦C6.1⟧
All 8 new tests exercise the float-bias path only under LPBQ, i.e. per-channel weight (`weights_scales.size() > 1`). The `weights_scales.size() == 1` per-tensor branch of the new float-bias code (`conv_op_builder.cc:620` and `QuantizeFloatBiasTensor`'s single-scale broadcast) is never executed by any test. Since D-1 establishes this path is general (not LPBQ-only), the untested per-tensor variant is a real coverage gap.
*Better design:* add one test — float bias + non-LPBQ per-tensor int8 weight — to cover the `weights_scales.size() == 1` branch end-to-end.

huaychou:
Just wondering, is it possible to get BQ Conv with float bias under current QDQ selector?

qti-ashimaj:
> Just wondering, is it possible to get BQ Conv with float bias under current QDQ selector?

@huaychou, The current QDQ selector only fuses the nodes when all 3 inputs have their corresponding DQ nodes. To enable float bias support, the changes have been done in https://github.com/onnxruntime/onnxruntime-qnn/pull/481

huaychou:
> > Just wondering, is it possible to get BQ Conv with float bias under current QDQ selector?
> 
> @huaychou, The current QDQ selector only fuses the nodes when all 3 inputs have their corresponding DQ nodes. To enable float bias support, the changes have been done in #481

Could we move the float bias support, float bias QDQ selector, and float bias UT to https://github.com/onnxruntime/onnxruntime-qnn/pull/481 and keep only LPBQ support in this PR?
That would make the PR cleaner and easier for reviewers to read.

qti-ashimaj:
> ### 🟪 Architecture Fit
> **[D-3]** `🔵 Improvement` `[Arch]` — `qnn_quant_params_wrapper.cc` (Init) + `conv_op_builder.cc:250` ⟦C1.2⟧ Whether the LPBQ path actually engages depends on conditions spread across two files with no single owner: `QnnQuantParamsWrapper::Init` silently `return`s the float-BQ fallback on three conditions (`enable_block_quant_weight_optimization` off, non-int4, asymmetric zero-points), while `conv_op_builder` gates `is_lpbq_weight` on three more (NPU backend, 16-bit activation, initializer). A maintainer debugging "why did my model take the float BQ path instead of LPBQ" has to correlate VERBOSE logs across both files to reconstruct the decision. This is an undocumented implicit precondition split across a module boundary. _Better design:_ emit one VERBOSE log at the conv-builder decision point stating the chosen path and the deciding condition (e.g. `is_lpbq_weight=0: act_dtype not 16-bit`), so the path selection is diagnosable from a single line.
> 
> ### 🟩 Maintainability
> **[D-2]** `🟠 Concern` `[Maint]` — `conv_op_builder.cc:521-642` (ProcessConv2D3DInputs) ⟦C1.1⟧ `ProcessConv2D3DInputs` is ~430 lines; the bias section alone is ~120 lines nested 5–6 levels deep (`has_bias → is_initializer → both-quantized → is_quantized/float → needs_requant → per-channel loop`). This PR adds to that depth rather than extracting. The function now carries three distinct bias strategies (BQ FP16 dequant, requantize-if-mismatch, quantize-float) inline. The utils extraction elsewhere in this same PR proves the author knows the pattern — the bias block is the natural next candidate. _Better design:_ extract `ProcessConvBias(qnn_model_wrapper, inputs, is_lpbq_weight, input_names, ...)` mirroring the weight/activation helpers already pulled into `qnn_utils`. This makes the two-file control flow far easier to follow and shrinks the function back toward a readable size.
> 
> **[D-4]** `🔵 Improvement` `[Maint]` — `conv_op_builder.cc:250` ⟦C1.3⟧ `is_lpbq_weight = IsNpuBackend && IsQuant16bit(act) && is_initializer && IsLPBQ()`. The name reads as a property of the weight, but it actually means "take the LPBQ code path" and depends on backend + activation dtype + weight. A reader hitting `is_bq_weight && !is_lpbq_weight` has to reverse-engineer that an 8-bit-activation LPBQ-quant weight deliberately falls back to the BW_FLOAT_BLOCK path. _Better design:_ rename to `use_lpbq_path` and add a one-line comment at the definition enumerating the four conditions and why each is required.
> 
> ### 🟩 Simplicity
> **[D-1]** `🟠 Concern` `[Simplicity]` — `conv_op_builder.cc:599-634`, `qnn_utils.cc` (QuantizeFloatBiasTensor) ⟦C8.4⟧ (minimal-change discipline) `QuantizeFloatBiasTensor` and the "bias is float → quantize with `activation_scale · weight_scale`" branch are **not** LPBQ-specific — they fire for any quantized-activation + quantized-weight Conv that has a float bias initializer. Bundling a general new capability into a PR scoped to "LPBQ encodings for Conv" inflates the diff's stated boundary and couples two independently-testable, independently-revertable features into one unit: if either the float-bias path or the LPBQ path later needs reverting, they can't be separated, and blame/bisect get noisier. _Better design:_ ideally split into two PRs (LPBQ weight path; general float-bias quantization). If kept together, state it explicitly in the title/description and close the coverage gap called out in D-5.
> 
> ### 🔵 Completeness
> **[D-5]** `🔵 Improvement` `[Completeness]` — `conv_test.cc` (new LPBQ tests) ⟦C6.1⟧ All 8 new tests exercise the float-bias path only under LPBQ, i.e. per-channel weight (`weights_scales.size() > 1`). The `weights_scales.size() == 1` per-tensor branch of the new float-bias code (`conv_op_builder.cc:620` and `QuantizeFloatBiasTensor`'s single-scale broadcast) is never executed by any test. Since D-1 establishes this path is general (not LPBQ-only), the untested per-tensor variant is a real coverage gap. _Better design:_ add one test — float bias + non-LPBQ per-tensor int8 weight — to cover the `weights_scales.size() == 1` branch end-to-end.

**[D-3]** Added verbose log which tells which path (bq/lpbq) is taken.

**[D-2]** A new function `ProcessConvBias` with multiple helper functions have been added.

**[D-4]** Updated to `use_lpbq_path`.

**[D-1]** Removed the float bias support from this PR.

**[D-5]** See above.

qti-ashimaj:
> > > Just wondering, is it possible to get BQ Conv with float bias under current QDQ selector?
> > 
> > 
> > @huaychou, The current QDQ selector only fuses the nodes when all 3 inputs have their corresponding DQ nodes. To enable float bias support, the changes have been done in #481
> 
> Could we move the float bias support, float bias QDQ selector, and float bias UT to #481 and keep only LPBQ support in this PR? That would make the PR cleaner and easier for reviewers to read.

Removed the float bias support from this PR.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -13,6 +13,53 @@
 namespace onnxruntime {
 namespace qnn {
 
+namespace {
+inline bool IsQuant16bit(Qnn_DataType_t qnn_data_type) {
+  return qnn_data_type == QNN_DATATYPE_UFIXED_POINT_16 || qnn_data_type == QNN_DATATYPE_SFIXED_POINT_16;
```
qti-kromero: should this be in a more general utility location for widespread usage?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -13,6 +13,53 @@
...
+}
+
+// Extracts the per-tensor activation scale from a QNN quant params wrapper.
+// Returns 1.0f for unrecognized encodings (treated as unquantized).
+inline float GetActivationScale(const QnnQuantParamsWrapper& quant_params) {
+  const auto& qp = quant_params.Get();
+  if (qp.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+    return qp.scaleOffsetEncoding.scale;
+  }
+  if (qp.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) {
```
qti-kromero: this is a minor rant but I don't like how every time I see "BW" I want it to be bitwidth

not really something in scope for this PR

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -177,6 +235,133 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
   return Ort::Status();
 }
 
+// Dequantize INT32 bias to FP16 for BW_FLOAT_BLOCK Conv.
```
qti-kromero, line 233: is this not supported by blockwise encodings yet or why is this generally not able to be processed as-is?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -177,6 +235,133 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
...
+static Ort::Status ProcessBqFp16Bias(QnnModelWrapper& qnn_model_wrapper,
+                                     const OrtNodeUnitIODef& bias_def,
+                                     std::vector<std::string>& input_names) {
+  TensorInfo bias_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(bias_def, bias_info));
+  RETURN_IF(!bias_info.is_initializer, "QNN EP: BQ Conv bias must be a constant initializer");
```
qti-kromero: nit: could use RETURN_IF_NOT

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -177,6 +235,133 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
...
+                                              was_requantized));
+        if (was_requantized) {
+          return Ort::Status();
+        }
+        // Else scales already match, process bias normally.
+      }
+    }
+  }
+
+  // Process bias normally: non-initializer, or activation/weight not quantized, or scales already match.
```
qti-kromero, line 394: can DISABLED_ConvLPBQ_U16Int4_1x1_WithFloatBias_BS8 and DISABLED_ConvLPBQ_U16Int4_1x1_WithFloatBias_BS32 be enabled with these changes?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -13,6 +13,53 @@
...
+// Returns 1.0f for unrecognized encodings (treated as unquantized).
+inline float GetActivationScale(const QnnQuantParamsWrapper& quant_params) {
+  const auto& qp = quant_params.Get();
+  if (qp.quantizationEncoding == QNN_QUANTIZATION_ENCODING_SCALE_OFFSET) {
+    return qp.scaleOffsetEncoding.scale;
+  }
+  if (qp.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_SCALE_OFFSET) {
+    return qp.bwScaleOffsetEncoding.scale;
+  }
+  return 1.0f;
```
qti-kromero, line 25: do we want to just return 1.0f here? why not just fail if the quantization encoding type is unrecognized? I'm also not sure that the claim "treated as unquantized" is true

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
quic-calvnguy: INT32 bias is not supported by HTP and results in an unsupported node IIRC

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -3471,6 +3492,124 @@ TEST_F(QnnHTPBackendTests, ConvBQ_U16UInt8_1x1_BlockSize4) {
...
+                                      /*weight_bits=*/4,
+                                      /*weight_is_unsigned=*/false,
+                                      /*bias_per_channel=*/true),
+                  GetLPBQConvProviderOptions(),
+                  /*opset=*/21,
+                  EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(2e-2f)});
+}
+
+// LPBQ: 1x1 Conv, INT4 weight, block_size=8, float (unquantized) bias.
+TEST_F(QnnHTPBackendTests, DISABLED_ConvLPBQ_U16Int4_1x1_WithFloatBias_BS8) {
```
huaychou: Can we enable this unit test back?

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -3471,6 +3492,124 @@ TEST_F(QnnHTPBackendTests, ConvBQ_U16UInt8_1x1_BlockSize4) {
...
+                                      /*weight_is_unsigned=*/false,
+                                      /*bias_per_channel=*/false,
+                                      /*is_bias_quantized=*/false),
+                  GetLPBQConvProviderOptions(),
+                  /*opset=*/21,
+                  EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(2e-2f)});
+}
+
+// LPBQ: 1x1 Conv, INT4 weight, block_size=32, float (unquantized) bias.
+TEST_F(QnnHTPBackendTests, DISABLED_ConvLPBQ_U16Int4_1x1_WithFloatBias_BS32) {
```
huaychou: Can we enable this unit test back?

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1741,6 +1741,90 @@ bool CheckBiasScaleMatch(float bias_scale, float weights_scale, float activation
   return std::abs(bias_scale - expected_scale) <= tolerance;
 }
 
+Ort::Status GetWeightQuantScales(const QnnQuantParamsWrapper& weight_quant_param,
```
huaychou, line 1744: Please update `onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc` with `GetWeightQuantScales`

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -177,6 +235,133 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
...
+
+  RETURN_IF_ERROR(AddStaticBiasTensor(qnn_model_wrapper, bias_def.name, bias_info.shape,
+                                      bias_info.qnn_data_type,
+                                      BuildBiasQuantParams(new_scales, new_offsets, bias_quant_axis),
+                                      std::move(new_bias_data), input_names));
+  was_requantized = true;
+  return Ort::Status();
+}
+
+Ort::Status ConvOpBuilder::ProcessConvBias(QnnModelWrapper& qnn_model_wrapper,
```
huaychou, line 347: Please update with `ProcessFloatBias` in mainline.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Added in qnn_utils, similar api has been added in matmulnbits PR https://github.com/onnxruntime/onnxruntime-qnn/pull/625 as well which removes the existing function in other op builders too.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Done

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: With PR https://github.com/onnxruntime/onnxruntime-qnn/pull/481 merged, these tests have been enabled back.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Added a check that activation can be per tensor quantized only

onnxruntime/test/providers/qnn/conv_test.cc
qti-ashimaj: With PR https://github.com/onnxruntime/onnxruntime-qnn/pull/481 merged, these tests have been enabled back.

onnxruntime/test/providers/qnn/conv_test.cc
qti-ashimaj: With PR https://github.com/onnxruntime/onnxruntime-qnn/pull/481 merged, these tests have been enabled back.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
qti-ashimaj: @huaychou, `conv_op_builder.cc` already uses `GetWeightQuantScales`. Is there any additional place that you are referring to use it?

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: After rebasing this branch with mainline, `ProcessFloatBias` has been incorporated in `conv_op_builder.cc`

quic-calvnguy (approved):

qti-kromero (approved):

qti-kromero:
@qti-ashimaj FYI the force push above prevents people from viewing from last change in review UI

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -177,6 +230,89 @@ Ort::Status ConvOpBuilder::GetInputChannelNumber(QnnModelWrapper& qnn_model_wrap
...
+  std::vector<uint8_t> new_bias_data;
+  std::vector<float> new_scales;
+  std::vector<int32_t> new_offsets;
+  const auto axis_opt = (num_channels > 1) ? std::optional<int64_t>(bias_quant_axis) : std::nullopt;
+  RETURN_IF_ERROR(utils::RequantizeBiasTensor(
+      original_bias_data, bias_info.shape, current_scales, current_offsets,
+      weights_scales, activation_scale, bias_info.qnn_data_type,
+      new_bias_data, new_scales, new_offsets, axis_opt));
+
+  RETURN_IF_ERROR(AddStaticBiasTensor(qnn_model_wrapper, bias_def.name, bias_info.shape,
```
huaychou: Should we append a suffix like "_rq" for requantized bias to prevent name collisions?

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
huaychou: No, I think the refactor of `ProcessConvBias` is fine.

huaychou (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
qti-ashimaj: Done

PR #486, opened 2026-06-05
[QNN EP] Negotiate ORT API version at plugin load

### LINKED ISSUE #492: QnnEp::QnnEp needs to explicitly set ort_version_supported for compatibility with ORT 1.27?
I was trying to test the current QNN EP distributed through WinML with an ORT 1.27 candidate and was hitting an error during ORT session creation when CreateEp is called on the plugin EP. 

Error creating execution provider: [ONNXRuntimeError] : 1 : FAIL : ep_plugin_provider_interfaces.cc:82 onnxruntime::SanityCheckOrtEp OrtEp has invalid ort_version_supported=0 (expected at least 22).

ORT 1.27 includes a new sanity check via the PR below, requiring ort_version_supported >= 22: https://github.com/microsoft/onnxruntime/pull/28601/changes

I was looking through the QnnEp constructor and ort_version_supported seems 0-initialized from the base class, but I couldn't spot where it's initialized in the function. 
https://github.com/onnxruntime/onnxruntime-qnn/blob/main/onnxruntime/core/providers/qnn/qnn_execution_provider.cc#L344

Other places in the same file set it to ORT_API_VERSION, so I'm wondering if that needs to be added here.

### LINKED ISSUE #564: onnxruntime-qnn==2.3.0 (Linux x86-64): QNN EP version handshake fails (ort_version_supported=0)
### Summary

The **2.3.0 Linux x86-64 wheel's QNN execution provider fails its version handshake** — `ort_version_supported=0`, and no QNN context binary is produced — so QNN compilation/inference fails. The same pipeline works when pinned to the previous Linux x86-64 build, `1.25.0.dev20260330004` (ORT-Nightly).

`2.3.0` (published 2026-06-22) is the **first and only `2.x` release to ship a Linux x86-64 wheel** (`manylinux_2_35_x86_64`); earlier `2.x` releases were aarch64/Windows-only, so on x86-64 the resolver previously stayed on the `1.25.0.dev*` nightly line. When `2.3.0` appeared, an unpinned best-match install floated up to it and broke.

### Environment

| | |
|---|---|
| OS / arch | Linux x86-64 (`manylinux`) |
| Python | 3.12 (cp312) |
| `onnxruntime-qnn` | **2.3.0** (PyPI `manylinux_2_35_x86_64`) |
| `onnxruntime` (co-resolved) | `1.28.0.dev20260623002` |
| Last known-good | `onnxruntime-qnn==1.25.0.dev20260330004` (ORT-Nightly, cp312 linux_x86_64) |

### What happens

We compile YOLO models to a QNN context binary on the host via the ONNX Runtime QNN Execution Provider (the `onnxruntime-qnn` package, which bundles QAIRT) — see the [Ultralytics QNN export docs](https://docs.ultralytics.com/integrations/qnn). With `onnxruntime-qnn==2.3.0` on Linux x86-64, the QNN EP reports the running ONNX Runtime version as unsupported:

```
QNN provider: ort_version_supported=0
```

No `*_qnn.onnx` context binary is generated and the export/session fails. This looks like a version mismatch between the QNN EP packaged in the `2.3.0` x86-64 wheel and the `onnxruntime==1.28.0.dev…` it resolves alongside.

### Reproduce

```bash
python3.12 -m venv venv && . venv/bin/activate
pip install "onnxruntime-qnn==2.3.0" \
  --extra-index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/ \
  --prerelease=allow
# create an InferenceSession with the QNN EP / run a QNN context-binary export
# -> QNN EP init reports ort_version_supported=0, no context binary produced
```

Pinning back makes it work again:

```bash
pip install "onnxruntime-qnn==1.25.0.dev20260330004" \
  --extra-index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/ --prerelease=allow
```

### Expected

The `2.3.0` Linux x86-64 wheel's QNN EP should accept the ONNX Runtime version it ships with (or pin/declare a compatible `onnxruntime`), so `ort_version_supported` is non-zero and QNN context-binary compilation succeeds.

### Notes / context

- Seen in production while building YOLO → QNN exports for the [Ultralytics Platform](https://platform.ultralytics.com); the Ultralytics QNN export path lives in [ultralytics/ultralytics](https://github.com/ultralytics/ultralytics) (`ultralytics/utils/export/qnn.py`).
- Workaround we adopted: pin `onnxruntime-qnn==1.25.0.dev20260330004`.

Happy to provide additional logs from the failing session if helpful.

qti-niscmami:
### Description

`CreateEpFactories` in `qnn_provider_factory.cc` now negotiates the ORT API version with the host at plugin load instead of demanding `ORT_API_VERSION`

Verified locally with four venvs against the same EP wheel (and a second wheel built with a simulated lower ceiling)

### Motivation and Context

QNN EP is deployed into fixed host environments. The old code crashed (null deref) or emitted an opaque host message when the host ORT was older than the EP was compiled against. This converts that into a clear, actionable error.

### DIFF
--- a/onnxruntime/core/providers/qnn/ort_api_version_parser.h
+++ b/onnxruntime/core/providers/qnn/ort_api_version_parser.h
@@ -27,12 +27,12 @@ inline uint32_t ParseRuntimeOrtApiVersion(const char* version_str) {
     return 0;
   }
 
-  int minor = 0;
+  uint32_t minor = 0;
   auto [p2, ec2] = std::from_chars(p1 + 1, end, minor);
-  if (ec2 != std::errc{} || minor < 0) {
+  if (ec2 != std::errc{}) {
     return 0;
   }
-  return static_cast<uint32_t>(minor);
+  return minor;
 }
 
 }  // namespace onnxruntime::qnn::detail
--- a/onnxruntime/core/providers/qnn/ort_api_version_parser.h
+++ b/onnxruntime/core/providers/qnn/ort_api_version_parser.h
@@ -1,5 +1,5 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
 
 #pragma once
 
--- a/onnxruntime/core/providers/qnn/ort_api_version_parser.h
+++ b/onnxruntime/core/providers/qnn/ort_api_version_parser.h
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
+
+#pragma once
+
+#include <charconv>
+#include <cstdint>
+#include <string_view>
+#include <system_error>
+
+namespace onnxruntime::qnn::detail {
+
+// Parse the host ORT version string ("1.X.Y") and return the API version (the
+// minor component). Returns 0 if the string is null, malformed, or major != 1.
+// The "minor == API version" contract is documented for ORT 1.x; rejecting
+// non-1 majors keeps a hypothetical ORT 2.x from being silently misparsed.
+inline uint32_t ParseRuntimeOrtApiVersion(const char* version_str) {
+  if (version_str == nullptr) {
+    return 0;
+  }
+  std::string_view sv{version_str};
+  const char* end = sv.data() + sv.size();
+
+  int major = 0;
+  auto [p1, ec1] = std::from_chars(sv.data(), end, major);
+  if (ec1 != std::errc{} || p1 == end || *p1 != '.' || major != 1) {
+    return 0;
+  }
+
+  int minor = 0;
+  auto [p2, ec2] = std::from_chars(p1 + 1, end, minor);
+  if (ec2 != std::errc{} || minor < 0) {
+    return 0;
+  }
+  return static_cast<uint32_t>(minor);
+}
+
+}  // namespace onnxruntime::qnn::detail
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -3,7 +3,10 @@
 
 #include "core/providers/qnn/qnn_provider_factory.h"
 
+#include <algorithm>
 #include <cassert>
+#include <cstdint>
+#include <cstdio>
 #include <iostream>
 #include <optional>
 
@@ -465,6 +468,27 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetHardwareDeviceIncompatibilityDetailsImp
 
 }  // namespace onnxruntime
 
+namespace {
+// Parse the host ORT version string ("1.X.Y") and return the API version (the
+// minor component). Returns 0 if the string is null, malformed, or major != 1.
+// The "minor == API version" contract is documented for ORT 1.x; rejecting
+// non-1 majors keeps a hypothetical ORT 2.x from being silently misparsed.
+uint32_t ParseRuntimeOrtApiVersion(const char* version_str) {
+  if (version_str == nullptr) {
+    return 0;
+  }
+  int major = 0;
+  int minor = 0;
+  if (std::sscanf(version_str, "%d.%d", &major, &minor) < 2) {
+    return 0;
+  }
+  if (major != 1 || minor < 0) {
+    return 0;
+  }
+  return static_cast<uint32_t>(minor);
+}
+}  // namespace
+
 extern "C" {
 //
 // Public symbols
@@ -476,12 +500,38 @@ OrtStatus* CreateEpFactories(const char* registration_name,
                              size_t max_factories,
                              size_t* num_factories) {
   if (ort_api_base == nullptr) {
-    return nullptr;  // Cannot create status without API base
+    return nullptr;
   }
 
-  const OrtApi* ort_api = ort_api_base->GetApi(ORT_API_VERSION);
+  // kMinOrtApiVersion must equal the highest \since version of any ORT API
+  // method this EP calls. Below this floor, GetApi() returns a function table
+  // that lacks members the EP would dereference.
+  constexpr uint32_t kMinOrtApiVersion = 24;
+
+  const char* version_str = ort_api_base->GetVersionString();
+  const uint32_t runtime_api_version = ParseRuntimeOrtApiVersion(version_str);
+
+  if (runtime_api_version < kMinOrtApiVersion) {
+    const uint32_t fallback_version = runtime_api_version > 0 ? runtime_api_version : 1;
+    const OrtApi* fallback_api = ort_api_base->GetApi(fallback_version);
+    if (fallback_api == nullptr) {
+      return nullptr;
+    }
+    char msg[256];
+    std::snprintf(msg, sizeof(msg),
+                  "QNN EP requires ORT >= 1.%u (API %u). "
+                  "Host ORT is %s (API %u).",
+                  kMinOrtApiVersion, kMinOrtApiVersion,
+                  version_str != nullptr ? version_str : "unknown",
+                  runtime_api_version);
+    return fallback_api->CreateStatus(ORT_FAIL, msg);
+  }
+
+  const uint32_t requested_api_version =
+      std::min(runtime_api_version, static_cast<uint32_t>(ORT_API_VERSION));
+  const OrtApi* ort_api = ort_api_base->GetApi(requested_api_version);
   if (ort_api == nullptr) {
-    return nullptr;  // Cannot create status without ORT API
+    return nullptr;
   }
 
   // Manual init for the C++ API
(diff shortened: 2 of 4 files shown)


onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -465,6 +471,33 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetHardwareDeviceIncompatibilityDetailsImp
 
 }  // namespace onnxruntime
 
+namespace {
+// Parse the host ORT version string ("1.X.Y") and return the API version (the
+// minor component). Returns 0 if the string is null, malformed, or major != 1.
+// The "minor == API version" contract is documented for ORT 1.x; rejecting
+// non-1 majors keeps a hypothetical ORT 2.x from being silently misparsed.
+uint32_t ParseRuntimeOrtApiVersion(const char* version_str) {
```
qti-kromero: this function seems pretty unit-testable and could maybe be moved to a utils file? the latter I don't think necessarily needs to be addressed here

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -476,12 +509,38 @@ OrtStatus* CreateEpFactories(const char* registration_name,
                              size_t max_factories,
                              size_t* num_factories) {
   if (ort_api_base == nullptr) {
-    return nullptr;  // Cannot create status without API base
+    return nullptr;
   }
 
-  const OrtApi* ort_api = ort_api_base->GetApi(ORT_API_VERSION);
+  // kMinOrtApiVersion must equal the highest \since version of any ORT API
```
qti-kromero: nit: having a hard time parsing this comment, specifically the " \since" part. not sure if I'm just brain farting or if it is typo

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -476,12 +509,38 @@ OrtStatus* CreateEpFactories(const char* registration_name,
...
+  // method this EP calls. Below this floor, GetApi() returns a function table
+  // that lacks members the EP would dereference.
+  constexpr uint32_t kMinOrtApiVersion = 24;
+
+  const char* version_str = ort_api_base->GetVersionString();
+  const uint32_t runtime_api_version = ParseRuntimeOrtApiVersion(version_str);
+
+  if (runtime_api_version < kMinOrtApiVersion) {
+    const uint32_t fallback_version = runtime_api_version > 0 ? runtime_api_version : 1;
```
qti-kromero: do runtime minor version 0 & 1 share API version 1? I guess it kind of doesn't matter given this is an error but am slightly confused by this

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -476,12 +509,38 @@ OrtStatus* CreateEpFactories(const char* registration_name,
...
+    if (fallback_api == nullptr) {
+      return nullptr;
+    }
+    char msg[256];
+    std::snprintf(msg, sizeof(msg),
+                  "QNN EP requires ORT >= 1.%u (API %u). "
+                  "Host ORT is %s (API %u).",
+                  kMinOrtApiVersion, kMinOrtApiVersion,
+                  version_str != nullptr ? version_str : "unknown",
+                  runtime_api_version);
```
qti-kromero: if version_str is nullptr then runtime_api_version gets printed as "Host ORT is ... (API 0)" which is kind of confusing

would prefer having some kind of additional error message that says something like "unrecognized version format or unsupported major version" on failure to parse version string

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-niscmami: Not a typo.  "\since" is the Doxygen tag on ORT's API methods.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-niscmami: Done. Added unit tests to verify.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-niscmami: It doesn't matter but yeah this should be changed. Updated the code.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-niscmami: Updated the code to give a more verbose error message.

onnxruntime/core/providers/qnn/ort_api_version_parser.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
qti-kromero: this copyright and below for ort_api_version_parser_test.cc should be Qualcomm as it is added by us, see other files added by us or go/markings guidance for open source

qti-kromero: sorry missed this on last review

qti-niscmami: Fixed.

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:L489`
`kMinOrtApiVersion = 24` currently equals `ORT_API_VERSION` (also 24) but there is no `static_assert` to enforce this at compile time. On future API bumps — either `ORT_API_VERSION` advancing when the EP starts using a new API function, or someone lowering `kMinOrtApiVersion` for backward-compat without realizing the EP already depends on API-24 features — the two constants can silently diverge and cause a null-deref crash.

```suggestion
  // CreateExternalResourceImporterForDevice and the full EP bridge were added in API 24.
  constexpr uint32_t kMinOrtApiVersion = 24;
  static_assert(kMinOrtApiVersion <= ORT_API_VERSION,
                "kMinOrtApiVersion must not exceed ORT_API_VERSION");
```

---

### 🔵 Minor

**[N-1]** `onnxruntime/test/providers/qnn/ort_api_version_parser_test.cc:L1`
The new test file uses a Microsoft copyright header while the corresponding `ort_api_version_parser.h` (and QNN EP convention [B1]) requires Qualcomm copyright for new files added by Qualcomm. Note that `qti-kromero` already flagged this for the header file; the test file still needs the same fix.

```suggestion
// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
// SPDX-License-Identifier: MIT
```

**[N-2]** `onnxruntime/core/providers/qnn/ort_api_version_parser.h:L36`
`minor` is declared as `int` and then guarded with `minor < 0` before casting to `uint32_t`. `std::from_chars` on `uint32_t` already returns `std::errc::invalid_argument` for negative inputs and `std::errc::result_out_of_range` for overflow, so the `int` intermediate and the extra guard are unnecessary. Functionally correct, style issue only.

```suggestion
  uint32_t minor = 0;
  auto [p2, ec2] = std::from_chars(p1 + 1, end, minor);
  if (ec2 != std::errc{}) {
    return 0;
  }
  return minor;
```

qti-kromero (approved):

qti-niscmami:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/qnn_provider_factory.cc:L489` `kMinOrtApiVersion = 24` currently equals `ORT_API_VERSION` (also 24) but there is no `static_assert` to enforce this at compile time. On future API bumps — either `ORT_API_VERSION` advancing when the EP starts using a new API function, or someone lowering `kMinOrtApiVersion` for backward-compat without realizing the EP already depends on API-24 features — the two constants can silently diverge and cause a null-deref crash.
> 
> ```
>   // CreateExternalResourceImporterForDevice and the full EP bridge were added in API 24.
>   constexpr uint32_t kMinOrtApiVersion = 24;
>   static_assert(kMinOrtApiVersion <= ORT_API_VERSION,
>                 "kMinOrtApiVersion must not exceed ORT_API_VERSION");
> ```
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/test/providers/qnn/ort_api_version_parser_test.cc:L1` The new test file uses a Microsoft copyright header while the corresponding `ort_api_version_parser.h` (and QNN EP convention [B1]) requires Qualcomm copyright for new files added by Qualcomm. Note that `qti-kromero` already flagged this for the header file; the test file still needs the same fix.
> 
> ```
> ```
> 
> **[N-2]** `onnxruntime/core/providers/qnn/ort_api_version_parser.h:L36` `minor` is declared as `int` and then guarded with `minor < 0` before casting to `uint32_t`. `std::from_chars` on `uint32_t` already returns `std::errc::invalid_argument` for negative inputs and `std::errc::result_out_of_range` for overflow, so the `int` intermediate and the extra guard are unnecessary. Functionally correct, style issue only.
> 
> ```
>   uint32_t minor = 0;
>   auto [p2, ec2] = std::from_chars(p1 + 1, end, minor);
>   if (ec2 != std::errc{}) {
>     return 0;
>   }
>   return minor;
> ```

Addressed all comments

tirupath-qti (approved):

PR #487, opened 2026-06-05
[QNN EP] Build-time check for minimum ORT API version

qti-niscmami:
### Description

  Extends #486. Replaces the hardcoded `kMinOrtApiVersion = 24` literal with a `QNN_EP_MIN_ORT_API_VERSION` macro generated each build, plus a lint gate that catches silent drift.

  - `compute_min_ort_api_version.py` scans ORT headers for `\since` annotations, scans QNN EP source for `ort_api->X()` / `ep_api.X()` call sites, emits `max(\since)`.
  - CMake `add_custom_target(ALL)` writes `qnn_ep_min_ort_api_version.h` before compile.
  - `qnn_provider_factory.cc` includes that header and uses the macro.
  - Lint gate (`MIN-ORT-API-VERSION`) compares the computed floor to `MIN_ORT_API_VERSION.txt`; fails with a remediation command on drift, making floor bumps a visible diff in the same PR.

  ### Motivation and Context

  Without this, a future PR adding an `ort_api->NewMethod1_25()` call would silently pass the negotiation check on a 1.24 host and UB-deref. The baseline file (`MIN_ORT_API_VERSION.txt`) makes floor bumps a required, reviewable part of the PR diff.


qti-chuteng:
### 🟠 Major

**[M-1]** `qcom/scripts/all/compute_min_ort_api_version.py:L84`
`build_since_map` keeps the **minimum** `\since` version per member, contradicting its own docstring ("record the highest version found"). On duplicate member names — or when a `\since`-less doc block (treated as version 0) pairs to the same name — the floor is understated, which defeats the entire purpose of this PR: the runtime negotiation then accepts an older ORT API table missing members the EP dereferences.
```suggestion
                    if name not in result or result[name] < version:
```

**[M-2]** `qcom/linters/check_min_ort_api_version.py:L64`
The adapter calls the compute script in `--check` mode without `--ort-header-root`, so it falls back to globbing `build/*/.../_deps/ort_core-src/include`. On a clean checkout or a lint-only CI runner (no build), the compute script returns exit 2 ("ORT header root not found"). The adapter treats only `returncode == 0` as a pass, so it reports this environmental failure as a lint error. Distinguish exit 1 (genuine drift) from exit 2 (cannot compute → skip with a clear message), or pass the header root explicitly.

**[M-3]** `qcom/scripts/all/compute_min_ort_api_version.py → build_since_map() / compute_floor()`
264-line parsing script added with no unit test, even though `qcom/scripts/all/tests/` already has pytest coverage for sibling scripts. M-1 is exactly the kind of bug a 5-line test would catch. Add `test_compute_min_ort_api_version.py` covering max-of-duplicate-`\since`, no-`\since`→0, each `_DECL_RE`/`_CALL_RE` form, and `--check`/`--update-baseline`.

qti-niscmami:
> ### 🟠 Major
> **[M-1]** `qcom/scripts/all/compute_min_ort_api_version.py:L84` `build_since_map` keeps the **minimum** `\since` version per member, contradicting its own docstring ("record the highest version found"). On duplicate member names — or when a `\since`-less doc block (treated as version 0) pairs to the same name — the floor is understated, which defeats the entire purpose of this PR: the runtime negotiation then accepts an older ORT API table missing members the EP dereferences.
> 
> ```
>                     if name not in result or result[name] < version:
> ```
> 
> **[M-2]** `qcom/linters/check_min_ort_api_version.py:L64` The adapter calls the compute script in `--check` mode without `--ort-header-root`, so it falls back to globbing `build/*/.../_deps/ort_core-src/include`. On a clean checkout or a lint-only CI runner (no build), the compute script returns exit 2 ("ORT header root not found"). The adapter treats only `returncode == 0` as a pass, so it reports this environmental failure as a lint error. Distinguish exit 1 (genuine drift) from exit 2 (cannot compute → skip with a clear message), or pass the header root explicitly.
> 
> **[M-3]** `qcom/scripts/all/compute_min_ort_api_version.py → build_since_map() / compute_floor()` 264-line parsing script added with no unit test, even though `qcom/scripts/all/tests/` already has pytest coverage for sibling scripts. M-1 is exactly the kind of bug a 5-line test would catch. Add `test_compute_min_ort_api_version.py` covering max-of-duplicate-`\since`, no-`\since`→0, each `_DECL_RE`/`_CALL_RE` form, and `--check`/`--update-baseline`.

Resolved all comments.

qcom/scripts/all/tests/test_compute_min_ort_api_version.py
```diff
@@ -0,0 +1,600 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
```
qti-kromero, line 2: does this get called as part of some CI pytest run?

qcom/scripts/all/compute_min_ort_api_version.py
```diff
@@ -0,0 +1,350 @@
+#!/usr/bin/env python3
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+"""Compute the minimum ORT API version QNN EP requires.
+
+Walks the ORT C API headers for ``\\since Version 1.X`` annotations on every
```
qti-kromero, line 6: maybe I need more context on something - am slightly confused why we need to store MIN_ORT_API_VERSION.txt and have the linting for it if we are running this analysis at CMake time anyways? can we not just derive this information at CMake-time based on the map built in this script?

qcom/scripts/all/tests/test_compute_min_ort_api_version.py
qti-niscmami: I don't think so. They don't run in any of the CI workflows.

qcom/scripts/all/compute_min_ort_api_version.py
qti-niscmami: The baseline file + linter makes floor bumps show up as visible diff in the PRs. This is to help reviewers see when EP starts requiring a newer ORT API version.

qcom/scripts/all/tests/test_compute_min_ort_api_version.py
qti-kromero: let's create a follow-up ticket for this

qcom/linters/check_min_ort_api_version.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
qti-kromero: can the functionality of this script be incorporated into the compute script such that we only have one script? i.e. fold some behaviors from thsi script into compute_min_ort_api... and call compute_min_ort_api... directly in the lintrunner with --check option

qti-niscmami: Was able to consolidate it into just the compute script.

qti-kromero (approved):

qti-chuteng:
### 🟪 Architecture Fit

**[D-3]** `🟠 Concern` `[Arch]` — `cmake/onnxruntime_providers_qnn.cmake:67`
In `onnxruntime_ORT_HOME` (MS-prebuilt) mode the codegen command scans `--ort-header-root ${ORT_SOURCE_DIR}/include` (the `FetchContent_Populate(ort_core)` tree, pinned by `cmake/deps.txt` to v1.26.0), but the EP is *compiled* against `${onnxruntime_ORT_HOME}/include` — a separate header set that need not match the pin. `FetchContent_Populate(ort_core)` runs unconditionally, so the scan silently succeeds against the wrong headers, and the floor (plus the surviving `static_assert`) is derived from a different ORT version than the function table actually dereferenced — breaking the exact invariant the PR protects.
*Better design:* point `--ort-header-root` at the same include root the target compiles against (`${ONNXRUNTIME_APPLICATION_INCLUDES}` / the ORT_HOME include when set), or assert the two roots are the same version.

**[D-5]** `🔵 Improvement` `[Arch]` — `qcom/scripts/all/compute_min_ort_api_version.py:225`
The `build_since_map` doc-block→next-declaration pairing assigns each `/* */` block's `\since` to the first following decl. A no-`\since` doc block immediately ahead of a real API decl records that decl at version 0 → if the EP calls it, the floor is silently too low (the UB the PR prevents). Running the branch script against the real v1.26.0 headers yields 24 with no unknown-API raise, so this is latent fragility for future header restructurings, not a present defect.
*Better design:* anchor `\since` to the specific member (parse the decl name, look *backwards* for its own doc block), or assert every EP-used member resolved from a block that actually contained a `\since`; add a header-form regression fixture.

### 🟩 Simplicity

**[D-2]** `🟠 Concern` `[Simplicity]` — `qcom/scripts/all/compute_min_ort_api_version.py:354`
`fetch_ort_headers` / `parse_deps_txt` / `_cache_root` re-roll `urllib`+`hashlib`+`zipfile`+cache-dir logic that the repo already has: `package_manager.py::FileCache.fetch(expected_sha1=...)` (~:62) and `fetch_cmake_deps.py::_parse_deps_txt` (:57). deps.txt is now parsed in 3 places, with a 4th cache convention and a new `QNN_EP_LINT_CACHE` env var. Critically, the hand-rolled `urllib.request.urlopen(url)` uses **no** SSL context, whereas `FileCache` wires `ssl.create_default_context(cafile=certifi.where())` — a real SSL-verify regression on machines with an incomplete trust store (CI / corporate proxies).
*Better design:* delete the three helpers and reuse `FileCache.fetch` + `fetch_cmake_deps._parse_deps_txt`; if avoiding `package_manager`'s heavy imports is the reason, state it in a comment and still route through the certifi CA bundle. Removes ~80 lines, a failure class, and the SSL regression.

### 🔵 Completeness

**[D-1]** `🟠 Concern` `[Completeness]` — `qcom/scripts/all/compute_min_ort_api_version.py:272`
The success criterion is that the floor can never *silently* lag behind a newly-used newer ORT API. But `_CALL_RE` matches only direct C-table calls (`ort_api->X`, `ep_api.X`, `model_editor_api.`, `compile_api.`). QNN EP also calls versioned C APIs through the `Ort::` C++ wrapper layer (`onnxruntime_cxx_inline.h`) — 136 `Ort::Const*` wrapper sites across 25 files (e.g. `Ort::ConstNode(...).GetAttributeByName` → `Node_GetAttributeByName`, `Ort::ConstValueInfo(...).GetConsumers()` → `ValueInfo_GetValueConsumers`) — all invisible to the scanner. The computed floor is `max(\since)` over only the direct-call subset. Today's result (24) is correct by *luck*: every wrapper's backing C API is ≤1.23 and direct calls independently push it to 24. But the headers already ship 45 APIs at 1.25 and 8 at 1.26. The moment someone wraps a 1.25/1.26 call via `Ort::`, the floor stays too low, lint stays green, and the build ships an under-declared floor — the exact UB-on-old-hosts failure the PR exists to prevent. A docstring calling this "out of scope" re-introduces the silent-lag path in prose.
*Better design:* turn the blind spot into an *enforced tripwire* — fail the check if any `Ort::Const*`/wrapper method call appears in EP source that the scanner can't account for, so the blind spot can't silently grow. (Full solution: resolve each wrapper method to its backing C API and treat it as a call to that member.)

**[D-4]** `🟠 Concern` `[Completeness]` — `cmake/onnxruntime_providers_qnn.cmake:66`
The codegen custom command invokes `${Python3_EXECUTABLE}`, populated only by `find_package(Python3 ...)` in the build-from-source branch of `onnxruntime_prebuilt.cmake:30`. The `onnxruntime_ORT_HOME` branch skips it; the only other discovery (top-level `find_package(Python 3.10 ...)`, `CMakeLists.txt:284`) sets `Python_EXECUTABLE`, **not** `Python3_EXECUTABLE`. So in ORT_HOME mode `${Python3_EXECUTABLE}` can expand empty and the codegen command degrades to running the script with no interpreter → configure/build error. The existing prebuilt code guards this with `FATAL_ERROR`; the new consumer has no guard.
*Better design:* add an unconditional `find_package(Python3 REQUIRED COMPONENTS Interpreter)` before the custom command, or reuse the already-found `Python_EXECUTABLE`.

**[D-6]** `🔵 Improvement` `[Completeness]` — `.lintrunner.toml:170`
The `MIN-ORT-API-VERSION` lint command passes `--fetch-from-deps-txt` (and `run_lintrunner` auto-enables it), so on a clean checkout with no build tree the linter downloads the ort_core archive from `github.com`. Lint gates are normally hermetic; this couples PR-blocking status to GitHub availability and is flaky offline. It degrades gracefully (build-tree headers tried first; offline fetch failure → exit 2 → `ort-api-floor-check-error` finding, not a crash), and fixing D-2 mitigates it.
*Better design:* prefer build-tree headers (already the first choice in `run()`); document that CI lint must run after a configure, or note the offline expectation.

**[D-7]** `🔵 Improvement` `[Completeness]` — `qcom/scripts/all/tests/test_compute_min_ort_api_version.py`
The PR adds `pytest>=7.0` and a 658-line suite, but no `PyTestTask`/runner targets `qcom/scripts/all/tests/`, so scanner regressions aren't caught automatically. This is a pre-existing gap (the sibling `test_archive_tests.py` is also unwired); the actual production guard is the CI-wired `MIN-ORT-API-VERSION` lint.
*Better design:* add a runner for `qcom/scripts/all/tests` (a `conftest.py` already sets up import paths), or document these as developer-run-only.

**[D-8]** `🔵 Improvement` `[Completeness]` — `docs/`
The new generated header, the checked-in baseline, and the `--update-baseline` workflow are undocumented outside the script; a contributor who trips `ort-api-floor-drift` learns the fix only from stderr. The in-script remediation is thorough, so this is a discoverability gap, not correctness.
*Better design:* one `build.md`/development.md paragraph describing the generated header, the baseline file, and the bump-in-the-same-PR rule.

qti-niscmami:
> ### 🟪 Architecture Fit
> **[D-3]** `🟠 Concern` `[Arch]` — `cmake/onnxruntime_providers_qnn.cmake:67` In `onnxruntime_ORT_HOME` (MS-prebuilt) mode the codegen command scans `--ort-header-root ${ORT_SOURCE_DIR}/include` (the `FetchContent_Populate(ort_core)` tree, pinned by `cmake/deps.txt` to v1.26.0), but the EP is _compiled_ against `${onnxruntime_ORT_HOME}/include` — a separate header set that need not match the pin. `FetchContent_Populate(ort_core)` runs unconditionally, so the scan silently succeeds against the wrong headers, and the floor (plus the surviving `static_assert`) is derived from a different ORT version than the function table actually dereferenced — breaking the exact invariant the PR protects. _Better design:_ point `--ort-header-root` at the same include root the target compiles against (`${ONNXRUNTIME_APPLICATION_INCLUDES}` / the ORT_HOME include when set), or assert the two roots are the same version.
> 
> **[D-5]** `🔵 Improvement` `[Arch]` — `qcom/scripts/all/compute_min_ort_api_version.py:225` The `build_since_map` doc-block→next-declaration pairing assigns each `/* */` block's `\since` to the first following decl. A no-`\since` doc block immediately ahead of a real API decl records that decl at version 0 → if the EP calls it, the floor is silently too low (the UB the PR prevents). Running the branch script against the real v1.26.0 headers yields 24 with no unknown-API raise, so this is latent fragility for future header restructurings, not a present defect. _Better design:_ anchor `\since` to the specific member (parse the decl name, look _backwards_ for its own doc block), or assert every EP-used member resolved from a block that actually contained a `\since`; add a header-form regression fixture.
> 
> ### 🟩 Simplicity
> **[D-2]** `🟠 Concern` `[Simplicity]` — `qcom/scripts/all/compute_min_ort_api_version.py:354` `fetch_ort_headers` / `parse_deps_txt` / `_cache_root` re-roll `urllib`+`hashlib`+`zipfile`+cache-dir logic that the repo already has: `package_manager.py::FileCache.fetch(expected_sha1=...)` (~:62) and `fetch_cmake_deps.py::_parse_deps_txt` (:57). deps.txt is now parsed in 3 places, with a 4th cache convention and a new `QNN_EP_LINT_CACHE` env var. Critically, the hand-rolled `urllib.request.urlopen(url)` uses **no** SSL context, whereas `FileCache` wires `ssl.create_default_context(cafile=certifi.where())` — a real SSL-verify regression on machines with an incomplete trust store (CI / corporate proxies). _Better design:_ delete the three helpers and reuse `FileCache.fetch` + `fetch_cmake_deps._parse_deps_txt`; if avoiding `package_manager`'s heavy imports is the reason, state it in a comment and still route through the certifi CA bundle. Removes ~80 lines, a failure class, and the SSL regression.
> 
> ### 🔵 Completeness
> **[D-1]** `🟠 Concern` `[Completeness]` — `qcom/scripts/all/compute_min_ort_api_version.py:272` The success criterion is that the floor can never _silently_ lag behind a newly-used newer ORT API. But `_CALL_RE` matches only direct C-table calls (`ort_api->X`, `ep_api.X`, `model_editor_api.`, `compile_api.`). QNN EP also calls versioned C APIs through the `Ort::` C++ wrapper layer (`onnxruntime_cxx_inline.h`) — 136 `Ort::Const*` wrapper sites across 25 files (e.g. `Ort::ConstNode(...).GetAttributeByName` → `Node_GetAttributeByName`, `Ort::ConstValueInfo(...).GetConsumers()` → `ValueInfo_GetValueConsumers`) — all invisible to the scanner. The computed floor is `max(\since)` over only the direct-call subset. Today's result (24) is correct by _luck_: every wrapper's backing C API is ≤1.23 and direct calls independently push it to 24. But the headers already ship 45 APIs at 1.25 and 8 at 1.26. The moment someone wraps a 1.25/1.26 call via `Ort::`, the floor stays too low, lint stays green, and the build ships an under-declared floor — the exact UB-on-old-hosts failure the PR exists to prevent. A docstring calling this "out of scope" re-introduces the silent-lag path in prose. _Better design:_ turn the blind spot into an _enforced tripwire_ — fail the check if any `Ort::Const*`/wrapper method call appears in EP source that the scanner can't account for, so the blind spot can't silently grow. (Full solution: resolve each wrapper method to its backing C API and treat it as a call to that member.)
> 
> **[D-4]** `🟠 Concern` `[Completeness]` — `cmake/onnxruntime_providers_qnn.cmake:66` The codegen custom command invokes `${Python3_EXECUTABLE}`, populated only by `find_package(Python3 ...)` in the build-from-source branch of `onnxruntime_prebuilt.cmake:30`. The `onnxruntime_ORT_HOME` branch skips it; the only other discovery (top-level `find_package(Python 3.10 ...)`, `CMakeLists.txt:284`) sets `Python_EXECUTABLE`, **not** `Python3_EXECUTABLE`. So in ORT_HOME mode `${Python3_EXECUTABLE}` can expand empty and the codegen command degrades to running the script with no interpreter → configure/build error. The existing prebuilt code guards this with `FATAL_ERROR`; the new consumer has no guard. _Better design:_ add an unconditional `find_package(Python3 REQUIRED COMPONENTS Interpreter)` before the custom command, or reuse the already-found `Python_EXECUTABLE`.
> 
> **[D-6]** `🔵 Improvement` `[Completeness]` — `.lintrunner.toml:170` The `MIN-ORT-API-VERSION` lint command passes `--fetch-from-deps-txt` (and `run_lintrunner` auto-enables it), so on a clean checkout with no build tree the linter downloads the ort_core archive from `github.com`. Lint gates are normally hermetic; this couples PR-blocking status to GitHub availability and is flaky offline. It degrades gracefully (build-tree headers tried first; offline fetch failure → exit 2 → `ort-api-floor-check-error` finding, not a crash), and fixing D-2 mitigates it. _Better design:_ prefer build-tree headers (already the first choice in `run()`); document that CI lint must run after a configure, or note the offline expectation.
> 
> **[D-7]** `🔵 Improvement` `[Completeness]` — `qcom/scripts/all/tests/test_compute_min_ort_api_version.py` The PR adds `pytest>=7.0` and a 658-line suite, but no `PyTestTask`/runner targets `qcom/scripts/all/tests/`, so scanner regressions aren't caught automatically. This is a pre-existing gap (the sibling `test_archive_tests.py` is also unwired); the actual production guard is the CI-wired `MIN-ORT-API-VERSION` lint. _Better design:_ add a runner for `qcom/scripts/all/tests` (a `conftest.py` already sets up import paths), or document these as developer-run-only.
> 
> **[D-8]** `🔵 Improvement` `[Completeness]` — `docs/` The new generated header, the checked-in baseline, and the `--update-baseline` workflow are undocumented outside the script; a contributor who trips `ort-api-floor-drift` learns the fix only from stderr. The in-script remediation is thorough, so this is a discoverability gap, not correctness. _Better design:_ one `build.md`/development.md paragraph describing the generated header, the baseline file, and the bump-in-the-same-PR rule.

Resolved all issues

D-7 is currently being handled in a follow-up PR https://github.com/onnxruntime/onnxruntime-qnn/pull/570

PR #488, opened 2026-06-05
[QNN EP] Framework op tracing follow-up: profiling integration and AOT sidecar

kuanyul-qti:
Add two follow-up features to the framework op tracing implementation:

1. Profiling <-> trace auto-merge (JIT and AOT Phase 2)
   - Extend profiling output with onnx op column when the profiling level is detailed or optrace.

2. AOT Phase 2 sidecar discovery
   - Check and load a Phase 1 trace JSON placed alongside the context model.

Tests and documentation
 - Add unit tests.
 - New `## QNN EP Framework Op Tracing` section in `QNN-ExecutionProvider.md` covering output file naming and the zero-step Phase 1->2 sidecar workflow.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -31,6 +31,7 @@
 #include "core/providers/qnn/rpcmem_library.h"
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/op_package/op_package.h"
+#include "core/providers/qnn/builder/op_tracing/qnn_op_tracing_types.h"
 #include "core/providers/qnn/builder/qnn_context_mem_handle_manager.h"
 #include "core/providers/qnn/builder/qnn_def.h"
 #include "core/providers/qnn/builder/qnn_htp_power_config_manager.h"
@@ -122,6 +123,11 @@ struct QnnBackendManagerConfig {
   uint32_t soc_model;
   std::vector<OpPackage> op_packages;
   bool skip_qnn_version_check;
+  // When true, framework op tracing is active for the session and the profiling
+  // CSV gains the `ONNX Source Ops` column. Provided here alongside the other
+  // profiling settings so it is set once at backend-manager construction and
+  // remains constant for the manager's lifetime.
+  bool enable_framework_op_trace = false;
 };
 
 class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager> {
@@ -146,6 +152,7 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
         profiling_level_etw_(config.profiling_level_etw),
         profiling_level_(config.profiling_level),
         profiling_file_path_(config.profiling_file_path),
+        enable_framework_op_trace_(config.enable_framework_op_trace),
         context_priority_(config.context_priority),
         qnn_serializer_config_(config.qnn_serializer_config),
         device_id_(config.device_id),
@@ -231,6 +238,25 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
 
   Ort::Status ExtractBackendProfilingInfo(qnn::profile::ProfilingInfo& profiling_info);
 
+  // Framework op tracing: profiling enrichment lookup, shared across all
+  // QnnModels in this session. Populated by:
+  //   - JIT path:  ComposeGraph() merges each per-graph lookup via
+  //                MergeOpTraceLookup() after the trace collector finalizes.
+  //   - AOT path:  CompileContextModel() loads the sidecar JSON via
+  //                SetOpTraceLookup() before any context binary is loaded.
+  // Self-attached to the local ProfilingInfo inside ExtractBackendProfilingInfo
+  // when HasNodeLevelProfiling() is true and op tracing is enabled.
+  void SetOpTraceLookup(qnn::OpTraceLookup lookup) { op_trace_lookup_ = std::move(lookup); }
+  // Merges `other` into the session-wide lookup. On key collision the entry
+  // from `other` wins (last-write-wins), matching the operator[] semantics
+  // already used by OpTraceCollector::Finalize and LoadTraceLookupFromFile
+  // when they populate a lookup. `other` is consumed.
+  void MergeOpTraceLookup(qnn::OpTraceLookup&& other) {
+    for (auto& kv : other) {
+      op_trace_lookup_[kv.first] = std::move(kv.second);
+    }
+  }
+
   Ort::Status ExtractProfilingSubEvents(QnnProfile_EventId_t profile_event_id, profile::Serializer& profile_writer,
                                         bool backendSupportsExtendedEventData);
 
@@ -295,6 +321,14 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
 #ifdef QNN_SYSTEM_PROFILE_API_ENABLED
   bool ProfilingEnabled() { return profiling_enabled_; }
 #endif
+
+  // Returns true when profiling is at DETAILED or OPTRACE level, meaning per-node
+  // (QNN_PROFILE_EVENTTYPE_NODE) events are generated. Used to gate ONNX source
+  // annotation in the profiling CSV.
+  bool HasNodeLevelProfiling() const {
+    return profiling_level_merge_ == ProfilingLevel::DETAILED ||
+           profiling_level_merge_ == ProfilingLevel::OPTRACE;
+  }
 
   bool IsBackendSetup() { return backend_setup_completed_; }
   bool FileMappingIsEnabled() {
@@ -636,6 +670,13 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   ProfilingLevel profiling_level_;
   ProfilingLevel profiling_level_merge_;
   const std::string profiling_file_path_;
+  // Framework op trace -> profiling enrichment lookup. See accessor comments
+  // for population rules.
+  qnn::OpTraceLookup op_trace_lookup_;
+  // True when framework op tracing is active for the session. Fixed at
+  // construction, so the profiling CSV's `ONNX Source Ops` column is present
+  // for every graph's events and its header matches all NODE rows.
+  const bool enable_framework_op_trace_ = false;
   bool system_lib_loaded_ = false;
 
 #ifdef QNN_SYSTEM_PROFILE_API_ENABLED
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -247,7 +247,7 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   // Self-attached to the local ProfilingInfo inside ExtractBackendProfilingInfo
   // when profiling is at DETAILED/OPTRACE level (per-NODE events) and op
   // tracing is enabled.
-  void SetOpTraceLookup(OpTraceLookup lookup) { op_trace_lookup_ = std::move(lookup); }
+  void SetOpTraceLookup(OpTraceLookup&& lookup) { op_trace_lookup_ = std::move(lookup); }
   // Merges `other` into the session-wide lookup. On key collision the entry
   // from `other` wins (last-write-wins), matching the operator[] semantics
   // already used by OpTraceCollector::Finalize and LoadTraceLookupFromFile
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
@@ -236,7 +236,7 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   // NOTE: This function locks the internal `logger_recursive_mutex_`.
   Ort::Status ResetQnnLogLevel(std::optional<OrtLoggingLevel> ort_log_level = std::nullopt);
 
-  Ort::Status ExtractBackendProfilingInfo(qnn::profile::ProfilingInfo& profiling_info);
+  Ort::Status ExtractBackendProfilingInfo(profile::ProfilingInfo& profiling_info);
 
   // Framework op tracing: profiling enrichment lookup, shared across all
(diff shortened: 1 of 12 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -636,6 +670,13 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   ProfilingLevel profiling_level_;
   ProfilingLevel profiling_level_merge_;
   const std::string profiling_file_path_;
+  // Framework op trace -> profiling enrichment lookup. See accessor comments
```
minfhong-qti: Move to a separated section with description commented.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -296,6 +322,14 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   bool ProfilingEnabled() { return profiling_enabled_; }
 #endif
 
+  // Returns true when profiling is at DETAILED or OPTRACE level, meaning per-node
+  // (QNN_PROFILE_EVENTTYPE_NODE) events are generated. Used to gate ONNX source
+  // annotation in the profiling CSV.
+  bool HasNodeLevelProfiling() const {
```
minfhong-qti: Prefer directly flatten in code instead of defining a function.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -231,6 +238,25 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
 
+  // Framework op tracing: profiling enrichment lookup, shared across all
+  // QnnModels in this session. Populated by:
+  //   - JIT path:  ComposeGraph() merges each per-graph lookup via
+  //                MergeOpTraceLookup() after the trace collector finalizes.
+  //   - AOT path:  CompileContextModel() loads the sidecar JSON via
+  //                SetOpTraceLookup() before any context binary is loaded.
+  // Self-attached to the local ProfilingInfo inside ExtractBackendProfilingInfo
+  // when HasNodeLevelProfiling() is true and op tracing is enabled.
+  void SetOpTraceLookup(qnn::OpTraceLookup lookup) { op_trace_lookup_ = std::move(lookup); }
```
minfhong-qti:
```suggestion
  void SetOpTraceLookup(OpTraceLookup lookup) { op_trace_lookup_ = std::move(lookup); }
```
Already in `qnn` namespace. Please check for others.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -308,7 +308,11 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
 
   // Collect framework op trace after graph composition
   if (trace_collector) {
-    trace_collector->Finalize(graph_name, qnn_model_wrapper, *context.op_trace_output);
+    qnn::OpTraceLookup per_graph_lookup;
+    trace_collector->Finalize(graph_name, qnn_model_wrapper, *context.op_trace_output, per_graph_lookup);
+    // Hand the per-graph lookup off to the backend manager, which holds the
+    // session-wide lookup that ExtractBackendProfilingInfo reads from.
+    qnn_backend_manager_->MergeOpTraceLookup(std::move(per_graph_lookup));
```
minfhong-qti, line 315: Since `QnnBackendManager.ExtractBackendProfilingInfo` is only called during `QnnBackendManager.LoadCachedQnnContextFromBuffer`, when will the lookup set in `QnnModel.ComposeGraph` be used?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -636,6 +670,13 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   ProfilingLevel profiling_level_;
   ProfilingLevel profiling_level_merge_;
   const std::string profiling_file_path_;
+  // Framework op trace -> profiling enrichment lookup. See accessor comments
+  // for population rules.
+  qnn::OpTraceLookup op_trace_lookup_;
```
minfhong-qti: As discussed offline, I don't like treating `QnnBackendManager` as the carrier for non-backend related information. But I do not have a better suggestion now so please evaluate whether it is possible not to record this here.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Done. Moved to a dedicated section after system_lib_loaded_ with a section-header comment explaining what each field controls and how it's read.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Removed. Also swept the rest of the new code:
- qnn_backend_manager.h (qnn::profile::ProfilingInfo), qnn_backend_manager.h (MergeOpTraceLookup parameter), qnn_backend_manager.h (member field), and qnn_model.cc: were all redundant inside namespace qnn { ... } — fixed.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: `ExtractBackendProfilingInfo` is actually called from four places, not just `LoadCachedQnnContextFromBuffer`:
 - `qnn_model.cc:283` — end of `ComposeGraph` (COMPOSE_GRAPHS phase)
 - `qnn_model.cc:371` — end of `FinalizeGraphs` (FINALIZE phase)
 - `qnn_model.cc:592` — end of `ExecuteGraph` (EXECUTE phase, per inference)
 - `qnn_backend_manager.cc:1807` — inside `LoadCachedQnnContextFromBuffer` (AOT only, CREATE_FROM_BINARY)
 
The merge at `qnn_model.cc:315` is consumed by the JIT-path `FinalizeGraphs` and `ExecuteGraph` extractions above — that's where DETAILED/OPTRACE-level per-NODE events actually appear and need annotation. AOT uses the separate `SetOpTraceLookup` (overwrite) call from qnn_execution_provider.cc instead, which is what then feeds `LoadCachedQnnContextFromBuffer`'s extraction.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Acknowledged. I evaluated three alternatives:

1. Pass the lookup as an arg to `ExtractBackendProfilingInfo` — all four callsites (`qnn_model.cc`, `qnn_backend_manager.cc`) would have to know about session-scoped tracing state. Pushes the coupling further out without removing it.
2. Park on `QnnExecutionProvider`, give `BackendManager` a non-owning pointer — same coupling, just relocated. `BackendManager` still has to pass and read it. The lookup still lives on a non-backend object that BackendManager holds a reference to. Doesn't actually decouple.
3. A new session-scoped service object injected into `BackendManager` at construction — cleanest, but introduces a third object whose only role today is the CSV ONNX Source Ops annotation. I think this is the right shape if/when more session-scoped non-backend state shows up.

The constraint that pinned me to `BackendManager` is that `ExtractBackendProfilingInfo` is the only writer of the profiling CSV (header + rows), so the lookup has to be reachable from this at that point. I left it on `BackendManager` for this PR, but happy to refactor to (3) in a follow-up if more state lands here.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -229,7 +236,27 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+  // Framework op tracing: profiling enrichment lookup, shared across all
+  // QnnModels in this session. Populated by:
+  //   - JIT path:  ComposeGraph() merges each per-graph lookup via
+  //                MergeOpTraceLookup() after the trace collector finalizes.
+  //   - AOT path:  CompileContextModel() loads the sidecar JSON via
+  //                SetOpTraceLookup() before any context binary is loaded.
+  // Self-attached to the local ProfilingInfo inside ExtractBackendProfilingInfo
+  // when profiling is at DETAILED/OPTRACE level (per-NODE events) and op
+  // tracing is enabled.
+  void SetOpTraceLookup(OpTraceLookup lookup) { op_trace_lookup_ = std::move(lookup); }
```
minfhong-qti:
```suggestion
  void SetOpTraceLookup(OpTraceLookup&& lookup) { op_trace_lookup_ = std::move(lookup); }
```
Not sure.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -229,7 +236,27 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+  // Merges `other` into the session-wide lookup. On key collision the entry
+  // from `other` wins (last-write-wins), matching the operator[] semantics
+  // already used by OpTraceCollector::Finalize and LoadTraceLookupFromFile
+  // when they populate a lookup. `other` is consumed.
+  void MergeOpTraceLookup(OpTraceLookup&& other) {
+    for (auto& kv : other) {
+      op_trace_lookup_[kv.first] = std::move(kv.second);
```
minfhong-qti, line 257: Will there be key collision?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1890,6 +1891,28 @@ OrtStatus* QnnEp::CompileContextModel(const OrtGraph** graphs,
   std::basic_string<ORTCHAR_T> context_model_path;
   GetContextOnnxModelFilePath(context_cache_path_cfg_, model_path, context_model_path);
 
+  // AOT Phase 2 sidecar discovery for profiling enrichment. Loaded here
+  // (before any context binary is restored) so the lookup is on the backend
+  // manager when the first profile extraction runs, ensuring InitCsvFile()
+  // emits the `ONNX Source Ops` column and every NODE event row is annotated.
+  if (enable_framework_op_trace_) {
+    auto trace_path = qnn::DeriveTracePathFromContextModel(std::filesystem::path(context_model_path));
```
minfhong-qti, line 1890: Just a thought for removing tracing from backend manager.
For `QnnModel`, I believe look up table can be recorded inside it and pass to `ExtractProfilingXXX` when ever needed, so this is not the problem.
For loading context model back, I guess we can pass the `trace_path` around and loaded before `ExtractProfilingXXX` and pass into it, and then no need to record inside backend manager. However, the problem in this proposal is that this trace path needs to pass through several functions including the ones in `onnx_ctx_model_helper.cc`.

Please evaluate this proposal if it works or I am as well fine with current implementation.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Practically, no. The lookup key is qnn_op_name (the QNN node name). All QNN node names are minted by `utils::UniqueNameGenerator()`, which is a process-wide static singleton with a `counter_` map; duplicates are renamed with a `_N` suffix on every call. So, across all subgraphs composed in a session, the keys we feed into `op_trace_lookup_` are unique by construction.
The merge code uses last-write-wins as a **defensive fallback in case a future refactor regresses the unique-name guarantee**. If you'd prefer hard-failing on collision instead of last-write-wins I can swap to try_emplace + warning log.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Walked through both halves and have a concrete cost.
JIT half (lookup on `QnnModel`) — workable. Each `QnnModel` already owns its `OpTraceCollector`, and the four `ExtractBackendProfilingInfo` callsites are all on `QnnModel` or on `QnnBackendManager`. We can add an `OpTraceLookup* lookup` parameter to `ExtractBackendProfilingInfo` and pass `&this->op_trace_lookup_` from each `QnnModel` site. One signature change, three forwarders, zero callsite logic changes. This part is clean.

AOT half (thread `trace_path` through) — this is where it gets expensive. The chain is:

> qnn_execution_provider.cc  LoadQnnCtxFromOnnxGraph(...)
> onnx_ctx_model_helper.cc   LoadQnnCtxFromOnnxGraph -> LoadCachedQnnContextFromBuffer (3 callsites)
> qnn_backend_manager.cc   LoadCachedQnnContextFromBuffer -> ExtractBackendProfilingInfo

So, `trace_path` (or the loaded lookup) has to be added as a parameter to `LoadQnnCtxFromOnnxGraph`, propagated through three internal `LoadCachedQnnContextFromBuffer` forwarders, and held until the extraction point. `onnx_ctx_model_helper.cc` doesn't currently know anything about framework op tracing — adding this parameter is purely plumbing for one optional, profiling-only feature.

Trade-off:
- Current: one extra non-backend field on `QnnBackendManager`; lookup is loaded once at the EP level (qnn_execution_provider.cc) before any context binary is touched.
- Proposal: removes the field from `QnnBackendManager`, but adds an optional parameter to four function signatures that span three translation units (`qnn_execution_provider.cc`, `onnx_ctx_model_helper.cc`, `qnn_backend_manager.h/.cc`).

The proposal cleans up one coupling (lookup on `QnnBackendManager`) by introducing plumbing coupling across four signatures, which feels like a lateral move for a single optional feature. If a second piece of session-scoped non-backend state shows up I'd revisit and lift this into a small session-scoped service object — that would carry both, and `BackendManager` could hold a non-owning ref to it instead of the field today. I'd rather do that as a follow-up than do half of it now.

minfhong-qti (approved):

PR #490, opened 2026-06-05
Add DqlDqFusion: fuse DynamicQuantizeLinear+DequantizeLinear into identity Transpose

yuhuchua-qti:
**Summary**

- Add DqlDqFusion to fuse DynamicQuantizeLinear + DequantizeLinear into an identity-permutation Transpose on QNN EP
- Fixes CPU fallback caused by QNN EP's lack of native DynamicQuantizeLinear support

**Background**

Some models contain a fake-quantize identity pattern where DynamicQuantizeLinear (DQL) computes dynamic quant params
that are immediately reversed by DequantizeLinear (DQ), with no integer kernel (e.g. MatMulInteger) ever consuming the
quantized tensor:

x (float32) ──► DynamicQuantizeLinear ──► (y_uint8, y_scale, y_zp)
				ALL 3 outputs exclusively ──► DequantizeLinear ──► x_approx (float32)
																	 └──► Gemm (float32)

This is a leftover artifact from an incomplete quantization conversion. Since QNN EP does not support
DynamicQuantizeLinear, the entire subgraph falls back to CPU, impacting performance.

**Solution**

Replace the DQL + DQ pair with an identity-permutation Transpose (zero cycles on HTP). The round-trip DQ(DQL(x)) ≈ x —
replacing it with identity actually improves numerical accuracy (removes quantization noise) since the downstream op
is float32.

Guard conditions (all must hold to fuse):
1. DQL is a SingleNode with 1 float32 input and 3 outputs
2. All 3 DQL outputs consumed exclusively by DequantizeLinear nodes
3. The DQ consuming DQL.y also receives DQL.y_scale (via quant_param) and optionally DQL.y_zero_point — same DQ node,
no other consumers
4. DQ output is float32
5. QNN validates the replacement Transpose

Files Changed

  - builder/qnn_node_group/dql_dq_fusion.h — New: DqlDqFusion class declaration
  - builder/qnn_node_group/dql_dq_fusion.cc — New: fusion implementation
  - builder/qnn_node_group/qnn_node_group.cc — Register DqlDqFusion::TryFusion for DynamicQuantizeLinear

Implementation Note

OrtNodeUnit::Inputs() embeds scale/zero_point in quant_param rather than exposing them as separate input entries. The
DQ verification therefore uses quant_param->scale and quant_param->zero_point instead of raw input indices.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/dql_dq_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/dql_dq_fusion.cc
@@ -75,9 +75,10 @@ std::unique_ptr<IQnnNodeGroup> DqlDqFusion::TryFusion(
     return reject("DQL.output[0] is not exclusively consumed by a standalone DequantizeLinear");
   }
 
-  // y_scale must have exactly one consumer and it must be the fused DQ node.
-  // y_zero_point is optional (DQ may omit it); if present it must also belong solely to this DQ.
-  // These checks prevent fusion when a second DQ node shares y_scale or y_zp: in that case the
+  // y_scale and y_zero_point must each have exactly one consumer and it must be the fused DQ node.
+  // y_zero_point must not be absent: DQL's y_zp is dynamic and typically non-zero, so allowing DQ
+  // to omit it (defaulting to zero_point=0) would break the identity round-trip.
+  // These checks also prevent fusion when a second DQ node shares y_scale or y_zp: in that case the
   // second DQ would lose its producer after fusion, corrupting the graph.
   {
     auto has_single_dq_consumer = [&](const Ort::ConstValueInfo& vi) -> bool {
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/builder/qnn_node_group/dql_dq_fusion.cc
```diff
@@ -0,0 +1,237 @@
...
+  const OrtNodeUnit* dq = GetOnlyChildOfOutput(
+      qnn_model_wrapper, dql_node_unit, dql_outputs[0],
+      node_to_node_unit, node_unit_to_qnn_node_group);
+  if (dq == nullptr || dq->OpType() != DEQUANTIZE_LINEAR ||
+      dq->UnitType() != OrtNodeUnit::Type::SingleNode) {
+    return reject("DQL.output[0] is not exclusively consumed by a standalone DequantizeLinear");
+  }
+
+  // y_scale must have exactly one consumer and it must be the fused DQ node.
+  // y_zero_point is optional (DQ may omit it); if present it must also belong solely to this DQ.
```
huaychou: Could we have a unit test for y_zero_point is not used by DQ?

huaychou (commented):
Just wondering, is it possible to add no-op in QNN? or we can only add dummy transpose or reshape to represent no-op 🤔

yuhuchua-qti:
> Just wondering, is it possible to add no-op in QNN? or we can only add dummy transpose or reshape to represent no-op 🤔

In theory we could avoid the extra op by adding a tensor name remapping mechanism in QnnModelWrapper, but that would be an invasive change to the core tensor management logic. The identity-permutation Transpose is already the established pattern in this codebase for this use case and compiles to zero cycles on HTP, so we followed the existing convention.

onnxruntime/core/providers/qnn/builder/qnn_node_group/dql_dq_fusion.cc
yuhuchua-qti: We hadn't explicitly verified this before — the original comment "DQ may omit it" was technically correct per ONNX spec (x_zero_point is optional for DequantizeLinear), but we didn't validate whether it was safe in this specific fusion context.

The new test exposed the problem: DQL computes a dynamic y_zp that is typically non-zero (e.g. ~128 for float32 inputs in [-1, 1]), so DQ omitting zero_point silently defaults to 0, making DQ(y, scale, 0) ≠ x and breaking the identity round-trip.

This also contradicts the fusion pattern's own definition — all three DQL outputs must feed the same DQ — so y_zp must always have exactly one consumer. Fixed the guard to reject zp_consumers.empty() and require quant_param->zero_point to be present. The test is updated to DqlDqFusion_NoZeroPoint_NoFusion to verify fusion is correctly rejected in this case.

huaychou (approved):
LGTM. I have one question: if the input for DQL and DQ is a constant value (for example, weight), will the fusion be triggered, or will ORT simplify the graph first?

qti-chuteng (approved):

yuhuchua-qti:
> LGTM. I have one question: if the input for DQL and DQ is a constant value (for example, weight), will the fusion be triggered, or will ORT simplify the graph first?

Confirmed: ORT L1 constant folding runs before TryFusion (inference_session.cc:1535 vs 1619). DQL with constant input
gets folded — its outputs become initializers. The downstream DQ is preserved (skip_dequantize_linear), but DQL is
gone, so our fusion entry point never matches.

PR #491, opened 2026-06-05
[QNN EP] Framework op tracing follow-up: source->optimized ONNX matcher

kuanyul-qti:
Closes the original-ONNX -> optimized-ONNX provenance gap left by the existing framework op trace. Combined with qnn_op_trace.json, gives end-to-end original ONNX -> optimized ONNX -> QNN op mappings.

Three pieces:

1.  dump_qnn_ep_input_graph session config — serializes the ONNX graph the EP receives in GetCapability (post-Level-1, pre-partition) as a QNN-Netron-schema JSON. Needed because optimized_model_filepath rejects graphs containing compiled nodes, and the plugin EP cannot link protobuf to write real .onnx.
2. source_to_optimized_matcher.py — six-strategy structural matcher (name / I/O / initializer name+hash / fusion pattern / lineage) between source.onnx and the EP-input dump (or an optimized.onnx). Optionally extends qnn_op_trace.json with original_sources[].
3. enrich_profiling_csv.py — appends an Original ONNX Source Ops column to the QNN EP profiling CSV; adapts to four input shapes.

Tests: add new tests.
Docs updated end-to-end, including the highest-<n>-wins rule for picking the right dump file when ORT runs GetCapability multiple times in one session.

### DIFF
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -21,6 +21,7 @@ ONNX Runtime QNN EP can be used on Windows devices with Qualcomm Snapdragon SOC'
 - [Running an LLM model with QNN EP's Genie backend](#running-an-llm-model-with-qnn-eps-genie-backend)
 - [QNN context binary cache feature](#qnn-context-binary-cache-feature)
 - [QNN EP Framework Op Tracing](#qnn-ep-framework-op-tracing)
+- [QNN EP Input Graph Dump](#qnn-ep-input-graph-dump)
 - [QNN EP Profiling](#qnn-ep-profiling)
 - [QNN EP weight sharing](#qnn-ep-weight-sharing)
 - [Usage](#usage)
@@ -234,6 +235,15 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
 |`"json_qnn_graph_dir"`|Description|
 |---|---|
 |Directory path (string)|Directory path for dumping QNN JSON graphs. Only effective when `dump_json_qnn_graph` is enabled.|
+
+|`"dump_qnn_ep_input_graph"`|Description|
+|---|---|
+|'0'|Default. Disabled.|
+|'1'|Dump the ONNX graph the QNN EP receives at compile time (after ORT Level 1 optimizations, before partitioning) as a JSON file. The output uses the same schema as `dump_json_qnn_graph`, so it opens in QNN Netron. Unlike `dump_json_qnn_graph` (which dumps the post-compile QNN graph), this captures the pre-compile ONNX graph. See `dump_qnn_ep_input_graph_dir` to control the output location.|
+
+|`"dump_qnn_ep_input_graph_dir"`|Description|
+|---|---|
+|Directory path (string)|Directory path for the QNN EP input graph dump. Only effective when `dump_qnn_ep_input_graph` is `'1'`. Defaults to the current working directory if not set. The directory is created automatically if it does not exist. If the directory cannot be created or written to, the dump is disabled with a WARNING log. Output files use the pattern `<graph_name>.<n>_qnn_ep_input_graph.json`, where `<n>` is a per-EP counter that increments for each dumped graph. When the source graph has no name, `<graph_name>` is the literal `graph`.|
 
 |`"dump_qnn_ir_dlc"`|Description|
 |---|---|
@@ -806,6 +816,89 @@ ort.unregister_execution_provider_library(ep_registration_name)
 ```
 
 For an out-of-the-box example, see [`qcom/samples/test_genie.py`](../../qcom/samples/test_genie.py).
+
+## QNN EP Input Graph Dump
+
+Setting `dump_qnn_ep_input_graph` to `'1'` writes the ONNX graph the QNN EP
+receives at compile time — after ORT Level 1 optimizations, before partitioning
+and EP compile — to a JSON file (one per graph, named
+`<graph_name>.<n>_qnn_ep_input_graph.json`, where `<n>` is a per-EP counter
+that disambiguates collisions when two graph names sanitize to the same
+string). Use `dump_qnn_ep_input_graph_dir` to
+choose the output directory (defaults to the current working directory).
+
+This is useful for inspecting exactly what the EP sees: that graph is **not** the
+user's source `.onnx` (ORT has run constant folding, QDQ duplication via
+`EnsureUniqueDQForNodeUnit`, Transpose optimization, etc.), and it is otherwise
+hard to observe — `SessionOptions::optimized_model_filepath` refuses to serialize
+a graph once an EP has compiled part of it into opaque compiled nodes ("Unable to
+serialize model as it contains compiled nodes").
+
+### Why the dump is QNN-Netron JSON, not an ONNX model
+
+The dump deliberately uses the same JSON schema as `dump_json_qnn_graph` rather
+than emitting a real `.onnx`. The QNN EP ships as a **standalone ABI plugin** that
+links only the public ONNX Runtime C/C++ API, Abseil, and nlohmann/json — it does
+**not** link protobuf or the ONNX proto definitions. Constructing or serializing
+an `onnx::ModelProto` therefore is not possible from the EP without adding a
+protobuf dependency to every shipped `onnxruntime_providers_qnn` binary, which
+would break the plugin's minimal-dependency contract.
+
+Reusing the QNN-Netron JSON schema instead means:
+
+- **No new dependencies** — nlohmann/json is already used by the EP.
+- **Opens directly in QNN Netron** — same schema Netron already understands.
+- **One graph-dump format** across the EP: `dump_json_qnn_graph` for the
+  post-compile QNN graph, `dump_qnn_ep_input_graph` for the pre-compile ONNX graph.
+- **Consumable by tooling** — the offline `qcom/tools/op_trace_matcher/source_to_optimized_matcher.py`
+  reads this JSON directly as its `--optimized-model`, recovering
+  `original ONNX -> EP-input -> QNN op` provenance.
+
+The schema carries node name/op_type/input_names/output_names and a tensor table
+flagging each tensor's role (graph input / output / intermediate / initializer),
+which is everything downstream consumers need.
+
+**Known limitation:** the dump does **not** include initializer tensor data
+bytes. Only the tensor name, dtype, shape, and role are recorded. Tools
+consuming the dump (for example `source_to_optimized_matcher.py`'s
+`--optimized-model` input) must therefore rely on initializer **names** or
+graph **topology** — not data hashes — when correlating tensors across the
+source and the EP-input graph. The matcher uses a name-derived sentinel hash
+for initializers from this input so name-based and topology-based matching
+work unchanged.
+
+### Multiple dumps per graph name
+
+A single inference session can produce **more than one** dump for the same
+graph name, because ORT may invoke `GetCapability` more than once:
+
+- **Layout transformation 2-pass cycle.** When the graph contains
+  layout-sensitive ops (Conv, Resize, Pool, …), ORT runs a 1st pass to
+  collect EP claims, applies an NCHW->NHWC rewrite for nodes the EP
+  accepted, and runs a 2nd pass on the transformed graph. The two dumps
+  differ — the 2nd has Conv in `kMSInternalNHWCDomain` and includes
+  inserted Transpose ops.
+- **Control-flow subgraphs.** Each `If` / `Loop` / `Scan` subgraph is a
+  separate `OrtGraph*` and gets its own `GetCapability` invocation, so
+  models with control flow legitimately produce multiple dumps with
+  different graph names.
+- **EPContext model loads.** Loading a previously-cached EPContext model
+  triggers another `GetCapability` pass on that context graph.
+
+Filenames append a per-EP atomic counter (`<graph_name>.<n>_…json`) so each
+pass gets a distinct file, and an explicit overwrite warning is logged if
+the same name is ever about to be rewritten.
+
+**Rule for the offline matcher (`source_to_optimized_matcher.py`):** for
+each unique `<graph_name>` in the dump directory, feed the file with the
+**highest `<n>`** as `--optimized-model`. That is the graph the EP actually
+compiled, and its node names are exactly the names that appear in
+`qnn_op_trace.json`'s `sources[]`. Lower-numbered files are intermediate
+snapshots (useful for debugging the layout transformer itself, not for
+matcher input).
+
+If a model has subgraphs (different graph names), pick the highest-numbered
+file *per unique graph name* and run the matcher per graph.
 
 ## QNN context binary cache feature
 There's a QNN context which contains QNN graphs after converting, compiling, finalizing the model. QNN can serialize the context into binary file, so that user can use it for futher inference directly (without the QDQ model) to improve the model loading cost.
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -836,8 +836,15 @@ serialize model as it contains compiled nodes").
 
 ### Why the dump is QNN-Netron JSON, not an ONNX model
 
-The dump deliberately uses the same JSON schema as `dump_json_qnn_graph` rather
-than emitting a real `.onnx`. The QNN EP ships as a **standalone ABI plugin** that
+Netron is the open-source neural-net viewer that already understands the
+JSON shape produced by QNN's existing `dump_json_qnn_graph`
+(the post-compile QNN graph). The schema is not formally specified,
+but it is stable and observable in this repository at
+`onnxruntime/core/providers/qnn/builder/qnn_utils.cc` (`QnnJSONGraph`). The
+EP-input dump deliberately reuses that schema rather than emitting a real
+`.onnx`.
+
+The QNN EP ships as a **standalone ABI plugin** that
 links only the public ONNX Runtime C/C++ API, Abseil, and nlohmann/json — it does
 **not** link protobuf or the ONNX proto definitions. Constructing or serializing
 an `onnx::ModelProto` therefore is not possible from the EP without adding a
--- a/onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
@@ -1,0 +1,304 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "core/providers/qnn/builder/qnn_ep_input_graph_dumper.h"
+
+#include <algorithm>
+#include <cctype>
+#include <fstream>
+#include <set>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include "nlohmann/json.hpp"
+
+#include "QnnTypes.h"
+
+#include "core/providers/qnn/builder/qnn_utils.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+namespace {
+
+// QNN tensor-type integers Netron uses to classify a tensor. Mirrors the
+// values emitted by the post-compile QnnJSONGraph path (qnn_utils.cc), which
+// serializes the raw Qnn_TensorType_t enum.
+constexpr int kQnnTensorTypeAppWrite = QNN_TENSOR_TYPE_APP_WRITE;  // graph input  = 0
+constexpr int kQnnTensorTypeAppRead = QNN_TENSOR_TYPE_APP_READ;    // graph output = 1
+constexpr int kQnnTensorTypeNative = QNN_TENSOR_TYPE_NATIVE;       // intermediate = 3
+constexpr int kQnnTensorTypeStatic = QNN_TENSOR_TYPE_STATIC;       // initializer  = 4
+
+// Assigns a stable, incrementing integer id per tensor name. The public
+// ValueInfo API exposes no per-value id, and Netron only needs uniqueness, so
+// a name->int allocator suffices.
+class TensorIdAllocator {
+ public:
+  TensorIdAllocator() = default;
+
+  int GetOrAdd(const std::string& name) {
+    auto it = ids_.find(name);
+    if (it != ids_.end()) {
+      return it->second;
+    }
+    int id = next_id_++;
+    ids_.emplace(name, id);
+    return id;
+  }
+
+ private:
+  std::unordered_map<std::string, int> ids_;
+  int next_id_ = 0;
+
+  ORT_DISALLOW_COPY_AND_ASSIGNMENT(TensorIdAllocator);
+};
+
+// Classify a value info into the QNN tensor-type integer Netron expects.
+// Order matters: an initializer that is also an optional graph input is
+// classified as STATIC (its data is what Netron should show).
+int ClassifyTensorType(const Ort::ConstValueInfo& value_info) {
+  if (value_info.IsConstantInitializer()) {
+    return kQnnTensorTypeStatic;
+  }
+  if (value_info.IsRequiredGraphInput() || value_info.IsOptionalGraphInput()) {
+    return kQnnTensorTypeAppWrite;
+  }
+  if (value_info.IsGraphOutput()) {
+    return kQnnTensorTypeAppRead;
+  }
+  return kQnnTensorTypeNative;
+}
+
+// Build the per-tensor JSON entry. `raw_value_info` may be null (an optional,
+// not-provided node input); the caller skips those before calling this.
+// `type_info_failures` is incremented when the value's type info cannot be
+// read; the dumper logs a single aggregated WARNING at the end so silent
+// degradation (every tensor reported as FLOAT_32 + empty dims) is observable.
+nlohmann::json BuildTensorJson(const Ort::ConstValueInfo& value_info, int id,
+                               size_t& type_info_failures) {
+  using json = nlohmann::json;
+  json tensor_json = json::object();
+
+  tensor_json["id"] = id;
+  tensor_json["type"] = ClassifyTensorType(value_info);
+  tensor_json["dataFormat"] = 0;
+  tensor_json["src_axis_format"] = "NOT_YET_DEFINED";
+  tensor_json["axis_format"] = "NOT_YET_DEFINED";
+
+  // Non-quantized placeholder: this dumper records ONNX-level structure, not
+  // QDQ scale/zero-point, so the encoding is reported as undefined.
+  tensor_json["quant_params"] = {
+      {"definition", QNN_DEFINITION_UNDEFINED},
+      {"encoding", QNN_QUANTIZATION_ENCODING_UNDEFINED},
+      {"scale_offset", {{"scale", 0.0}, {"offset", 0}}},
+  };
+
+  // data_type + dims are read from the value's type info, which may be absent.
+  // Default to FLOAT_32 + empty dims when unavailable, and treat dynamic dims
+  // as "no shape" (empty array) rather than emitting negative sizes.
+  int data_type = QNN_DATATYPE_FLOAT_32;
+  json dims = json::array();
+
+  const OrtValueInfo* raw = value_info;
+  if (raw != nullptr) {
+    try {
+      Ort::ConstTypeInfo type_info = value_info.TypeInfo();
+      Ort::ConstTensorTypeAndShapeInfo shape_info = type_info.GetTensorTypeAndShapeInfo();
+
+      ONNXTensorElementDataType elem_type = shape_info.GetElementType();
+      Qnn_DataType_t qnn_dt = QNN_DATATYPE_FLOAT_32;
+      if (utils::OnnxDataTypeToQnnDataType(elem_type, qnn_dt, /*is_quantized=*/false)) {
+        data_type = qnn_dt;
+      } else {
+        // The mapping helper does not yet cover this ONNX dtype (e.g. some
+        // 4-bit / brain-float variants). Reuse the same accumulator as the
+        // type-info exception path so the user still sees a single
+        // aggregated WARNING from the dumper instead of a silent FLOAT_32
+        // substitution in the JSON.
+        ++type_info_failures;
+      }
+
+      if (shape_info.HasShape()) {
+        std::vector<int64_t> shape = shape_info.GetShape();
+        bool has_dynamic = false;
+        for (int64_t s : shape) {
+          if (s < 0) {
+            has_dynamic = true;
+            break;
+          }
+        }
+        if (!has_dynamic) {
+          for (int64_t s : shape) {
+            dims.push_back(s);
+          }
+        }
+      }
+    } catch (const Ort::Exception&) {
+      ++type_info_failures;
+    }
+  }
+
+  tensor_json["data_type"] = data_type;
+  tensor_json["dims"] = std::move(dims);
+  return tensor_json;
+}
+
+// Collect the input/output tensor-name array for a node, recording each tensor
+// in `tensors_json` (deduped) and skipping null optional inputs.
+nlohmann::json CollectNodeTensorNames(const std::vector<Ort::ConstValueInfo>& value_infos,
+                                      TensorIdAllocator& id_alloc,
+                                      std::unordered_set<std::string>& seen_tensors,
+                                      nlohmann::json& tensors_json,
+                                      size_t& type_info_failures) {
+  nlohmann::json names = nlohmann::json::array();
+  for (const Ort::ConstValueInfo& vi : value_infos) {
+    const OrtValueInfo* raw = vi;
+    if (raw == nullptr) {
+      continue;  // optional input not provided
+    }
+    std::string name = std::string(vi.GetName());
+    if (name.empty()) {
+      continue;
+    }
+    names.push_back(name);
+    if (!seen_tensors.count(name)) {
+      seen_tensors.insert(name);
+      tensors_json[name] = BuildTensorJson(vi, id_alloc.GetOrAdd(name), type_info_failures);
+    }
+  }
+  return names;
+}
+
+}  // namespace
+
+bool DumpQnnEpInputGraphToJson(const OrtGraph* graph,
+                               const std::filesystem::path& output_path,
+                               const Ort::Logger& logger) {
+  using json = nlohmann::json;
+
+  json root = {
+      // Dummy model.cpp / model.bin: not required for Netron visualization.
+      {"model.cpp", "N/A"},
+      {"model.bin", "N/A"},
+      {"converter_command", ""},
+      {"copyright_str", "Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries."},
+      {"op_types", json::array()},
+      {"Total parameters", ""},
+      {"Total MACs per inference", ""},
+      {"graph", {{"tensors", json::object()}, {"nodes", json::object()}}},
+  };
+
+  json& tensors_json = root["graph"]["tensors"];
+  json& nodes_json = root["graph"]["nodes"];
+
+  TensorIdAllocator id_alloc;
+  std::unordered_set<std::string> seen_tensors;
+  // Sorted set so the serialized op_types array order is stable across runs
+  // and platforms (downstream diff/hash tools will not see hash-iteration
+  // noise).
+  std::set<std::string> seen_op_types;
+  size_t type_info_failures = 0;
+
+  Ort::ConstGraph ort_graph{graph};
+
+  // Nodes: one entry per ONNX node, keyed by node name (fall back to
+  // op_type + index when unnamed so the JSON object key stays unique).
+  std::vector<Ort::ConstNode> nodes = ort_graph.GetNodes();
+  for (size_t i = 0; i < nodes.size(); ++i) {
+    const Ort::ConstNode& node = nodes[i];
+
+    std::string op_type = std::string(node.GetOperatorType());
+    std::string node_name = std::string(node.GetName());
+    if (node_name.empty()) {
+      node_name = op_type + "_" + std::to_string(i);
+    }
+
+    json node_json = json::object();
+    node_json["package"] = "onnx";
+    node_json["type"] = op_type;
+    node_json["input_names"] =
+        CollectNodeTensorNames(node.GetInputs(), id_alloc, seen_tensors, tensors_json, type_info_failures);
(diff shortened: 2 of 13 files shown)


qti-chuteng (approved):

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -792,6 +802,89 @@ ort.unregister_execution_provider_library(ep_registration_name)
...
+choose the output directory (defaults to the current working directory).
+
+This is useful for inspecting exactly what the EP sees: that graph is **not** the
+user's source `.onnx` (ORT has run constant folding, QDQ duplication via
+`EnsureUniqueDQForNodeUnit`, Transpose optimization, etc.), and it is otherwise
+hard to observe — `SessionOptions::optimized_model_filepath` refuses to serialize
+a graph once an EP has compiled part of it into opaque compiled nodes ("Unable to
+serialize model as it contains compiled nodes").
+
+### Why the dump is QNN-Netron JSON, not an ONNX model
```
huaychou, line 837: Is QNN-NETRON a public or well-defined schema?

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -792,6 +802,89 @@ ort.unregister_execution_provider_library(ep_registration_name)
...
+**Known limitation:** the dump does **not** include initializer tensor data
+bytes. Only the tensor name, dtype, shape, and role are recorded. Tools
+consuming the dump (for example `source_to_optimized_matcher.py`'s
+`--optimized-model` input) must therefore rely on initializer **names** or
+graph **topology** — not data hashes — when correlating tensors across the
+source and the EP-input graph. The matcher uses a name-derived sentinel hash
+for initializers from this input so name-based and topology-based matching
+work unchanged.
+
+### Multiple dumps per graph name
```
huaychou, line 877: Could we just dump the last graph before QNN EP?

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
```diff
@@ -0,0 +1,304 @@
...
+
+namespace {
+
+// QNN tensor-type integers Netron uses to classify a tensor. Mirrors the
+// values emitted by the post-compile QnnJSONGraph path (qnn_utils.cc), which
+// serializes the raw Qnn_TensorType_t enum.
+constexpr int kQnnTensorTypeAppWrite = QNN_TENSOR_TYPE_APP_WRITE;  // graph input  = 0
+constexpr int kQnnTensorTypeAppRead = QNN_TENSOR_TYPE_APP_READ;    // graph output = 1
+constexpr int kQnnTensorTypeNative = QNN_TENSOR_TYPE_NATIVE;       // intermediate = 3
+constexpr int kQnnTensorTypeStatic = QNN_TENSOR_TYPE_STATIC;       // initializer  = 4
```
huaychou, line 30: Should we define these in header file?

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
```diff
@@ -0,0 +1,304 @@
...
+
+// Assigns a stable, incrementing integer id per tensor name. The public
+// ValueInfo API exposes no per-value id, and Netron only needs uniqueness, so
+// a name->int allocator suffices.
+class TensorIdAllocator {
```
huaychou, line 35: ditto

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -217,6 +218,17 @@ class QnnEp : public OrtEp, public ApiPtrs {
...
+  // === Qnn Ep Input Graph ===
+  // Dumps the ONNX graph the EP receives in GetCapabilityImpl (compile-time,
+  // pre-partition) as a QNN-Netron-schema JSON. Distinct from
+  // dump_json_qnn_graph_ above, which dumps the post-compile QNN graph.
+  bool dump_qnn_ep_input_graph_ = false;
+  std::string dump_qnn_ep_input_graph_dir_;
+  // Per-EP counter that disambiguates output filenames. Always appended to
+  // the filename so two graphs whose names sanitize to the same string still
+  // produce two distinct files.
+  std::atomic<size_t> dump_qnn_ep_input_graph_count_{0};
```
huaychou: Why do we need to use atomic here? Is there any multi-thread use case?

qcom/tools/op_trace_matcher/source_to_optimized_matcher.py
```diff
@@ -0,0 +1,1197 @@
...
+# Sequences are listed in source-graph topological order (input -> output),
+# which matches how the matcher walks back from the optimized node's input.
+#
+# Only LINEAR source chains can be expressed here — the walker follows one
+# input edge per step. Fusions with branching source patterns (LayerNorm,
+# EmbedLayerNorm, Attention, GroupQueryAttention, the Tanh-form FastGelu, …)
+# typically rely on the matcher's initializer-anchoring or lineage-extension
+# strategies instead. Patterns derived by reading the corresponding fusion
+# files in onnxruntime/core/optimizer/.
+FUSION_PATTERNS: dict[str, list[list[str]]] = {
```
huaychou, line 289: Do we need to update this when adding new fusion patterns? Is it possible to put them in the same place to reduce maintenance effort?

qcom/tools/op_trace_matcher/test_op_trace_matcher.py
```diff
@@ -0,0 +1,422 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+"""Unit tests for `source_to_optimized_matcher` and `enrich_profiling_csv`.
```
huaychou, line 3: Do we run this unit test in CI?

docs/execution_providers/QNN-ExecutionProvider.md
kuanyul-qti: Yes, Netron is the open-source neural-net viewer that already understands the JSON shape produced by QNN's existing `dump_json_qnn_graph`.

docs/execution_providers/QNN-ExecutionProvider.md
kuanyul-qti: No. ORT's plugin EP API gives no signal from inside `GetCapabilityImpl` to predict whether a subsequent call will come (layout-transform 2nd pass / EPContext / subgraphs). The EP would need to dump every call and filter post-hoc.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
kuanyul-qti: They live in the anonymous namespace of the .cc — translation unit-local helpers consumed only by BuildTensorJson / ClassifyTensorType in the same translation unit. Exposing in the header would add unused public surface.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
kuanyul-qti: Ditto

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Changed to `size_t`.

qcom/tools/op_trace_matcher/test_op_trace_matcher.py
kuanyul-qti: No. There will be a separate PR to manage the testing of all test_* files in CI.

qcom/tools/op_trace_matcher/source_to_optimized_matcher.py
kuanyul-qti: Yes. It only exists in this place.

huaychou: Do we need to manually add fusion patterns if a new pattern is introduced in onnxruntime/core/optimizer?

kuanyul-qti: Yes — manual update to FUSION_PATTERNS in source_to_optimized_matcher.py, single place. Failure mode is graceful, not silent:

  - New fused op falls through to initializer-name / hash anchoring → still gets a Medium-confidence multi-source match in most cases (weights usually survive fusion verbatim).
  - Anything truly unmapped lands in unmapped_optimized_nodes in the output JSON — one grep tells the contributor what pattern to add.

  We don't auto-derive because ORT fusions are hand-rolled C++ pattern matchers with no machine-readable schema.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.h
```diff
@@ -0,0 +1,91 @@
...
+// the result is never `..` or a hidden-file form), and trailing dots and
+// spaces are dropped (a Windows-reserved suffix). Windows reserved device
+// names (CON, PRN, AUX, NUL, COM1..9, LPT1..9, case-insensitive) are
+// suffixed with `_` so they no longer match the device-name pattern. An
+// input that sanitizes to the empty string returns "graph", giving callers a
+// non-empty fallback they can disambiguate further.
+//
+// Defined inline so the unit tests in onnxruntime_provider_test (which does
+// not link onnxruntime_providers_qnn) can call it directly.
+inline std::string SanitizeGraphNameForFilename(const std::string& graph_name) {
```
huaychou: **[N-1] (Minor)** SanitizeGraphNameForFilename defined inline in a header that pulls in private ort_api.h, violating the Pure-Data + Free-Function Header Split pattern

`SanitizeGraphNameForFilename` is a pure string-processing function (needs only STL), but it is defined inline in `qnn_ep_input_graph_dumper.h`, which `#include`s the private EP-internal `core/providers/qnn/ort_api.h`. The test file `qnn_basic_test.cc` includes this header (see T-2) solely to call this pure function, but is forced to transitively compile the full private ORT API wrapper.

**Impact:** Violates the Pure-Data + Free-Function Header Split boundary (B7), spreading private EP implementation includes into the test binary. The header comment at line 45 acknowledges the design trade-off, but the structural coupling remains. Flagged in v1 review; not addressed by the "Fix upon comments" commit.

**Suggested fix:**

Extract `SanitizeGraphNameForFilename` and its five STL-only includes into a new header (e.g. `qnn_ep_sanitize_utils.h`) that does NOT include `ort_api.h`. Have `qnn_ep_input_graph_dumper.h` include the new header. The test file can then include only the clean, dependency-free header, removing the private-header coupling from the test boundary.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1689,6 +1743,28 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
...
+  // Best-effort diagnostic: failures are logged inside the dumper and never
+  // abort compilation. The filename always carries a per-EP counter so two
+  // graphs whose names sanitize to the same string produce two distinct
+  // files; the sanitized name (when present) is used as a human-readable
+  // prefix.
+  if (ep->dump_qnn_ep_input_graph_) {
+    Ort::ConstGraph dump_graph{graph};
+    std::string raw_name = std::string(dump_graph.GetName());
+    std::string graph_name;
+    if (raw_name.empty()) {
```
huaychou: **[N-2] (Minor)** Redundant empty-string guard in GetCapabilityImpl duplicates SanitizeGraphNameForFilename's built-in fallback

The call site reads: `if (raw_name.empty()) { graph_name = "graph"; } else { graph_name = qnn::SanitizeGraphNameForFilename(raw_name); }`. However, `SanitizeGraphNameForFilename` already returns `"graph"` when the input sanitizes to empty, making the outer guard redundant.

**Impact:** Adds cognitive overhead — a reader must understand both the call-site guard and the function's internal fallback. Flagged in v1 review; not addressed.

**Suggested fix:**

Replace the two-branch block with a single call: `std::string graph_name = qnn::SanitizeGraphNameForFilename(std::string(dump_graph.GetName()));`

qcom/tools/op_trace_matcher/enrich_profiling_csv.py
```diff
@@ -0,0 +1,481 @@
...
+    input has no NODE-level events (BASIC profiling — nothing to enrich).
+    `merged_trace` is the parsed JSON dict (loaded from disk in Mode A,
+    computed in-memory in Mode B). Returns a stats dict.
+
+    Raises ValueError if the input CSV is empty, missing the
+    `Event Identifier` column, or already carries the `Original ONNX Source Ops`
+    column (i.e. is itself an enricher output)."""
+    originals_lookup, optimized_lookup = build_lookups(merged_trace)
+
+    stats = {
```
huaychou, line 175: **[N-3] (Minor)** enrich() stats dict missing 'filled_existing_onnx_column' key — early-return path causes KeyError for callers

The `stats` dict initializer (lines 175–183) does not include `'filled_existing_onnx_column'`. The key is written dynamically at line 269, but the early return at lines 248–253 (`if not has_node_events: shutil.copyfile(...); return stats`) exits before reaching line 269.

**Impact:** Any caller that accesses `stats["filled_existing_onnx_column"]` will get a `KeyError` for BASIC-profiling CSVs (no node events), creating an inconsistent return contract. Flagged in v1 review; not addressed.

**Suggested fix:**

Add `"filled_existing_onnx_column": False` to the `stats` dict initializer at lines 175–183 alongside the other keys. The dynamic assignment at line 269 can remain as an update.

qcom/tools/op_trace_matcher/source_to_optimized_matcher.py
```diff
@@ -0,0 +1,1197 @@
...
+            # guaranteed deterministic across re-serializations, but the
+            # underlying numpy buffer is — this keeps the hash stable when the
+            # source and optimized graphs were each round-tripped through
+            # different ONNX load/save passes.
+            if init.raw_data:
+                payload = init.raw_data
+            elif _onnx_numpy_helper is not None:
+                try:
+                    payload = _onnx_numpy_helper.to_array(init).tobytes()
+                except Exception:
```
huaychou: **[N-4] (Minor)** GraphIndex.build uses overly broad except Exception for numpy conversion failure

Line 158 uses `except Exception:` as a fallback when `_onnx_numpy_helper.to_array(init).tobytes()` fails, reverting to `SerializeToString()`. While the intent is documented, bare `except Exception` also swallows `MemoryError` and unexpected attribute errors on `_onnx_numpy_helper` itself.

**Impact:** Masks genuine programming bugs, causing the fallback path to silently activate instead of surfacing the problem. Flagged in v1 review (as T-1); not addressed.

**Suggested fix:**

Narrow to `except (ValueError, TypeError, RuntimeError):` — the concrete types that `numpy_helper.to_array` raises for unsupported or malformed tensor data. Add a comment enumerating the covered types.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1689,6 +1743,28 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
...
+      graph_name = "graph";
+    } else {
+      graph_name = qnn::SanitizeGraphNameForFilename(raw_name);
+    }
+    size_t count = ep->dump_qnn_ep_input_graph_count_++;
+    std::filesystem::path out_path =
+        std::filesystem::path(ep->dump_qnn_ep_input_graph_dir_) /
+        (graph_name + "." + std::to_string(count) + "_qnn_ep_input_graph.json");
+    qnn::DumpQnnEpInputGraphToJson(graph, out_path, ep->logger_);
```
huaychou: **[T-1] (Nit)** Return value of DumpQnnEpInputGraphToJson (bool) silently discarded in GetCapabilityImpl

The call `qnn::DumpQnnEpInputGraphToJson(graph, out_path, ep->logger_)` discards its `bool` return value without a `(void)` cast or explanatory comment. The function logs a WARNING and returns `false` on failure, so discarding is consistent with the best-effort diagnostic design, but the intent is not immediately clear from the code.

**Impact:** Static analysis tools will flag this site; a future `[[nodiscard]]` attribute would produce a warning here.

**Suggested fix:**

Add a comment `// Best-effort: failures are already logged inside the dumper.` above the call, or cast to void: `(void)qnn::DumpQnnEpInputGraphToJson(graph, out_path, ep->logger_);`

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -14,6 +14,7 @@
 #include "onnxruntime_session_options_config_keys.h"
 
 #include "core/providers/qnn/builder/op_package/op_package_parser.h"
+#include "core/providers/qnn/builder/qnn_ep_input_graph_dumper.h"
```
huaychou: **[T-2] (Nit)** Test file includes private provider header qnn_ep_input_graph_dumper.h — B7 boundary violation, direct consequence of N-1

`qnn_basic_test.cc` adds `#include "core/providers/qnn/builder/qnn_ep_input_graph_dumper.h"` at line 17. This header pulls in private EP-internal symbols via `ort_api.h`, violating B7. This is a direct consequence of N-1: extracting `SanitizeGraphNameForFilename` to a clean header would let this include be replaced.

**Impact:** Extends the test binary's private include path. Root fix is N-1.

**Suggested fix:**

After applying N-1, replace this include with the new dependency-free header (e.g. `#include "core/providers/qnn/builder/qnn_ep_sanitize_utils.h"`).

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
```diff
@@ -0,0 +1,304 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "core/providers/qnn/builder/qnn_ep_input_graph_dumper.h"
+
+#include <algorithm>
```
huaychou: **[T-3] (Nit)** qnn_ep_input_graph_dumper.cc redundantly re-includes STL headers already provided by its own header

`qnn_ep_input_graph_dumper.cc` explicitly includes `<algorithm>`, `<cctype>`, `<string>`, `<unordered_set>` (lines 6–13), already transitively provided by its own header. The `.cc` body does not call any symbol from these headers directly — they are only used by the inline function in the header.

**Impact:** Creates false documentation of direct dependencies; will be flagged by IWYU / clang-include-cleaner.

**Suggested fix:**

Remove `#include <algorithm>`, `<cctype>`, `<string>`, `<unordered_set>` from the `.cc` file.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <atomic>
```
huaychou, line 6: **[T-4] (Nit)** #include <atomic> added by this PR but not needed by the new dump fields — missing explanatory comment

This PR adds `#include <atomic>` to `qnn_execution_provider.h`, but the three new dump fields (`bool`, `std::string`, `size_t`) do not use `std::atomic`. The only `std::atomic` usage is the pre-existing `mutable std::atomic<uint64_t> genie_kv_cache_rewind_`.

**Impact:** Making an implicit dependency explicit is good hygiene, but without a comment readers may incorrectly assume it is needed for the new dump fields.

**Suggested fix:**

Verify whether this include was previously provided transitively (try removing it). If genuinely new, add a comment: `// for std::atomic (genie_kv_cache_rewind_)`.

qcom/tools/op_trace_matcher/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
huaychou, line 1: **[T-5] (Nit)** __init__.py defines no __all__ and re-exports no public symbols — low discoverability

`__init__.py` is 7 lines of module docstring and version string with no `__all__` and no re-exports from submodules. Both submodules define their own `__all__` correctly, but the package provides no stable top-level API surface.

**Impact:** Users must spell out full submodule paths (e.g. `from qcom.tools.op_trace_matcher.source_to_optimized_matcher import Matcher`), reducing discoverability.

**Suggested fix:**

Either add top-level re-exports with a combined `__all__`, or document in the package docstring that users should import from submodules directly.

qcom/tools/op_trace_matcher/source_to_optimized_matcher.py
```diff
@@ -0,0 +1,1197 @@
...
+            }
+        )
+        if i < len(source_nodes) - 1:
+            connecting = _connecting_tensor(src_index, name, source_nodes[i + 1])
+            if connecting:
+                chain.append({"name": connecting, "type": _TRACE_TYPE_TENSOR})
+    return chain
+
+
+def join_qnn_trace(
```
huaychou, line 962: **[T-6] (Nit)** join_qnn_trace second return value (stats dict) has no documented key schema

`join_qnn_trace` is typed as `tuple[dict, dict]` but the stats dict key schema (e.g. `op_sources_unresolved`, `tensor_sources_unresolved`) is absent from any `Returns:` section in the docstring. Callers must read `main()` to infer the keys.

**Impact:** As a public API exported in `__all__`, the return contract is opaque to library users without reading the function body.

**Suggested fix:**

Add a `Returns:` section to the `join_qnn_trace` docstring enumerating stats dict keys and types, or define a `JoinStats` TypedDict as the return annotation.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
```diff
@@ -0,0 +1,304 @@
...
+bool DumpQnnEpInputGraphToJson(const OrtGraph* graph,
+                               const std::filesystem::path& output_path,
+                               const Ort::Logger& logger) {
+  using json = nlohmann::json;
+
+  json root = {
+      // Dummy model.cpp / model.bin: not required for Netron visualization.
+      {"model.cpp", "N/A"},
+      {"model.bin", "N/A"},
+      {"converter_command", ""},
```
huaychou, line 183: Are these required for QNN-Netron? Is there any document for QNN-Netron format?

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.h
kuanyul-qti: Fixed. Extracted to new `qnn_ep_sanitize_utils.h` with only STL deps.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Fixed. Collapsed to single call with explanatory comment.

qcom/tools/op_trace_matcher/enrich_profiling_csv.py
kuanyul-qti: Fixed. Added `"filled_existing_onnx_column": False` to initializer.

qcom/tools/op_trace_matcher/source_to_optimized_matcher.py
kuanyul-qti: Fixed. Narrowed to `except (ValueError, TypeError, RuntimeError)` with a comment explaining each.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Fixed. Added comment "Best-effort diagnostic: failures are already logged inside the dumper, so the bool return is intentionally discarded."

onnxruntime/test/providers/qnn/qnn_basic_test.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
kuanyul-qti: Fixed. Removed <algorithm> and <cctype> only; kept <string> and <unordered_set> because they are direct dependencies of the .cc body

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Fixed.

qcom/tools/op_trace_matcher/__init__.py
kuanyul-qti: Fixed (docs). Expanded the docstring to explicitly state "Public APIs are intentionally exposed only on the submodules" with example imports. No re-exports added (intentional, to keep submodule boundaries clean and let Mode A skip the onnx dependency).

qcom/tools/op_trace_matcher/source_to_optimized_matcher.py
kuanyul-qti: Fixed. Added a `Returns` section enumerating the 6 actual stats keys

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
kuanyul-qti: These are not all strictly required by Netron's parser, but they are present in the schema produced by the existing `dump_json_qnn_graph` path (see `QnnJSONGraph` at `onnxruntime/core/providers/qnn/builder/qnn_utils.cc`). The dumper here intentionally mirrors that schema 1:1 so the same JSON shape works in both code paths and Netron sees a familiar structure.

There is no formal QNN-Netron schema specification published. The de-facto schema is observable in two places:
- Producer side: `qnn_utils.cc::QnnJSONGraph` — what QNN EP emits.
- Consumer side: Netron's QNN parser at `https://github.com/lutzroeder/netron/blob/main/source/qnn.js` — what Netron reads.

The dummy values ("N/A", "") for fields that don't apply to the EP-input graph (no compiled model.cpp, no MAC count) match what the existing `dump_json_qnn_graph` path emits when the corresponding QNN data isn't available. Netron handles them as opaque metadata fields, so emitting placeholders keeps the JSON loadable without needing to mock real values.

onnxruntime/core/providers/qnn/builder/qnn_ep_sanitize_utils.h
```diff
@@ -0,0 +1,74 @@
...
+
+  size_t first = 0;
+  while (first < result.size() && (result[first] == '.' || result[first] == '-')) {
+    ++first;
+  }
+  if (first > 0) {
+    result.erase(0, first);
+  }
+
+  while (!result.empty() && (result.back() == '.' || result.back() == ' ')) {
```
huaychou: How can result have space after safeguard? Should it be '-'?

onnxruntime/core/providers/qnn/builder/qnn_ep_sanitize_utils.h
```diff
@@ -0,0 +1,74 @@
...
+
+  if (result.empty()) {
+    return "graph";
+  }
+
+  std::string stem = result.substr(0, result.find('.'));
+  std::string upper_stem = stem;
+  std::transform(upper_stem.begin(), upper_stem.end(), upper_stem.begin(),
+                 [](unsigned char ch) { return static_cast<char>(std::toupper(ch)); });
+  static const std::unordered_set<std::string> kReserved = {
```
huaychou, line 73: What is the kReserved used for?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1030,6 +1032,59 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    if (!probe_ok) {
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
+                  ("QNN EP input graph dump directory not writable: " + probe_dir.string() +
+                   "; the dump will be disabled.")
+                      .c_str());
+      dump_qnn_ep_input_graph_ = false;
+    } else {
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_INFO,
+                  ("QNN EP input graph dump enabled. Output dir: " + dump_qnn_ep_input_graph_dir_).c_str());
+    }
```
huaychou, line 1098: Can we make these a helper function to create directories and apply to json_qnn_graph_dir_?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1689,6 +1744,27 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
     }
   }
 
+  // Dump the EP-input ONNX graph (pre-partition) as a QNN-Netron-schema JSON.
+  // Best-effort diagnostic: failures are logged inside the dumper and never
+  // abort compilation. The filename always carries a per-EP counter so two
+  // graphs whose names sanitize to the same string produce two distinct
+  // files; the sanitized name (when present) is used as a human-readable
+  // prefix.
+  if (ep->dump_qnn_ep_input_graph_) {
```
huaychou: Shouldn't we dump the graph right after GetCapabilityImpl?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -2837,6 +2838,268 @@ TEST_F(QnnHTPBackendTests, ExtendedUdmaModeTest) {
...
+  // `..foo` must not survive as a parent-dir reference. Leading `-` is
+  // stripped so the result cannot resemble an argument flag if it ever
+  // round-trips through a CLI.
+  EXPECT_EQ(qnn::SanitizeGraphNameForFilename("..foo"), "foo");
+  EXPECT_EQ(qnn::SanitizeGraphNameForFilename("-rf"), "rf");
+  EXPECT_EQ(qnn::SanitizeGraphNameForFilename(".."), "graph");
+  EXPECT_EQ(qnn::SanitizeGraphNameForFilename("."), "graph");
+}
+
+TEST(QnnEpInputGraphDumperTest, SanitizeGraphName_DropsTrailingDotAndSpace) {
```
huaychou: There is no Space in the test cases.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
```diff
@@ -0,0 +1,302 @@
...
+    const Ort::ConstNode& node = nodes[i];
+
+    std::string op_type = std::string(node.GetOperatorType());
+    std::string node_name = std::string(node.GetName());
+    if (node_name.empty()) {
+      node_name = op_type + "_" + std::to_string(i);
+    }
+
+    json node_json = json::object();
+    node_json["package"] = "onnx";
```
huaychou: Do we need to identify contrib ops from onnx ops?

huaychou (commented):
Could we add a requirements.txt under `qcom/tools/op_trace_matcher`?

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/qnn_ep_sanitize_utils.h
kuanyul-qti: Added a 4-line block comment directly above `kReserved` explaining these are Windows-reserved device names, the case-insensitivity, stem-only check, and the "suffix unconditionally for portability" reasoning.

onnxruntime/core/providers/qnn/builder/qnn_ep_sanitize_utils.h
kuanyul-qti: Fixed. Removed the unreachable ' '. Trailing "-" doesn't need stripping.

The reason we strip trailing "." is Windows-specific: `foo.` and `foo` resolve to the same file on NTFS (the OS silently drops trailing dots), so leaving them would make two distinct sanitized graph names collide on disk. Trailing dashes have no such quirk — `foo-` is a perfectly valid, distinct filename on every supported platform.

Leading "-" is a different case and is handled separately (earlier in the function, in the leading-strip loop) — that one is stripped to prevent the sanitized name from looking like a CLI flag if it ever round-trips through a tool that parses argv. Trailing "-" carries no such risk.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done. Extracted `ProbeDumpDirectoryWritable` as a file-local static helper.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
kuanyul-qti: Remove the "space" string in the test name.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Moved.

kuanyul-qti:
> Could we add a requirements.txt under `qcom/tools/op_trace_matcher`?

Added.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
```diff
@@ -0,0 +1,312 @@
...
+  // Nodes: one entry per ONNX node, keyed by node name (fall back to
+  // op_type + index when unnamed so the JSON object key stays unique).
+  std::vector<Ort::ConstNode> nodes = ort_graph.GetNodes();
+  for (size_t i = 0; i < nodes.size(); ++i) {
+    const Ort::ConstNode& node = nodes[i];
+
+    std::string op_type = std::string(node.GetOperatorType());
+    std::string node_name = std::string(node.GetName());
+    if (node_name.empty()) {
+      node_name = op_type + "_" + std::to_string(i);
```
huaychou: One corner case is, one node name is add_1, and another add node doesn't have name with i = 1. We handle this case in qcom/tools/op_trace_matcher/source_to_optimized_matcher.py by adding "__dup{idx}. Do we need to handle it here?

onnxruntime/core/providers/qnn/builder/qnn_ep_sanitize_utils.h
```diff
@@ -0,0 +1,84 @@
...
+#include <string>
+#include <unordered_set>
+
+namespace onnxruntime {
+namespace qnn {
+
+// Returns a filename-safe form of `graph_name`: any character outside
+// [A-Za-z0-9._-] is replaced with `_`, leading dots/hyphens are stripped (so
+// the result is never `..` or a hidden-file form), and trailing dots and
+// spaces are dropped (a Windows-reserved suffix). Windows reserved device
```
huaychou: Only trailing dots are dropped.

huaychou (commented):
Since this is backend agnostic, I think we don't need HTP backend tests, right?

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/qnn_ep_sanitize_utils.h
kuanyul-qti: Fixed.

kuanyul-qti (commented):
Right — no HTP test needed. The dump runs in `GetCapabilityImpl` before any backend work and reads only ONNX-level structure (node names, op types, domains, tensor info) via `Ort::ConstGraph`. That output is identical across CPU/HTP/GPU/IR, so the three `QnnCPUBackendTests cases` cover it fully — CPU also runs on dev/CI hosts without an NPU. An HTP variant would re-exercise the same backend-agnostic path while adding a device dependency for zero new coverage.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    // `package` distinguishes contrib / internal-domain ops (e.g.
+    // `com.microsoft`, `com.ms.internal.nhwc` introduced by ORT's layout
+    // transformer) from the default ONNX op set. Empty / `ai.onnx` is
+    // normalized to "onnx" so the field matches what the post-compile
+    // `dump_json_qnn_graph` path emits for native ONNX nodes.
+    std::string node_domain = std::string(node.GetDomain());
+    json node_json = json::object();
+    if (node_domain.empty() || node_domain == "ai.onnx") {
```
huaychou, line 236: Just wondering, what is the node_domain of EPContext? I think we support EPContext for dump json, right?

qcom/tools/op_trace_matcher/requirements.txt
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+
+# Runtime requirements for source_to_optimized_matcher.py and
+# enrich_profiling_csv.py. For unit tests, additionally `pip install pytest`
+# (or use the repo-wide requirements-dev.txt which already includes all three).
+numpy
+onnx
```
huaychou: nit, add minimum version constraints

huaychou (approved):
LGTM

qcom/tools/op_trace_matcher/requirements.txt
kuanyul-qti: Added.

onnxruntime/core/providers/qnn/builder/qnn_ep_input_graph_dumper.cc
kuanyul-qti: Yes — EPContext models are dumped. The dump now fires at the very top of `GetCapabilityImpl` (right after the ep cast), before the `is_qnn_ctx_model` / Genie / shared-context early-returns, so a session loading a cached EPContext model produces a dump too.
The EPContext node's domain is `kMSDomain` (`com.microsoft`) — it's created with that domain in `onnx_ctx_model_helper.cc:CreateEPContextNodes`. With the domain-aware package field, the dumped node will show `"package": "com.microsoft", "type": "EPContext"`, so it's distinguishable from native ONNX ops in both Netron and the matcher.

PR #494, opened 2026-06-08
[QNN EP] Add support for QLinearConv Op

qti-mbadnara:
### Description

Adds support for the ONNX `QLinearConv` operator (opset 10) in the QNN Execution Provider, targeting the QNN HTP and CPU backends via the native `QNN_OP_CONV_2D`, `QNN_OP_DEPTH_WISE_CONV_2D`, and `QNN_OP_CONV_3D` ops.

---

### Motivation & Context

`QLinearConv` is the opset-10 explicit-quantization form of convolution, produced by PyTorch's static quantization export path. Prior to this change, any model containing `QLinearConv` nodes would silently fall back to CPU EP, preventing HTP acceleration for a common class of pre-quantized models.

ONNX defines `QLinearConv` at opset 10: https://onnx.ai/onnx/operators/onnx__QLinearConv.html

Unlike the standard `Conv` op (which relies on surrounding QDQ nodes for quantization metadata), `QLinearConv` encodes scale and zero-point as 8 explicit op inputs (`x_scale`, `x_zero_point`, `w_scale`, `w_zero_point`, `y_scale`, `y_zero_point`, plus
an optional `B` bias). This requires a dedicated builder that extracts quant params directly from initializer inputs rather than from QDQ `NodeUnit` metadata.

**QNN op mapping:**

| ONNX op | Condition | QNN op |
|---------|-----------|--------|
| `QLinearConv` | 1D or 2D, non-depthwise | `QNN_OP_CONV_2D` |
| `QLinearConv` | 2D, `group == in_channels == out_channels` | `QNN_OP_DEPTH_WISE_CONV_2D` |
| `QLinearConv` | 3D | `QNN_OP_CONV_3D` |

Because the activation is a bare quantized integer tensor with no quant metadata, ORT's layout transformer cannot safely insert an unquantized Transpose on it (HTP rejects an unquantized uint8/int8 Transpose). `ShouldConvertDataLayoutForOpImpl` suppresses the layout transform for `QLinearConv`; the builder performs the
NCHW↔NHWC transposition internally with quant params attached to every tensor (same strategy as `DQConvIntegerFusion` for `ConvInteger`).

---

### Supported Configuration

| Attribute / Input | Supported values |
|-------------------|-----------------|
| `x` dtype | int8, uint8 |
| `w` dtype | int8, uint8 |
| `y` dtype | int8, uint8 |
| Activation quant | Per-tensor only (scalar `x_scale` / `y_scale`) |
| Weight quant | Per-tensor or per-channel (1-D `w_scale` of size M, axis 0) |
| Bias | Optional INT32; `bias_scale = x_scale * w_scale` |
| Input rank | 3 (1D), 4 (2D), 5 (3D) |
| `strides` | Any |
| `pads` | Any (explicit or `auto_pad=SAME_UPPER`) |
| `dilations` | Any |
| `group` | Any (depthwise automatically detected) |

---

### Unsupported Configuration (falls back to CPU EP)

- Dynamic `x_scale`, `x_zero_point`, `w_scale`, `w_zero_point`, `y_scale`, or `y_zero_point` — QNN requires compile-time constant quant params.
- Per-channel activation or output quantization — QNN only supports per-channel on weights (axis 0).
- GPU backend — `QLinearConv` is HTP/CPU only.

---

### Data Types

| Backend | Activation (x/y) | Weight (w) |
|---------|-----------------|------------|
| CPU     | uint8           | uint8      |
| HTP     | int8, uint8     | int8, uint8 (per-tensor or per-channel) |

Mixed types (e.g. uint8 activation + int8 weight) are supported on HTP.

---

### Files Changed

| File | Change |
|------|--------|
| `builder/opbuilder/qlinear_conv_op_builder.cc` | **New** — `QLinearConvOpBuilder` |
| `builder/op_builder_factory.h` | Added `CreateQLinearConvOpBuilder` declaration |
| `builder/op_builder_factory.cc` | Registered `CreateQLinearConvOpBuilder("QLinearConv", *this)` |
| `qnn_execution_provider.cc` | Added `ShouldConvertDataLayoutForOpImpl` suppression for `QLinearConv` to enable internal layout handling |
| `test/providers/qnn/qlinear_conv_test.cc` | **New** — HTP accuracy tests (u8, s8, mixed, per-channel, bias, 1D/2D/3D, depthwise, strides, pads, SAME_UPPER) + rejection tests for dynamic scale/zp |
| `docs/execution_providers/QNN-ExecutionProvider.md` | Added `ai.onnx:QLinearConv` to Supported Operators table |


docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -435,6 +435,7 @@ ort.unregister_execution_provider_library(ep_registration_name)
 |ai.onnx:PRelu|fp16, int32 supported since 1.18.0|
 |ai.onnx:Pad||
 |ai.onnx:Pow||
+|ai.onnx:QLinearConv|All scale/zero-point inputs must be initializers; per-tensor activation/output quant; per-channel weight quant (1-D w_scale/w_zp) supported; input types int8 or uint8; optional int32 bias; 1D, 2D, and 3D convolution|
```
qti-yuduo, line 585: My main question is mostly why we need to support this? Currently QDQ is our canonical path, and well supported and tested. Adding each Q variant (parallel-builder route) does not sounds scalable and easy to maintain (would duplicate many code), have we considered unfused at graph step and let existing QDQ support does their job?

qti-yuduo: Also, would like to get opinion from @yath1

yath1: I agree with Yuduo. @qti-mbadnara what is the motivation to add this op? If its to support a model which has `QLinearConv` node, then it can be converted to represent using QDQs and `Conv` op which would work through our current code. If we go this path, we would have to repeat this code for fusions, any changes/fixes we make for the Conv op needs to be repeated for `QLinearConv` op which is unnecessary since we do not see this op much. 

PS: You can use onnxscript to update this op to QDQ+Conv before taking it through EP to get the op to work.

qti-mbadnara: Yes agreed. Changes have been made to implement this Op as a node group fusion.

yath1:
Hello @qti-mbadnara I see you have changed the implementation from your earlier commit of op-builder. But, could you explain the high-level motivation for adding support for this op? Personally I feel, users going through this flow `ONNX model -> quantize -> EP` or `Pytorch model -> quantize -> export to QDQ ONNX -> EP` would not see this op.

tirupath-qti:
> Hello @qti-mbadnara I see you have changed the implementation from your earlier commit of op-builder. But, could you explain the high-level motivation for adding support for this op? Personally I feel, users going through this flow `ONNX model -> quantize -> EP` or `Pytorch model -> quantize -> export to QDQ ONNX -> EP` would not see this op.

QDQ ONNX is one way to represent quant graphs. There is another way with using Quant operators like QLinearMatMul and QLinearConv. We received some models with these operators in the past for Native QAIRT converter support. Since these are standard ONNX operators, they will likely present in any future customer models. Instead of waiting for pressing customer request, we need to add these operators support upfront. Btw this also achieves parity with Native QAIRT Converters too.

tirupath-qti (commented):
This should be implemented as Op builder like other operators rather than node group. We should not make any changes to Conv op builder for this operator.

docs/execution_providers/QNN-ExecutionProvider.md
yath1: @qti-mbadnara just an FYI. When I added this above "I agree with Yuduo. @qti-mbadnara what is the motivation to add this op? If its to support a model which has QLinearConv node, then it can be converted to represent using QDQs and Conv op which would work through our current code. If we go this path, we would have to repeat this code for fusions, any changes/fixes we make for the Conv op needs to be repeated for QLinearConv op which is unnecessary since we do not see this op much." 
I do not mean we need to make this QLinearConv Op a fusion pass. My concern was about code maintenance which needs to be done for QLinearConv along with Conv op - we have discussed this topic in detail offline.

qti-kromero:
@qti-mbadnara can we move this to draft for the time being until we move to op builder as described above by Tiru?

qti-mbadnara:
@qti-kromero @tirupath-qti @yath1 @qti-yuduo The Op has been reimplemented as a standalone op

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
```diff
@@ -0,0 +1,713 @@
...
+  out_quant_param = QnnQuantParamsWrapper::PerTensor(scale, offset);
+  return Ort::Status();
+}
+
+Ort::Status QLinearConvOpBuilder::BuildWeightQuantParam(const QnnModelWrapper& qnn_model_wrapper,
+                                                        const OrtNodeUnitIODef& scale_input,
+                                                        const OrtNodeUnitIODef& zp_input,
+                                                        uint32_t num_output_channels,
+                                                        QnnQuantParamsWrapper& out_quant_param) {
+  RETURN_IF(!scale_input.Exists(), "QLinearConv: w_scale input does not exist.");
```
qti-kromero, line 167: nit: slight personal preference for RETURN_IF_NOT in these simple ! cases

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
```diff
@@ -0,0 +1,713 @@
...
+    std::vector<int32_t> bias_offsets(w_scales.size(), 0);
+    out_quant_param = QnnQuantParamsWrapper::PerChannel(gsl::span<const float>(bias_scales),
+                                                        gsl::span<const int32_t>(bias_offsets),
+                                                        /*axis=*/0);
+  }
+  return Ort::Status();
+}
+
+// ---------------------------------------------------------------------------
+// Layout helpers
```
qti-kromero: these layout helpers seem kind of like common utility functions that should exist elsewhere already or in some common location

onnxruntime/test/providers/qnn/qlinear_conv_test.cc
```diff
@@ -0,0 +1,336 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "qnn_test_utils.h"
```
qti-kromero: doesn't this and line 13 point to the same file

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
```diff
@@ -0,0 +1,713 @@
...
+
+    const OrtApi& ort_api = qnn_model_wrapper.GetOrtApi();
+    const OrtTypeInfo* type_info = nullptr;
+    ORT_CXX_RETURN_ON_API_FAIL(ort_api.GetValueInfoTypeInfo(
+        static_cast<const OrtValueInfo*>(w_info.initializer_tensor), &type_info));
+    const OrtTensorTypeAndShapeInfo* type_shape = nullptr;
+    ORT_CXX_RETURN_ON_API_FAIL(ort_api.CastTypeInfoToTensorInfo(type_info, &type_shape));
+    ONNXTensorElementDataType elem_type = ONNX_TENSOR_ELEMENT_DATA_TYPE_UNDEFINED;
+    ORT_CXX_RETURN_ON_API_FAIL(ort_api.GetTensorElementType(type_shape, &elem_type));
+    const size_t elem_byte_size = utils::GetElementSizeByType(elem_type);
```
qti-kromero: why can't inputs[kIdxW].type be used here?

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
```diff
@@ -0,0 +1,713 @@
...
+    RETURN_IF_ERROR(quant_w.HandleTranspose<size_t>(perm_inv));
+  }
+
+  const std::string w_hwcn_name = node_base + "_w_hwcn";
+  {
+    QnnTensorWrapper w_tensor(w_hwcn_name, QNN_TENSOR_TYPE_STATIC, qnn_dtype_w, quant_w.Copy(),
+                              std::vector<uint32_t>(hwcn_shape), std::vector<uint8_t>(w_bytes));
+    if (do_op_validation) {
+      // For validation we still register the tensor so QNN can type-check the Conv node.
+      RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(w_tensor)), "Failed to add weight tensor.");
```
qti-kromero: these branches appear to be the same

onnxruntime/test/providers/qnn/qlinear_conv_test.cc
```diff
@@ -0,0 +1,336 @@
...
+    builder.MakeScalarInitializer<AType>("x_zp", qp_x.zero_point);
+
+    // w / w_scale / w_zp — per-tensor or per-channel (axis 0 = M).
+    std::vector<WType> q_w(num_w);
+    if (per_channel_weight) {
+      const size_t per_oc = num_w / static_cast<size_t>(M);
+      std::vector<float> w_scales(static_cast<size_t>(M));
+      std::vector<WType> w_zps(static_cast<size_t>(M));
+      for (int64_t oc = 0; oc < M; ++oc) {
+        const auto qp_w = ComputeQuantParams<WType>(-0.3f, 0.3f);
```
qti-kromero: it might be good to vary this per channel such that we are testing the math on a per-channel basis vs. here they could be jumbled and we wouldn't know

onnxruntime/test/providers/qnn/qlinear_conv_test.cc
qti-mbadnara: Thanks for catching!

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
qti-mbadnara: Fixed

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
qti-mbadnara: Noted!

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
qti-mbadnara: Will use them

onnxruntime/core/providers/qnn/builder/opbuilder/qlinear_conv_op_builder.cc
qti-mbadnara: Moved to utils

onnxruntime/core/providers/qnn/builder/op_builder_factory.h
```diff
@@ -63,6 +63,7 @@ void CreateCastOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_
 void CreateClipOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
 void CreateConcatOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
 void CreateConvOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
+void CreateQLinearConvOpBuilder(const std::string& op_type, OpBuilderRegistrations& op_registrations);
```
qti-kromero: nit: this ordering is putting the op with Conv but in docs it is doing in alpha order according to QLinear

tirupath-qti (approved):

PR #495, opened 2026-06-08
[QNN EP] Enable framework op trace for FCB multi-SoCs

kuanyul-qti:
## Summary

Extends the framework op trace (PR #345) to support FCB multi-SoC compilation, and wires it through the FCB compile path so tracing is actually emitted for multi-SoC sessions (previously it was force-disabled whenever multi-SoC EP context was requested).

`FrameworkOpTrace.compilation_target` (single) is replaced by `soc_traces: vector<SocTrace>`, with one entry per SoC iteration. `subgraph_traces` stays at root because op-builder output is SoC-agnostic; it is collected once rather than duplicated per SoC. `unsupported_nodes` also stays at root, but is now merged into one row per ONNX node with each rejection reason attributed to the SoC that produced it.

### Schema change

```json
{
  "soc_traces": [{ "compilation_target": { "htp_arch": ..., "soc_model": ... } }, ...],
  "subgraph_traces": [...],
  "unsupported_nodes": [...],
  "summary": { "total_socs": N, ... }
}
```

Single-SoC sessions degenerate to a length-1 soc_traces array — uniform JSON shape for both single- and multi-SoC traces.

### Multi-SoC wiring

- `CompileMultiSocOnnxModel` appends one `SocTrace` per iteration, then finalizes and writes the trace exactly once per session. The single-SoC path appends one `SocTrace` from the current backend target; both converge on the same Finalize / Write stages.
- `subgraph_traces` are collected only on the first SoC iteration (op-builder output is driven by `ModelSettings`, not HTP graph configs, so it is identical across SoCs).
- The empty-partition early-flush path records every configured SoC, so `summary.total_socs` stays correct even when ORT skips `CompileImpl` because no node is supported.

### Unsupported-node attribution

A node rejected on multiple SoCs is emitted as a single row whose reason prefixes each distinct cause with the SoC that produced it, e.g. "V73: <reason>; V81: <reason>". Identical reasons across SoCs are deduplicated. Keeping one row per node keeps `summary.unsupported_nodes` and `total_onnx_nodes` as true distinct-node counts rather than (SoC × node) counts.

### Breaking change

Consumers reading `compilation_target` at root must migrate to `soc_traces[0].compilation_target`. Migration note added to `docs/execution_providers/QNN-ExecutionProvider.md`.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -16,6 +16,7 @@
 #include <thread>
 #include <unordered_map>
 #include <unordered_set>
+#include <utility>
 #include <vector>
 
 #ifdef _WIN32
@@ -1170,15 +1171,6 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     framework_op_trace_dir_ = trace_dir_str;
   }
 
-  // TODO: Re-enable framework op trace for multi-SoC preparation.
-  if (enable_framework_op_trace_ && enable_multi_soc_ep_context_) {
-    ORT_CXX_LOG(logger_,
-                ORT_LOGGING_LEVEL_WARNING,
-                "Option 'enable_framework_op_trace' is unsupported in multi-SoC EP context.");
-    enable_framework_op_trace_ = false;
-    framework_op_trace_dir_ = "";
-  }
-
   if (enable_framework_op_trace_) {
     if (framework_op_trace_dir_.empty()) {
       framework_op_trace_dir_ = std::filesystem::current_path().string();
@@ -1572,7 +1564,15 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
 OrtStatus* QnnEp::GetMultiSocSupportedNodes(const OrtGraph* graph,
                                             const std::unordered_map<const OrtNode*, const OrtNodeUnit*>& node_unit_map,
                                             const size_t node_unit_size,
-                                            std::vector<const OrtNode*>& supported_nodes) const {
+                                            std::vector<const OrtNode*>& supported_nodes,
+                                            std::vector<qnn::UnsupportedNodeInfo>& unsupported_nodes) const {
+  // Collect each SoC's rejections tagged with its SoC label, then merge into one
+  // row per ONNX node via MergePerSocUnsupportedNodes.
+  std::vector<qnn::PerSocUnsupportedNodes> per_soc_unsupported;
+  if (enable_framework_op_trace_) {
+    per_soc_unsupported.reserve(htp_arch_per_soc_.size());
+  }
+
   // Iterate each SoC and intersect supported nodes.
   for (size_t idx = 0; idx < htp_arch_per_soc_.size(); ++idx) {
     ORT_CXX_LOG(logger_,
@@ -1585,15 +1585,22 @@ OrtStatus* QnnEp::GetMultiSocSupportedNodes(const OrtGraph* graph,
     ScopedPerSocQnnBackendSetup scoped_backend_setup(*this);
     RETURN_IF_NOT_OK(scoped_backend_setup.Init(idx));
 
-    // Get SoC-specific supported nodes.
+    // Get SoC-specific supported and unsupported nodes.
     std::vector<const OrtNode*> supported_nodes_per_soc;
-    // Dummy for now as framework op trace is temporarily disabled for multi-SoC preparation.
-    std::vector<qnn::UnsupportedNodeInfo> unsupported_nodes;
+    std::vector<qnn::UnsupportedNodeInfo> unsupported_nodes_per_soc;
     RETURN_IF_NOT_NULL(GetSupportedNodes(graph,
                                          node_unit_map,
                                          node_unit_size,
                                          supported_nodes_per_soc,
-                                         unsupported_nodes));
+                                         unsupported_nodes_per_soc));
+
+    // Tag this SoC's rejections with its (htp_arch, soc_model) label for merging.
+    // Identical (arch, soc_model) configs share a label and merge — acceptable,
+    // they are indistinguishable targets.
+    if (enable_framework_op_trace_) {
+      per_soc_unsupported.emplace_back(htp_arch_per_soc_[idx], soc_model_per_soc_[idx],
+                                       std::move(unsupported_nodes_per_soc));
+    }
 
     if (idx == 0) {
       // Directly move for the first SoC.
@@ -1612,6 +1619,13 @@ OrtStatus* QnnEp::GetMultiSocSupportedNodes(const OrtGraph* graph,
       supported_nodes.erase(std::remove_if(supported_nodes.begin(), supported_nodes.end(), not_contains),
                             supported_nodes.end());
     }
+  }
+
+  if (enable_framework_op_trace_) {
+    std::vector<qnn::UnsupportedNodeInfo> merged = qnn::MergePerSocUnsupportedNodes(per_soc_unsupported);
+    unsupported_nodes.insert(unsupported_nodes.end(),
+                             std::make_move_iterator(merged.begin()),
+                             std::make_move_iterator(merged.end()));
   }
 
   return nullptr;
@@ -2016,6 +2030,13 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
     return nullptr;
   }
 
+  // GetCapability runs up to twice. Clear before pass 2 so its collection
+  // supersedes pass 1; pass 1's data is kept for the all-unsupported flush below
+  // (reached only on pass 1, since pass 2 doesn't run when nothing is supported).
+  if (ep->enable_framework_op_trace_ && ep->is_post_layout_transform_) {
+    ep->op_trace_builder_.Reset();
+  }
+
   // Store original graph I/O order for use in Compile.
   // The partitioning process may reorder inputs, so we capture the original order here.
   {
@@ -2052,9 +2073,13 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
                                              node_unit_map,
                                              node_unit_holder.size(),
                                              supported_nodes,
-                                             ep->trace_.unsupported_nodes));
+                                             ep->op_trace_builder_.UnsupportedNodes()));
   } else {
-    RETURN_IF_NOT_NULL(ep->GetMultiSocSupportedNodes(graph, node_unit_map, node_unit_holder.size(), supported_nodes));
+    RETURN_IF_NOT_NULL(ep->GetMultiSocSupportedNodes(graph,
+                                                     node_unit_map,
+                                                     node_unit_holder.size(),
+                                                     supported_nodes,
+                                                     ep->op_trace_builder_.UnsupportedNodes()));
   }
 
   // Helper function that returns a string that lists all unsupported nodes.
@@ -2084,8 +2109,15 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
     // unsupported list collected in GetSupportedNodes would otherwise be
     // discarded. Flush a trace here so the diagnostic record (why every node
     // was rejected) is preserved.
-    if (ep->enable_framework_op_trace_ && !ep->trace_.unsupported_nodes.empty()) {
-      ep->CollectAndWriteFrameworkOpTrace(graph);
+    if (ep->enable_framework_op_trace_ && !ep->op_trace_builder_.UnsupportedNodes().empty()) {
+      // CompileImpl is skipped here, so record every configured SoC (multi-SoC)
+      // or the live target (single-SoC) before writing.
+      if (ep->enable_multi_soc_ep_context_) {
+        ep->AppendMultiSocTraces();
+      } else {
+        ep->AppendSingleSocTrace();
+      }
+      ep->WriteFrameworkOpTrace(graph);
     }
     return nullptr;
   }
@@ -2141,7 +2173,8 @@ OrtStatus* QnnEp::CompileOnnxModel(const OrtGraph** graphs,
                                    const OrtNode** fused_nodes,
                                    size_t count,
                                    OrtNodeComputeInfo** node_compute_infos,
-                                   const qnn::HtpGraphConfigs_t& htp_graph_configs) {
+                                   const qnn::HtpGraphConfigs_t& htp_graph_configs,
+                                   bool collect_subgraph_traces) {
 #if defined(_WIN32) && (defined(__aarch64__) || defined(_M_ARM64))
   // Initialize now for possible reuse in loop
   auto finalize_start = std::chrono::steady_clock::time_point::min();
@@ -2222,10 +2255,10 @@ OrtStatus* QnnEp::CompileOnnxModel(const OrtGraph** graphs,
         /*tensor_name_overrides=*/&tensor_name_overrides_,
         /*json_qnn_graph_path=*/std::string{}};
 
-    // Each ComposeGraph writes its mappings into a slot in trace_.subgraph_traces.
-    if (enable_framework_op_trace_) {
-      trace_.subgraph_traces.push_back(qnn::OpTraceInfo{});
-      context.op_trace_output = &trace_.subgraph_traces.back();
+    // subgraph_traces are SoC-agnostic, so the FCB multi-SoC loop collects them
+    // only on its first iteration (collect_subgraph_traces=false thereafter).
+    if (enable_framework_op_trace_ && collect_subgraph_traces) {
+      context.op_trace_output = op_trace_builder_.NewSubgraphSlot();
     }
 
     if (dump_json_qnn_graph_) {
@@ -2318,12 +2351,14 @@ OrtStatus* QnnEp::CompileMultiSocOnnxModel(const OrtGraph** graphs,
     ScopedPerSocQnnBackendSetup scoped_backend_setup(*this);
     RETURN_IF_NOT_OK(scoped_backend_setup.Init(idx));
 
-    // Compile for current SoC.
+    // Collect subgraph_traces once (on idx 0): the ONNX->QNN op mapping is the
+    // same across SoCs, since which QNN op a node lowers to does not depend on arch.
     RETURN_IF_NOT_NULL(CompileOnnxModel(graphs,
                                         fused_nodes,
                                         count,
                                         node_compute_infos,
-                                        htp_graph_configs_per_soc_[idx]));
+                                        htp_graph_configs_per_soc_[idx],
+                                        /*collect_subgraph_traces=*/idx == 0));
 
     if (idx != htp_arch_per_soc_.size() - 1) {
       // `qnn_models_` and `node_compute_infos` are repeatedly set in each `CompileOnnxModel` call, where previous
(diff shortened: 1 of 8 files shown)


qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc → CompileImpl / GetCapabilityImpl`

The PR title claims "Extend framework op trace schema for FCB multi-SoC", and the header comment says FCB's `CompileOnnxModelMultiSoc` will call `AppendSocTrace` once per iteration — but this diff contains no modifications to `CompileOnnxModelMultiSoc`. As a result, `trace_.soc_traces` will always be empty for FCB sessions, `total_socs` will be 0, and `ValidateTraceStructure`'s `ASSERT_GE(j["soc_traces"].size(), 1u)` will fail when run against a FCB trace. The `TODO(FCB)` comment in `CompileImpl` confirms the wire-up is incomplete.

If this is an intentional two-phase approach (Phase 1: schema migration; Phase 2: FCB wire-up), please update the PR description to say so explicitly (e.g., "Phase 1 — schema only; FCB AppendSocTrace wire-up to follow in a subsequent PR") and soften the PR title (e.g., "Prepare op trace schema for FCB multi-SoC") to avoid reviewer misunderstanding of scope.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc → FinalizeFrameworkOpTrace()`

The `unsupported_nodes` dedup logic uses only `node_index` as the key. If the same node is rejected in different SoC iterations for different reasons (e.g., V73 rejects for data type, V81 rejects for rank), only the first record is kept and others are silently discarded. This is harmless now (FCB path is not yet wired), but will cause diagnostic information loss once M-1 is fixed and FCB multi-SoC sessions are active.

Consider using a `(node_index, reason)` composite key, or moving per-SoC rejection lists inside `SocTrace` rather than sharing a root-level list. If the design intent is deliberately cross-SoC merge (only tracking *which* nodes, not *why* per SoC), please document that decision in a comment.

**[N-2]** `onnxruntime/test/providers/qnn/framework_op_trace_test.cc`

`FrameworkOpTrace_SingleSoc_SocTracesHasOneEntry`, `FrameworkOpTrace_NoRootCompilationTarget`, `FrameworkOpTrace_SubgraphTracesAtRoot`, and `FrameworkOpTrace_SummaryTotalSocs` all build the exact same `Add` model and call the same `RunModelWithTracing` (full session creation + inference), validating different JSON fields. Consider consolidating into a single test `FrameworkOpTrace_SingleSoc_NewSchema` that verifies all four schema contract properties in one pass, reducing CI execution time and maintenance cost.

kuanyul-qti:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc → CompileImpl / GetCapabilityImpl`
> 
> The PR title claims "Extend framework op trace schema for FCB multi-SoC", and the header comment says FCB's `CompileOnnxModelMultiSoc` will call `AppendSocTrace` once per iteration — but this diff contains no modifications to `CompileOnnxModelMultiSoc`. As a result, `trace_.soc_traces` will always be empty for FCB sessions, `total_socs` will be 0, and `ValidateTraceStructure`'s `ASSERT_GE(j["soc_traces"].size(), 1u)` will fail when run against a FCB trace. The `TODO(FCB)` comment in `CompileImpl` confirms the wire-up is incomplete.
> 
> If this is an intentional two-phase approach (Phase 1: schema migration; Phase 2: FCB wire-up), please update the PR description to say so explicitly (e.g., "Phase 1 — schema only; FCB AppendSocTrace wire-up to follow in a subsequent PR") and soften the PR title (e.g., "Prepare op trace schema for FCB multi-SoC") to avoid reviewer misunderstanding of scope.
> 
> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/qnn_execution_provider.cc → FinalizeFrameworkOpTrace()`
> 
> The `unsupported_nodes` dedup logic uses only `node_index` as the key. If the same node is rejected in different SoC iterations for different reasons (e.g., V73 rejects for data type, V81 rejects for rank), only the first record is kept and others are silently discarded. This is harmless now (FCB path is not yet wired), but will cause diagnostic information loss once M-1 is fixed and FCB multi-SoC sessions are active.
> 
> Consider using a `(node_index, reason)` composite key, or moving per-SoC rejection lists inside `SocTrace` rather than sharing a root-level list. If the design intent is deliberately cross-SoC merge (only tracking _which_ nodes, not _why_ per SoC), please document that decision in a comment.
> 
> **[N-2]** `onnxruntime/test/providers/qnn/framework_op_trace_test.cc`
> 
> `FrameworkOpTrace_SingleSoc_SocTracesHasOneEntry`, `FrameworkOpTrace_NoRootCompilationTarget`, `FrameworkOpTrace_SubgraphTracesAtRoot`, and `FrameworkOpTrace_SummaryTotalSocs` all build the exact same `Add` model and call the same `RunModelWithTracing` (full session creation + inference), validating different JSON fields. Consider consolidating into a single test `FrameworkOpTrace_SingleSoc_NewSchema` that verifies all four schema contract properties in one pass, reducing CI execution time and maintenance cost.

[M-1] This is intentionally Phase 1 (schema migration only); the FCB CompileOnnxModelMultiSoc AppendSocTrace wire-up lands in a follow-up PR. Update title and description accordingly.
[N-1] Switched to `std::set<std::pair<size_t, std::string>>` with (node_index, reason) as the composite key. Single-SoC dedup remains a no-op (one entry per node_index, one reason); multi-SoC sessions now preserve distinct reasons across SoCs (V73 data-type rejection AND V81 rank rejection both kept). Comment at `qnn_execution_provider.cc` now spells out the V73/V81 example you raised.
[N-2] Done — collapsed into F`rameworkOpTrace_SingleSoc_NewSchema` at `framework_op_trace_test.cc`. Single inference run, four section-header comments preserve each assertion's purpose (soc_traces shape / no root compilation_target / subgraph_traces location / summary.total_socs).

qti-chuteng (approved):

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2005,6 +2052,17 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
...
   }
 
+  // Reset the trace for this GetCapabilityImpl invocation. The ORT graph
+  // partitioner calls GetCapabilityImpl twice per graph: once to identify
+  // supported nodes and apply layout transformation, then a second time to
+  // process newly-introduced nodes and perform QDQ/op fusions. Without a
+  // reset, the second pass accumulates soc_traces and unsupported_nodes on
+  // top of the first pass, producing doubled soc_traces counts and
+  // inflated unsupported_nodes lists.
+  if (ep->enable_framework_op_trace_) {
```
minfhong-qti: In which call of `GetCapability` call should framework op trace collect?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2043,7 +2101,8 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
                                              supported_nodes,
                                              ep->trace_.unsupported_nodes));
   } else {
-    RETURN_IF_NOT_NULL(ep->GetMultiSocSupportedNodes(graph, node_unit_map, node_unit_holder.size(), supported_nodes));
+    RETURN_IF_NOT_NULL(ep->GetMultiSocSupportedNodes(graph, node_unit_map, node_unit_holder.size(), supported_nodes,
+                                                     ep->trace_.unsupported_nodes));
```
```suggestion
    RETURN_IF_NOT_NULL(ep->GetMultiSocSupportedNodes(graph,
                                                     node_unit_map,
                                                     node_unit_holder.size(),
                                                     upported_nodes,
                                                     ep->trace_.unsupported_nodes));
```

minfhong-qti (commented):
Per-offline discussion, I think it's op trace's responsibility to construct trace structure instead of performing them in QnnEp.
In addition, I think Claude's code comments are too excessive and in fact not providing any useful comments. Prefer to concise them or at least reduce to one or two lines in general.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Collection lands on the pass that reaches compilation; the reset prevents pass-1/pass-2 double-counting. At present, we record the trace for GetCapability pass 2 (post-Layout-Transform).

minfhong-qti: `QnnEp` now has a member `is_post_layout_transform_` used to indicate whether current GetCapability call is the second one. Prefer to leverage this flag and only collect trace when true instead of reset.

onnxruntime/test/providers/qnn/qnn_ep_context_multi_soc_test.cc
```diff
@@ -257,6 +260,218 @@ TEST_F(QnnHTPBackendTests, EPContextMultiSoc_HtpArch_68_73_81_NotAllArchSupporte
...
+      << "Multi-SoC trace must have one soc_traces entry per htp_arch";
+  ASSERT_TRUE(j.contains("summary"));
+  EXPECT_EQ(j["summary"]["total_socs"].get<uint32_t>(), static_cast<uint32_t>(num_socs));
+  EXPECT_FALSE(j.contains("compilation_target"))
+      << "compilation_target must not be at root (moved to soc_traces[i])";
+}
+
+}  // namespace
+
+// FCB multi-SoC framework op trace E2E test.
```
minfhong-qti: Not sure but I prefer to collect all trace testcases in one file (i.e., `framework_op_trace_test.cc`).

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -19,7 +19,7 @@
 #include "core/providers/qnn/builder/qnn_def.h"
 #include "core/providers/qnn/builder/qnn_model.h"
 #include "core/providers/qnn/builder/qnn_model_wrapper.h"
-#include "core/providers/qnn/builder/op_tracing/qnn_op_tracing_types.h"
+#include "core/providers/qnn/builder/op_tracing/qnn_op_tracing.h"
```
minfhong-qti, line 22: Explicitly include `qnn_op_tracing_types` as well since there is usage.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1572,7 +1564,17 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
 OrtStatus* QnnEp::GetMultiSocSupportedNodes(const OrtGraph* graph,
                                             const std::unordered_map<const OrtNode*, const OrtNodeUnit*>& node_unit_map,
                                             const size_t node_unit_size,
-                                            std::vector<const OrtNode*>& supported_nodes) const {
+                                            std::vector<const OrtNode*>& supported_nodes,
+                                            std::vector<qnn::UnsupportedNodeInfo>& unsupported_nodes) const {
+  // Collect each SoC's rejections tagged with its SoC label, then merge into one
+  // row per ONNX node via MergePerSocUnsupportedNodes (a pure, unit-tested helper).
```
minfhong-qti: The comment here should be concise. Why `unit-tested helper` appears here?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1585,15 +1587,25 @@ OrtStatus* QnnEp::GetMultiSocSupportedNodes(const OrtGraph* graph,
...
+
+    // Stash this SoC's rejections with a label for post-loop merging. The label
+    // combines htp_arch and soc_model so that same-arch/different-model SoCs
+    // (e.g. two V73 devices with distinct soc_model) keep separate attribution
+    // in MergePerSocUnsupportedNodes. Fully identical (arch, soc_model) configs
+    // still share a label and merge together, which is acceptable — they are
+    // indistinguishable targets.
+    if (enable_framework_op_trace_) {
+      std::string soc_label = qnn::MakeSocLabel(htp_arch_per_soc_[idx], soc_model_per_soc_[idx]);
+      per_soc_unsupported.push_back({std::move(soc_label), std::move(unsupported_nodes_per_soc)});
```
minfhong-qti: Could `PerSocUnsupportedNodes` have a constructor converting the given info into its required type?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1585,15 +1587,25 @@ OrtStatus* QnnEp::GetMultiSocSupportedNodes(const OrtGraph* graph,
...
-    std::vector<qnn::UnsupportedNodeInfo> unsupported_nodes;
+    std::vector<qnn::UnsupportedNodeInfo> unsupported_nodes_per_soc;
     RETURN_IF_NOT_NULL(GetSupportedNodes(graph,
                                          node_unit_map,
                                          node_unit_size,
                                          supported_nodes_per_soc,
-                                         unsupported_nodes));
+                                         unsupported_nodes_per_soc));
+
+    // Stash this SoC's rejections with a label for post-loop merging. The label
```
minfhong-qti: Ditto.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2685,40 +2728,33 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
...
-    QnnHtpDevice_Arch_t htp_arch = qnn_backend_manager_->GetHtpArch();
-    if (htp_arch != QNN_HTP_DEVICE_ARCH_NONE) {
-      trace_.compilation_target.htp_arch = "V" + std::to_string(htp_arch);
-    }
+    op_trace_builder_.AppendSoc(qnn_backend_manager_->GetHtpArch(),
+                                qnn_backend_manager_->GetSocModel(),
+                                device_id_);
+  } else {
+    // Non-NPU backends have no arch/soc_model; the unknown sentinels are omitted.
+    op_trace_builder_.AppendSoc(qnn::kSocUnknown, qnn::kSocUnknown, device_id_);
```
minfhong-qti: Use `kSocUnknown` for both HTP arch and SoC model is a bit weird. Why not use QNN API's unknown macro?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2318,12 +2355,18 @@ OrtStatus* QnnEp::CompileMultiSocOnnxModel(const OrtGraph** graphs,
     ScopedPerSocQnnBackendSetup scoped_backend_setup(*this);
     RETURN_IF_NOT_OK(scoped_backend_setup.Init(idx));
 
-    // Compile for current SoC.
+    // Collect subgraph_traces once (idx 0): valid only while op lowering is
+    // arch-invariant (today only support predicates are arch-gated, not lowering).
```
minfhong-qti: What does "valid only while op lowering is arch-invariant (today only support predicates are arch-gated, not lowering)" mean?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2748,10 +2789,13 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
   // below to serialize the io_name_overrides attribute into the EPContext model.
   ep->onnx_graph_io_names_.reset();
 
-  // Framework op trace: serialize and write JSON.
-  // When multiple graphs are compiled, all subgraph traces are collected into a single file.
+  // Framework op trace: single-SoC records its one SocTrace here; the FCB
+  // multi-SoC path already appended one per iteration in CompileMultiSocOnnxModel.
```
minfhong-qti: Can FCB append SoCs here instead of in `CompileMultiSocOnnxModel`?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2685,40 +2728,33 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
...
 
-void QnnEp::CollectAndWriteFrameworkOpTrace(const OrtGraph* primary_graph) {
-  trace_.model_name = std::filesystem::path(GetModelPathString(primary_graph, ort_api)).filename().u8string();
-  if (trace_.model_name.empty()) {
-    trace_.model_name = "<in-memory>";
-  }
-  trace_.backend_type = qnn::QnnBackendTypeToString(qnn_backend_manager_->GetQnnBackendType());
-  trace_.compilation_target.device_id = device_id_;
-
+void QnnEp::AppendCurrentSocTrace() {
```
```suggestion
void QnnEp::AppendSingleSocTrace() {
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2685,40 +2728,33 @@ OrtStatus* QnnEp::CompileDlcContextModel(OrtEp* this_ptr,
...
-  // The unsupported-only branch covers the "entire model is unsupported"
-  // case, where ORT may not invoke CompileImpl at all but the trace still
-  // has diagnostic value (it records why every node was rejected).
-  // A wholly empty trace can still occur on EPContext cached runs
-  // - skip the file write there.
-  if (!trace_.subgraph_traces.empty() || !trace_.unsupported_nodes.empty()) {
-    qnn::WriteTraceToFile(trace_,
-                          std::filesystem::path(framework_op_trace_dir_) / "qnn_op_trace.json",
-                          logger_);
+void QnnEp::AppendConfiguredSocTraces() {
```
```suggestion
void QnnEp::AppendMultiSocTraces() {
```

onnxruntime/test/providers/qnn/qnn_ep_context_multi_soc_test.cc
kuanyul-qti: Moved to `framework_op_trace_test.cc`.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Added.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Added `is_post_layout_transform_` on flag, but keep the usage of reset so that the empty-flush path is still working (pass 1).

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Reworded plainly.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Now uses QNN_HTP_DEVICE_ARCH_NONE, QNN_SOC_MODEL_UNKNOWN.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Done.

minfhong-qti (approved):

PR #496, opened 2026-06-08
[QNN EP] Fix cmake deps mirror on Windows: resilience + ort_core propagation

qti-kromero:
### Description

Two fixes to make the cmake dependency mirror work reliably on Windows machines where libcurl (used by CMake's FetchContent) cannot follow HTTPS redirects due to certificate revocation check failures (`CRYPT_E_NO_REVOCATION_CHECK`), but `urllib`/WinINet can.

**1. `fetch_cmake_deps.py`: resilient per-dep download**

Previously, a single failed `cache.fetch()` call would abort the entire script, leaving the mirror directory unpopulated — even for deps that were already in the local cache (`~/.ort-package-cache`). Wrapped each fetch in a try/except so cached deps still get mirrored when others fail to download. Logs a warning per failure and an error summary at the end.

**2. `build.ps1`: surface `fetch_cmake_deps.py` failures**

The script was called without `Assert-Success`, so failures were silently ignored and the build continued with an empty mirror. Wrapped with `Assert-Success` so failures are visible in CI logs.

**3. `onnxruntime_prebuilt.cmake`: forward mirror dir to ort_core build**

The ARM64X build invokes ort_core as a separate cmake process via ORT's `build.py`. That process computes its own default `onnxruntime_CMAKE_DEPS_MIRROR_DIR` relative to its own source tree (pointing at a non-existent `_deps/mirror`), so it bypasses the mirror and attempts live downloads. Passes `--cmake_deps_mirror_dir` explicitly so the ort_core cmake uses the same pre-populated mirror as the parent build.

### Motivation and Context

On certain Windows CI runners, CMake's bundled libcurl fails to follow GitHub release asset redirects (`github.com` → `release-assets.githubusercontent.com`) due to corporate certificate revocation policy. `Invoke-WebRequest` and Python's `urllib` (via `fetch_cmake_deps.py`) work because they use the WinINet stack which respects system trust settings. This change ensures the mirror populated by `fetch_cmake_deps.py` is used end-to-end for all cmake FetchContent downloads.

### DIFF
--- a/qcom/scripts/all/fetch_cmake_deps.py
+++ b/qcom/scripts/all/fetch_cmake_deps.py
@@ -25,9 +25,17 @@ def main(deps_dir: Path, mirror_dir: Path) -> None:
 
     logging.info(f"Downloading dependencies into {deps_dir}")
     cached_paths: list[tuple[str, Path]] = []
+    failed: list[str] = []
     for name, url, sha1 in deps:
-        cached_paths.append((url, cache.fetch(name, url, expected_sha1=sha1)))
-        logging.debug(f"{name} --> {cached_paths[-1]}")
+        try:
+            cached_paths.append((url, cache.fetch(name, url, expected_sha1=sha1)))
+            logging.debug(f"{name} --> {cached_paths[-1]}")
+        except Exception as e:
+            logging.warning(f"Failed to fetch {name} ({url}): {e}")
+            failed.append(name)
+
+    if failed:
+        logging.error(f"Could not fetch {len(failed)} dep(s): {', '.join(failed)}. CMake will attempt to download them directly.")
 
     logging.info(f"Making dependencies available in {mirror_dir}")
     if mirror_dir.exists():
--- a/qcom/scripts/all/fetch_cmake_deps.py
+++ b/qcom/scripts/all/fetch_cmake_deps.py
@@ -35,7 +35,9 @@ def main(deps_dir: Path, mirror_dir: Path) -> None:
             failed.append(name)
 
     if failed:
-        logging.error(f"Could not fetch {len(failed)} dep(s): {', '.join(failed)}. CMake will attempt to download them directly.")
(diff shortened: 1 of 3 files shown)


qti-mbadnara (approved):
LGTM

qcom/scripts/all/fetch_cmake_deps.py
```diff
@@ -25,9 +25,19 @@ def main(deps_dir: Path, mirror_dir: Path) -> None:
...
-        logging.debug(f"{name} --> {cached_paths[-1]}")
+        try:
+            cached_paths.append((url, cache.fetch(name, url, expected_sha1=sha1)))
+            logging.debug(f"{name} --> {cached_paths[-1]}")
+        except Exception as e:
+            logging.warning(f"Failed to fetch {name} ({url}): {e}")
+            failed.append(name)
+
+    if failed:
+        logging.error(
```
qti-ashwshan, line 38: logging.error here only logs? it doesn't raise or sys.exit, so the script still exits 0 on a dep-fetch failure. Is that intended? If so, the Assert-Success wrapper in build.ps1 won't actually catch download failures (only hard crashes), and the build will proceed to CMake and fail later. Is the intention to hard fail here ?

qti-kromero: that’s intentional since it **can** pull in the external dependency from the network (not the mirror) at build time on some machines so i opted to let it fall back to it and possibly succeed. on machines that don’t get it from the mirror **and** fail during build (basically due to cert issue this PR is intended to resolve) we will see the same error this PR is intended to resolve

qti-kromero: the reason i’m ok with leaving this fix at this PR is that i believe all the machines have the right cert configuration to not fail the fetch in our script

qti-ashwshan (approved):

PR #497, opened 2026-06-08
Allow BatchNorm in float when its output has no Q on HTP

qti-yuduo:
AIMET emits QDQ graphs where BatchNormalization is intentionally left with no Quantize on its output, signaling the result flows in float into a downstream island (e.g. BN -> Relu -> Conv in NASNet's reduction cell) before being re-quantized. The QNN EP's BN QDQ selector required a Q on the output, so these BNs dropped to a SingleNode unit where the DQ-fed scale looked dynamic, and the BN plus its neighbors fell back to CPU. That fragments the graph; AI Hub then reports "cannot capture the entire model".

Treat a quantized-input / no-output-Q BN as a float island and execute it in FP:
- Selector (qnn_ep_utils.cc): allow the empty-Q group and skip the input/output dtype-match check when the output is not quantized.
- Op builder (batchnormalization_op_builder.cc): reuse the existing float-param path (dequantize input, run BN in f32) but emit the BN's float result directly with no trailing Quantize; the island re-quantizes downstream at the next Q edge.
- CheckHtpDataTypes: accept quantized input with float output.

Adds QnnHTPBackendTests.BatchNorm_FloatIsland_NoOutputQ reproducing DQ -> BN -> Relu -> Conv -> Q and asserting whole-graph QNN assignment.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
@@ -489,6 +489,30 @@ void OverrideParamTypeForRequantize(Qnn_DataType_t x_dtype,
   }
 }
 
+// Single source of truth for float execution, shared by ProcessInputs (stores params) and
+// ProcessAttributesAndOutputs (emits the op).
+//   - has_float_output: quantized input, no output Q -> float island; BN emits float directly.
+//   - use_float_params: also covers u8/u16 input with per-channel scale, whose fused weight
+//     gamma/sqrt(var+eps) can overflow a single per-tensor requant scale.
+struct BatchNormFloatExecution {
+  bool has_float_output;
+  bool use_float_params;
+};
+
+BatchNormFloatExecution GetBatchNormFloatExecution(const TensorInfo& input_info,
+                                                   const TensorInfo& scale_info,
+                                                   const TensorInfo& output_info) {
+  const bool is_quantized_op = input_info.quant_param.IsQuantized();
+  const bool has_float_output = is_quantized_op && !output_info.quant_param.IsQuantized();
+  const bool use_float_params =
+      has_float_output ||
+      (is_quantized_op &&
+       (input_info.qnn_data_type == QNN_DATATYPE_UFIXED_POINT_8 ||
+        input_info.qnn_data_type == QNN_DATATYPE_UFIXED_POINT_16) &&
+       scale_info.quant_param.IsPerChannel());
+  return {has_float_output, use_float_params};
+}
+
 }  // namespace
 
 // BatchNorm is sensitive with data layout, no special validation so far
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -587,6 +587,12 @@ Ort::Status BatchNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_mode
     RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
     const bool is_quantized_op = input_info.quant_param.IsQuantized();
 
+    // Quantized input but a float output: BN runs in float, so params are stored as float and the
+    // input is dequantized in ProcessAttributesAndOutputs.
+    TensorInfo float_output_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Outputs()[0], float_output_info));
+    const bool has_float_output = is_quantized_op && !float_output_info.quant_param.IsQuantized();
```
yath1: This part is handling outputs in ProcessInputs(...) method. Do you think there is an easy way to separate it out without a large refactoring?

yath1: Not a blocker for this PR if there none. We can refactor separately...

onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -587,6 +587,12 @@ Ort::Status BatchNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_mode
     RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], input_info));
     const bool is_quantized_op = input_info.quant_param.IsQuantized();
 
+    // Quantized input but a float output: BN runs in float, so params are stored as float and the
+    // input is dequantized in ProcessAttributesAndOutputs.
+    TensorInfo float_output_info = {};
```
yath1: This is only checking input and output activations. But if both these are int16, but scale, var, mean are in fp (dont have a qdq), then we need to move the op to fp as well. This can be handled as a universal rule through a refactor later on.

onnxruntime/test/providers/qnn/batchnormalization_test.cc
```diff
@@ -786,6 +786,133 @@ TEST_F(QnnHTPBackendTests, BatchNorm2D_NearZeroVariance_U8) {
...
+    builder.AddNode("conv", "Conv", {"relu_out", "conv_w"}, {"conv_out"}, "", conv_attrs);
+
+    AddQDQNodePairWithOutputAsGraphOutput<InputQType>(builder, "output_qdq", "conv_out",
+                                                      output_qparams[0].scale, output_qparams[0].zero_point);
+  };
+}
+
+// A BatchNormalization with no Q on its output must be captured by QNN (entire graph), not fall back
+// to CPU. Accuracy is compared against the float reference.
+TEST_F(QnnHTPBackendTests, BatchNorm_NoOutputQ) {
```
yath1, line 887: Nit: Maybe add a `NoInputQ` test as well

yath1 (approved):
lgtm (other than the refactor items which are probably out of scope)

onnxruntime/test/providers/qnn/batchnormalization_test.cc
qti-yuduo: No input Q does not hit any logic here (no behavior change).. skipping this test.

onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
qti-yuduo: DOne.

PR #498, opened 2026-06-09
[QNN EP] Add hidden_size supplemental attribute to QNN LSTM and GRU op builders

kuanyul-qti:
The QAIRT SDK LSTM and GRU opdef do not declare hidden_size as a formal QNN parameter. The native QAIRT converter emits it as a supplemental attribute (assertAttrExists, defined in IrOpDefs/LstmOp.cpp) required by the IR/DLC backend. The QNN EP path was missing this attribute, causing DLC generation to fail with a hidden_size mismatch error.

Emit hidden_size as a supplemental uint32 scalar param, gated on IsIrBackend() (QnnBackendType::SERIALIZER only). The QNN CPU and HTP backends do not expect supplemental params and must not receive it.

Tests added:
- QnnIRBackendTests.LSTM_HiddenSize_DLC (verifies DLC generation succeeds)
- QnnIRBackendTests.GRU_HiddenSize_DLC

Kill-test confirmed: removing the IsIrBackend() guard causes both DLC tests to fail with a hidden_size mismatch error.


kuanyul-qti:
Close this PR as the fix needs to be implemented on the native tool side instead of the ORT side.

PR #499, opened 2026-06-09
[FCB] Add deploy_multi_soc_ep_context tool for editing FCB DLCs

minfhong-qti:
### Description

Add a Python utility to inspect and prune the per-SoC HTP cache records inside a Flexible Context Binary (FCB) DLC, and keep the accompanying EP-context ONNX model's compatibility metadata in sync.

The tool supports two actions, both restricted to HTP_CACHE_RECORD:
  * --list:          list each HTP cache record with its DSP arch, SoC
                     model and (first-graph) VTCM size.
  * --remove_record: remove one or more records by name and drop the
                     matching per-SoC entries from the
                     'ep_compatibility_info.QNNExecutionProvider' value
                     in the ONNX model (--onnx_model, required). A record
                     is matched to the metadata entry at the same index it
                     occupies in the sorted record-name list. The ONNX
                     metadata is updated first; the DLC record is only
                     removed on success so the two stay consistent.

Packaging:
  * cmake: copy the script into the onnxruntime_qnn wheel package, and bundle the QAIRT SDK's libDlModelToolsPy extension from the dlc_utils/<platform> dir matching QNN_ARCH_ABI so the dependency imports out of the box.
  * setup.py: add the libDlModelToolsPy* glob to package_data.

The tool relies on libDlModelToolsPy, which the QAIRT SDK compiles only for Python 3.10 and 3.12; the import selects libDlModelToolsPy310 or libDlModelToolsPy accordingly.

Docs: document the FCB feature, its per-SoC provider options, and the tool's usage in QNN-ExecutionProvider.md.

### Motivation and Context

- Provide a Python utility for user to strip redundant record from FCB before deployment.
- Document FCB usage.

### DIFF
--- a/cmake/onnxruntime_providers_qnn.cmake
+++ b/cmake/onnxruntime_providers_qnn.cmake
@@ -204,6 +204,58 @@
         $<TARGET_FILE_DIR:${onnxruntime_providers_qnn_target}>/onnxruntime_qnn
     COMMENT "Copying version, license, README, and release notes files"
   )
+
+  # Bundle libDlModelToolsPy (used by deploy_multi_soc_ep_context.py) from the
+  # QAIRT SDK's dlc_utils into the wheel. The dlc_utils platform sub-directory
+  # name differs from QNN_ARCH_ABI, so map it explicitly (see the SDK's
+  # lib/python/qti/aisw/dlc_utils/__init__.py for the canonical mapping).
+  set(QNN_DLC_UTILS_SUBDIR "")
+  if (QNN_ARCH_ABI STREQUAL "x86_64-linux-clang")
+    set(QNN_DLC_UTILS_SUBDIR "linux-x86_64")
+  elseif (QNN_ARCH_ABI STREQUAL "aarch64-oe-linux-gcc11.2")
+    set(QNN_DLC_UTILS_SUBDIR "linux-aarch64-oe-gcc11.2")
+  elseif (QNN_ARCH_ABI STREQUAL "x86_64-windows-msvc")
+    set(QNN_DLC_UTILS_SUBDIR "windows-x86_64")
+  elseif (QNN_ARCH_ABI STREQUAL "arm64x-windows-msvc")
+    set(QNN_DLC_UTILS_SUBDIR "windows-arm64ec")
+  endif()
+
+  if (NOT QNN_DLC_UTILS_SUBDIR)
+    message(WARNING "dlc_utils sub-directory not found for QNN_ARCH_ABI=${QNN_ARCH_ABI}; "
+                    "libDlModelToolsPy will not be bundled and deploy_multi_soc_ep_context will be unusable")
+  elseif (NOT (Python_VERSION_MAJOR EQUAL 3 AND Python_VERSION_MINOR EQUAL 12))
+    # QAIRT SDK ships libDlModelToolsPy in limited Python version. Considering
+    # QNN-EP's supported Python version, only bundle it in Python 3.12 wheel.
+    message(STATUS "libDlModelToolsPy is only bundled for the Python 3.12 wheel; "
+                   "skipping for Python ${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}")
+  elseif (EXISTS "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}")
+    file(GLOB QNN_DLMODELTOOLS_LIBS LIST_DIRECTORIES false
+         "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}/libDlModelToolsPy312.so"
+         "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}/libDlModelToolsPy312.pyd")
+    if (NOT QNN_DLMODELTOOLS_LIBS)
+      message(WARNING "libDlModelToolsPy312 not found under "
+                      "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}; "
+                      "deploy_multi_soc_ep_context.py will not be bundled and will be unusable")
+    else()
+      foreach(QNN_DLMODELTOOLS_LIB ${QNN_DLMODELTOOLS_LIBS})
+        add_custom_command(
+          TARGET ${onnxruntime_providers_qnn_target} POST_BUILD
+          COMMAND ${CMAKE_COMMAND} -E copy_if_different "${QNN_DLMODELTOOLS_LIB}" $<TARGET_FILE_DIR:${onnxruntime_providers_qnn_target}>/onnxruntime_qnn
+          COMMENT "Copying ${QNN_DLMODELTOOLS_LIB} to onnxruntime_qnn for Python Wheel"
+        )
+      endforeach()
+      # Only package the deploy tool alongside its libDlModelToolsPy312 dependency,
+      # so the extension (3.12) and the script always ship together.
+      add_custom_command(
+        TARGET ${onnxruntime_providers_qnn_target} POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            ${ONNXRUNTIME_ROOT}/python/multi_soc/deploy_multi_soc_ep_context.py
+            $<TARGET_FILE_DIR:${onnxruntime_providers_qnn_target}>/onnxruntime_qnn
+        COMMENT "Copying deploy_multi_soc_ep_context.py to onnxruntime_qnn for Python Wheel"
+      )
+    endif()
+  endif()
+
   # Create document destination directory first to ensure it exists
   add_custom_command(
     TARGET ${onnxruntime_providers_qnn_target} POST_BUILD
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -357,6 +357,83 @@ ep_options = {
 # ep.context_enable must be set to 1; enable_htp_prepare_only is enabled automatically.
 ```
 
+#### Stripping an FCB for deployment (`deploy_multi_soc_ep_context.py`)
+After an FCB is generated, you may want to **prune** the SoCs it targets — for example, to ship a smaller artifact that only covers the devices you actually support. The `deploy_multi_soc_ep_context.py` helper (shipped in the `onnxruntime-qnn` Python wheel) lists and removes the per-SoC HTP cache records inside the FCB, and keeps the accompanying ONNX model's compatibility metadata in sync.
+
+The script works on both EP context model layouts, selected by which input you pass:
+
+* **Non-embed mode** (`ep.context_embed_mode = "0"`, the default): the context binary is a standalone `*_qnn.bin` file alongside the ONNX model. Pass it via `--input_bin` (and `--input_onnx` for the metadata). The stripped binary is written to `--output_bin` and the updated model to `--output_onnx`; both inputs are left untouched. Pass `--inplace` to overwrite the inputs instead.
+* **Embed mode** (`ep.context_embed_mode = "1"`): the context binary is embedded inside the EP context node of the ONNX model. Omit `--input_bin` and pass the model via `--input_onnx`; the tool extracts the embedded binary to a temporary file, edits it, and re-embeds the result into `--output_onnx`. The input model is left untouched. Pass `--inplace` to overwrite `--input_onnx` instead.
+
+Run it as a module from an installed wheel:
+
+```bash
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context --input_bin model_ctx_qnn.bin --list
+```
+
+|Argument|Description|
+|---|---|
+|`--input_bin`|Path to a standalone context binary / EP-context file (**non-embed mode**). Omit it for an embed-mode model, in which case the embedded binary is read from `--input_onnx`.|
+|`--input_onnx`|Path to the input EP context ONNX model. **Required with `--remove_record`** so its `ep_compatibility_info.QNNExecutionProvider` metadata is kept in sync. In embed mode this is also the source of the embedded binary, so it is always required there.|
+|`-l`, `--list`|List every HTP cache record with its DSP architecture, SoC model and (first-graph) VTCM size.|
+|`-r`, `--remove_record`|Comma-separated HTP cache record name(s) to remove (e.g. `-r backend.metadata0,backend.metadata1`).|
+|`--output_onnx`|**Required with `--remove_record`** (both modes) unless `--inplace` is given. Destination path for the updated ONNX model. In embed mode the stripped binary is re-embedded into it; in non-embed mode it carries the updated metadata. The input `--input_onnx` is left untouched.|
+|`--output_bin`|**Required with `--remove_record` in non-embed mode** unless `--inplace` is given. Destination path for the stripped binary. The input `--input_bin` is left untouched.|
+|`--temp_bin`|*Embed mode only, optional.* Path to stage the extracted binary while it is edited. When omitted, an auto-generated temporary file is used. The file is removed when the tool finishes either way.|
+|`--inplace`|*Optional.* Overwrite the input model (and, in non-embed mode, the input binary) in place instead of writing to new paths. Mutually exclusive with `--output_onnx` / `--output_bin`.|
+
+**Listing the packaged SoCs:**
+
+```bash
+# Non-embed mode (standalone context binary)
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context --input_bin model_ctx_qnn.bin --list
+
+# Embed mode (context binary embedded in the ONNX model)
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context --input_onnx model_ctx.onnx --list
+```
+
+**Removing a SoC (non-embed mode)** — writes the pruned binary and updated model to new paths, leaving the inputs untouched:
+
+```bash
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context \
+    --input_bin model_ctx_qnn.bin \
+    --input_onnx model_ctx.onnx \
+    --output_bin model_ctx_qnn.pruned.bin \
+    --output_onnx model_ctx.pruned.onnx \
+    -r backend.metadata0
+```
+
+**Removing a SoC (embed mode)** — extracts, edits, and re-embeds the binary into a new model:
+
+```bash
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context \
+    --input_onnx model_ctx.onnx \
+    --output_onnx model_ctx.pruned.onnx \
+    -r backend.metadata0 \
+    --list
+```
+
+**Removing a SoC in place** — overwrite the inputs instead of writing new files (works in both modes):
+
+```bash
+# Non-embed mode: overwrites both model_ctx_qnn.bin and model_ctx.onnx
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context \
+    --input_bin model_ctx_qnn.bin \
+    --input_onnx model_ctx.onnx \
+    --inplace \
+    -r backend.metadata0
+
+# Embed mode: overwrites model_ctx.onnx
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context \
+    --input_onnx model_ctx.onnx \
+    --inplace \
+    -r backend.metadata0
+```
+
+> **Note**: `deploy_multi_soc_ep_context.py` depends on the `libDlModelToolsPy` extension from the QAIRT SDK. The `onnxruntime-qnn` wheel bundles it next to the script, so `python -m onnxruntime_qnn.deploy_multi_soc_ep_context` works out of the box. When running the script directly from a source checkout instead, the SDK's `lib/python/qti/aisw/dlc_utils/<platform>` directory must be on `PYTHONPATH`.
+
+> **Note**: Only **Python 3.12** is supported. The `onnxruntime-qnn` wheel bundles the `libDlModelToolsPy` extension (compiled for Python 3.12) only into the 3.12 wheel, and the script imports it directly. The script is thus not shipped in the wheels other than Python 3.12 one.
+
 ### Run Options
 
 Run options can be set dynamically at runtime using the ORT Run API. These options allow you to configure QNN EP behavior on a per-inference basis.
--- a/onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
+++ b/onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
@@ -1,0 +1,573 @@
+#!/usr/bin/env python3
+# -*- mode: python -*-
+# -------------------------------------------------------------------------
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""List and remove HTP_CACHE_RECORD entries inside a FCB / EP-context file.
+
+A trimmed-down sibling of ``qairt-dlc-deploy`` that operates only on HTP cache
+records:
+
+  * ``--list``          : list each HTP cache record with its DSP arch, SoC
+                          model and VTCM size (read via ``libDlModelToolsPy``).
+  * ``--remove_record`` : remove records by name and drop the matching SoC
+                          entries from the accompanying ONNX model's QNN EP
+                          compatibility metadata, keeping the two in sync.
+
+The metadata lives in ``metadata_props`` under key
+``ep_compatibility_info.QNNExecutionProvider``. For a Flexible Context Binary the
+QNN EP writes the V2 layout, a single ':'-separated string::
+
+    v2:<BackendId>:<SDK>:<BackendApi>:<HtpArch1>,...:<SoCModel1>,...:<VtcmMb1>,...:<IsHtpUsrDrv>
+
+Fields 4-6 (HtpArch, SoCModel, VtcmMb) are parallel comma-separated lists: entry
+*i* of each is one SoC. A removed record is matched to the entry at the same
+index *i* it holds in the sorted HTP cache record name list, and that index is
+dropped from all three lists.
+"""
+
+import argparse
+import os
+import re
+import sys
+import tempfile
+import traceback
+
+import onnx
+
+# libDlModelToolsPy312 is a QAIRT SDK pybind extension copied flat into this
+# package at wheel-build time; make its directory importable regardless of cwd.
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+# Only the Python 3.12 build is bundled into the wheel, so it is the sole module
+# imported here; all classes/enums live at its top level.
+try:
+    import libDlModelToolsPy312 as modeltools
+except ImportError as import_error:
+    # This is the QAIRT SDK's libDlModelToolsPy extension. It fails to import in
+    # two common situations; spell both out so the user knows how to fix it
+    # instead of getting a bare ModuleNotFoundError traceback.
+    running_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+    sys.exit(
+        f"Failed to import 'libDlModelToolsPy312' ({import_error}).\n"
+        f"This tool requires Python 3.12 (currently running {running_version}); the extension is\n"
+        "bundled only in the onnxruntime-qnn Python 3.12 wheel. To fix:\n"
+        "  - Run under Python 3.12 from an installed onnxruntime-qnn wheel, e.g.\n"
+        "    'python -m onnxruntime_qnn.deploy_multi_soc_ep_context ...'.\n"
+        "  - When running from a source checkout instead, put the QAIRT SDK's\n"
+        "    'lib/python/qti/aisw/dlc_utils/<platform>' directory on PYTHONPATH."
+    )
+
+
+# Key under which the QNN EP stores per-SoC compatibility info in metadata_props.
+_EP_COMPAT_KEY = "ep_compatibility_info.QNNExecutionProvider"
+
+# Positions of the parallel (DspArch, SoCModel, VTCMSize) lists in the
+# ':'-separated value, and the expected field count.
+_LIST_FIELD_INDICES = (4, 5, 6)
+_NUM_FIELDS = 8
+
+# V2-format field layout: the value is only editable when field 0 is "v2".
+_VERSION_FIELD_INDEX = 0
+_EXPECTED_VERSION = "v2"
+
+# Attribute names on the EPContext node, mirroring onnx_ctx_model_helper.h.
+_EPCONTEXT_OP = "EPContext"
+_EMBED_MODE_ATTR = "embed_mode"
+_EP_CACHE_CONTEXT_ATTR = "ep_cache_context"
+
+
+class DeployError(Exception):
+    """Raised for an expected, user-facing failure. ``main`` catches it and prints
+    the message (no traceback), then exits non-zero. Distinct from an incidental
+    ``Exception`` (e.g. from onnx/modeltools or a bug), which keeps its traceback."""
+
+
+def _safe(fn, *args, default="N/A"):
+    """Call a reader accessor, returning ``default`` on error or empty value so
+    one bad record can't abort the listing."""
+    try:
+        value = fn(*args)
+        return value if value not in (None, "") else default
+    except Exception:
+        return default
+
+
+def get_htp_cache_record_names(reader):
+    """Return the HTP_CACHE_RECORD names sorted by embedded number (natural
+    order, e.g. ``cache_2`` before ``cache_10``) rather than lexicographically."""
+    # Split into non-digit / digit runs; compare digit runs as ints.
+    natural_sort_key = lambda name: [int(p) if p.isdigit() else p for p in re.split(r"(\d+)", name)]  # noqa: E731
+    names = reader.get_record_names_by_record_type(modeltools.DlcRecordType.HTP_CACHE_RECORD, False)
+    return sorted(names, key=natural_sort_key)
+
+
+def list_records(reader):
+    """Print a table of HTP cache records with DSP arch, SoC model and VTCM size."""
+    record_names = get_htp_cache_record_names(reader)
+
+    if not record_names:
+        print("No HTP_CACHE_RECORD entries found in the FCB.")
+        return
+
+    rows = []
+    for name in record_names:
+        dsp_arch = _safe(reader.get_cache_dsp_arch, name)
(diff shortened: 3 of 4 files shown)


qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🔴 Blocker` `[Arch]` — `deploy_multi_soc_ep_context.py:267` (`_resolve_removed_indices` / `get_htp_cache_record_names:71-74` / `_apply_metadata_removal:153-208`)
The whole tool's correctness rests on: *lexicographic order of HTP_CACHE_RECORD names in the DLC* == *the per-SoC order the QNN EP wrote into the metadata lists*. Checked against the writer, this doesn't hold:
- Metadata order is the **user's comma-separated config order** (`qnn_execution_provider.cc:2597-2611` ← `ParsePerSocHtpConfigs:361-373`).
- Record names are minted inside the QAIRT SDK's `contextAddToDlc` (`qnn_backend_system_dlc_plugin.cc:44`); the EP never assigns them, and `QnnSystemDlc_getRecordsByType` (QnnSystemDlc.h:295-321) guarantees **no ordering**.
- Records *are* added in config order (insertion order aligns), but the tool **discards** that by calling `sorted()`. And with `backend.metadata0..N` names, ≥10 records sort `metadata10` before `metadata2`.

On any divergence the tool removes SoC A's DLC record but drops SoC B's metadata entry — silent data corruption; none of the length/bounds guards catch it because they never verify the record actually describes that SoC.
*Better design:* Match by a content key, not sorted position. The tool already calls `get_cache_soc_model(name)` / `get_cache_dsp_arch(name)` per record (L87-88, only used for `--list`). On removal, look up the removed record's `(soc_model, dsp_arch, vtcm)` and delete the metadata entry whose triple matches; error out on no-match/multi-match.

**[D-2]** `🟠 Concern` `[Arch]` — `deploy_multi_soc_ep_context.py:434-442` (+ `remove_records:134`)
The non-embed flow comments that editing the ONNX first and the DLC on success keeps them consistent, but that ordering creates inconsistency in two cases. **`--inplace`**: `output_onnx == input_onnx`, so the input ONNX is overwritten (metadata dropped) before a DLC `save()` that can `sys.exit` on failure — leaving a corrupted model, an unchanged DLC, and no originals. **Multi-record**: `remove_records` saves once per record (L134) while the metadata drops all indices at once, so `-r a,b` failing on `b` leaves the DLC with only `a` removed but the ONNX with both dropped — inconsistent even in non-inplace mode.
*Better design:* Make the two-artifact write all-or-nothing: stage all DLC + ONNX edits, validate, then flush; save the DLC once (out of the per-record loop), and for `--inplace` write temp files and rename-over only after both succeed.

### 🟩 Simplicity

Two small points: (1) remove_records calls save() once per record (L134), rewriting the whole DLC repeatedly for multi-record removal and feeding the D-2 partial-failure window; save once after all removals instead. (2) Several arguments use metavar="\b" (a backspace char) to suppress argparse's metavar display — a fragile hack that can drift across formatter/argparse versions. 

### 🔵 Completeness

**[D-3]** `🔵 Improvement` `[Completeness]` — `deploy_multi_soc_ep_context.py:172-177`
The writer has V1 (no prefix, 6 fields) and V2 (`v2:`, 8 fields), distinguished by the first slice (`qnn_cache_compatibility_manager.cc:170-185`). The tool only checks `len(fields) != 8` and never validates `fields[0] == "v2"`. A V1 model is rejected with a misleading "expected 8 fields, got 6" message (real reason: it's a single-SoC legacy binary), and a future 8-field vN with different semantics would be silently mis-parsed. (The V2 8-field/(4,5,6) layout itself is verified to match the writer exactly, so this is a robustness gap, not a current mis-parse.)
*Better design:* Check `fields[0] == "v2"` before splitting the lists and reject non-v2 with a clear message; gate `_NUM_FIELDS`/`_LIST_FIELD_INDICES` behind that version discrimination.

minfhong-qti:
- [D-1] Since the DLC order depends on the given config during generating FCB, I believe exploiting that model compatibility has the same order is fine. But original sorting is incorrect, revise the sorting to natural order.
- [D-2] Since the true atomicity requires temporary file saving which violates the purpose of in-place modify. Revise the execution order to guarantee ONNX model is saved only after DLC is saved successfully, which minimizes the risk of misalignment.
- [D-3] Add explicit check for version 2.
- [S-1] Revise to save DLC only after all records are removed.
- [S-2] Revise to "metavar=''".

qti-chuteng (approved):

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,537 @@
...
+        return
+
+    entry = next((p for p in model.metadata_props if p.key == _EP_COMPAT_KEY), None)
+    if entry is None:
+        sys.exit(f"Metadata key '{_EP_COMPAT_KEY}' not found in the ONNX model.")
+
+    fields = entry.value.split(":")
+    if len(fields) != _NUM_FIELDS:
+        sys.exit(
+            f"Unexpected '{_EP_COMPAT_KEY}' value: expected {_NUM_FIELDS} "
```
huaychou, line 188: **[M-3] (Major)** sys.exit() called inside library function _apply_metadata_removal()

`_apply_metadata_removal()` calls `sys.exit()` at multiple validation failure points (lines 161, 169, 177, 184). Same problem as M-2: this function is indirectly called from the `try/finally` block in `run_embed_mode()`. If `sys.exit()` fires, the `finally` clause cleans up the temp file, but the caller cannot catch the error or do any recovery.

**Impact**: In embed mode, `_apply_metadata_removal()` may exit before it has finished modifying the in-memory model or before `updater.save()` runs, leaving state inconsistent. More critically: the function is also called via `update_onnx_model()` in non-embed mode — there is no `try/finally` there, so the ONNX file is never written on exit, but the error message gives no indication of which steps completed.

**Suggested fix:**

Same as M-2: replace sys.exit() with RuntimeError so the caller handles the error uniformly in main()'s outer except block.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,537 @@
...
+            with open(tmp_path, "rb") as stripped:
+                cache_attr.s = stripped.read()
+            onnx.save(model, args.output_onnx)
+            print(f"Wrote re-embedded EP-context model to {args.output_onnx}.")
+
+        # List last so a combined --list reflects the post-removal state.
+        if args.list:
+            list_records(updater)
+    finally:
+        os.remove(tmp_path)
```
huaychou: **[M-4] (Major)** finally block unconditionally removes temp file without existence check

`finally: os.remove(tmp_path)` (line 349) unconditionally removes the temp path. But when `--temp_dlc` is given and an `OSError` occurs during the initial write (e.g. disk full), `tmp_path` may never have been created, and `os.remove()` raises a secondary `FileNotFoundError` that obscures the original error.

**Impact**: If the initial `open(tmp_path, "wb")` fails on the `--temp_dlc` path, the `finally` clause raises a secondary `FileNotFoundError`, hiding the original `OSError` from the user.

**Suggested fix:**

Add an existence check:
```
finally:
    if os.path.exists(tmp_path):
        os.remove(tmp_path)
```
Or more Pythonically:
```
finally:
    try:
        os.remove(tmp_path)
    except FileNotFoundError:
        pass
```

setup.py
```diff
@@ -226,6 +226,18 @@ def finalize_options(self):
 # Extra files such as EULA and ThirdPartyNotices (and Qualcomm License, only for QNN release packages)
 extra = ["LICENSE", "ThirdPartyNotices.txt", "Privacy.md", "Qualcomm_LICENSE.pdf"]
 
+# QAIRT SDK pybind extensions bundled next to deploy_multi_soc_ep_context.py
+# (.so on Linux, .pyd on Windows). The SDK ships one extension per Python
+# version (libDlModelToolsPy312 for 3.12, libDlModelToolsPy310 for 3.10) but the
+# tool is only packaged into the Python 3.12 wheel, so the 312 module is the only
+# one we ever bundle. cmake copies just that file into the package directory for
+# 3.12 builds; for any other version nothing is copied and this glob matches
```
huaychou: **[M-5] (Major)** setup.py package_data lists both .so and .pyd regardless of current platform

When Python 3.12 is active, `dlc_utils_libs` always contains `["libDlModelToolsPy312.so", "libDlModelToolsPy312.pyd"]` regardless of whether the current platform is Linux or Windows. Listing filenames in `package_data` that will never exist in the package directory on the current platform is misleading: on Linux there is never a `.pyd`; on Windows there is never a `.so`.

**Impact**: Some versions of the `wheel` / MANIFEST tools warn or error on missing `package_data` entries. More importantly, this is asymmetric with the cmake copy logic (which copies only the platform-appropriate file), creating an inconsistency that is likely to confuse future maintainers.

**Suggested fix:**

Select the correct extension based on the current platform:
```
import platform as _platform

if sys.version_info[:2] == (3, 12):
    _ext = ".pyd" if _platform.system() == "Windows" else ".so"
    dlc_utils_libs = [f"libDlModelToolsPy312{_ext}"]
else:
    dlc_utils_libs = []
```

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -204,6 +204,52 @@
...
+    foreach(QNN_DLMODELTOOLS_LIB ${QNN_DLMODELTOOLS_LIBS})
+      add_custom_command(
+        TARGET ${onnxruntime_providers_qnn_target} POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different "${QNN_DLMODELTOOLS_LIB}" $<TARGET_FILE_DIR:${onnxruntime_providers_qnn_target}>/onnxruntime_qnn
+        COMMENT "Copying ${QNN_DLMODELTOOLS_LIB} to onnxruntime_qnn for Python Wheel"
+      )
+    endforeach()
+    # Only package the deploy tool alongside its libDlModelToolsPy312 dependency,
+    # so the extension (3.12) and the script always ship together.
+    add_custom_command(
```
huaychou: **[M-6] (Major)** CMake: deploy script is copied even when GLOB returns empty, breaking the "ship together with dependency" invariant

Inside the `elseif (EXISTS ...)` block (line 231), `file(GLOB QNN_DLMODELTOOLS_LIBS ...)` (lines 232–234) lists both `libDlModelToolsPy312.so` and `libDlModelToolsPy312.pyd`. After `endforeach()` at line 241, the `add_custom_command` that copies `deploy_multi_soc_ep_context.py` (lines 244–250) still executes within the same `elseif` block, **unconditionally with respect to the GLOB result**.

If the directory exists but neither `libDlModelToolsPy312` file is present (e.g. after an SDK upgrade that renamed the file to `libDlModelToolsPy313`), GLOB returns empty, `foreach` does nothing, but `deploy_multi_soc_ep_context.py` is still copied into the wheel. This directly contradicts the comment at lines 242–243: "Only package the deploy tool alongside its libDlModelToolsPy312 dependency, so the extension (3.12) and the script always ship together."

**Impact**: The wheel ships the deploy script without its pybind extension, causing an immediate `ModuleNotFoundError: No module named 'libDlModelToolsPy312'` at runtime. The setup.py package_data glob will also silently skip the missing file, so no build-time warning is generated.

**Suggested fix:**

After endforeach(), gate the deploy script copy on the GLOB result:

endforeach()
if (QNN_DLMODELTOOLS_LIBS)
  add_custom_command(
    TARGET ${onnxruntime_providers_qnn_target} POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy_if_different
        ${ONNXRUNTIME_ROOT}/python/multi_soc/deploy_multi_soc_ep_context.py
        $/onnxruntime_qnn
    COMMENT "Copying deploy_multi_soc_ep_context.py to onnxruntime_qnn for Python Wheel"
  )
endif()

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,537 @@
...
+        # No standalone DLC -> embed mode (binary lives in the ONNX model).
+        # Each mode validates its own arguments.
+        if not args.input_dlc:
+            run_embed_mode(args)
+        else:
+            run_non_embed_mode(args)
+
+    except SystemExit:
+        raise
+    except Exception:
```
huaychou, line 567: **[N-2] (Minor)** Outer except Exception in main() swallows pybind init errors

`main()`'s `except Exception: traceback.print_exc(); sys.exit(-2)` is better than silently swallowing, but [P07] requires specific exception types. More concretely: if `libDlModelToolsPy312` loading raises `ImportError` (after the try/except guard from M-1 is applied), this `except Exception` would convert it to a traceback dump + exit code -2 rather than a clean version message.

**Impact**: Exit code -2 is not a standard POSIX exit code; shell scripts checking `$?` will see 254 (`256 - 2`), which is unexpected and hard to handle.

**Suggested fix:**

Use standard exit code 1, and catch specific exception types with corresponding messages; or at minimum change sys.exit(-2) to sys.exit(1). If keeping a generic fallback, explicitly exclude cases already handled by M-1 / M-2 / M-3.

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -204,6 +204,52 @@
...
+    # QAIRT SDK ships libDlModelToolsPy in limited Python version. Considering
+    # QNN-EP's supported Python version, only bundle it in Python 3.12 wheel.
+    message(STATUS "libDlModelToolsPy is only bundled for the Python 3.12 wheel; "
+                   "skipping for Python ${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}")
+  elseif (EXISTS "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}")
+    file(GLOB QNN_DLMODELTOOLS_LIBS LIST_DIRECTORIES false
+         "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}/libDlModelToolsPy312.so"
+         "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}/libDlModelToolsPy312.pyd")
+    foreach(QNN_DLMODELTOOLS_LIB ${QNN_DLMODELTOOLS_LIBS})
+      add_custom_command(
```
huaychou, line 249: **[N-6] (Minor)** CMake GLOB at configure time silently produces empty list when SDK files are absent

`file(GLOB QNN_DLMODELTOOLS_LIBS ...)` runs at cmake configure time. If the SDK's `.so` / `.pyd` files are absent (e.g. SDK was upgraded but the directory structure changed), GLOB silently returns an empty list, the `foreach` loop does nothing, and no warning or error is emitted. This means `libDlModelToolsPy312` will not be packaged into the wheel, with no notification to the developer.

**Impact**: If a developer switches QAIRT SDK versions and the DLC utils file path has changed, the build silently produces a wheel without `libDlModelToolsPy312`. This is only discovered by end users when they try to run the tool.

**Suggested fix:**

After the GLOB, check whether the result is empty and emit a WARNING:

file(GLOB QNN_DLMODELTOOLS_LIBS LIST_DIRECTORIES false
     "...libDlModelToolsPy312.so"
     "...libDlModelToolsPy312.pyd")
if (NOT QNN_DLMODELTOOLS_LIBS)
  message(WARNING "libDlModelToolsPy312 not found at "
                  "${onnxruntime_QNN_HOME}/lib/python/qti/aisw/dlc_utils/${QNN_DLC_UTILS_SUBDIR}; "
                  "deploy_multi_soc_ep_context will be unusable in the built wheel.")
else()
  foreach(...) ... endforeach()
endif()

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,537 @@
...
+import re
+import sys
+import tempfile
+import traceback
+
+import onnx
+
+# libDlModelToolsPy312 is a QAIRT SDK pybind extension copied flat into this
+# package at wheel-build time; make its directory importable regardless of cwd.
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
```
huaychou, line 42: **[N-3] (Minor)** sys.path.insert(0, ...) mutates import path without sufficient documentation

`sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` prepends the script's own directory to `sys.path[0]` to locate `libDlModelToolsPy312`. Per [P06], `pathlib.Path` should be preferred over `os.path`. In an installed wheel environment, the `__file__` directory is already the `onnxruntime_qnn/` package directory and is already on `sys.path`, making this line sometimes redundant.

**Impact**: Inserting at index 0 makes this directory take priority over all other paths, which could shadow identically-named modules elsewhere (unlikely given the unique name, but not ideal). The comment explaining why this is necessary should be more explicit about the source-checkout vs. installed-wheel distinction.

**Suggested fix:**

Use pathlib.Path and add a more explicit comment:
```
import sys
from pathlib import Path

# When running from a source checkout, libDlModelToolsPy312 is not on sys.path;
# add this file's directory so the co-located extension can be found.
# In an installed wheel this is usually a no-op (the package dir is already on sys.path).
_pkg_dir = Path(__file__).parent
if str(_pkg_dir) not in sys.path:
    sys.path.insert(0, str(_pkg_dir))
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -353,6 +353,84 @@ ep_options = {
...
+
+The script works on both EP context model layouts, selected by which input you pass:
+
+* **Non-embed mode** (`ep.context_embed_mode = "0"`, the default): the DLC is a standalone `*_qnn.bin` file alongside the ONNX model. Pass it via `--input_dlc` (and `--input_onnx` for the metadata). The stripped DLC is written to `--output_dlc` and the updated model to `--output_onnx`; both inputs are left untouched. Pass `--inplace` to overwrite the inputs instead.
+* **Embed mode** (`ep.context_embed_mode = "1"`): the DLC is embedded inside the EP context node of the ONNX model. Omit `--input_dlc` and pass the model via `--input_onnx`; the tool extracts the embedded DLC to a temporary file, edits it, and re-embeds the result into `--output_onnx`. The input model is left untouched. Pass `--inplace` to overwrite `--input_onnx` instead.
+
+Run it as a module from an installed wheel:
+
+```bash
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context --input_dlc model_ctx_qnn.bin --list
```
huaychou: I think the --input_dlc with `.bin` file is quite confusing. How does user identify the `.bin` is a context binary or DLC file?

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,538 @@
...
+
+# Only the Python 3.12 build is bundled into the wheel, so it is the sole module
+# imported here; all classes/enums live at its top level.
+import libDlModelToolsPy312 as modeltools
```
huaychou: **[M-1] (Major)** Top-level import of optional native extension without try/except

`import libDlModelToolsPy312 as modeltools` sits at module top-level with no `try/except ImportError` guard. On any environment other than Python 3.12 (or whenever the native extension has not been copied into the package directory), any `import onnxruntime_qnn.deploy_multi_soc_ep_context` or `from onnxruntime_qnn import ...` will raise `ModuleNotFoundError` at *import time* instead of giving a helpful message when the tool is actually invoked.

**Impact**: If the wheel's `__init__.py` or any other module indirectly imports this file via `import onnxruntime_qnn.multi_soc`, the entire package import fails and affects users on all Python versions. Even without an indirect import, the bare `ModuleNotFoundError: No module named 'libDlModelToolsPy312'` gives no indication that the tool is Python 3.12-only, making the failure hard to diagnose.

**Suggested fix:**

Move the import inside a try/except and provide an actionable version message:
```
try:
    import libDlModelToolsPy312 as modeltools
except ImportError as e:
    modeltools = None
    _MODELTOOLS_IMPORT_ERROR = (
        "libDlModelToolsPy312 is not available. "
        "This tool requires Python 3.12 and the onnxruntime-qnn wheel "
        "built for Python 3.12. "
        f"Original error: {e}"
    )
```
Then at the start of main() (or each entry point):
```
if modeltools is None:
    sys.exit(_MODELTOOLS_IMPORT_ERROR)
```

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,538 @@
...
+    is all-or-nothing."""
+    record_name = record_name.strip()
+    if not record_name:
+        return
+
+    remove_result = updater.remove_record_and_get_names(record_name, modeltools.DlcRecordType.HTP_CACHE_RECORD)
+    remove_success, removed_record_names = remove_result
+
+    if not remove_success:
+        sys.exit(
```
huaychou: **[M-2] (Major)** sys.exit() called inside library function remove_records()

`remove_records()` calls `sys.exit()` on failure (lines 127, 132). This function is called inside the `try/finally` block in `run_embed_mode()`. `sys.exit()` raises `SystemExit`, a `BaseException` subclass, so the `finally` clause does run (the temp DLC is cleaned up, which is correct). However, any work that was supposed to happen after `remove_records()` returns — notably reading back the saved DLC bytes and re-embedding them into the output ONNX model — is silently skipped, with no way for the caller to distinguish success from early termination.

**Impact**: In `run_embed_mode()`, when `remove_records()` is called in a loop, a failure on any one record causes all subsequent removals and the re-embed step to be skipped, leaving the output model in an inconsistent state (ONNX metadata already updated, DLC records not fully removed).

**Suggested fix:**

Replace sys.exit() with a raised exception (e.g. RuntimeError or a custom exception class) so callers can handle the error explicitly. For example:
```
raise RuntimeError(
    f"Removing record {record_name} unsuccessful. ..."
)
```

Callers (run_embed_mode, run_non_embed_mode) can then propagate through main()'s outer except, which converts it to sys.exit().

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,538 @@
...
+        list_records(updater)
+
+
+def main():
+    try:
+        parser = argparse.ArgumentParser(
+            description="List and remove HTP_CACHE_RECORD entries in a DLC / EP-context file.",
+            formatter_class=argparse.RawTextHelpFormatter,
+        )
+        parser._action_groups.pop()
```
huaychou: **[T-1] (Nit)** argparse._action_groups is a private attribute and should not be accessed

`parser._action_groups.pop()` accesses a private attribute of `argparse.ArgumentParser` to remove the default positional/optional groups in order to use custom group names. This is a common workaround but could break on future Python versions if the internal structure changes.

**Suggested fix:**

If the goal is cleaner help output, consider simply adding a named argument group without popping the defaults:
```
optional = parser.add_argument_group('arguments')
```
If the pop is truly needed, at minimum add a TODO comment noting the Python version compatibility risk.

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -204,6 +204,52 @@
...
+  # lib/python/qti/aisw/dlc_utils/__init__.py for the canonical mapping).
+  set(QNN_DLC_UTILS_SUBDIR "")
+  if (QNN_ARCH_ABI STREQUAL "x86_64-linux-clang")
+    set(QNN_DLC_UTILS_SUBDIR "linux-x86_64")
+  elseif (QNN_ARCH_ABI STREQUAL "aarch64-oe-linux-gcc11.2")
+    set(QNN_DLC_UTILS_SUBDIR "linux-aarch64-oe-gcc11.2")
+  elseif (QNN_ARCH_ABI STREQUAL "x86_64-windows-msvc")
+    set(QNN_DLC_UTILS_SUBDIR "windows-x86_64")
+  elseif (QNN_ARCH_ABI STREQUAL "aarch64-windows-msvc" OR QNN_ARCH_ABI STREQUAL "arm64x-windows-msvc")
+    set(QNN_DLC_UTILS_SUBDIR "windows-arm64ec")
```
huaychou, line 220: I think this is not correct. Arm64ec is used for x64 python on Arm64 device. I think we don't have arm64 native pybind library in QAIRT sdk.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: The impact described above is incorrect. The original model/DLC are intentional left unmodified if anything goes wrong in the middle. Only everything is correctly modified then they will be saved. This is to keep they are either all non-modified or all modified.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: I think the statement is incorrect.
> In an installed wheel environment, the __file__ directory is already the onnxruntime_qnn/ package directory and is already on sys.path, making this line sometimes redundant.

If execute the script through `python -m onnxruntime_qnn.deploy_multi_soc_ep_context`, this insert is still necessary.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: Like the [comment](https://github.com/onnxruntime/onnxruntime-qnn/pull/499#discussion_r3611769488), I think current implementation is correct.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: Added check.

setup.py
minfhong-qti: Added platform check.

cmake/onnxruntime_providers_qnn.cmake
minfhong-qti: Added empty check.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: The import error from `libDlModelToolsPy312` is now explicitly handled with message. Thus, I think the concern here no longer exists.

cmake/onnxruntime_providers_qnn.cmake
minfhong-qti: Fixed along with above similar comment.

docs/execution_providers/QNN-ExecutionProvider.md
minfhong-qti: Rename to `--input_bin` and rephrase all the wording in the script. Use wording "binary" when not sure it is context binary or FCB and use wording "FCB" when known for sure is FCB.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: Add `try-except ` and provide helpful message. But I directly raise here instead of in main.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: Revised to not using private attribute.

cmake/onnxruntime_providers_qnn.cmake
minfhong-qti: Removed `aarch64-windows-msvc` part.

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -204,6 +204,58 @@
         $<TARGET_FILE_DIR:${onnxruntime_providers_qnn_target}>/onnxruntime_qnn
     COMMENT "Copying version, license, README, and release notes files"
   )
+
+  # Bundle libDlModelToolsPy (used by deploy_multi_soc_ep_context.py) from the
+  # QAIRT SDK's dlc_utils into the wheel. The dlc_utils platform sub-directory
```
huaychou, line 209: Just wondering, do we release libDlModelToolsPy with archive or just wheel?

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -204,6 +204,58 @@
...
+  # name differs from QNN_ARCH_ABI, so map it explicitly (see the SDK's
+  # lib/python/qti/aisw/dlc_utils/__init__.py for the canonical mapping).
+  set(QNN_DLC_UTILS_SUBDIR "")
+  if (QNN_ARCH_ABI STREQUAL "x86_64-linux-clang")
+    set(QNN_DLC_UTILS_SUBDIR "linux-x86_64")
+  elseif (QNN_ARCH_ABI STREQUAL "aarch64-oe-linux-gcc11.2")
```
huaychou, line 215: I think we don't build release wheel for aarch64-oe-linux-gcc11.2 target. We build wheel for aarch64_manylinux_2_34 with aarch64-oe-linux-gcc11.2 QAIRT artifacts.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
huaychou: Even though there is no impact, I think using runtime error with error handling is better than sys.exit() 🤔

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -355,6 +355,84 @@ ep_options = {
...
+|`--output_onnx`|**Required with `--remove_record`** (both modes) unless `--inplace` is given. Destination path for the updated ONNX model. In embed mode the stripped binary is re-embedded into it; in non-embed mode it carries the updated metadata. The input `--input_onnx` is left untouched.|
+|`--output_bin`|**Required with `--remove_record` in non-embed mode** unless `--inplace` is given. Destination path for the stripped binary. The input `--input_bin` is left untouched.|
+|`--temp_bin`|*Embed mode only, optional.* Path to stage the extracted binary while it is edited. When omitted, an auto-generated temporary file is used. The file is removed when the tool finishes either way.|
+|`--inplace`|*Optional.* Overwrite the input model (and, in non-embed mode, the input binary) in place instead of writing to new paths. Mutually exclusive with `--output_onnx` / `--output_bin`.|
+
+**Listing the packaged SoCs:**
+
+```bash
+# Non-embed mode (standalone context binary)
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context --input_bin model_ctx_qnn.bin --list
```
kuanyul-qti, line 389: Just want to check whether we can leave out `--input_onnx` here?

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -355,6 +355,84 @@ ep_options = {
...
+**Removing a SoC (non-embed mode)** — writes the pruned binary and updated model to new paths, leaving the inputs untouched:
+
+```bash
+python -m onnxruntime_qnn.deploy_multi_soc_ep_context \
+    --input_bin model_ctx_qnn.bin \
+    --input_onnx model_ctx.onnx \
+    --output_bin model_ctx_qnn.pruned.bin \
+    --output_onnx model_ctx.pruned.onnx \
+    -r backend.metadata0 \
+    --list
```
kuanyul-qti: Is `--list` required in this case, or can it be left as an optional argument?

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,563 @@
...
+_EP_COMPAT_KEY = "ep_compatibility_info.QNNExecutionProvider"
+
+# Positions of the parallel (DspArch, SoCModel, VTCMSize) lists in the
+# ':'-separated value, and the expected field count.
+_LIST_FIELD_INDICES = (4, 5, 6)
+_NUM_FIELDS = 8
+
+# V2-format field layout: the value is only editable when field 0 is "v2".
+_VERSION_FIELD_INDEX = 0
+_EXPECTED_VERSION = "v2"
```
kuanyul-qti: Prefer to place these variables before any function definitions. (L64).

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
```diff
@@ -0,0 +1,563 @@
...
+            )
+        names_to_remove.append(name)
+        removed_indices.append(name_to_index[name])
+    return names_to_remove, removed_indices
+
+
+# Attribute names on the EPContext node, mirroring onnx_ctx_model_helper.h.
+_EPCONTEXT_OP = "EPContext"
+_EMBED_MODE_ATTR = "embed_mode"
+_EP_CACHE_CONTEXT_ATTR = "ep_cache_context"
```
kuanyul-qti: Ditto.

huaychou (commented):
Do we have a test plan for this tool?

docs/execution_providers/QNN-ExecutionProvider.md
minfhong-qti: Yes. Binary file only is enough for listing. ONNX model is only required if want to remove.

cmake/onnxruntime_providers_qnn.cmake
minfhong-qti: No. Archive packaging is explicitly listed in `tools/ci_build/pkg_assets.py`.

cmake/onnxruntime_providers_qnn.cmake
minfhong-qti: Target `build_ort_linux_aarch64_manylinux_2_34` uses `QNN_ARCH_ABI=aarch64-oe-linux-gcc11.2` as well.
https://github.com/onnxruntime/onnxruntime-qnn/blob/main/qcom/build_and_test.py#L252

docs/execution_providers/QNN-ExecutionProvider.md
minfhong-qti: Not required. Removed to avoid ambiguity.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: Moved.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: Moved.

onnxruntime/python/multi_soc/deploy_multi_soc_ep_context.py
minfhong-qti: Revised to raising a custom error to differentiate with runtime error with unknown root cause.

minfhong-qti:
> Do we have a test plan for this tool?

I would like to have a Python UT for this but I think current CI is not fully ready for QNN EP's Python UT. I prefer a follow-up PR for this.

cmake/onnxruntime_providers_qnn.cmake
huaychou: Acknowledge

huaychou (approved):

kuanyul-qti (approved):

PR #500, opened 2026-06-09
[QNN EP] Stop advertising QNN CPU backend by default

qti-niscmami:
### Description

  Hide the QNN CPU `OrtHardwareDevice` by default. CPU is advertised (and
  `backend_type=cpu` accepted) only when `ORT_QNN_ENABLE_CPU_BACKEND` is set,
  via a new `QnnCpuBackendEnabled()` gate in `qnn_provider_factory.cc`. The
  gtest harness sets the flag for itself, so all CPU/HTP/GPU tests still pass.
  No backend code is removed.

  ### Motivation and Context

  `GetSupportedDevicesImpl` claimed any CPU device, so `GetEpDevices()` returned
  a `(QNN, CPU)` entry everywhere and `SetEpPolicy(PREFER_CPU)` routed to QNN EP,
  which then failed loading `libQnnCpu`. This is the small, reversible interim
  fix ahead of the full removal in #470.

Setting the env var restores the old behavior.


qti-mattsinc:
@minfhong-qti, I remember we had a discussion about this advertised CPU device previously, and you had mentioned it was needed for Linux HTP emulator support. I don't follow the HTP side of things very closely, but I saw this PR and figured I'd ping you here in case you want to check this change if you haven't already.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -37,6 +38,25 @@ static const std::unordered_map<OrtHardwareDeviceType, std::string> kSupportedBa
     {OrtHardwareDeviceType_GPU, "gpu"},
 };
 
+// QNN CPU is not advertised by default; set ORT_QNN_ENABLE_CPU_BACKEND to re-enable it (used by tests).
+static bool QnnCpuBackendEnabled() {
```
minfhong-qti, line 46: I think CPU should always be enabled on x86 environments as they are for HTP emulation.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
```diff
@@ -27,6 +27,18 @@
 namespace onnxruntime {
 namespace test {
 
+// Tests rely on the QNN CPU device being advertised; opt in before any EP registration (pre-main).
+namespace {
+[[maybe_unused]] const bool g_enable_qnn_cpu_backend_for_tests = []() {
```
minfhong-qti: Should we set it in main instead of here which may be a bit misleading?

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-niscmami: Sure. Will make QnnCpuBackendEnabled() auto-enable on x86 and keep the env var for arm64 override.

onnxruntime/test/providers/qnn/qnn_test_utils.cc
qti-niscmami: Will move it into main

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-niscmami: Done

onnxruntime/test/providers/qnn/qnn_test_utils.cc
qti-niscmami: Done

qti-chuteng:
### 🟠 Major

**[M-1]** `setup.py → qnn_deps` (lines ~154 and ~183, not in this PR's diff)
`pkg_assets.py` and `generate_nuspec_for_native_nuget.py` both drop `QnnCpu.dll` / `libQnnCpu.so`, but `setup.py` still lists them in both the Linux and Windows `qnn_deps` lists. All three packaging scripts must be kept in sync — if `QnnCpu.*` is not present in the build environment, Python wheel packaging will fail; if it is present, the wheel will ship the CPU backend DLL while the NuGet/zip do not, producing inconsistent release artifacts.

Suggested fix: remove `"libQnnCpu.so"` from the Linux `qnn_deps` list (~line 154) and `"QnnCpu.dll"` from the Windows `qnn_deps` list (~line 183) in `setup.py`, mirroring the changes already made in `pkg_assets.py`.

---

### 🔵 Minor

**[N-1]** `docs/execution_providers/QNN-ExecutionProvider.md → backend_path table`
Two documentation gaps: (1) The `backend_path` table at line 90 still lists `QnnCpu.dll` / `libQnnCpu.so` as available packaged options, but those files are now removed from the release package. (2) The new `ORT_QNN_ENABLE_CPU_BACKEND` env var is not documented anywhere in the EP docs, leaving arm64 developers without guidance on how to opt in to the CPU backend.

Suggested fix: (1) Add a note in the CPU row of the `backend_path` table explaining the file is no longer in the release package (still available from the SDK directly). (2) Add a short paragraph near the Provider Options section documenting `ORT_QNN_ENABLE_CPU_BACKEND`: always-enabled on x86; opt-in on arm64.

qti-niscmami:
> ### 🟠 Major
> **[M-1]** `setup.py → qnn_deps` (lines ~154 and ~183, not in this PR's diff) `pkg_assets.py` and `generate_nuspec_for_native_nuget.py` both drop `QnnCpu.dll` / `libQnnCpu.so`, but `setup.py` still lists them in both the Linux and Windows `qnn_deps` lists. All three packaging scripts must be kept in sync — if `QnnCpu.*` is not present in the build environment, Python wheel packaging will fail; if it is present, the wheel will ship the CPU backend DLL while the NuGet/zip do not, producing inconsistent release artifacts.
> 
> Suggested fix: remove `"libQnnCpu.so"` from the Linux `qnn_deps` list (~line 154) and `"QnnCpu.dll"` from the Windows `qnn_deps` list (~line 183) in `setup.py`, mirroring the changes already made in `pkg_assets.py`.
> 
> ### 🔵 Minor
> **[N-1]** `docs/execution_providers/QNN-ExecutionProvider.md → backend_path table` Two documentation gaps: (1) The `backend_path` table at line 90 still lists `QnnCpu.dll` / `libQnnCpu.so` as available packaged options, but those files are now removed from the release package. (2) The new `ORT_QNN_ENABLE_CPU_BACKEND` env var is not documented anywhere in the EP docs, leaving arm64 developers without guidance on how to opt in to the CPU backend.
> 
> Suggested fix: (1) Add a note in the CPU row of the `backend_path` table explaining the file is no longer in the release package (still available from the SDK directly). (2) Add a short paragraph near the Provider Options section documenting `ORT_QNN_ENABLE_CPU_BACKEND`: always-enabled on x86; opt-in on arm64.

Addressed feedback.

minfhong-qti (approved):
Kindly remind there is merge conflict.

qti-chuteng (approved):

PR #502, opened 2026-06-09
[QNN EP] Revert "[OpBuilder] Fix MatMul uint16 handling."

qti-mbadnara:
This PR reverts #300 as part of the mainline


tirupath-qti (approved):
Reverting for recovering accuracy for a critical customer usecase.

tirupath-qti:
Let us wait for some more analysis from the team to see if there is a way to limit the revert to a specific configuration to unblock the customer usecase.

PR #504, opened 2026-06-09
Add cosine similarity verifier

qti-mattsinc:
### Description
* Extend `EPVerificationParams` to support multiple verifiers, and add a cosine similarity verifier.
* Switch `EPVerificationParams` to be a direct arg to `RunQnnModelTest` instead of having multiple args for each struct member. This means that this commit is quite large, as all existing callsites must be updated. But, this will simplify future additions to `EPVerificationParams` without having to pollute the signature of `RunQnnModelTest` with more optional args.

### Motivation and Context
* Enables better testing for ops where CS is a better metric that elementwise comparison. This verifier is not currently used in any of the op tests, but the plan is to use it in GroupQueryAttention UTs, which will be added later.


onnxruntime/test/util/test_utils.cc
```diff
@@ -17,10 +18,70 @@
...
+  if (verifier.index() == 0) {
+    VerifyElementwiseAbsoluteOutput(output_name, expected_data, actual_data, element_count,
+                                    std::get<ElementwiseAbsoluteVerifier>(verifier));
+  } else if (verifier.index() == 1) {
+    VerifyCosineSimilarityOutput(output_name, expected_data, actual_data, element_count,
+                                 std::get<CosineSimilarityVerifier>(verifier));
+  } else {
+    QNN_ASSERT(!"Unknown verifier type");
+  }
+}
```
qti-mattsinc: Alternatively, we could use `std::visit` w/ template magic to avoid manual checks against the variant index.

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/test/util/test_utils.cc:L56`
`VerifyFloatOutput` dispatches on `verifier.index()` (raw ordinal) instead of using `std::visit` or `std::holds_alternative`. If a future contributor inserts or reorders types in `TensorVerifier`, the dispatch silently swaps without any compiler error. Additionally, `QNN_ASSERT` is a no-op in Release builds — replace with `FAIL()` so an unknown verifier type always produces a GTest failure.
```suggestion
  std::visit(onnxruntime::test::overloaded{
    [&](const ElementwiseAbsoluteVerifier& v) {
      VerifyElementwiseAbsoluteOutput(output_name, expected_data, actual_data, element_count, v);
    },
    [&](const CosineSimilarityVerifier& v) {
      VerifyCosineSimilarityOutput(output_name, expected_data, actual_data, element_count, v);
    }
  }, verifier);
```

**[M-2]** `onnxruntime/test/providers/qnn/qnn_test_utils.h → VerifyQDQOutput():L872`
`VerifyQDQOutput` constructs `EPVerificationParams{ExpectedEPNodeAssignment::All, ...}` just to call `VerifyOutput`, but `ep_node_assignment` has no effect there (the author even added a comment acknowledging this). `VerifyOutput` should accept `TensorVerifier` directly rather than the full `EPVerificationParams`, or provide an overload that takes only `TensorVerifier` so callers without an EP session context are not forced to supply a dummy field.

---

### 🔵 Minor

**[N-1]** `onnxruntime/test/providers/qnn/logical_compare_test.cc → EqualToCast4D():L793`
The PR documents that the `1` passed as `fp32_abs_err` was a historical accident (originally `num_nodes_in_ep`) but still migrates it as `ElementwiseAbsoluteVerifier(1)` — a tolerance of 1.0, which is far too loose for a Cast op whose outputs are 0.0 or 1.0 exactly. The bug is documented but not fixed. Either fix the tolerance (e.g., `1e-5f` with an explicit `graph_verifier` to check node count), or revert the change to this test until it can be properly addressed.

chunghow-qti:
Could you also include GQA test here or somewhere?

qti-mattsinc:
> ### 🟠 Major
> 
> **[M-1]** `onnxruntime/test/util/test_utils.cc:L56` `VerifyFloatOutput` dispatches on `verifier.index()` (raw ordinal) instead of using `std::visit` or `std::holds_alternative`. If a future contributor inserts or reorders types in `TensorVerifier`, the dispatch silently swaps without any compiler error. Additionally, `QNN_ASSERT` is a no-op in Release builds — replace with `FAIL()` so an unknown verifier type always produces a GTest failure.
> 
> ```
>   std::visit(onnxruntime::test::overloaded{
>     [&](const ElementwiseAbsoluteVerifier& v) {
>       VerifyElementwiseAbsoluteOutput(output_name, expected_data, actual_data, element_count, v);
>     },
>     [&](const CosineSimilarityVerifier& v) {
>       VerifyCosineSimilarityOutput(output_name, expected_data, actual_data, element_count, v);
>     }
>   }, verifier);
> ```

See https://github.com/onnxruntime/onnxruntime-qnn/pull/504#discussion_r3384484882. I want to know if others think making use the variadic `overloaded` template is stylistically ok in this codebase, not just what Claude's opinion is.

> **[M-2]** `onnxruntime/test/providers/qnn/qnn_test_utils.h → VerifyQDQOutput():L872` `VerifyQDQOutput` constructs `EPVerificationParams{ExpectedEPNodeAssignment::All, ...}` just to call `VerifyOutput`, but `ep_node_assignment` has no effect there (the author even added a comment acknowledging this). `VerifyOutput` should accept `TensorVerifier` directly rather than the full `EPVerificationParams`, or provide an overload that takes only `TensorVerifier` so callers without an EP session context are not forced to supply a dummy field.

Done

> ### 🔵 Minor
> 
> **[N-1]** `onnxruntime/test/providers/qnn/logical_compare_test.cc → EqualToCast4D():L793` The PR documents that the `1` passed as `fp32_abs_err` was a historical accident (originally `num_nodes_in_ep`) but still migrates it as `ElementwiseAbsoluteVerifier(1)` — a tolerance of 1.0, which is far too loose for a Cast op whose outputs are 0.0 or 1.0 exactly. The bug is documented but not fixed. Either fix the tolerance (e.g., `1e-5f` with an explicit `graph_verifier` to check node count), or revert the change to this test until it can be properly addressed.

Out of scope of this PR.

qti-mattsinc:
> Could you also include GQA test here or somewhere?

The GQA tests will be in a separate change, because they depend on both the OpBuilder and this PR. See https://github.com/onnxruntime/onnxruntime-qnn/pull/424#issuecomment-4662683807.

onnxruntime/test/util/test_utils.cc
```diff
@@ -17,10 +18,70 @@
...
+  EXPECT_GE(cosine_similarity, verifier.fp32_cs_threshold) << "Cosine similarity below threshold for " << output_name;
+}
+
+template <typename T>
+void VerifyFloatOutput(const std::string& output_name,
+                       const T* expected_data,
+                       const T* actual_data,
+                       size_t element_count,
+                       const TensorVerifier& verifier) {
+  if (verifier.index() == 0) {
```
huaychou:
```suggestion
  if (std::holds_alternative<ElementwiseAbsoluteVerifier>(verifier)) {
```

onnxruntime/test/util/test_utils.cc
```diff
@@ -17,10 +18,70 @@
...
+    VerifyElementwiseAbsoluteOutput(output_name, expected_data, actual_data, element_count,
+                                    std::get<ElementwiseAbsoluteVerifier>(verifier));
+  } else if (verifier.index() == 1) {
```
huaychou:
```suggestion
  } else if (std::holds_alternative<CosineSimilarityVerifier>(verifier)) {
```

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1946,6 +1942,32 @@ TEST_F(QnnGPUBackendTests, AutoEp_GpuOnly) {
   ASSERT_ORTSTATUS_OK(Ort::GetApi().UnregisterExecutionProviderLibrary(*ort_env, kQnnExecutionProvider));
 }
 
+TEST_F(QnnGPUBackendTests, ElementWiseAbsoluteVerifier) {
```
huaychou:
```suggestion
TEST_F(QnnGPUBackendTests, ElementwiseAbsoluteVerifier) {
```

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1946,6 +1942,32 @@ TEST_F(QnnGPUBackendTests, AutoEp_GpuOnly) {
...
+  ProviderOptions options;
+  options["backend_type"] = "gpu";
+
+  auto input_defs = {TestInputDef<float>({1, 3, 4, 4}, false, -10.0f, 10.0f),
+                     TestInputDef<float>({1, 3, 4, 4}, false, -10.0f, 10.0f)};
+
+  RunQnnModelTest(BuildOpTestCase<float>("Add_node", "Add", input_defs, {}, {}, kOnnxDomain),
+                  options,
+                  13,
+                  EPVerificationParams{ExpectedEPNodeAssignment::All, CosineSimilarityVerifier(0.99)});
```
huaychou:
```suggestion
                  EPVerificationParams{ExpectedEPNodeAssignment::All, CosineSimilarityVerifier(0.99f)});
```

onnxruntime/test/util/include/test_utils.h
```diff
@@ -35,14 +35,27 @@ enum class ExpectedEPNodeAssignment { None,
...
 
   // Some EP may use different rounding than ORT CPU EP, which may cause a bigger abs error than
   // the default of 1e-5f, especially for scenarios such as [Q -> Quantized op -> DQ]
   // Set this only if this is necessary
-  float fp32_abs_err = 1e-5f;
+  float fp32_abs_err;
+};
+
+struct CosineSimilarityVerifier {
+  explicit CosineSimilarityVerifier(float fp32_cs_threshold) : fp32_cs_threshold(fp32_cs_threshold) {}
```
huaychou, line 48: Should we have a default value for fp32_cs_threshold?

onnxruntime/test/util/test_utils.cc
```diff
@@ -17,10 +18,70 @@
...
+    const float expected_float = static_cast<float>(expected_data[element_index]);
+    const float actual_float = static_cast<float>(actual_data[element_index]);
+    dot += expected_float * actual_float;
+    expected_norm_sq += expected_float * expected_float;
+    actual_norm_sq += actual_float * actual_float;
+  }
+  const float denom = std::sqrt(expected_norm_sq) * std::sqrt(actual_norm_sq);
+  const float cosine_similarity = denom > 0.0f
+                                      ? dot / denom
+                                      : ((expected_norm_sq == 0.0f && actual_norm_sq == 0.0f) ? 1.0f : 0.0f);
```
huaychou, line 58: Could we add a comment for this check? I think it is only triggered when all the expected and actual value are 0.0f

onnxruntime/test/providers/qnn/qnn_basic_test.cc
qti-mattsinc: Done

onnxruntime/test/providers/qnn/qnn_basic_test.cc
qti-mattsinc: Done

onnxruntime/test/util/include/test_utils.h
qti-mattsinc: We could, although I don't necessarily think we should. Requiring a value to be provided isn't much of a burden, and `CosineSimilarityVerifier(0.99f)` is clearer than  `CosineSimilarityVerifier()` at a glance. I guess the same could be said of `ElementwiseAbsoluteVerifier` though.

onnxruntime/test/util/test_utils.cc
qti-mattsinc: It's triggers if at least one of the two are all 0, and makes the cs value reasonable in that case. Added a comment for it in latest commit.

(An alternative to this would be always adding a small epsilon to `denom`, but for our use-case of just verifying the EP output, it's simplest just to check for the null vector case.)

onnxruntime/test/util/test_utils.cc
qti-mattsinc: Done

onnxruntime/test/util/test_utils.cc
qti-mattsinc: Done

onnxruntime/test/util/include/test_utils.h
huaychou: Acknowledge

onnxruntime/test/util/test_utils.cc
huaychou: It’s ok to check for a null vector (when all values in the vector are zero). One additional improvement would be to log a warning if a null vector is detected.

onnxruntime/test/providers/qnn/logical_compare_test.cc
```diff
@@ -176,11 +176,15 @@ TEST_F(QnnHTPBackendTests, EqualToCast4D) {
...
+  // RunQnnModelTest expected an arg for `num_nodes_in_ep`. This "1" has appeared to have survived
+  // changes to RunQnnModelTest and has now erroneously ended up as the fp32_abs_err arg, which means
+  // this test is probably not testing for the correct thing anymore.
+  // TODO: Fix this test, perhaps by using `EPVerificationParams::graph_verifier`?
   RunQnnModelTest(build_qdq_equal_to_cast,
                   provider_options,
                   17,  // opset
-                  ExpectedEPNodeAssignment::All,
-                  1);  // expected nodes in graph
+                  EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(1)});
```
huaychou, line 187: Remember to fix this as follow up

huaychou (approved):

qti-chuteng (approved):

PR #505, opened 2026-06-09
Add rank 5 to rank 4 Reshape - Transpose - Reshape folding

quic-muchhsu:
### Description

QNN HTP with rank-5 tensor around fc performs poorly likely due to some pattern matching for optimization on HTP. This fusion sidesteps that limitation by rewriting a 
`t0 -> Reshape(R1) -> t1 -> Transpose(T1, perm) -> t2 -> Reshape(R2) -> t3`
so that the Transpose (and the tensors it touches) are rank-4 instead of rank-5, while preserving the end-to-end semantics.
with the preconditions:
  - rank(t1) == rank(t2) == 5
  - Both Reshape shape inputs are constants
  - There exists an adjacent position p in the rank-5 perm where perm[p+1] == perm[p] + 1 — meaning two adjacent input dims are moved together as a contiguous block by the Transpose

### Motivation and Context

We have observed in mobilesam encoder that fc layer with 5d tensor around it failed to match optimized pattern in HTP that cause attention layer is 3x slower than 4d tensor ones. 
QAIRT job 48.6 ms : https://workbench.aihub.qualcomm.com/jobs/jpyvjkw7p/ 
EP job 112.5 ms: https://workbench.aihub.qualcomm.com/jobs/jp1qw9z2g/
Job with this PR 68.3 ms: https://dev.aihub.qualcomm.com/jobs/jgdo1j36p/

There are other 5d tensor exists in the graph so feature branch job is still slower, but when compare the affected attention block, its QAIRT [37M cycles] vs EP [103M cycles] vs PR [36M cycles]. Saying this PR does fix one of the issue about 5d tensor.


qti-yuduo:
If we know the exact pattern we're looking for, can we match larger exact pattern?

quic-muchhsu:
> If we know the exact pattern we're looking for, can we match larger exact pattern?

We don't know the larger exact pattern, only some insights about 5d tensor shouldn't hanging around attention or FCs, and it starts from 3 tensors above LN. The wording is vague, because I don't have critical evidence on what's the pattern, its just some insights.

But for what happened around MLP (WIP PR), we does know the exact pattern for it.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_transpose_rank5_to_rank4.cc
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  // Create Reshape1 output tensor wrapper (rank-4).
+  TensorInfo reshape1_output_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(reshape1_output, reshape1_output_info));
+
+  QnnTensorWrapper reshape1_output_tensor_wrapper(reshape1_output.name,
+                                                  QNN_TENSOR_TYPE_NATIVE,
+                                                  reshape1_output_info.qnn_data_type,
+                                                  std::move(reshape1_output_info.quant_param),
```
qti-yuduo, line 328: per-axis QDQ quant params are unsafe with rank collapse. reshape1_output and transpose_output are rebuilt as rank-4 tensors but reuse quant params initialized from rank-5 metadata. QnnQuantParamsWrapper stores per-channel axes directly, and this fusion does not remap/reject axes. If the quant axis is after the merged axes it must shift; if it is one of the merged axes it may be unrepresentable. Simplest robust fix: reject per-channel/per-axis QDQ for this fusion, or add a dedicated collapse-axis remap helper and tests.

onnxruntime/test/providers/qnn/qnn_node_group/reshape_transpose_rank5_to_rank4_test.cc
```diff
@@ -0,0 +1,202 @@
...
+                      /*reshape2_shape=*/{1, 12, 12, 8}),
+                  provider_options,
+                  /*opset_version=*/13,
+                  /*expected_ep_assignment=*/ExpectedEPNodeAssignment::All,
+                  /*fp32_abs_err=*/1e-2f);
+
+  // After fusion the Transpose runs on rank-4 tensors. The two Reshapes survive (one before and
+  // one after the Transpose). The fusion does not change op counts in the QNN graph, only ranks.
+  AssertOpInQnnGraph(json_qnn_graph_dir, "Transpose", 1);
+  AssertOpInQnnGraph(json_qnn_graph_dir, "Reshape", 2);
```
qti-yuduo, line 89: this doesn't guaranteed we got rank-4. Better to have stronger checks.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_transpose_rank5_to_rank4.cc
```diff
@@ -0,0 +1,389 @@
...
+  if (merge_input_idx_b != merge_input_idx_a + 1) {
+    return Ort::Status("Merge indices must be consecutive in input space", OrtErrorCode::ORT_FAIL);
+  }
+
+  // Build the rank-4 t1 shape by merging t1_rank5[merge_input_idx_a] and t1_rank5[merge_input_idx_b].
+  std::vector<uint32_t> t1_rank4_dims;
+  t1_rank4_dims.reserve(kRank4);
+  for (size_t i = 0; i < t1_rank5_dims.size(); ++i) {
+    if (static_cast<int64_t>(i) == merge_input_idx_a) {
+      t1_rank4_dims.push_back(t1_rank5_dims[i] * t1_rank5_dims[i + 1]);
```
qti-yuduo: Likely overflow.. use SaftInt?

qti-chuteng:
## [M-1] `reshape_transpose_rank5_to_rank4.cc` L1

New files in this PR (`reshape_transpose_rank5_to_rank4.cc` and `reshape_transpose_rank5_to_rank4.h`) use the Microsoft copyright header, but per team policy [B1][C10], **all newly added files must use the Qualcomm copyright header**.

**Suggested fix**: Replace the header in both new files with:
```
```

---

## [M-2] Missing negative test for perm with no adjacent consecutive pair

`reshape_transpose_rank5_to_rank4_test.cc` — file level

The `FindAdjacentMergeIndex` rejection path (returns `std::nullopt` when no adjacent consecutive pair exists in perm) is not covered by any test. A regression that causes it to incorrectly find a pair would fire the fusion on unsupported perm patterns with no test to catch it.

**Suggested fix**: Add a negative test with rank-5 intermediate tensors and a perm with no adjacent pair (e.g., `perm=[0,2,4,1,3]`), asserting `ExpectedEPNodeAssignment::All` still passes (ops fall through to standalone builders):

```cpp
TEST_F(QnnHTPBackendTests, Rank5ToRank4Fusion_Float_NoPairInPerm_NoFusion) {
  SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V68);
  RunQnnModelTest(BuildRank5ToRank4FloatTestCase(
      /*input_shape=*/{6, 4, 5, 6},
      /*reshape1_shape=*/{2, 3, 4, 5, 6},
      /*perm=*/{0, 2, 4, 1, 3},
      /*reshape2_shape=*/{6, 4, 5, 6}),
      GetProviderOptions(),
      /*opset_version=*/13,
      /*expected_ep_assignment=*/ExpectedEPNodeAssignment::All,
      /*fp32_abs_err=*/1e-2f);
}
```

---

## [N-1] `reshape_transpose_rank5_to_rank4.cc` L64

`GetTensorShape(const OrtApi&, const OrtValueInfo*)` is duplicated verbatim between `reshape_transpose_rank5.cc` (line 40) and this new file (line 64). Consider moving it to `utils.h`/`utils.cc`, or at minimum add a `// TODO: consolidate with GetTensorShape in reshape_transpose_rank5.cc` comment.

---

## [N-2] `reshape_transpose_rank5_to_rank4.cc` L236

The check `if (merge_input_idx_b != merge_input_idx_a + 1)` at line 236 can never fire: `FindAdjacentMergeIndex` already guarantees `perm[p+1] == perm[p] + 1`, making this invariant a pre-condition. The error message is misleading for maintainers. Consider replacing with a `DCHECK`/`ORT_ENFORCE` and a comment explaining the invariant.

---

qti-shubham (approved):

PR #338, opened 2026-05-05
Add SimplifiedLayerNormalization to QDQ node group selector

simejanko:
### Description
Add `SimplifiedLayerNormalization` to the RMSNorm QDQ node group selector registration.

The QNN EP's op builder factory and op type mapping already support `SimplifiedLayerNormalization` (mapping it to QNN RMSNorm op), but the QDQ node group selector only registered `RMSNormalization` ONNX op.

### Motivation and Context
`DQ` -> `SimplifiedLayerNormalization` -> `Q` patterns were never grouped into QDQ NodeUnits, and the op was forced to execute in float with explicit Dequantize/Quantize conversion nodes. This caused significant slowdowns on a quanitized model we were testing on HTP backend, affecting not only RMSNorm operations but neighboring operations as well (possibly suboptimal fusion).

PR #506, opened 2026-06-09
[CI] Add fork-PR stub jobs to unblock required status checks

qti-yuduo:
## Problem

Fork PRs (e.g. #338) are permanently blocked even when all builds pass. The root cause:

- `test` and `test-wheel` jobs in `qualcomm-internal-build-and-test-single-os.yml` require Artifactory secrets for the cross-runner hand-off, so they are gated with `github.event.pull_request.head.repo.fork != true` and produce a **`skipped`** conclusion on fork PRs.
- GitHub branch protection does **not** accept `skipped` as satisfying a required status check. The check context is simply never reported as `success`, so the PR is permanently blocked.

## Fix

Add two complementary stub jobs — `test-fork-stub` and `test-wheel-fork-stub` — with:

- The **same templated `name:`** as the real `test` / `test-wheel` jobs (so branch protection sees the exact check context it requires).
- A **flipped `if` condition** (`fork == true`) so they run only when the real jobs don't.
- `needs: [build]` so they run after the build succeeds.
- A single `echo` step — no secrets needed.

Non-fork PRs are completely unaffected: the real test jobs still run, the stubs are skipped.

## To unblock #338

Once this lands on `main`, rebase #338 onto `main` so CI re-runs with the new workflow logic.

### DIFF
--- a/.github/workflows/qualcomm-internal-qdc-test.yml
+++ b/.github/workflows/qualcomm-internal-qdc-test.yml
@@ -58,7 +58,7 @@ jobs:
     runs-on: ["self-hosted", "linux", "X64", "Build"]
     timeout-minutes: 10
     steps:
-      - run: echo "QDC tests skipped on fork PR (no Artifactory/QDC access); build validated the change."
+      - run: echo "QDC device tests skipped on fork PRs (need hardware + Artifactory secrets); change is still compiled, full device suite runs post-merge on main."
         shell: bash
 
   test:
--- a/.github/workflows/qualcomm-internal-qdc-test.yml
+++ b/.github/workflows/qualcomm-internal-qdc-test.yml
@@ -50,9 +50,20 @@ env:
   BUILD_ARTIFACTORY_REPO: "${{ secrets.BUILD_ARTIFACTORY_REPO }}"
 
 jobs:
+  # Fork PRs cannot access Artifactory secrets or the QDC API token, so the real test job is skipped for them.
+  # This stub carries the same `name:` and runs only on fork PRs so branch protection sees a green check.
+  test-fork-stub:
+    name: "test-${{inputs.target_os}}-${{inputs.target_arch}}-pynone"
+    if: ${{ github.event.pull_request.head.repo.fork == true }}
+    runs-on: ["self-hosted", "linux", "X64", "Build"]
+    timeout-minutes: 10
+    steps:
+      - run: echo "QDC tests skipped on fork PR (no Artifactory/QDC access); build validated the change."
+        shell: bash
+
   test:
     name: "test-${{inputs.target_os}}-${{inputs.target_arch}}-pynone"
-    # QDC tests pull artifacts from Artifactory; skip on fork PRs where secrets are unavailable.
+    # QDC tests pull artifacts from Artifactory; fork stub above handles the fork case.
     if: ${{ github.event.pull_request.head.repo.fork != true }}
     runs-on: ["self-hosted", "Linux", "QDC"]
     defaults:
--- a/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
+++ b/.github/workflows/qualcomm-internal-build-and-test-single-os.yml
@@ -268,6 +268,28 @@ jobs:
           name: "${{ inputs.target_os }}-${{ inputs.target_arch }}-py${{ inputs.target_py_vsn }}-pdb-zip"
           src_pattern: "${{inputs.target_os}}-${{ inputs.target_arch }}/${{ env.BUILD_CONFIG_DIR }}/dist/onnxruntime-qnn-*-win-*-pdb.zip"
 
+  # Fork PRs cannot access Artifactory secrets, so the real test/test-wheel jobs are skipped for them.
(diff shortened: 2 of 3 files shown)


qti-mbadnara (approved):
LGTM!

.github/workflows/qualcomm-internal-qdc-test.yml
```diff
@@ -42,9 +42,19 @@ env:
...
+    runs-on: ["self-hosted", "linux", "X64", "Build"]
+    steps:
+      - run: echo "QDC tests skipped on fork PR (no Artifactory/QDC access); build validated the change."
+        shell: bash
+
   test:
     name: "test-${{inputs.target_os}}-${{inputs.target_arch}}-pynone"
-    # QDC tests pull artifacts from Artifactory; skip on fork PRs where secrets are unavailable.
+    # QDC tests pull artifacts from Artifactory; fork stub above handles the fork case.
     if: ${{ github.event.pull_request.head.repo.fork != true }}
```
qti-mbadnara, line 67: Should this be changed?

qti-chuteng:
### 🟠 Major

**[M-1]** `.github/workflows/qualcomm-internal-build-and-test-single-os.yml:L267`
The second commit's message states "Add timeout-minutes: 10 to both stubs in single-os workflow" but neither `test-fork-stub` nor `test-wheel-fork-stub` has a `timeout-minutes` field in the actual diff. The stub in `qualcomm-internal-qdc-test.yml` is also missing it. Without this bound, a saturated runner pool can block a fork PR for up to 6 hours — the exact risk the commit message acknowledges but fails to address.

```suggestion
  test-fork-stub:
    name: "test-${{inputs.target_os}}-${{inputs.target_arch}}-py${{ inputs.target_py_vsn != 'None' && inputs.target_py_vsn || 'none'}}"
    if: ${{ inputs.run_tests && !inputs.build_test_same_runner && github.event.pull_request.head.repo.fork == true }}
    needs: [build]
    runs-on: ["self-hosted", "linux", "X64", "Build"]
    timeout-minutes: 10
    steps:
      - run: echo "Tests skipped on fork PR (no Artifactory access); build validated the change."
        shell: bash
```

.github/workflows/qualcomm-internal-qdc-test.yml
qti-yuduo: This line is the real `test` job, which must stay gated on `fork != true` because it needs the Artifactory/QDC secrets a fork PR can't access. The new `test-fork-stub` job right above (flipped condition `fork == true`, same `name:`) is what satisfies the required check on fork PRs. Changing this guard would make the real secret-dependent job attempt to run on forks and fail.

qti-yuduo:
**[M-1]** Fixed. `timeout-minutes: 10` is now on all three stubs — `test-fork-stub` and `test-wheel-fork-stub` in `qualcomm-internal-build-and-test-single-os.yml`, and `test-fork-stub` in `qualcomm-internal-qdc-test.yml`. The bound was genuinely absent before despite the commit message; it's folded into the `Fix fork-PR stub coverage gaps and bound stub timeouts` commit (ad2bdf6) so the message now matches the diff.

Also rebased onto latest `main` to resolve the conflict: the two caller jobs in `build-and-test.yml` keep main's `path-filter` dependency and `has_code_changes` input while dropping the fork guard, and the qdc-test stub coexists cleanly with main's `has_code_changes` step gating.

.github/workflows/qualcomm-internal-qdc-test.yml
```diff
@@ -50,9 +50,20 @@ env:
...
 jobs:
+  # Fork PRs cannot access Artifactory secrets or the QDC API token, so the real test job is skipped for them.
+  # This stub carries the same `name:` and runs only on fork PRs so branch protection sees a green check.
+  test-fork-stub:
+    name: "test-${{inputs.target_os}}-${{inputs.target_arch}}-pynone"
+    if: ${{ github.event.pull_request.head.repo.fork == true }}
+    runs-on: ["self-hosted", "linux", "X64", "Build"]
+    timeout-minutes: 10
+    steps:
+      - run: echo "QDC tests skipped on fork PR (no Artifactory/QDC access); build validated the change."
```
qti-kromero: what does it mean here and in other places by "build validated the change"? it seems like it's trying to say the build ran this testing but I don't believe it did? it is slightly concerning that fork PRs appear to get less testing than non-fork PRs but unsure if there is a way to resolve that

qti-yuduo: Wording is too strong here, i'll amend. The build job does run on forks (only Artifactory uploads no-op, gated on BUILD_ARTIFACTORY_REPO != ''), and it compiles test binaries — but it does not run the tests these three checks require QDC or ARM64 runners. We only get full test coverage from post-merge CIs. This change won't bring full coverage back just like non-fork initiated PR.

qti-yuduo: The constraint is: GitHub never gives repo secrets to a pull_request run from a fork — by design. 

I'll see if we can workaround this as follow up.

qti-kromero (approved):

PR #509, opened 2026-06-10
[QNN EP] Add support to build Windows ARM64EC, x64 Zip archives as part of CI

qti-mbadnara:
This PR adds support to build the Zip archives for Windows ARM64EC and Windows x86_64 variants as part of the release pipeline. This allows us to ship these artifacts as part of external release.


huaychou:
Why do we need to release arm64ec and x64 zip separately? Previously, we would include them all in a single delivery, such as onnxruntime_qnn-2.2.0-cp314-cp314-win_amd64.whl, which has both x64 and arm64ec libraries.

qti-mbadnara:
> Why do we need to release arm64ec and x64 zip separately? Previously, we would include them all in a single delivery, such as onnxruntime_qnn-2.2.0-cp314-cp314-win_amd64.whl, which has both x64 and arm64ec libraries.

I believe for future releases we might also ship the Zip files across different Windows architectures to maintain parity with ORT

.github/workflows/qualcomm-internal-release.yml
```diff
@@ -331,6 +330,16 @@ jobs:
         with:
           name: windows-arm64-py3.11-zip
 
+      - name: Download Windows AR64EC Zip Package from Artifactory
```
huaychou:
```suggestion
      - name: Download Windows ARM64EC Zip Package from Artifactory
```

.github/workflows/qualcomm-internal-upload-artifactory-from-github.yml
```diff
@@ -470,29 +493,29 @@ jobs:
...
 
-          # Upload Windows ARM64EC PDB archive
+          # Upload Linux Aarch64 manylinux_2_34 zip
           ./qcom/scripts/upleveling/upload_zip_to_artifactory.sh \
-            "${{ github.workspace }}/build/windows-arm64ec/Release/Release/dist" \
+            "${{ github.workspace }}/build/linux-aarch64_manylinux_2_34/Release/dist" \
             "$NETRC_FILE"
 
-          # Upload Windows x86_64 PDB archive
+          # Upload Linux x86_6422.04 Ubuntu 22.04 zip
```
huaychou:
```suggestion
          # Upload Linux x86_64 Ubuntu 22.04 zip
```

kuanyul-qti (approved):
The pdb string should be removed from the title since the pdb archives are already built, and this PR is only adding additional zip archive builds.

huaychou (approved):
LGTM. Remember to update the docs once Arm64ec/x64 zips are available.

qti-chuteng (approved):

qti-mbadnara:
> LGTM. Remember to update the docs once Arm64ec/x64 zips are available.

Yes will be done along with 2.4.0 docs!

PR #510, opened 2026-06-10
[OpBuilder] Add Shape op builder for QNN EP

kuanyul-qti:
Add Shape op builder for QNN EP.
Downcasts the int64 output to int32 (same mechanism as ArgMax/ArgMin) and forwards the start/end attributes to the QNN start/end scalar params

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
@@ -13,7 +13,7 @@ namespace qnn {
 
 // Shape op builder.
 // Maps ONNX Shape -> QNN Shape. ONNX Shape produces an int64 1-D tensor; QNN requires the
-// output to be uint32/int32, so the int64 output is downcast to int32 (BaseOpBuilder::ProcessOutputs
+// output to be int32, so the int64 output is downcast to int32 (BaseOpBuilder::ProcessOutputs
 // inserts a Cast back to int64 when the output is a graph output). For intermediate outputs,
 // the QNN tensor type remains INT_32 and downstream ops consume it directly.
 // The ONNX `start`/`end` attributes are mapped to the QNN `start`/`end` scalar params (uint32).
@@ -34,14 +34,12 @@ class ShapeOpBuilder : public BaseOpBuilder {
 };
 
 Qnn_DataType_t ShapeOpBuilder::GetSupportedOutputDataType(size_t index, Qnn_DataType_t qnn_data_type) const {
-  // ONNX Shape has an int64 output, but QNN requires uint32 or int32.
-  // If this node produces a graph output, BaseOpBuilder::ProcessOutputs() adds a Cast node after the Shape op.
-  // Otherwise, it just sets the output type to uint32 or int32.
+  // ONNX Shape always produces an int64 output (no unsigned variant per the ONNX spec), but QNN
+  // requires int32. If this node produces a graph output, BaseOpBuilder::ProcessOutputs() adds a
+  // Cast node after the Shape op. Otherwise, it just sets the output type to int32.
   ORT_UNUSED_PARAMETER(index);
   if (qnn_data_type == QNN_DATATYPE_INT_64) {
     return QNN_DATATYPE_INT_32;
-  } else if (qnn_data_type == QNN_DATATYPE_UINT_64) {
-    return QNN_DATATYPE_UINT_32;
   }
 
   return qnn_data_type;
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
@@ -58,8 +58,9 @@ Ort::Status ShapeOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mod
   const int64_t rank = static_cast<int64_t>(input_shape.size());
   RETURN_IF(rank < 1, "QNN Shape requires an input of rank >= 1.");
 
-  // ONNX `start`/`end` attributes (opset >= 15). Negative values index from the end and are
-  // clamped to [0, rank], matching the ONNX Shape spec.
+  // Step 1: Resolve `start`/`end` per the ONNX Shape spec (opset >= 15). The reference op is
+  // `data.shape[start:end]`: negative values count from the end (add rank), then both are clamped
+  // to [0, rank]. The resolved output length is max(0, end - start) (start >= end => empty shape).
   OrtNodeAttrHelper node_helper(node_unit);
   int64_t start = node_helper.Get("start", static_cast<int64_t>(0));
   int64_t end = node_helper.Get("end", rank);
@@ -72,13 +73,19 @@ Ort::Status ShapeOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mod
   }
   start = std::min<int64_t>(std::max<int64_t>(start, 0), rank);
   end = std::min<int64_t>(std::max<int64_t>(end, 0), rank);
+  const int64_t output_length = std::max<int64_t>(0, end - start);
 
-  // QNN constraints: start must be in [0, rank] and end must be in [start+1, rank]
-  // (start == rank is valid for an empty slice at the tail end).
-  RETURN_IF_NOT(start >= 0 && start <= rank,
-                "QNN Shape requires the resolved `start` axis to be in range [0, rank].");
-  RETURN_IF_NOT(end >= start + 1 && end <= rank,
-                "QNN Shape requires the resolved `end` axis to be in range [start+1, rank].");
+  // Step 2: Postprocess to match the QNN op definition. Per QnnOpDef (MasterOpDef "Shape"), the
+  // output is a 1-D tensor of shape [M] with M = end - start, and QNN constrains `end` to
+  // [start + 1, N] -- i.e. M >= 1. QNN cannot represent a zero-length output, so the ONNX-valid
+  // empty-slice case (output_length == 0) has no QNN equivalent. Clamping `end` up to `start + 1`
+  // would silently produce a length-1 output and corrupt results, so instead we reject the op here.
+  // Returning an error from IsOpSupported() leaves the node unassigned, and it falls back to CPU EP.
+  RETURN_IF(output_length < 1,
+            "QNN Shape produces a 1-D output of length (end - start) and requires end >= start + 1; "
+            "the ONNX empty-slice case (start >= end) is not supported by QNN and falls back to CPU.");
+  // With output_length >= 1 we have end >= start + 1 and end <= rank, which also implies
+  // start in [0, rank - 1] (QNN's `start` constraint), so no separate start-range check is needed.
 
   std::vector<std::string> param_tensor_names;
 
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
@@ -1,0 +1,112 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include <algorithm>
+
+#include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/op_builder_factory.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+// Shape op builder.
+// Maps ONNX Shape -> QNN Shape. ONNX Shape produces an int64 1-D tensor; QNN requires the
+// output to be uint32/int32, so the int64 output is downcast to int32 (BaseOpBuilder::ProcessOutputs
+// inserts a Cast back to int64 when the output is a graph output).
+// The ONNX `start`/`end` attributes are mapped to the QNN `start`/`end` scalar params (uint32).
+class ShapeOpBuilder : public BaseOpBuilder {
+ public:
+  ShapeOpBuilder() : BaseOpBuilder("ShapeOpBuilder") {}
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(ShapeOpBuilder);
+
+ protected:
+  Qnn_DataType_t GetSupportedOutputDataType(size_t index,
+                                            Qnn_DataType_t qnn_data_type) const override ORT_MUST_USE_RESULT;
+
+  Ort::Status ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                          const OrtNodeUnit& node_unit,
+                                          std::vector<std::string>&& input_names,
+                                          const Ort::Logger& logger,
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+};
+
+Qnn_DataType_t ShapeOpBuilder::GetSupportedOutputDataType(size_t index, Qnn_DataType_t qnn_data_type) const {
+  // ONNX Shape has an int64 output, but QNN requires uint32 or int32.
+  // If this node produces a graph output, BaseOpBuilder::ProcessOutputs() adds a Cast node after the Shape op.
+  // Otherwise, it just sets the output type to uint32 or int32.
+  ORT_UNUSED_PARAMETER(index);
+  if (qnn_data_type == QNN_DATATYPE_INT_64) {
+    return QNN_DATATYPE_INT_32;
+  } else if (qnn_data_type == QNN_DATATYPE_UINT_64) {
+    return QNN_DATATYPE_UINT_32;
+  }
+
+  return qnn_data_type;
+}
+
+Ort::Status ShapeOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                                        const OrtNodeUnit& node_unit,
+                                                        std::vector<std::string>&& input_names,
+                                                        const Ort::Logger& logger,
+                                                        bool do_op_validation) const {
+  const auto& inputs = node_unit.Inputs();
+  std::vector<uint32_t> input_shape;
+  RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[0].shape, input_shape),
+                "QNN EP: Cannot get input shape for Shape op.");
+  const int64_t rank = static_cast<int64_t>(input_shape.size());
+  RETURN_IF(rank < 1, "QNN Shape requires an input of rank >= 1.");
+
+  // ONNX `start`/`end` attributes (opset >= 15). Negative values index from the end and are
+  // clamped to [0, rank], matching the ONNX Shape spec.
+  OrtNodeAttrHelper node_helper(node_unit);
+  int64_t start = node_helper.Get("start", static_cast<int64_t>(0));
+  int64_t end = node_helper.Get("end", rank);
+
+  if (start < 0) {
+    start += rank;
+  }
+  if (end < 0) {
+    end += rank;
+  }
+  start = std::min<int64_t>(std::max<int64_t>(start, 0), rank);
+  end = std::min<int64_t>(std::max<int64_t>(end, 0), rank);
+
+  // QNN constraints: start must be in [0, rank] and end must be in [start+1, rank]
+  // (start == rank is valid for an empty slice at the tail end).
+  RETURN_IF_NOT(start >= 0 && start <= rank,
+                "QNN Shape requires the resolved `start` axis to be in range [0, rank].");
+  RETURN_IF_NOT(end >= start + 1 && end <= rank,
+                "QNN Shape requires the resolved `end` axis to be in range [start+1, rank].");
+
+  std::vector<std::string> param_tensor_names;
+
+  Qnn_Scalar_t start_qnn_scalar = QNN_SCALAR_INIT;
+  start_qnn_scalar.dataType = QNN_DATATYPE_UINT_32;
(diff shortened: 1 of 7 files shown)


qti-chuteng:
### 🟠 Major

**[M-1]** `shape_op_builder.cc → ProcessAttributesAndOutputs()`

The `RETURN_IF_NOT(end >= start + 1 ...)` check is stricter than the ONNX Shape spec. Per ONNX `defs/tensor/defs.cc` line 480: `output_length = max(0, end - start)`, so `start == end` is a valid empty-slice (output length 0). This check rejects models with `start == end` and contradicts the code's own comment that "start == rank is valid for an empty slice at the tail end."

```suggestion
  // QNN allows end >= start (empty slice when end == start, output length 0)
  RETURN_IF_NOT(end >= start && end <= rank,
                "QNN Shape requires the resolved end axis to be in range [start, rank].");
```

---

**[M-2]** `shape_op_builder.cc → ProcessAttributesAndOutputs()`

After the `std::min/max` clamping, both `start` and `end` are already guaranteed to be in `[0, rank]`, so the subsequent `RETURN_IF_NOT(start >= 0 && start <= rank, ...)` is unreachable dead code. Remove it; the only check that needs to survive (after the M-1 fix) is `end >= start`.

---

### 🔵 Minor

**[N-1]** `shape_op_test.cc`

The empty-slice case (`start == end`) has no test coverage. After fixing M-1, add a CPU backend test to provide a kill-test fence:

```cpp
TEST_F(QnnCPUBackendTests, Shape_EmptySlice_Float) {
  RunShapeOpTest(TestInputDef<float>({2, 3, 4, 5}, false, -10.0f, 10.0f),
                 {test::MakeAttribute("start", static_cast<int64_t>(2)),
                  test::MakeAttribute("end",   static_cast<int64_t>(2))},
                 ExpectedEPNodeAssignment::All, "cpu", 15);
}
```

kuanyul-qti:
> ### 🟠 Major
> **[M-1]** `shape_op_builder.cc → ProcessAttributesAndOutputs()`
> 
> The `RETURN_IF_NOT(end >= start + 1 ...)` check is stricter than the ONNX Shape spec. Per ONNX `defs/tensor/defs.cc` line 480: `output_length = max(0, end - start)`, so `start == end` is a valid empty-slice (output length 0). This check rejects models with `start == end` and contradicts the code's own comment that "start == rank is valid for an empty slice at the tail end."
> 
> ```
>   // QNN allows end >= start (empty slice when end == start, output length 0)
>   RETURN_IF_NOT(end >= start && end <= rank,
>                 "QNN Shape requires the resolved end axis to be in range [start, rank].");
> ```
> 
> **[M-2]** `shape_op_builder.cc → ProcessAttributesAndOutputs()`
> 
> After the `std::min/max` clamping, both `start` and `end` are already guaranteed to be in `[0, rank]`, so the subsequent `RETURN_IF_NOT(start >= 0 && start <= rank, ...)` is unreachable dead code. Remove it; the only check that needs to survive (after the M-1 fix) is `end >= start`.
> 
> ### 🔵 Minor
> **[N-1]** `shape_op_test.cc`
> 
> The empty-slice case (`start == end`) has no test coverage. After fixing M-1, add a CPU backend test to provide a kill-test fence:
> 
> ```c++
> TEST_F(QnnCPUBackendTests, Shape_EmptySlice_Float) {
>   RunShapeOpTest(TestInputDef<float>({2, 3, 4, 5}, false, -10.0f, 10.0f),
>                  {test::MakeAttribute("start", static_cast<int64_t>(2)),
>                   test::MakeAttribute("end",   static_cast<int64_t>(2))},
>                  ExpectedEPNodeAssignment::All, "cpu", 15);
> }
> ```

[M-1]: Split the logic into two explicit phases:
Step 1. ONNX resolution. Verified against the authoritative ONNX spec.
Step 2. QNN postprocess. Verified against Op Def from QAIRT SDK.
[M-2]: Removed.
[N-1]: Added a new UT.

qti-chuteng:
### 🔵 Minor

**[N-2]** `onnxruntime/test/providers/qnn/shape_op_test.cc` (file naming)

Per [T19], QNN EP test files should be named `{op_lowercase}_test.cc` (e.g., `selu_test.cc`, `gemm_test.cc`). The file name `shape_op_test.cc` has an extra `_op` segment inconsistent with the majority convention. Rename to `shape_test.cc`.

---

**[N-3]** `shape_op_test.cc` — HTP tests lack quantized input coverage

Both `Shape_Default_Float_HTP` and `Shape_StartEnd_Float_HTP` use non-quantized FP32 input. Per [C06], new op additions should include HTP u8/u16 QDQ tests. Shape op's output is int64 (not quantized), but in real HTP inference the input is typically quantized. Consider adding a test with u8 quantized input (e.g., `Shape_Default_U8_HTP`) or documenting why FP32-only coverage is sufficient for HTP.

---

**[N-4]** `shape_op_builder.cc:L15`

The class comment only mentions Cast-back for graph outputs, but omits the intermediate output case. When the Shape op's output is a non-graph-output tensor, it remains INT_32 and downstream ops consume it directly (no Cast inserted). Add a sentence: `// For intermediate outputs, the QNN tensor type remains INT_32 and downstream ops consume it directly.`

kuanyul-qti:
> ### 🔵 Minor
> **[N-2]** `onnxruntime/test/providers/qnn/shape_op_test.cc` (file naming)
> 
> Per [T19], QNN EP test files should be named `{op_lowercase}_test.cc` (e.g., `selu_test.cc`, `gemm_test.cc`). The file name `shape_op_test.cc` has an extra `_op` segment inconsistent with the majority convention. Rename to `shape_test.cc`.
> 
> **[N-3]** `shape_op_test.cc` — HTP tests lack quantized input coverage
> 
> Both `Shape_Default_Float_HTP` and `Shape_StartEnd_Float_HTP` use non-quantized FP32 input. Per [C06], new op additions should include HTP u8/u16 QDQ tests. Shape op's output is int64 (not quantized), but in real HTP inference the input is typically quantized. Consider adding a test with u8 quantized input (e.g., `Shape_Default_U8_HTP`) or documenting why FP32-only coverage is sufficient for HTP.
> 
> **[N-4]** `shape_op_builder.cc:L15`
> 
> The class comment only mentions Cast-back for graph outputs, but omits the intermediate output case. When the Shape op's output is a non-graph-output tensor, it remains INT_32 and downstream ops consume it directly (no Cast inserted). Add a sentence: `// For intermediate outputs, the QNN tensor type remains INT_32 and downstream ops consume it directly.`


[N-2] Done.
[N-3] Added.
[N-4] Added.

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
```diff
@@ -0,0 +1,114 @@
...
+  Ort::Status ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                          const OrtNodeUnit& node_unit,
+                                          std::vector<std::string>&& input_names,
+                                          const Ort::Logger& logger,
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+};
+
+Qnn_DataType_t ShapeOpBuilder::GetSupportedOutputDataType(size_t index, Qnn_DataType_t qnn_data_type) const {
+  // ONNX Shape has an int64 output, but QNN requires uint32 or int32.
+  // If this node produces a graph output, BaseOpBuilder::ProcessOutputs() adds a Cast node after the Shape op.
```
huaychou: The logic of ProcessOutputs is weird after https://github.com/microsoft/onnxruntime/pull/25745. For example, if we return SupportedOutputDataType as INT64, it still inserts a cast node even it is not graph output.
Could you elaborate why we need to override GetSupportedOutputDataType and make ProcessOutputs easy to maintain in another PR?

onnxruntime/test/providers/qnn/shape_test.cc
```diff
@@ -0,0 +1,192 @@
...
+                                         {TestInputDef<float>({3, 4, 5}, false, -10.0f, 10.0f)},
+                                         {}, {}),
+                  provider_options,
+                  15,
+                  ExpectedEPNodeAssignment::All);
+}
+
+// Test that Shape with start=1 and end=3 works on QNN HTP backend (FP32 input).
+// Input: float32 [2, 3, 4, 5]. Expected output: int64 [3, 4].
+TEST_F(QnnHTPBackendTests, Shape_StartEnd_Float_HTP) {
```
huaychou, line 336: **[M-2] (Major)** Missing HTP rank > 4 rejection test (HTP Shape input rank limit is 4)

The HTP opdef supplement explicitly states: *"Shape: Max supported input rank is 4."* The analogous ArgMax/ArgMin builders include an `ArgMaxMinU8_RankGreaterThan4_Unsupported` test that verifies rank > 4 inputs on HTP fall back to CPU EP with `ExpectedEPNodeAssignment::None`. Shape's HTP tests only go up to rank 4 (`{2, 3, 4, 5}`); no rank-5 fallback test exists.

**Impact**: If a user applies Shape to a rank-5 tensor on HTP, QNN backend validation will fail with a hard error rather than cleanly falling back to CPU EP. The absence of a test means this behavior has never been verified.

**Suggested fix:**

Add a test inside the HTP section (`#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)`) of `shape_test.cc`:

```
TEST_F(QnnHTPBackendTests, Shape_RankGreaterThan4_Unsupported) {
  ProviderOptions provider_options;
  provider_options["backend_type"] = "htp";
  provider_options["offload_graph_io_quantization"] = "0";

  RunQnnModelTest(BuildOpTestCase<float>("shape_node", "Shape",
                                          {TestInputDef<float>({2,3,4,5,6}, false, -10.0f, 10.0f)},
                                          {}, {}),
                  provider_options,
                  15,
                  ExpectedEPNodeAssignment::None);
}
```

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
kuanyul-qti: The override is needed for the intermediate-output case: when Shape's output is consumed by a downstream QNN op (not a graph output), BaseOpBuilder::ProcessOutputs falls through to the else branch and sets the tensor type to whatever GetSupportedOutputDataType returns. Without the override that would leave the tensor as INT_64, which QNN Shape doesn't accept.
For the graph-output case, agreed the duplication makes ProcessOutputs harder to maintain. The cleanup affects all ops and risks regressions. I'd keep this PR scoped to Shape.

onnxruntime/test/providers/qnn/shape_test.cc
kuanyul-qti: Added

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
huaychou: Another question, if the consumer of shape op takes int64 input, but we pass int32 output for shape op, what happens to the consumer input dtype?

kuanyul-qti: The consumer doesn't end up with an INT_64 input either way — BaseOpBuilder::OnAddToBuilder invokes ProcessInt64Tensors on every op's inputs, which auto-inserts an INT_64 → INT_32 Cast on any INT_64 input. So:
- Without our override, Shape would produce an INT_64 tensor, and the downstream op's ProcessInt64Tensors would insert a Cast back down to INT_32 before consumption. (Plus QNN_OP_SHAPE itself doesn't accept INT_64 output, so this path fails op validation anyway.)
- With our override, Shape produces INT_32 directly; the downstream op's ProcessInt64Tensors is a no-op for that input.

Either way the consumer sees INT_32 — the override just avoids a redundant up/down Cast pair (and keeps QNN_OP_SHAPE valid).

huaychou: I think you mean BaseOpBuilder::AddToModelBuilder? In addition, the output dtype of shape should only be int64 according to ONNX document.

kuanyul-qti: Yes, it's AddToModelBuilder. Sorry for misleading.
Drop the uint64 branch.

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
```diff
@@ -0,0 +1,112 @@
...
+  RETURN_IF(rank < 1, "QNN Shape requires an input of rank >= 1.");
+
+  // Step 1: Resolve `start`/`end` per the ONNX Shape spec (opset >= 15). The reference op is
+  // `data.shape[start:end]`: negative values count from the end (add rank), then both are clamped
+  // to [0, rank]. The resolved output length is max(0, end - start) (start >= end => empty shape).
+  OrtNodeAttrHelper node_helper(node_unit);
+  int64_t start = node_helper.Get("start", static_cast<int64_t>(0));
+  int64_t end = node_helper.Get("end", rank);
+
+  if (start < 0) {
```
yath1: One of the main functionalities of this op is that the start/end can take multiple forms and they need to be pre-processed to provide absolute start/end indices to QNN. What do you think of making it a separate method, lets say  `resolve_shape_bounds`, move this logic there and separately unit test for all the scenarios instead of relying on op level tests like you have done in this PR? 

Below are the test scenarios claude came up with: 

Defaults / omitted
  1. start=None, end=None → (0, r) — full shape
  2. start=None, end=2 → (0, 2)
  3. start=1, end=None → (1, r)

  Basic positive
  4. start=1, end=3 → (1, 3)
  5. start=0, end=r → (0, r)

  Negative axes
  6. start=-1 → (r-1, …) counts back
  7. end=-1 → (…, r-1) excludes last
  8. start=-r → (0, …)
  9. both negative: start=-2, end=-1 → (r-2, r-1)

  Clamping high
  10. end > r → clamped to r
  11. start > r → clamped to r (empty)

  Clamping low
  12. start < -r → 0
  13. end < -r → 0

  Empty result (start > end)
  14. start=3, end=1 → (3, 1) empty
  15. start=2, end=2 → (2, 2) empty (equal)

  Edge ranks
  16. r=0 (scalar): start=None,end=None → (0, 0)
  17. r=1: start=0, end=1 → (0, 1)

  Boundary
  18. start=r, end=r → (r, r) empty
  19. start=0, end=0 → (0, 0) empty
  20. negative exactly at -r-1 (out of range) → clamped to 0

onnxruntime/test/providers/qnn/shape_test.cc
```diff
@@ -0,0 +1,207 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#if !defined(ORT_MINIMAL_BUILD)
```
yath1, line 4: Can you add following two tests? 
1. shape op's output is fed to a different op in the graph. For example: input -> shape -> gather -> output
3. shape op's output is both a graph output and is consumed by a different node?

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
kuanyul-qti: `ResolveShapeBounds` is extracted into a new header and 20 unit tests added under `QnnResolveShapeBoundsTest` in `shape_test.cc`.

onnxruntime/test/providers/qnn/shape_test.cc
kuanyul-qti: `Shape_To_Gather_HTP` and `Shape_DualUse_HTP` are added.

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
```diff
@@ -0,0 +1,111 @@
...
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/op_builder_factory.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+// Shape op builder.
+// Maps ONNX Shape -> QNN Shape. ONNX Shape produces an int64 1-D tensor; QNN requires the
+// output to be int32, so the int64 output is downcast to int32 (BaseOpBuilder::ProcessOutputs
+// inserts a Cast back to int64 when the output is a graph output). For intermediate outputs,
```
yath1: Minor doc correction: 
` QNN requires the output to be int32, so the int64 output is downcast to int32` 

can be changed to
`QNN generates output in int32. If the output is not an intermediate tensor in the graph, then a Cast op is inserted to convert to int64.`

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.h
```diff
@@ -0,0 +1,35 @@
...
+namespace onnxruntime {
+namespace qnn {
+
+// Resolves the ONNX Shape `start`/`end` attributes against the input rank per the ONNX spec
+// (opset >= 15). Mirrors `data.shape[start:end]`: negative values count from the end (add rank),
+// then both are clamped to [0, rank]. Returns the resolved (start, end, output_length) where
+// output_length = max(0, end - start).
+//
+// Defined inline in this header so it can be unit-tested directly from the test binary; the
+// op-builder also includes this header to ensure a single source of truth.
```
yath1: Minor: Please remove this comment. I think this is self-explanatory.

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.h
```diff
@@ -0,0 +1,35 @@
...
+inline void ResolveShapeBounds(int64_t rank, int64_t start_attr, int64_t end_attr,
```
yath1: Nit: In these cases, cleaner to compute using the input vars and return them instead. 

For example, above can be changed to 

`std::pair<int64_t, int64_t> ResolveShapeBounds(int64_t rank, int64_t start_attr, int64_t end_attr)`

`output_length` can be moved out of the method. Its not doing what the func name `ResolveShapeBounds` conveys.

onnxruntime/test/providers/qnn/shape_test.cc
```diff
@@ -0,0 +1,470 @@
...
+  std::vector<Ort::Value> actual;
+  InferenceModel(model_data, "Shape_Composition_QNN", provider_options, expected_ep_assignment,
+                 helper.feeds_, actual);
+
+  ASSERT_EQ(expected.size(), actual.size());
+  for (size_t out_idx = 0; out_idx < expected.size(); ++out_idx) {
+    auto exp_shape = expected[out_idx].GetTensorTypeAndShapeInfo().GetShape();
+    auto act_shape = actual[out_idx].GetTensorTypeAndShapeInfo().GetShape();
+    ASSERT_EQ(exp_shape, act_shape) << "Shape mismatch for output " << out_idx;
+    const size_t element_count = expected[out_idx].GetTensorTypeAndShapeInfo().GetElementCount();
```
yath1: Minor: Check for expected output type as well

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
```diff
@@ -0,0 +1,111 @@
...
+                                         param_tensor_names));
+
+  RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(),
+                                         static_cast<uint32_t>(end), QNN_OP_SHAPE_PARAM_END,
+                                         param_tensor_names));
+
+  RETURN_IF_ERROR(ProcessOutputs(qnn_model_wrapper, node_unit,
+                                 std::move(input_names),
+                                 std::move(param_tensor_names),
+                                 logger, do_op_validation, QNN_OP_SHAPE));
```
yath1: Replace hardcoded op-type to `GetQnnOpType(node_unit.OpType())` instead

yath1 (approved):
Thanks for adding tests. Lgtm other than the comments below

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.h
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.h
kuanyul-qti: Done.

onnxruntime/test/providers/qnn/shape_test.cc
kuanyul-qti: Done.

huaychou:
Do we need to add Shape op into drop_dq_ops list?

onnxruntime/core/providers/qnn/builder/opbuilder/shape_op_builder.h
```diff
@@ -0,0 +1,31 @@
...
+#include <cstdint>
+#include <utility>
+
+namespace onnxruntime {
+namespace qnn {
+
+// Resolves the ONNX Shape `start`/`end` attributes against the input rank per the ONNX spec
+// (opset >= 15). Mirrors `data.shape[start:end]`: negative values count from the end (add rank),
+// then both are clamped to [0, rank]. Returns the resolved (start, end) pair.
+inline std::pair<int64_t, int64_t> ResolveShapeBounds(int64_t rank, int64_t start_attr, int64_t end_attr) {
```
minfhong-qti, line 16: Define in .cc's anonymous namespace.

PR #438, opened 2026-05-25
Load target backend validator for backendValidateOpConfig when dumping DLC

yuhuchua-qti:
## Summary

  Re-implements PR #218 (reverted in #413) with the acknowledged gap addressed.

  **What changed:**
  - When `dump_qnn_ir_dlc=1` and `backend_path` is set to a non-Ir backend (HTP/CPU/GPU), `backendValidateOpConfig` is now routed to the user-specified target backend instead of QnnIr.
  - Adds `CreateValidatorDevice()` that passes `soc_model`/`htp_arch` to `qnn_validator_interface_.deviceCreate`, so the HTP shim receives the correct arch context for validation. Previously these configs only reached
  `qnn_interface_.deviceCreate` (=QnnIr in serializer flow).
  - Refactors `CreateDevice`/`ReleaseDevice` into `CreateDeviceCommon`/`ReleaseDeviceCommon`, consistent with the existing `InitializeBackendCommon`/`ShutdownBackendCommon`/`InitializeQnnLogCommon` pattern.

  **Behavior before this PR:**
  | `backend_path` | DLC dump | Validator used |
  |---|---|---|
  | `QnnIr.dll` | ✓ | QnnIr |
  | `QnnHtp.dll` | ✓ | QnnIr ⚠️  (target-agnostic, no HTP-specific checks) |
  | `QnnCpu.dll` | ✓ | QnnIr ⚠️  |
  | `QnnGpu.dll` | ✓ | QnnIr ⚠️  |

  **Behavior after this PR:**
  | `backend_path` | DLC dump | Validator used |
  |---|---|---|
  | `QnnIr.dll` | ✓ | QnnIr (unchanged) |
  | `QnnHtp.dll` | ✓ | HTP ✓ |
  | `QnnCpu.dll` | ✓ | CPU ✓ |
  | `QnnGpu.dll` | ✓ | GPU ✓ |

  **Motivation:**
  When a user sets `backend_path=QnnHtp.dll`, they have explicitly opted into an HTP-targeted workflow — the same signal already used by `backend_id_` to drive HTP-specific op builder calls. Routing `backendValidateOpConfig` to QnnIr
  in this case only checks generic QNN spec compliance and provides a false sense of correctness for HTP-targeted DLC.

  **Why the original PR #218 was reverted (PR #413) and what's different now:**
[...]

PR #512, opened 2026-06-10
Add skip_backend_op_validation session option

qti-yuduo:
On device-less x86_64 hosts (cloud compilation), PR #438 routes DLC-dump op validation through the target backend, which runs against an arch-agnostic static op table and over-rejects v73+ kernels (error 3110). htp_arch/soc_model cannot fix this because QnnBackend_validateOpConfig takes only a backend handle -- the device-level arch never reaches the validation verdict without real hardware bridging device->backend via a context.

This adds a boolean EP session option `skip_backend_op_validation` that, when set, skips the target-backend validator initialization so that ValidateQnnNode falls through to the generic QnnIr serializer validation (pre-#438 behavior). The DLC serialization path is unaffected -- op lowering still uses the intended backend (e.g. HTP) and the IR serializer writes the DLC as before.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -1043,6 +1043,16 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                 false,
                                                 logger_);
 
+  // When dumping to DLC on a device-less host, the target backend validates op configs against an
+  // arch-agnostic op table and over-rejects arch-specific (v73+) ops. This option skips that
+  // validation, falling back to the serializer's generic checks. Default false keeps validation on.
+  static const std::string SKIP_BACKEND_OP_VALIDATION = "skip_backend_op_validation";
+  auto skip_backend_op_validation = ParseBoolOption(ort_api,
+                                                    session_options,
+                                                    FormatEPConfigKey(SKIP_BACKEND_OP_VALIDATION),
+                                                    false,
+                                                    logger_);
+
   // For context binary generation with weight sharing enabled, use the QnnBackendManager from the shared context if it exits
   // So that all graphs from later sessions will be compiled into the same QNN context
   if (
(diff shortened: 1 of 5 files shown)


yath1 (approved):
Lgtm. Thanks for the quick fix.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1037,6 +1037,16 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                 false,
                                                 logger_);
 
+  // When dumping to DLC on a device-less host, the target backend validates op configs against an
+  // arch-agnostic op table and over-rejects arch-specific (v73+) ops. This option skips that
+  // validation, falling back to the serializer's generic checks. Default false keeps validation on.
+  static const std::string SKIP_BACKEND_OP_VALIDATION = "skip_backend_op_validation";
```
tirupath-qti, line 1049: Documentation updates are missing about the new flag and when it is intended to be used.

qti-chuteng:
**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h:L607–L611`
The invariant comment above `qnn_validator_interface_` says "Both remain zero/null in all other flows", but this PR introduces a third state in the serializer flow: when `skip_backend_op_validation_=true`, `validator_backend_lib_handle_` is loaded (to detect backend type) while the interface/handle pair stays null. The comment should note this exception to prevent future maintainers from misreading this as a bug.

```suggestion
  // Invariant: qnn_validator_interface_ and validator_backend_handle_ are always set together
  // by LoadQnnSerializerBackend() (serializer flow: QnnIr). Both remain zero/null in all other flows.
  // Exception: when skip_backend_op_validation_=true, validator_backend_lib_handle_ is loaded
  // (to detect backend type) but qnn_validator_interface_/validator_backend_handle_ stay null,
  // so ValidateQnnNode() falls back to the serializer's generic checks.
  // QnnModelWrapper::ValidateQnnNode() uses validator_backend_handle_ != nullptr as the sole signal
  // to route backendValidateOpConfig to the validator interface rather than qnn_interface_.
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-yuduo: Good catch — added. Documented `skip_backend_op_validation` in `docs/execution_providers/QNN-ExecutionProvider.md` (EP Provider Options), alongside the other DLC-dump options, explaining what it does and the device-less-host scenario it's intended for. Also rebased the branch onto latest `main` to resolve the conflict with #488.

PR #525, opened 2026-06-15
Adopt AddQnnScalar helper across QNN EP op builders and node fusions

huaychou:
### Description

- Replace the verbose manual scalar-parameter construction (declare Qnn_Scalar_t, set dataType + value-union field, build QnnParamWrapper, push name, AddParamWrapper) with the AddQnnScalar<T> helper across all applicable op builders and node fusions.
- Fix two dtype/value-field mismatches
- Fix a use-after-move bug in channel_shuffle_fusion.cc

### Motivation and Context

- Found that most of the scalar-parameters are constructed manully

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
@@ -139,8 +139,7 @@ Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qnn_model_wrapper,
                                          einsum_node_unit.Name(),
                                          static_cast<uint32_t>(QNN_OP_DEPTH_TO_SPACE_MODE_DCR),
                                          QNN_OP_DEPTH_TO_SPACE_PARAM_MODE, mode_param_names));
-  RETURN_IF_NOT(mode_param_names.size() == 1, "Failed to add DepthToSpace mode param.");
-  const std::string mode_param_name = mode_param_names[0];
+  const std::string mode_param_name = std::move(mode_param_names.back());
 
   const OrtNodeUnitIODef& einsum_output = einsum_node_unit.Outputs()[0];
   TensorInfo einsum_output_info = {};
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
@@ -139,6 +139,7 @@ Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qnn_model_wrapper,
                                          einsum_node_unit.Name(),
                                          static_cast<uint32_t>(QNN_OP_DEPTH_TO_SPACE_MODE_DCR),
                                          QNN_OP_DEPTH_TO_SPACE_PARAM_MODE, mode_param_names));
+  RETURN_IF_NOT(mode_param_names.size() == 1, "Failed to add DepthToSpace mode param.");
   const std::string mode_param_name = mode_param_names[0];
 
   const OrtNodeUnitIODef& einsum_output = einsum_node_unit.Outputs()[0];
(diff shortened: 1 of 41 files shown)


minfhong-qti (approved):

onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
```diff
@@ -134,16 +134,13 @@ Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qnn_model_wrapper,
...
-                             mode_qnn_scalar);
-  const std::string mode_param_name = mode_param.GetParamTensorName();
-  RETURN_IF_NOT(qnn_model_wrapper.AddParamWrapper(std::move(mode_param)), "Failed to add param.");
+  std::vector<std::string> mode_param_names;
+  RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, einsum_node_unit.Index(),
+                                         einsum_node_unit.Name(),
+                                         static_cast<uint32_t>(QNN_OP_DEPTH_TO_SPACE_MODE_DCR),
+                                         QNN_OP_DEPTH_TO_SPACE_PARAM_MODE, mode_param_names));
+  RETURN_IF_NOT(mode_param_names.size() == 1, "Failed to add DepthToSpace mode param.");
+  const std::string mode_param_name = mode_param_names[0];
```
kuanyul-qti: **[N-2] (Minor)** reshape_einsum_reshape.cc uses a single-element-vector size-check workaround to obtain a single param name

Because `AddQnnScalar` takes a `std::vector<std::string>&` output and only appends, the caller takes a detour to fetch the single name it just pushed.

The reader has to first understand `AddQnnScalar`'s side effect (it appends one entry to the vector), then notice the `size()==1` invariant assertion, then realize this is just a single-name extraction. Since `AddQnnScalar` always pushes exactly one element on success and returns early on failure, `size()==1` is always true — the runtime check has no protective value and just adds noise.

Downstream usage of `mode_param_name` (passed to `CreateQnnNode`) still wants a raw string, so this dance is purely an impedance match with the helper's signature, not a real downstream need.

**Impact**: Readability only — no functional impact. But if this pattern propagates to other fusions, the readability tax compounds.

**Suggested fix:**

Option A: keep the existing `QnnParamWrapper` + `GetParamTensorName()` boilerplate at this call site rather than forcing the helper to fit, or introduce a single-name overload returning the param name by out-reference:

```
// In qnn_model_wrapper.h, add an overload:
template <typename T>
inline Ort::Status AddQnnScalar(QnnModelWrapper& qnn_model_wrapper,
                                const size_t& node_index, const std::string& node_name,
                                const T& scalar, const std::string& qnn_scalar_param_name,
                                std::string& param_name_out) {
  std::vector<std::string> tmp;
  RETURN_IF_ERROR(AddQnnScalar<T>(qnn_model_wrapper, node_index, node_name, scalar,
                                  qnn_scalar_param_name, tmp));
  param_name_out = std::move(tmp[0]);
  return Ort::Status();
}
```

Option B: use `mode_param_names.back()` at the call site and drop the size check:

```
std::vector<std::string> mode_param_names;
RETURN_IF_ERROR(AddQnnScalar<uint32_t>(...));
const std::string mode_param_name = std::move(mode_param_names.back());
```

Either choice removes the fake `size()==1` assertion line.

qti-chuteng:
**[T-1]** `scatterelements_op_builder.cc:L148–153`

The `axis_param` in `ScatterElementsOpBuilder::ProcessAttributesAndOutputs` was not migrated to `AddQnnScalar`, and its `AddParamWrapper` return value is still silently discarded — which is exactly the bug pattern the second commit fixes in `AddQnnScalar` itself.

```cpp
// Suggested replacement:
RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(),
                                       static_cast<uint32_t>(axis),
                                       QNN_OP_SCATTER_ELEMENTS_PARAM_AXIS, param_tensor_names));
```

huaychou:
> **[T-1]** `scatterelements_op_builder.cc:L148–153`
> 
> The `axis_param` in `ScatterElementsOpBuilder::ProcessAttributesAndOutputs` was not migrated to `AddQnnScalar`, and its `AddParamWrapper` return value is still silently discarded — which is exactly the bug pattern the second commit fixes in `AddQnnScalar` itself.
> 
> ```c++
> // Suggested replacement:
> RETURN_IF_ERROR(AddQnnScalar<uint32_t>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(),
>                                        static_cast<uint32_t>(axis),
>                                        QNN_OP_SCATTER_ELEMENTS_PARAM_AXIS, param_tensor_names));
> ```

I think the axis_param needs to be handled in another PR due to the ProcessAxisAttribute function. We will create another PR to cleanup axis_param.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc
huaychou: Updated.

qti-chuteng (approved):

PR #527, opened 2026-06-15
[QNN EP] Add nightly per-file coverage Excel report to Artifactory QA

yuhuchua-qti:
## Summary

  - Add `update_coverage_excel.py`: parses `coverage.xml` (Cobertura format) and produces a per-file line%/branch% Excel for the current day
  - Add `GenerateFileCoverageExcelTask` + `file_coverage_excel_linux_x86_64` build task (depends on `coverage_linux_x86_64`) 
  - Add `openpyxl` to `qcom/requirements.txt`
  - Update `qualcomm-internal-coverage.yml`:
    - Add `nightly_coverage` input to both `workflow_call` and `workflow_dispatch`
    - After coverage build: generate Excel → upload to Artifactory
    - Add `!cancelled()` to post-build steps so UT failures don't suppress uploads
  - Update `qualcomm-internal-release-nightly.yml`: add `coverage` job with `continue-on-error: true`

  ## How it works
  
  Each nightly run uploads an independent file to:
  {REPO}/qa/nightly-YYYY-MM-DD-{commit[:10]}/per_file_coverage.xlsx
  This follows the existing QA artifact naming convention. 30-day retention is handled by the existing `qualcomm-internal-clean-artifacts.yml` policy.

  | Scenario | Behavior |
  |---|---|
  | UT failure during coverage build | `coverage.xml` still generated; Excel and Artifactory upload still run (`!cancelled()`) |
  | Coverage job failure | `publish-artifacts` job unaffected (no `needs:` dependency; `continue-on-error: true` prevents nightly workflow from showing ❌) |

  ## How to validate
  
  Manually dispatch `qualcomm-internal-coverage.yml` from GitHub UI or via `gh` CLI with `nightly_coverage: true`, then verify `{REPO}/qa/nightly-YYYY-MM-DD-{commit[:10]}/per_file_coverage.xlsx` appears in Artifactory.
  
  ## Note
  
  Historical aggregation of daily snapshots into a 30-day report is tracked as a follow-up.

### DIFF
--- a/.github/workflows/qualcomm-internal-coverage.yml
+++ b/.github/workflows/qualcomm-internal-coverage.yml
@@ -19,6 +19,10 @@ on:
         description: "Set to true to upload the per-file coverage Excel to Artifactory"
         type: boolean
         default: false
+      qa_tag:
+        description: "Pre-computed QA tag from the caller; leave blank to auto-generate from today's date + SHA"
+        type: string
+        default: ''
   workflow_dispatch:
     inputs:
       target_py_vsn:
@@ -29,6 +33,10 @@ on:
         description: "Set to true to upload the per-file coverage Excel to Artifactory"
         type: boolean
         default: false
+      qa_tag:
+        description: "Pre-computed QA tag from the caller; leave blank to auto-generate from today's date + SHA"
+        type: string
+        default: ''
 
 jobs:
   coverage-linux-x86_64:
@@ -116,11 +124,34 @@ jobs:
           JF_URL: ${{ secrets.BUILD_ARTIFACTORY_URL }}
           JF_ACCESS_TOKEN: ${{ secrets.BUILD_ARTIFACTORY_TOKEN }}
 
+      - name: Set QA Build Tag
+        # Use the caller-provided qa_tag if present so coverage shares the exact same
+        # Artifactory directory as the nightly release upload. Falls back to auto-compute
+        # for standalone runs (workflow_dispatch / PR CI). Mirrors qualcomm-internal-release.yml.
+        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
+        run: |
+          if [ -z "${{ inputs.qa_tag }}" ]; then
+            run_date="$(date --rfc-3339=date)"
+            echo "QA_TAG=nightly-${run_date}-${GITHUB_SHA:0:10}" >> ${GITHUB_ENV}
+            echo "COVERAGE_RUN_DATE=${run_date}" >> ${GITHUB_ENV}
+          else
+            echo "QA_TAG=${{ inputs.qa_tag }}" >> ${GITHUB_ENV}
+            # Extract date portion from the canonical "nightly-YYYY-MM-DD-{sha}" format
+            # so the Excel's date column matches the directory name.
+            echo "COVERAGE_RUN_DATE=$(echo '${{ inputs.qa_tag }}' | sed -n 's/^nightly-\([0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}\)-.*/\1/p')" >> ${GITHUB_ENV}
+          fi
+
       - name: Update per-file coverage Excel
         # !cancelled(): run even when the coverage step failed due to UT failure.
         # coverage.xml is still produced by generate_coverage.sh before it propagates the exit code.
+        # File guard: build-stage failures (cmake, link, submodule) can leave coverage.xml absent — skip cleanly.
+        # update_coverage_excel.py picks up $COVERAGE_RUN_DATE for the Excel's date column.
         if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
         run: |
+          [ -f "${COVERAGE_DIR}/coverage.xml" ] || {
+            echo "::warning::coverage.xml missing — skipping nightly Excel generation"
+            exit 0
+          }
           python3 qcom/build_and_test.py \
             --target-py-version ${{ inputs.target_py_vsn }} \
             --only \
@@ -129,9 +160,13 @@ jobs:
       - name: Upload per-file coverage Excel to Artifactory
         if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
         run: |
+          [ -f "${COVERAGE_DIR}/per_file_coverage.xlsx" ] || {
+            echo "::warning::per_file_coverage.xlsx missing — skipping nightly Artifactory upload"
+            exit 0
+          }
           jf rt upload \
             "${COVERAGE_DIR}/per_file_coverage.xlsx" \
-            "${{ secrets.BUILD_ARTIFACTORY_REPO }}/qa/nightly-$(date -u +%Y-%m-%d)-${GITHUB_SHA:0:10}/" \
+            "${{ secrets.BUILD_ARTIFACTORY_REPO }}/qa/${{ env.QA_TAG }}/" \
             --flat=true
 
       - name: Post Diff Coverage to PR
--- a/.github/workflows/qualcomm-internal-coverage.yml
+++ b/.github/workflows/qualcomm-internal-coverage.yml
@@ -15,12 +15,20 @@ on:
         description: "Python version for the coverage build wheel (does not affect C++ coverage results)"
         type: string
         default: '3.10'
+      nightly_coverage:
+        description: "Set to true on nightly runs to update the per-file coverage Excel in Artifactory"
+        type: boolean
+        default: false
   workflow_dispatch:
     inputs:
       target_py_vsn:
         description: "Python version for the coverage build wheel (does not affect C++ coverage results)"
         type: string
         default: '3.10'
+      nightly_coverage:
+        description: "Set to true to test the nightly Artifactory upload path"
+        type: boolean
+        default: false
 
 jobs:
   coverage-linux-x86_64:
@@ -42,7 +50,8 @@ jobs:
         run: |
           set -euo pipefail
 
-          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ] || \
+             [ "${{ inputs.nightly_coverage }}" = "true" ]; then
             echo "run=true" >> "${GITHUB_OUTPUT}"
             exit 0
           fi
@@ -79,7 +88,7 @@ jobs:
 
       - name: Upload Coverage Report
         uses: actions/upload-artifact@v7
-        if: steps.gate.outputs.run == 'true'
+        if: steps.gate.outputs.run == 'true' && !cancelled()
         with:
           name: "coverage-linux-x86_64-${{ github.sha }}"
           retention-days: 3
@@ -90,7 +99,7 @@ jobs:
             ${{ env.COVERAGE_DIR }}/diff-coverage.txt
 
       - name: Show Diff Coverage Summary
-        if: steps.gate.outputs.run == 'true'
+        if: steps.gate.outputs.run == 'true' && !cancelled()
         run: |
           txt="${COVERAGE_DIR}/diff-coverage.txt"
           if [ -f "${txt}" ]; then
@@ -99,6 +108,31 @@ jobs:
             cat "${txt}" >> "${GITHUB_STEP_SUMMARY}"
             echo '```' >> "${GITHUB_STEP_SUMMARY}"
           fi
+
+      - name: Set up JFrog CLI
+        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
+        uses: jfrog/setup-jfrog-cli@v5
+        env:
+          JF_URL: ${{ secrets.BUILD_ARTIFACTORY_URL }}
+          JF_ACCESS_TOKEN: ${{ secrets.BUILD_ARTIFACTORY_TOKEN }}
+
+      - name: Update per-file coverage Excel
+        # !cancelled(): run even when the coverage step failed due to UT failure.
+        # coverage.xml is still produced by generate_coverage.sh before it propagates the exit code.
+        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
+        run: |
+          python3 qcom/build_and_test.py \
+            --target-py-version ${{ inputs.target_py_vsn }} \
+            --only \
+            file_coverage_excel_linux_x86_64
+
+      - name: Upload per-file coverage Excel to Artifactory
+        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
+        run: |
+          jf rt upload \
+            "${COVERAGE_DIR}/per_file_coverage.xlsx" \
+            "${{ secrets.BUILD_ARTIFACTORY_REPO }}/qa/nightly-$(date -u +%Y-%m-%d)-${GITHUB_SHA:0:10}/" \
+            --flat=true
 
       - name: Post Diff Coverage to PR
         # TODO: re-enable once the coverage pipeline is stable.
--- a/.github/workflows/qualcomm-internal-coverage.yml
+++ b/.github/workflows/qualcomm-internal-coverage.yml
@@ -32,6 +32,8 @@ jobs:
     runs-on: ["self-hosted", "X64", "Linux", "Build"]
     env:
       COVERAGE_DIR: build/linux-x86_64/RelWithDebInfo/coverage
+      # Consumed by upload-qa-artifact's publish_qa.py (reads os.environ["BUILD_ARTIFACTORY_REPO"]).
+      BUILD_ARTIFACTORY_REPO: "${{ secrets.BUILD_ARTIFACTORY_REPO }}"
     steps:
       - uses: actions/checkout@v6
         with:
@@ -126,7 +128,12 @@ jobs:
             echo "QA_TAG=${{ inputs.qa_tag }}" >> ${GITHUB_ENV}
             # Extract date portion from the canonical "nightly-YYYY-MM-DD-{sha}" format
             # so the Excel's date column matches the directory name.
-            echo "COVERAGE_RUN_DATE=$(echo '${{ inputs.qa_tag }}' | sed -n 's/^nightly-\([0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}\)-.*/\1/p')" >> ${GITHUB_ENV}
+            run_date="$(echo '${{ inputs.qa_tag }}' | sed -n 's/^nightly-\([0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}\)-.*/\1/p')"
+            if [ -z "${run_date}" ]; then
+              echo "::warning::qa_tag '${{ inputs.qa_tag }}' is not canonical (nightly-YYYY-MM-DD-*); falling back to today's UTC date for the Excel date column"
+              run_date="$(date --rfc-3339=date)"
+            fi
+            echo "COVERAGE_RUN_DATE=${run_date}" >> ${GITHUB_ENV}
           fi
 
       - name: Update per-file coverage Excel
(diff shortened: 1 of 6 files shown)


.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -15,12 +15,20 @@ on:
...
+        type: boolean
+        default: false
   workflow_dispatch:
     inputs:
       target_py_vsn:
         description: "Python version for the coverage build wheel (does not affect C++ coverage results)"
         type: string
         default: '3.10'
+      nightly_coverage:
+        description: "Set to true to test the nightly Artifactory upload path"
```
kuanyul-qti: Just curious, why is the description here different from the one above?

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -76,7 +85,7 @@ jobs:
 
       - name: Upload Coverage Report
         uses: actions/upload-artifact@v7
-        if: steps.gate.outputs.run == 'true'
+        if: steps.gate.outputs.run == 'true' && !cancelled()
```
kuanyul-qti, line 89: Is `cancelled()` a built-in GitHub function? I couldn't find its definition anywhere.

qcom/build_and_test.py
```diff
@@ -980,6 +982,23 @@ def extract_ort_windows_x86_64(self, plan: Plan) -> str:
             )
         )
 
+    if is_host_linux() and is_host_x86_64():
+
+        @public_task("Generate per-file coverage Excel from coverage.xml (Linux x86_64)")
+        @depends(["coverage_linux_x86_64"])
+        def file_coverage_excel_linux_x86_64(self, plan: Plan) -> str:
+            build_dir = REPO_ROOT / "build" / "linux-x86_64"
+            excel_file = build_dir / "RelWithDebInfo" / "coverage" / "per_file_coverage.xlsx"
```
kuanyul-qti, line 991: Just curious, why not use the existing output format from the coverage task and switch to xlsx instead?

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -97,6 +106,31 @@ jobs:
...
+          JF_ACCESS_TOKEN: ${{ secrets.BUILD_ARTIFACTORY_TOKEN }}
+
+      - name: Update per-file coverage Excel
+        # !cancelled(): run even when the coverage step failed due to UT failure.
+        # coverage.xml is still produced by generate_coverage.sh before it propagates the exit code.
+        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
+        run: |
+          python3 qcom/build_and_test.py \
+            --target-py-version ${{ inputs.target_py_vsn }} \
+            --only \
```
kuanyul-qti, line 153: As described in the `build_and_test.py` file, the `--only` argument skips all dependencies. Should this argument be used here, considering that `file_coverage_excel_linux_x86_64` depends on `coverage_linux_x86_64`?

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Fixed — both workflow_call and workflow_dispatch now use the same description: "Set to true to upload the per-file coverage Excel to Artifactory".
Thanks.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Yes, cancelled() is a built-in GitHub Actions status check function, in the same family as success(), failure(), and always(). See [GitHub Actions expressions docs](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions).

qcom/build_and_test.py
yuhuchua-qti: The existing outputs (lcov, Cobertura XML, HTML) are per-build formats designed for machine tooling or single-build inspection. 
The xlsx is designed for multi-day trend tracking — it can be opened directly in Excel/Spreadsheet, filtered by file or date, and used to visualize 30-day coverage trends. 
The two serve different purposes. Thanks~

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Intentional. Steps within a GitHub Actions job run strictly in order, so "Run Coverage and Diff Coverage" is always guaranteed to complete (or fail) before this step. coverage.xml is on disk regardless of UT failures because generate_coverage.sh produces it before propagating the exit code. --only avoids re-running the full coverage build (~15–30 min) and just runs the Excel generation.

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -97,6 +106,31 @@ jobs:
...
+            file_coverage_excel_linux_x86_64
+
+      - name: Upload per-file coverage Excel to Artifactory
+        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
+        run: |
+          jf rt upload \
+            "${COVERAGE_DIR}/per_file_coverage.xlsx" \
+            "${{ secrets.BUILD_ARTIFACTORY_REPO }}/qa/nightly-$(date -u +%Y-%m-%d)-${GITHUB_SHA:0:10}/" \
```
kuanyul-qti: In PR description, the target path is `qa/coverage-YYYY-MM-DD` which is different from what code shows. May I know which one is the right one?

qcom/ep_build/tasks/build.py
```diff
@@ -256,6 +256,40 @@ def __init__(
...
 
+class GenerateFileCoverageExcelTask(RunExecutablesWithVenvTask):
+    """Generate a per-day per-file coverage Excel snapshot from a Cobertura XML report.
+
+    Calls update_coverage_excel.py with the coverage.xml produced by
+    GenerateCoverageTask and writes a fresh Excel file containing today's
+    line% and branch% for every QNN EP source file.  Each nightly run
+    produces an independent snapshot; historical aggregation across days is
+    handled by merge_coverage_excel.py and the aggregate CI workflow
+    (future work, tracked in PR 2).
```
kuanyul-qti: The "PR 2" is not clear.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Sorry for the confusion. 
The path was updated during review to align with the existing QA artifact naming convention (nightly-{date}-{commit[:10]}). 
The code is correct; the PR description is outdated. 
The final upload path is qa/nightly-YYYY-MM-DD-{commit[:10]}/per_file_coverage.xlsx.

qcom/ep_build/tasks/build.py
yuhuchua-qti: Fixed — replaced with a TODO comment describing the intent directly: "enable the aggregate CI workflow to merge daily snapshots into a 30-day history report."

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -97,6 +106,31 @@ jobs:
...
+
+      - name: Update per-file coverage Excel
+        # !cancelled(): run even when the coverage step failed due to UT failure.
+        # coverage.xml is still produced by generate_coverage.sh before it propagates the exit code.
+        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
+        run: |
+          python3 qcom/build_and_test.py \
+            --target-py-version ${{ inputs.target_py_vsn }} \
+            --only \
+            file_coverage_excel_linux_x86_64
```
kuanyul-qti, line 154: Build-stage failures *before* `generate_coverage.sh` runs (submodule update, cmake configure, link error, coverage compile-flag failure) still leave `coverage.xml` absent, and the two new steps run anyway - turning one failure into three red steps in the run. `continue-on-error: true` at the caller prevents this from gating publish-artifacts, so this is dashboard noise, not correctness. Suggest a one-line gate at the top of each step's `run:` block:
```suggestion
        run: |
          [ -f "${COVERAGE_DIR}/coverage.xml" ] || {
            echo "::warning::coverage.xml missing — skipping nightly upload step"
            exit 0
          }
          python3 qcom/build_and_test.py \
            --target-py-version ${{ inputs.target_py_vsn }} \
            --only \
            file_coverage_excel_linux_x86_64
```

.github/workflows/qualcomm-internal-coverage.yml
kuanyul-qti: The Artifactory target uses bash `date -u +%Y-%m-%d` while the Excel `date` column uses Python `datetime.now(timezone.utc).strftime("%Y-%m-%d")` in the preceding step. Across UTC midnight these two clocks can disagree by one day, leaving a directory whose Excel `date` column is off-by-one. Pick one source of truth — e.g. compute `run_date` once in the gate step, expose via `$GITHUB_OUTPUT`, and explicitly pass `--date=${{ steps.gate.outputs.run_date }}` to `update_coverage_excel.py` (the script already accepts `--date`).

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Done — added [ -f "${COVERAGE_DIR}/coverage.xml" ] || { echo "::warning::..."; exit 0; } guard at the top of "Update per-file coverage Excel", and an analogous guard for per_file_coverage.xlsx at the top of "Upload per-file coverage Excel to Artifactory". Build-stage failures now show one red step instead of three.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Fixed — but went one step further: rather than just keeping the date consistent within the coverage workflow, we now compute the QA tag once in the parent qualcomm-internal-release-nightly.yml (new setup job) and pass it as qa_tag input to both publish-artifacts and coverage, so the coverage Excel ends up in the exact same Artifactory directory as the nightly release artifacts.

Changes:
- qualcomm-internal-release-nightly.yml: add setup job that runs echo "qa_tag=nightly-$(date --rfc-3339=date)-${GITHUB_SHA:0:10}" >> "$GITHUB_OUTPUT"; both child jobs needs: setup and pass qa_tag: ${{ needs.setup.outputs.qa_tag }}
- qualcomm-internal-coverage.yml: add qa_tag input (with auto-compute fallback for standalone runs); "Set QA Build Tag" step either uses the input or falls back, mirroring qualcomm-internal-release.yml
- update_coverage_excel.py: honor $COVERAGE_RUN_DATE env var (CLI --date still wins) so the Excel date column matches the QA tag's date even when CI passes the tag down

qcom/scripts/linux/update_coverage_excel.py
```diff
@@ -0,0 +1,119 @@
...
+        filename = cls.get("filename", "").strip()
+        if not filename:
+            continue
+        rows.append(
+            {
+                "file": filename,
+                "line_pct": round(float(cls.get("line-rate", "0")) * 100, 1),
+                "branch_pct": round(float(cls.get("branch-rate", "0")) * 100, 1),
+            }
+        )
```
kuanyul-qti, line 72: **[T-1] (Nit)** Empty <class> set in coverage.xml is silently treated as success

`parse_coverage_xml` (line 57-70) does not warn when `coverage.xml` exists but has zero `<class>` elements:

```
for cls in root.iter("class"):
    filename = cls.get("filename", "").strip()
    if not filename:
        continue
    rows.append({...})

rows.sort(...)
return rows
```

If the XML is class-empty for any reason (over-aggressive filter, upstream parse glitch, empty .gcda), `file_rows` is an empty list; `main` still saves a header-only .xlsx and exits 0. Downstream sees a "successful" upload of an empty Excel — harder to spot than a clean failure.

**Suggested fix:**

After computing `file_rows`, gate with:
```
if not file_rows:
    print("ERROR: coverage XML contains no <class> elements; refusing to write an empty workbook", file=sys.stderr)
    sys.exit(1)
```
or at least surface it as a `::warning::` while still writing the file.

.github/workflows/qualcomm-internal-release-nightly.yml
```diff
@@ -13,8 +13,21 @@ on:
     - cron: "0 9 * * *"  # Every day at 2am in San Diego/PDT (9am UTC)
 
 jobs:
+  setup:
+    name: Compute Shared QA Tag
+    runs-on: ubuntu-latest
```
kuanyul-qti: It should be `[Linux, self-hosted]` instead of `ubuntu-latest`.

.github/workflows/qualcomm-internal-coverage.yml
kuanyul-qti: Can be optimized to - `inputs: &common_inputs`

.github/workflows/qualcomm-internal-coverage.yml
```diff
@@ -15,12 +15,28 @@ on:
...
         type: string
         default: '3.10'
+      nightly_coverage:
+        description: "Set to true to upload the per-file coverage Excel to Artifactory"
+        type: boolean
+        default: false
+      qa_tag:
+        description: "Pre-computed QA tag from the caller; leave blank to auto-generate from today's date + SHA"
+        type: string
+        default: ''
```
kuanyul-qti: Can be optimized to `inputs: *common_inputs`

qcom/scripts/linux/update_coverage_excel.py
yuhuchua-qti: Fixed — added a warning when file_rows is empty. 
Chose the "warn + still write" approach rather than sys.exit(1), because an empty Excel is still a valid signal that something upstream went wrong (e.g. filter misconfiguration), and the Artifactory upload path makes it easy to spot. 
A hard failure here would block the workflow and hide the fact that coverage did run — just with no class data.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Good idea in principle, but GitHub Actions' workflow parser does not support YAML anchors/aliases (&/*) — the workflow will fail validation. 
This is a known limitation (see https://github.com/actions/runner/issues/1182).
Keeping the explicit duplication for now.

.github/workflows/qualcomm-internal-release-nightly.yml
yuhuchua-qti: Fixed — changed to [Linux, self-hosted]. Thanks.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Same as above — not supported by GitHub Actions. Leaving as-is.

.github/workflows/qualcomm-internal-coverage.yml
kuanyul-qti: But we can set it in https://github.com/onnxruntime/onnxruntime-qnn/blob/main/.github/workflows/qualcomm-internal-release.yml#L13

yuhuchua-qti: You're right — anchor/alias is supported (release.yml:13 already uses it), my earlier reply was wrong. Applied in 9e458ef4f7; L39 is covered too.
Verified via manual dispatch (run 28562875260) — all nightly steps green.

.github/workflows/qualcomm-internal-coverage.yml
yuhuchua-qti: Same fix (9e458ef4f7) — resolved together with the anchor above.

kuanyul-qti (approved):

qti-chuteng:
## [M-1] `.github/workflows/qualcomm-internal-coverage.yml:L155-158`

The raw `jf rt upload "...${BUILD_ARTIFACTORY_REPO}/qa/${QA_TAG}/"` reinvents an existing convention: the repo already has `.github/actions/upload-qa-artifact`, whose `publish_qa.py` → `QaArtifactory.artifact_root` produces exactly `{repo}/qa/{tag}` (`qcom/scripts/artifactory/artifactory.py:79`) — the same target. `qualcomm-internal-release.yml` (L519-523) uses that action. Two implementations of the QA path convention means a future QA-layout change must be made in both places or coverage lands in the wrong directory silently. Consider:

```yaml
      - name: Upload per-file coverage Excel to Artifactory
        if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
        uses: ./.github/actions/upload-qa-artifact
        with:
          src_pattern: "linux-x86_64/RelWithDebInfo/coverage/per_file_coverage.xlsx"
          qa_tag: ${{ env.QA_TAG }}
```

If a single-file upload genuinely can't reuse it, please note why in the PR description.

## [N-1] `.github/workflows/qualcomm-internal-release-nightly.yml:L46`

The coverage job is `continue-on-error: true` and both new steps warn-and-`exit 0` on missing files. Reasonable best-effort, but combined it means a persistently broken coverage build keeps nightly green while Artifactory silently gets no new Excel. Consider surfacing "no coverage Excel produced this nightly" via a `GITHUB_STEP_SUMMARY` annotation in the skip branches.

## [N-2] `.github/workflows/qualcomm-internal-coverage.yml:L129`

When a caller passes a non-canonical `qa_tag` (e.g. a custom `workflow_dispatch` tag), the `sed` extraction of `COVERAGE_RUN_DATE` yields an empty string and `update_coverage_excel.py` silently falls back to `datetime.now(UTC)`, so the Excel's date column can disagree with the directory name. The nightly path is always canonical so it's unaffected. Emit a `::warning::` on empty extraction, or make the job-computed `run_date` the single source of truth.

yuhuchua-qti:
> ## [M-1] `.github/workflows/qualcomm-internal-coverage.yml:L155-158`
> The raw `jf rt upload "...${BUILD_ARTIFACTORY_REPO}/qa/${QA_TAG}/"` reinvents an existing convention: the repo already has `.github/actions/upload-qa-artifact`, whose `publish_qa.py` → `QaArtifactory.artifact_root` produces exactly `{repo}/qa/{tag}` (`qcom/scripts/artifactory/artifactory.py:79`) — the same target. `qualcomm-internal-release.yml` (L519-523) uses that action. Two implementations of the QA path convention means a future QA-layout change must be made in both places or coverage lands in the wrong directory silently. Consider:
> 
> ```yaml
>       - name: Upload per-file coverage Excel to Artifactory
>         if: inputs.nightly_coverage && steps.gate.outputs.run == 'true' && !cancelled()
>         uses: ./.github/actions/upload-qa-artifact
>         with:
>           src_pattern: "linux-x86_64/RelWithDebInfo/coverage/per_file_coverage.xlsx"
>           qa_tag: ${{ env.QA_TAG }}
> ```
> 
> If a single-file upload genuinely can't reuse it, please note why in the PR description.
> 
> ## [N-1] `.github/workflows/qualcomm-internal-release-nightly.yml:L46`
> The coverage job is `continue-on-error: true` and both new steps warn-and-`exit 0` on missing files. Reasonable best-effort, but combined it means a persistently broken coverage build keeps nightly green while Artifactory silently gets no new Excel. Consider surfacing "no coverage Excel produced this nightly" via a `GITHUB_STEP_SUMMARY` annotation in the skip branches.
> 
> ## [N-2] `.github/workflows/qualcomm-internal-coverage.yml:L129`
> When a caller passes a non-canonical `qa_tag` (e.g. a custom `workflow_dispatch` tag), the `sed` extraction of `COVERAGE_RUN_DATE` yields an empty string and `update_coverage_excel.py` silently falls back to `datetime.now(UTC)`, so the Excel's date column can disagree with the directory name. The nightly path is always canonical so it's unaffected. Emit a `::warning::` on empty extraction, or make the job-computed `run_date` the single source of truth.

[M-1]
Agreed — switched to ./.github/actions/upload-qa-artifact so the qa/{tag} convention stays in publish_qa.py → QaArtifactory.artifact_root only. Added BUILD_ARTIFACTORY_REPO to the job env (read by publish_qa.py). File-existence moved to a guard step gating the action, preserving clean-skip. Fixed in 86f83c2157.

[N-1]
Done — both skip branches (coverage.xml / per_file_coverage.xlsx missing) now write to GITHUB_STEP_SUMMARY, so a broken coverage build leaves a visible trace. 86f83c2157

[N-2]
Done — empty sed extraction now emits a ::warning:: and falls back to today's UTC date, so the Excel date column is never blank. Nightly path is canonical so unaffected. 86f83c2157

qti-chuteng (approved):

PR #528, opened 2026-06-15
Add unit and integration tests for ort_api.cc (88.4% coverage)

yuhuchua-qti:
**Summary**
  
  Adds explicit component-level and pipeline integration tests for ort_api.cc, replacing coverage that was previously only a side-effect of op-builder tests (which will be
  deleted).

  **Component-level tests (unit/ort_api_test.cc, 44 tests, QnnUnit_*):**
  - Use a zero-init stub OrtApi with only the needed function pointers filled in, and fake opaque pointers (reinterpret_cast<const OrtNode*>(uintptr_t{0x100})) that stubs
  accept but never dereference. 
  - Cover: GetProviderOptionPrefix, ReadFileIntoBuffer, dynamic library helpers (OrtLoadDynamicLibrary / OrtUnloadDynamicLibrary / OrtGetSymbolFromLibrary),
  GetSessionConfigEntryOrDefault, OrtGetRuntimePath, GetDynamicLibraryLocationByAddress.
  - Cover: OrtNodeUnit SingleNode constructor — happy-path I/O parsing, GetInputEdgesCount, GetOutputNodes, and the error branch when InitForSingleNode returns non-null
  OrtStatus*.
  - Build minimal ONNX models inline via Ort::Model / Ort::Graph / Ort::Node and run them through Ort::Session with the QNN CPU backend, so GetCapability + Compile produces
  real OrtNode* objects.
  - Cover GetQDQIODefs + QDQ OrtNodeUnit constructor via a DQ→Relu→Q model.
  - Cover InitForSingleNode DQ branch (standalone DQ→Relu) and Q branch (Relu→standalone Q).
  - Cover OrtNodeAttrHelper "found" paths via: Transpose (GetInt64s), Conv with auto_pad (Get(string)), LeakyRelu (Get(float)), Pad opset-10 (GetInt64s + GetFloat), ArgMax
  (Get(int32_t)), ConvTranspose (Get(vector<int32_t>)).
  - Tests skip gracefully when the QNN CPU backend is unavailable.
  - No private ORT Core header dependency — uses onnxruntime_cxx_api.h (filename-only) and public EP registration API only.

  **CMake**: adds integration/* glob to onnxruntime_provider_test alongside the existing unit/* glob.

  **Coverage** (ort_api.cc, 516 executable lines):

 | Filter | Lines hit | % | 
 |---|---|---|  
 | QnnUnit_*:QnnInt_* (this PR's tests only) | 443 / 516 | 85.9% |
 | *Qnn* (all tests) | 456 / 516 | 88.4% |

  The remaining ~60 uncovered lines are the practical ceiling: GetQDQIODefs error branches requiring OS/API fault injection, OrtNodeAttrHelper overloads only used by UDO
  fusion, and platform-specific / file I/O error paths.

  **Test plan**

  - QnnUnit_OrtApiTest.* — 44 tests PASS (component-level, no backend needed)
  - QnnInt_OrtApiTest.* — 9 tests PASS (QNN CPU backend on Linux x86_64)
  - Coverage build (coverage_linux_x86_64): ort_api.cc 443/516 = 85.9% (unit+integration), 456/516 = 88.4% (all Qnn)
  - Linter: no private ORT Core header in either test file


onnxruntime/test/providers/qnn/integration/ort_api_test.cc
```diff
@@ -0,0 +1,710 @@
...
+extern std::unique_ptr<Ort::Env> ort_env;
+
+namespace onnxruntime {
+namespace test {
+
+// ============================================================
+// Helpers
+// ============================================================
+
+namespace {
```
minfhong-qti: I guess all the items in this namespace can be reused by further integration test files. Maybe consider moving them into separate files?

onnxruntime/test/providers/qnn/integration/ort_api_test.cc
```diff
@@ -0,0 +1,710 @@
...
+
+#include <cstdint>
+#include <cstring>
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "onnxruntime_cxx_api.h"
+#include "gtest/gtest.h"
```
minfhong-qti:
```suggestion
#include "gtest/gtest.h"
#include "onnxruntime_cxx_api.h"
```
Nit.

onnxruntime/test/providers/qnn/unit/ort_api_test.cc
```diff
@@ -0,0 +1,824 @@
...
+//   GetProviderOptionPrefix, ReadFileIntoBuffer, OrtLoadDynamicLibrary,
+//   OrtUnloadDynamicLibrary, OrtGetSymbolFromLibrary, GetSessionConfigEntryOrDefault,
+//   OrtGetRuntimePath, GetDynamicLibraryLocationByAddress.
+//
+// NOTE: OrtNodeAttrHelper, OrtNodeUnit, ParseOrtValueInfo, and GetQDQIODefs are
+// integration test territory — they require a real OrtNode* and would crash or
+// produce UB with fake/null pointers. Those functions are NOT covered here.
+// Component-level ceiling for ort_api.cc is ~15% for this reason.
+
+#include "core/providers/qnn/ort_api.h"
```
minfhong-qti: Just curious. Why include it here instead of below?

onnxruntime/test/providers/qnn/integration/ort_api_test.cc
yuhuchua-qti: Good idea — extracted into `integration/qnn_int_test_utils.h`: `RegisteredQnnEp`, `MakeQnnHtpSessionOptions`, and `MakeValueInfo1D/2D/3D/4D`.
README updated to document the shared utilities so future integration test files can reuse them directly.

onnxruntime/test/providers/qnn/integration/ort_api_test.cc
yuhuchua-qti: Done — swapped to `gtest/gtest.h` before `onnxruntime_cxx_api.h`.

onnxruntime/test/providers/qnn/unit/ort_api_test.cc
yuhuchua-qti: Good catch — moved inside the `QNN_EP_INTERNAL_SYMBOL_ACCESS` guard.
Per `unit/README.md` policy, only `gtest/gtest.h` should stay outside the guard; this was an inconsistency in the original file.

minfhong-qti: I'm curious why leave `gtest/gtest.h` outside the `QNN_EP_INTERNAL_SYMBOL_ACCESS` even though the example in README does it so.

onnxruntime/test/providers/qnn/unit/ort_api_test.cc
```diff
@@ -0,0 +1,824 @@
...
+//
+// All OrtApi calls in ParseOrtValueInfo and InitForSingleNode go through the
+// *parameter* api (not Ort::GetApi()), so every branch is stubable.
+//
+// Exception: DequantizeLinear / QuantizeLinear paths in InitForSingleNode call
+// OrtNodeAttrHelper which uses the global OrtApi — those paths are integration
+// test territory and are not covered here.
+// ============================================================
+
+namespace {
```
minfhong-qti, line 345: Not sure but is there any parts that can move to common file to be reused in other UTs?

onnxruntime/test/providers/qnn/integration/qnn_test_utils.h
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: At first glance, I thought the naming `int` means "integer". Is there a better naming or is `qnn_test_utils.h` enough as it is already under `integration` folder?

onnxruntime/test/providers/qnn/unit/ort_api_test.cc
yuhuchua-qti: Following unit/README.md:54 — the guard keeps EP-internal symbol access out of non-coverage builds while leaving the TU well-formed as a gtest file (empty body when guard is off). Same pattern in qnn_def_test.cc / qnn_model_wrapper_test.cc / qnn_utils_test.cc.

onnxruntime/test/providers/qnn/unit/ort_api_test.cc
yuhuchua-qti: Held off intentionally. Each Stub* function is scenario-specific (e.g. StubHasShape_True vs StubHasShape_False, StubGetDimsCount_Two vs _Zero) rather than a general OrtApi mock. 
Happy to extract when a second file needs the same stub shapes.

onnxruntime/test/providers/qnn/integration/qnn_test_utils.h
yuhuchua-qti: Agreed — "int" read as integer to me too. Renamed the header to qnn_test_utils.h. Also renamed the test prefix QnnInt_ → QnnInteg_ (same ambiguity) to match the original 5-tier plan naming. 
Applied in efaf9cd52b — 9 tests still green, ort_api.cc coverage unchanged at 88.4%.

minfhong-qti (approved):
Can be done later.
Regarding the testcase naming, how about `QnnUT_xxx` and `QnnIT_xxx`?

yuhuchua-qti:
> Can be done later. Regarding the testcase naming, how about `QnnUT_xxx` and `QnnIT_xxx`?

Sure, I will prepare another PR for this.
Thanks.

qti-chuteng (approved):

PR #532, opened 2026-06-16
[QNN EP] Include QNN error string in QNN API failure messages

qti-chuteng:
## Summary

Several QNN API call sites in the QNN EP returned an error (either an `OrtStatus` propagated to the caller, or an error log line) that **dropped the QNN error handle**, surfacing only a hardcoded English message or a bare integer code. When one of these calls fails at runtime, the user is left with a message like `"Failed to finalize QNN graph."` or `"QNN graph execute error. Error code: 6005"` — with no symbolic indication of *why* it failed (out-of-memory vs. an unsupported op vs. a driver/SSR fault all look the same).

This PR aligns four such call sites with the convention **already used throughout `qnn_backend_manager.cc`** (e.g. `backendCreate`, `profileCreate`, `contextCreate`, `contextCreateFromBinaryListAsync`), which append the QNN symbolic message via the existing helpers `utils::GetQnnErrorMessage(...)` / `QnnBackendManager::QnnErrorHandleToString(...)`.

## Why this is needed

The QNN SDK exposes `errorGetMessage` / `errorGetVerboseMessage`, and the EP already wraps it in `utils::GetQnnErrorMessage` (`qnn_utils.cc`). Most QNN API call sites already use it, but these four were inconsistent and discarded the handle. The result is that field failures on these specific paths (graph finalize, graph execute, graph create, and LoRA-adapter binary-section apply) produce error text that cannot be triaged without attaching a debugger or correlating raw integer codes against the SDK headers by hand. Enriching these messages brings them in line with the rest of the file and makes failures self-describing.

## Changes

| Call site | File | Before | After |
|-----------|------|--------|-------|
| `graphFinalize` | `qnn_model.cc` | logged the code, then returned a generic `"Failed to finalize QNN graph."` that discarded it | returns `utils::GetQnnErrorMessage(...)` + code in the `OrtStatus` |
| `graphExecute` | `qnn_model.cc` | returned `"... Error code: <int>"` (bare integer) | includes the symbolic QNN message + code; the existing `QNN_COMMON_ERROR_SYSTEM_COMMUNICATION` (SSR / NPU-crash) special case is **left unchanged** |
| `graphCreate` | `qnn_model_wrapper.cc` | error log carried only the graph name | log now includes the QNN message + code from the failing call (function still returns `bool`; control flow unchanged) |
| `contextApplyBinarySection` | `qnn_backend_manager.cc` | hardcoded `"Failed to apply binary section."` | appended with `QnnErrorHandleToString(...)` + code |

Message format follows the existing convention in the same files: `"<context>. Error: <message>, Code: <int>"`.

## Scope / risk

- **Error-path only.** No success-path behavior, control flow, or public API changes. Each hunk only enriches the text emitted when a QNN call has *already* failed.
- The `graphExecute` SSR/NPU-crash branch is intentionally preserved as-is.
- `graphCreate` returns `bool`; only its log is enriched, the early-return is untouched.

## Testing

These are error-path message enrichments that only trigger on real QNN API failures, which are not exercised by the standard unit suite. Verified by:
- `lintrunner` (clang-format) — clean, no formatting changes required.
- Type/linkage review: `utils::GetQnnErrorMessage(const QNN_INTERFACE_VER_TYPE&, Qnn_ErrorHandle_t)` matches the `GetQnnInterface()` return type; the helper is already included and used elsewhere in these translation units' sibling files.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -201,9 +201,8 @@ Ort::Status QnnBackendManager::ParseLoraConfig(std::string lora_config_path) {
           auto context_apply_binary_section_rt = qnn_interface_.contextApplyBinarySection(
               contexts_[cIdx], graph, QNN_CONTEXT_SECTION_UPDATABLE, &contextBuffer, profile_backend_handle_, nullptr);
           RETURN_IF(QNN_SUCCESS != context_apply_binary_section_rt,
-                    ("Failed to apply binary section. Error: " +
-                     QnnErrorHandleToString(context_apply_binary_section_rt) +
-                     ", Code: " + std::to_string(context_apply_binary_section_rt))
+                    ("Failed to apply binary section. " +
+                     utils::FormatQnnError(qnn_interface_, context_apply_binary_section_rt))
                         .c_str());
           break;
         }
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -200,7 +200,11 @@ Ort::Status QnnBackendManager::ParseLoraConfig(std::string lora_config_path) {
 
           auto context_apply_binary_section_rt = qnn_interface_.contextApplyBinarySection(
               contexts_[cIdx], graph, QNN_CONTEXT_SECTION_UPDATABLE, &contextBuffer, profile_backend_handle_, nullptr);
-          RETURN_IF(QNN_SUCCESS != context_apply_binary_section_rt, "Failed to apply binary section.");
+          RETURN_IF(QNN_SUCCESS != context_apply_binary_section_rt,
+                    ("Failed to apply binary section. Error: " +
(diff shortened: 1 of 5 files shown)


qti-yuduo (approved):

qti-yuduo:
@qti-chuteng could you sign CLA and close out this change? Thanks!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -200,7 +200,11 @@ Ort::Status QnnBackendManager::ParseLoraConfig(std::string lora_config_path) {
 
           auto context_apply_binary_section_rt = qnn_interface_.contextApplyBinarySection(
               contexts_[cIdx], graph, QNN_CONTEXT_SECTION_UPDATABLE, &contextBuffer, profile_backend_handle_, nullptr);
-          RETURN_IF(QNN_SUCCESS != context_apply_binary_section_rt, "Failed to apply binary section.");
+          RETURN_IF(QNN_SUCCESS != context_apply_binary_section_rt,
+                    ("Failed to apply binary section. Error: " +
+                     QnnErrorHandleToString(context_apply_binary_section_rt) +
```
minfhong-qti: Consider define a new function to format both error message and code.

PR #533, opened 2026-06-16
[QNN EP] Harden integer parsing of provider/run options against invalid input

qti-chuteng:
## Summary

User-supplied option strings in the QNN EP were parsed with bare `std::stoi` / `std::stoul` / `std::stoull`. These throw `std::invalid_argument` on non-numeric input and `std::out_of_range` on overflow. Because the calls run during **session creation** and **option handling**, a simple typo such as `vtcm_mb=abc` or `device_id=xyz` crashes the process with an uncaught C++ exception, instead of being reported and gracefully skipped.

This PR routes all such parse sites through a small non-throwing helper and gives each caller a sensible failure policy (warn-and-skip, return an error status, or mark the model not-applicable).

## Why this is needed

Provider options, run options, and dynamic options are external, user-controlled input. Today an invalid value does not produce an actionable diagnostic — it propagates a raw C++ exception out of session init. This is both a robustness gap (a single bad character takes down the session) and a usability gap (no clear message telling the user which option was malformed). The fix aligns these sites with the defensive `try / catch (const std::exception&)` style already used elsewhere in this file (e.g. the RPCMem library load path).

## Approach

A file-scope helper that parses without throwing:

```cpp
static bool TryParseInt64(const std::string& value, int64_t& out) {
  try {
    size_t pos = 0;
    out = std::stoll(value, &pos);
    return pos == value.size();  // reject trailing garbage
  } catch (const std::exception&) {
    return false;                // non-numeric or overflow
  }
}
```

Each caller keeps its own failure handling:

| Site | Input source | On invalid input |
|------|--------------|------------------|
| `rpc_control_latency`, `vtcm_mb`, `device_id`, `soc_model` | provider options | log warning, skip (keep existing default); prior `>=0` / `<=0` range checks preserved |
| `num_graph_prepare_threads` | provider option (Win ARM64) | see bug fix below; overflow now falls back to default |
| `rpc_control_latency` | run option | log warning, skip |
| `kvcache_rewind` | `SetDynamicOptions` | return `ORT_INVALID_ARGUMENT` instead of throwing |
| compiled-model compatibility info | model metadata string | mark `EP_NOT_APPLICABLE` (matches existing format-check behavior) |

### Bug fix: `is_valid_number` (num_graph_prepare_threads)

The validation lambda returned `true` on **non-digit** input (after logging an error), so the invalid string still reached `std::stoi`. Two issues fixed:
- Non-digit input now returns `false` (was `true`).
- Guarded the `s[0]` access against an empty string.

### Note on `htp_graph_finalization_optimization_mode`

The warning for an unrecognized value now states it falls back to the default. The member is already initialized to the default *before* parsing, so runtime behavior is unchanged — this is a message-clarity improvement only.

## Scope / risk

- **No success-path behavior changes.** Valid values parse exactly as before; only the invalid-input paths change (from "throw" to "report and degrade gracefully").
- All range checks that previously existed are retained.
- Out of scope (different concern / internal source, low risk): the SDK-reported version string parse in `qnn_cache_compatibility_manager.cc` and the Genie output-config parse in `genie_node_compute_info.cc`. Can be follow-ups.

## Testing

- `lintrunner` (clang-format) — clean.
- These are mostly error-path guards; a malformed value now produces a warning / error status instead of an uncaught exception. Recommend a quick negative check (e.g. `vtcm_mb=abc`) in a normal build environment to confirm session creation no longer aborts.


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -559,8 +594,14 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                  "0",
                                  rpc_control_latency_str);
   if (!rpc_control_latency_str.empty() && rpc_control_latency_str != "0") {
-    default_rpc_control_latency_ = static_cast<uint32_t>(std::stoul(rpc_control_latency_str));
-    ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, ("rpc_control_latency: " + rpc_control_latency_str).c_str());
+    uint32_t parsed = 0;
+    if (TryParseUint32(rpc_control_latency_str, parsed)) {
+      default_rpc_control_latency_ = parsed;
```
```suggestion
    if (TryParseUint32(rpc_control_latency_str, default_rpc_control_latency_ )) {
```
As well for others.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -607,10 +648,14 @@ QnnEp::QnnEp(QnnEpFactory& factory,
   std::string vtcm_mb_str;
   GetSessionConfigEntryOrDefault(ort_api, session_options_, FormatEPConfigKey("vtcm_mb"), "0", vtcm_mb_str);
   if (!vtcm_mb_str.empty() && vtcm_mb_str != "0") {
-    vtcm_size_in_mb_ = std::stoi(vtcm_mb_str);
-    ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, ("vtcm_mb: " + vtcm_mb_str).c_str());
-    if (vtcm_size_in_mb_ <= 0) {
-      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING, ("Skip invalid vtcm_mb: " + vtcm_mb_str).c_str());
+    int64_t parsed = 0;
+    if (TryParseInt64(vtcm_mb_str, parsed) && parsed > 0 &&
```
minfhong-qti: Why not defined another parse function for `int32_t`?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -607,10 +648,14 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+        parsed <= static_cast<int64_t>(std::numeric_limits<int32_t>::max())) {
+      vtcm_size_in_mb_ = static_cast<int>(parsed);
```
```suggestion
      vtcm_size_in_mb_ = static_cast<int32_t>(parsed);
```

minfhong-qti (commented):
Since this PR is fixing how options are parsed, maybe it's a good time considering move all option parsing into a separate file instead of keep growing in QnnEp's constructor.

PR #535, opened 2026-06-16
[QNN EP] Fix HardSigmoidMul fusion missing reversed Mul input ordering

qti-chuteng:
## Summary

The `HardSigmoidMulFusion` pattern matcher recognizes `HardSwish(x) = x * HardSigmoid(x)` and fuses the `HardSigmoid` + `Mul` pair into a single QNN `ElementWiseNeuron` (HardSwish) op. To accept the pair, it verifies that the `HardSigmoid` input is *also* one of the two `Mul` inputs. That check had a copy-paste bug — both sides of the `||` compared `Mul.Inputs()[0]`:

```cpp
const bool same_root_input = mul->Inputs()[0].name == hs_input ||
                             mul->Inputs()[0].name == hs_input;  // [0] on both sides
```

So only the **first** Mul input was ever inspected. When a graph uses `Mul(hardsigmoid_output, root_input)` — HardSigmoid output as `Inputs()[0]`, the root as `Inputs()[1]` — the matcher fails to recognize the root, and the pattern is silently left **unfused**, degrading to separate `HardSigmoid` + `Mul` ops on the backend.

## Fix

Change the second comparison to `Mul.Inputs()[1]` so both input orderings — `Mul(root, hsig_out)` and `Mul(hsig_out, root)` — fuse correctly.

## Test

Adds `qnn_node_group/hardsigmoid_mul_fusion_test.cc` with two cases (normal and reversed Mul input ordering). Each dumps the QNN JSON graph and asserts the fusion actually happened: the standalone `ElementWiseMultiply` (Mul) is gone (`count = 0`) and a single `ElementWiseNeuron` (HardSwish) remains (`count = 1`).

### Why this wasn't already caught

Existing accuracy tests (`HardSigmoidFusedIntoHardSwish_FP32_as_FP16`, `HardSigmoidFusedIntoHardSwish_FP16`) already use the reversed ordering `Mul(hsig_out, input)`, but they only assert `ExpectedEPNodeAssignment::All`. That passes whether or not the fusion happens — both ops are still assigned to the QNN EP either way — so the missed fusion went undetected. The new test asserts the graph topology, not just EP assignment.

## Scope / risk

- One-character logic fix (`[0]` → `[1]`) plus a new test file.
- The fix only *adds* matches that were previously (incorrectly) rejected; it cannot cause a non-HardSwish pattern to fuse, because the rest of the matcher (op types, single child, alpha/beta values, no graph-output) is unchanged.

## Testing

- `lintrunner` — clean.
- The new tests are HTP-backend fusion tests (guarded to arm64 / linux, skipped on HTP arch <= V68). Recommend running them on an HTP-capable environment to confirm both orderings fuse; under TDD the reversed-ordering case fails before this fix and passes after.


qti-yuduo (approved):

onnxruntime/test/providers/qnn/qnn_node_group/hardsigmoid_mul_fusion_test.cc
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Should existing testcases moved inside?

onnxruntime/test/providers/qnn/qnn_node_group/hardsigmoid_mul_fusion_test.cc
```diff
@@ -0,0 +1,110 @@
...
+
+  AssertOpInQnnGraph(json_qnn_graph_dir, "ElementWiseMultiply", /*count=*/0);
+  AssertOpInQnnGraph(json_qnn_graph_dir, "ElementWiseNeuron", /*count=*/1);
+}
+
+}  // namespace
+
+// HardSigmoid -> Mul(input, hsig_out): HardSigmoid output is the SECOND Mul input.
+// This is the ordering the original same_root_input check already handled.
+TEST_F(QnnHTPBackendTests, HardSigmoidMulFusion_NormalOrder_Fuses) {
```
minfhong-qti, line 87: CI failed for all `HardSigmoidMulFusion` testcases in accuracy.

PR #539, opened 2026-06-16
[QNN EP] Add graph-structure tests for DQQ and CastLoneQ fusions

qti-chuteng:
## Summary

Adds **graph-structure** unit tests for the two Q/DQ-to-`Convert` fusions in the QNN EP — `DQQFusion` and `CastLoneQFusion`. Both fusions previously had only accuracy / EP-assignment coverage; nothing asserted that the fused `Convert` actually appears in the compiled QNN graph. These tests close that gap.

No production code changes — test-only.

## Why

Both fusions rewrite a small Q/DQ pattern into a single `QNN_OP_CONVERT`:

- **`DQQFusion`**: a standalone `DequantizeLinear -> QuantizeLinear` (same scale type, scalar constant qparams) collapses into one `Convert`.
- **`CastLoneQFusion`**: a standalone `Cast` whose parent is **not** a `DequantizeLinear` and whose only child is a `QuantizeLinear` collapses into one `Convert`.

Standalone `Quantize` / `Dequantize` / `Cast` all have QNN op builders, so whether or not the fusion fires the nodes still land on the QNN EP (`ExpectedEPNodeAssignment::All`). EP assignment therefore cannot distinguish "fused" from "not fused" — the only reliable signal is the compiled-graph JSON. These tests dump the QNN graph (`dump_json_qnn_graph`) and assert on the op types via `AssertOpInQnnGraph`.

## Coverage

`onnxruntime/test/providers/qnn/qnn_node_group/dq_q_fusion_test.cc`

| Test | Asserts |
|------|---------|
| `DQQFusion_U8_Convert` | uint8 requantize at output → `Convert` count 1 |
| `DQQFusion_U16_Convert` | uint16 requantize at output → `Convert` count 1 |
| `DQQFusion_ExtraConsumer_NoFusion` | DQ has a 2nd consumer → no fusion, `Convert` count 0 |

`onnxruntime/test/providers/qnn/qnn_node_group/cast_lone_q_fusion_test.cc`

| Test | Asserts |
|------|---------|
| `CastLoneQFusion_U8_Convert` | `Cast -> Q` → `Convert` count 1, `Cast` count 0 |
| `CastLoneQFusion_NoQChild_NoFusion` | `Cast -> Add` (no Q child) → `Convert` count 0, `Cast` count 1 |

## Implementation notes

- Test harness (`ResetQnnGraphDir` / `HasQnnJsonGraph` / `GetHTPProviderOptions` / `RunFusionTest`) mirrors the existing `dql_dq_fusion_test.cc`.
- Model topologies reuse patterns already proven in `simple_op_test.cc` (`BuildDQQConvertAtOutputTestCase`, including the `scale *= 1.01f` nudge so the ORT QDQ optimizer does not fold the `DQ -> Q` pair away before the QNN EP sees it) and `qnn_basic_test.cc` (the `Cast -> Q -> DQ -> Add` topology).
- Platform guards (`#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)`) and `SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V68)` match sibling fusion tests.
- New files are picked up automatically by the `providers/qnn/qnn_node_group/*` CMake glob — no CMake change needed.

## Testing

- `lintrunner` — clean.
- These are HTP graph-dump tests gated behind the arm64/linux platform guard; run them with `--gtest_filter="QnnHTPBackendTests.DQQFusion*:QnnHTPBackendTests.CastLoneQFusion*"` on an HTP-capable environment (arm64 device or x86 HTP emulator).

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_node_group/cast_lone_q_fusion_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_node_group/cast_lone_q_fusion_test.cc
@@ -1,0 +1,211 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#if !defined(ORT_MINIMAL_BUILD)
+
+#include <filesystem>
+#include <string>
+#include <vector>
+
+#include "test/providers/qnn/qnn_node_group/qnn_graph_checker.h"
+#include "test/providers/qnn/qnn_test_utils.h"
+#include "test/unittest_util/qdq_test_utils.h"
+#include "gtest/gtest.h"
+
+namespace onnxruntime {
+namespace test {
+
+#if defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
+
+namespace {
+
+// ---------------------------------------------------------------------------
+// Helper: JSON graph directory setup / teardown
+// ---------------------------------------------------------------------------
+void ResetQnnGraphDir(const std::filesystem::path& dir) {
+  std::filesystem::remove_all(dir);
+  ASSERT_TRUE(std::filesystem::create_directory(dir));
+}
+
+// Returns true if at least one QNN JSON graph file exists in `dump_dir`.
+// Used to skip graph assertions when the test was not executed (e.g., FP32
+// HTP not available and no JSON dump is produced).
+bool HasQnnJsonGraph(const std::filesystem::path& dump_dir) {
+  if (!std::filesystem::exists(dump_dir)) return false;
+  for (const auto& entry : std::filesystem::directory_iterator{dump_dir}) {
+    if (entry.is_regular_file() && entry.path().extension() == ".json" &&
+        entry.path().filename().string().find("_tensor_log") == std::string::npos) {
+      return true;
+    }
+  }
+  return false;
+}
+
+ProviderOptions GetHTPProviderOptions() {
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = "htp";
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/qnn_node_group/dq_q_fusion_test.cc
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// ==========================================================================
+// No-fusion test — guard condition not met; no Convert produced
+// ==========================================================================
+
+// The DQ feeding the final Q has a second consumer (the Add branch), so the
+// DQ->Q pair is not a lone sequence and DQQFusion must be rejected. No Convert
+// must appear in the compiled graph.
+TEST_F(QnnHTPBackendTests, DQQFusion_ExtraConsumer_NoFusion) {
```
minfhong-qti, line 226: CI failed in this testcase.

onnxruntime/test/providers/qnn/qnn_node_group/cast_lone_q_fusion_test.cc
```diff
@@ -0,0 +1,211 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: According to the description,
```
Both fusions previously had only accuracy / EP-assignment coverage; nothing asserted that the fused Convert actually appears in the compiled QNN graph.
```
Should existing testcases as well move into these new test files?

PR #540, opened 2026-06-16
[QNN EP] Add op builder & fusion developer guidelines; fix cross-pass naming and validate-gate conformance

tirupath-qti:
## Description

Adds two developer-facing guideline documents (`docs/qnn_op_builder_guidelines.md` and `docs/qnn_node_group_fusion_guidelines.md`) that codify the full lifecycle of implementing an op builder or node-group fusion in the QNN EP. Alongside the documentation, applies the rules the guidelines describe to existing implementations that violated them: deterministic naming for cross-pass intermediate tensors, the mandatory `ValidateOnQnn` gate in `TryFusion`, and several minor null-safety and `do_op_validation` threading fixes.

---

### Why documentation

The QNN EP's op-builder and fusion machinery has accumulated a number of subtle contracts that
are not visible from the code alone. Examples found during review:

- Whether an intermediate tensor name must be stable across the two execution passes
  (GetCapability and Compile)
- When `ValidateOnQnn` in `TryFusion` is mandatory vs. optional
- The difference between `UniqueNameGenerator` (stateful global counter) and a deterministic
  `node_unit.Name() + suffix` rename
- When to override `IsOpSupported` vs. relying on the QNN validation API

Without written guidance these rules are rediscovered by each contributor from bug reports. The two new documents provide a single authoritative reference.

### Cross-pass tensor naming (`UniqueNameGenerator` misuse)

`utils::UniqueNameGenerator` is a **process-wide stateful counter**. Using it for intermediate tensor names that appear in *both* the GetCapability (Phase 1) and Compile (Phase 2) passes is wrong: the counter has advanced between the two phases, so Phase 2 generates a different name, creating a dangling reference instead of reusing the Phase 1 tensor. The fix for every affected callsite is to replace `UniqueNameGenerator().New(x, "_suffix")` with `x + "_suffix"` (or
`node_unit.Name() + "_suffix"`) for tensors that are referenced across passes. Node *names* (not tensor names) may still use `UniqueNameGenerator` because they are never cross-referenced.

Affected builders: `cast`, `gemm`, `resize`, `scatternd`, `simple_op` (DQ-chain and fold
intermediates), `topk`.

### Missing `ValidateOnQnn` gate in `TryFusion` (`reshape_gemm`, `reshape_einsum_reshape`)

The IQnnNodeGroup contract requires `TryFusion` to call `ValidateOnQnn` as its **last gate** before returning a group object. If a structurally valid pattern is claimed in Phase 1 but the backend can't actually compile it, Phase 2 `AddToModelBuilder` returns an error that **aborts the entire graph compilation** — there is no per-fusion fallback at that stage.

Both `ReshapeGemmFusionGroup::TryFusion2/3/4` and `ReshapeEinsumReshapeNodeGroup::TryFusion` were returning the group without this validation step.

**`reshape_gemm_fusion`** fix: construct the group tentatively, call `group->CreateOrValidateOnQnn(..., validate=true)`, and return `nullptr` on failure. The `ORT_UNUSED_PARAMETER(logger)` suppressions are removed because logger is now forwarded.

**`reshape_einsum_reshape`** fix: `CreateOrValidateOnQnn` previously had no true validate branch — the entire function built and registered tensors in the `QnnModelWrapper`, making it unsafe to call from `TryFusion` (which must not mutate the wrapper). A new validate-only code path is added (`do_op_validation=true`) that constructs stack-allocated `QnnTensorWrapper` objects and calls `ValidateQnnNode` directly without touching the wrapper's state. The create
path is restructured to share the pre-computed shapes.

### `qnn_node_group.cc` — add `Gemm` to QDQ-group exception list

`TryQnnFusions` skips NodeUnits that are `QDQGroup` type unless the op is on a hardcoded exception list (`Gather`, `MatMul`, `Erf`, `Reshape`). LPBQ Gemm can begin from a `QDQGroup` starting unit; without `Gemm` in the exception list `LowPowerBlockQuantizedGemmFusion::TryFusion` is never reached. Added `Gemm` to the list.

### `matmulnbits_op_builder.cc` — null guards for constant-tensor lookups

`GetConstantTensor` can return `nullptr` if the input is not a constant initializer (e.g. a dynamic weight). Three callsites in the LPBQ path were dereferencing the result without checking, which would crash on non-constant inputs. Added `RETURN_IF_NOT(ptr != nullptr, msg)` before each dereference to produce a clean fallback-to-CPU error instead.

### `simple_op_builder.cc` — thread `do_op_validation` into `AddCastNode`

`ProcessVariadicToBinaryChain` was passing literal `false` as the validate flag to `AddCastNode` for the int64-output cast, meaning the cast node was always materialized during the validation pass. Changed to forward `do_op_validation`.

---

## Files Changed

| File | Change |
|------|--------|
| `docs/qnn_op_builder_guidelines.md` | New: 200-line op builder developer guide covering setup, hooks, factory registration, decomposition, QDQ, advanced patterns, and common pitfalls. |
| `docs/qnn_node_group_fusion_guidelines.md` | New: 210-line node-group fusion developer guide covering the `IQnnNodeGroup` interface, discovery/registration, pattern matching, emission, validation idiom, and conformance pitfalls. |
| `cast_op_builder.cc` | Replace `UniqueNameGenerator` with deterministic names for `_notequal_zero`, `_signabs_sign`, `_signabs_abs` intermediates. |
| `gemm_op_builder.cc` | Replace `UniqueNameGenerator` with deterministic names for `_transpose`, `_fp16`, and `_fc` intermediates in both the float and BQ code paths. |
| `matmulnbits_op_builder.cc` | Add `RETURN_IF_NOT(ptr != nullptr)` null guards before dereferencing weight, scales, and zero-point constant-tensor lookups. |
| `resize_op_builder.cc` | Replace `UniqueNameGenerator` with deterministic name for the `_tfhp_resize_out` intermediate tensor. |
| `scatternd_op_builder.cc` | Replace `UniqueNameGenerator` with deterministic `_qnn_idx` rename for index tensor normalization. |
| `simple_op_builder.cc` | Deterministic names for DQ chain (`_to_f32`) and fold intermediates (`_fold_N`); thread `do_op_validation` into the int64-output `AddCastNode` call. |
| `topk_op_builder.cc` | Deterministic names for `_transpose` and `_cast` intermediates. |
| `qnn_node_group.cc` | Add `Gemm` to the QDQ-group exception list in `TryQnnFusions`. |
| `reshape_gemm_fusion.cc` | Add `ValidateOnQnn` gate in all three `TryFusion` variants (2/3/4); remove `ORT_UNUSED_PARAMETER(logger)`. |
| `reshape_einsum_reshape.cc` | Refactor `CreateOrValidateOnQnn` to add a proper validate-only branch; add `ValidateOnQnn` gate in `TryFusion`. |


qti-chuteng (approved):

docs/qnn_op_builder_guidelines.md
```diff
@@ -0,0 +1,208 @@
...
+- `tile_op_builder.cc` — inputs + params + qparam override (best starting point).
+- `argmax_argmin_op_builder.cc` — attributes → params.
+- `rmsnormalization_op_builder.cc` — one builder reused across equivalent ops.
+- `layernormalization_op_builder.cc` / `softmax_op_builder.cc` — decomposition into multiple QNN nodes.
+- `conv_op_builder.cc` / `pool_op_builder.cc` — layout transforms + per-channel quant.
+- `batchnormalization_op_builder.cc` — per-backend datatype allow-lists + `GetDQNodes()` reach-through.
+
+## Related Tooling
+
+- `/qnn_ep_op_builder_codegen <manifest>` — scaffold a builder from a YAML manifest, or `--from-onnx <OpName>` to generate a draft manifest first.
```
tirupath-qti: remove this line

docs/qnn_op_builder_guidelines.md
```diff
@@ -0,0 +1,208 @@
...
+## 10. Layout (NCHW → NHWC)
+
+- QNN spatial ops (Conv, Pool, etc.) are NHWC-native; ONNX is NCHW.
+- Layout-sensitive builders must branch on the domain in `IsOpSupported`: only call QNN validation when `node_unit.Domain() == kMSInternalNHWCDomain`; for the pre-transform form do manual checks and return OK.
+- After transform, channel is the **last** dimension.
+- Reorder ONNX pads to QNN order with `ReArrangePads` — forgetting this gives silently wrong results.
+
+## 10.5. The `do_op_validation` Pass
+
+- These two executions map to the EP's two phases: `do_op_validation=true` runs during **Phase 1 / GetCapability** (via `QnnNodeUnitWrapper::IsSupported` → `IsOpSupported`), and `do_op_validation=false` runs during **Phase 2 / Compile** (via `QnnNodeUnitWrapper::AddToModelBuilder`, [qnn_node_group.cc:61-66](../onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc#L61-L66)). This is the same Phase 1 / Phase 2 lifecycle the fusion doc describes ([qnn_node_group_fusion_guidelines.md](qnn_node_group_fusion_guidelines.md) §1).
```
tirupath-qti: remove reference to fusion guideline here.

docs/qnn_node_group_fusion_guidelines.md
```diff
@@ -0,0 +1,213 @@
...
+
+## 2. Discovery / Registration
+
+The registry is a static map from **starting op type** → an ordered list of fusion factory functions ([qnn_node_group.cc:82-104](../onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc#L82-L104)):
+
+```cpp
+static std::unordered_map<std::string, std::vector<FusionFunc>> fusions = {
+    {"DequantizeLinear", {DQQFusion::TryFusion}},
+    {"Gemm", {LowPowerBlockQuantizedGemmFusion::TryFusion, ReshapeGemmFusionGroup::TryFusion4,
+              ReshapeGemmFusionGroup::TryFusion3, ReshapeGemmFusionGroup::TryFusion2}},
```
huaychou, line 47: **[M-2] (Major)** Hard-coded line-number references in documentation will become stale

Both guide documents contain extensive hard-coded line-number references, for example:

```
[qnn_node_group.cc:82-104]
[base_op_builder.h:131-232]
[qnn_node_group.cc:133-160]
```

These line-number citations may be accurate when the document is written, but they become stale whenever the referenced source code is modified, reorganized, or has lines inserted. Readers following these links will land at incorrect locations.

**Impact**: Documentation trustworthiness degrades over time. Developers who follow the line-number references will see different content than intended, increasing learning overhead and risking misunderstanding. After a few rounds of code changes, most of these references will be wrong.

**Suggested fix:**

Use one or both of the following approaches: (1) Replace line-number links with GitHub permalinks in the format `https://github.com/onnxruntime/onnxruntime-qnn/blob/{SHA}/path/to/file.cc#L82-L104`, using the PR's headRefOid as the SHA (accurate at merge time at least); (2) Better: extract the referenced logic into well-named functions or constants and reference the function name in documentation (e.g., "see the implementation of `TryQnnFusions()`") rather than a line number. The latter is immune to line drift.

onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_gemm_fusion.cc
```diff
@@ -521,8 +525,12 @@ std::unique_ptr<IQnnNodeGroup> ReshapeGemmFusionGroup::TryFusion4(
...
   }
 
-  return std::make_unique<ReshapeGemmFusionGroup>(
+  auto group = std::make_unique<ReshapeGemmFusionGroup>(
       std::vector<const OrtNodeUnit*>{input_reshape, &gemm_node_unit, output_reshape1, output_reshape2});
+  if (!group->CreateOrValidateOnQnn(qnn_model_wrapper, logger, /*validate=*/true).IsOK()) {
+    return nullptr;
+  }
+  return group;
 }
```
huaychou, line 559: **[N-1] (Minor)** reshape_gemm_fusion construct-then-validate pattern cannot soft-fail if constructor throws

The new pattern in `TryFusion2/3/4` constructs `ReshapeGemmFusionGroup` first, then calls `group->CreateOrValidateOnQnn(..., validate=true)`:

```
auto group = std::make_unique<ReshapeGemmFusionGroup>(
    std::vector<const OrtNodeUnit*>{input_reshape, &gemm_node_unit});
if (!group->CreateOrValidateOnQnn(...).IsOK()) {
    return nullptr;
}
```

The fusion guidelines state: "Some constructors `ORT_CXX_API_THROW` on wrong NodeUnit count. `TryFusion` must fail soft (`nullptr`)". If `ReshapeGemmFusionGroup`'s constructor throws in any edge case, `std::make_unique` will propagate the exception rather than returning `nullptr`.

By contrast, the `reshape_einsum_reshape.cc` fix validates first and constructs after, entirely avoiding this concern.

**Impact**: Low risk in practice (each TryFusion variant passes a fixed and correct node count), but if `ReshapeGemmFusionGroup`'s constructor logic is extended in the future (e.g., to validate node types), it could cause TryFusion to throw where it should soft-fail, breaking the dispatcher's graceful fallback behavior.

**Suggested fix:**

If `ReshapeGemmFusionGroup`'s constructor is currently noexcept, add a `noexcept` annotation to make the assumption explicit for future maintainers. A more robust approach is to mirror the `reshape_einsum_reshape.cc` pattern: perform validation inside `TryFusion` first, then construct the group only after validation passes.

docs/qnn_node_group_fusion_guidelines.md
```diff
@@ -0,0 +1,213 @@
...
+6. **Ordering among fusions sharing a start op.** More specific patterns must precede general ones — first non-null wins.
+7. **Validation/creation IODef drift.** When the pattern is sandwiched in DQ/Q, validate against the *outer* (quantized) tensors but emit against the *inner* (float) tensors — keep two IODef sets like GELU.
+8. **Graph-output / multi-consumer leaks.** Don't hand-roll edge walks; the helpers reject nodes that produce graph outputs or have >1 consumer. Fusing a node whose intermediate output is consumed elsewhere corrupts the graph.
+9. **Constructor throwing.** Some constructors `ORT_CXX_API_THROW` on wrong NodeUnit count. `TryFusion` must fail soft (`nullptr`), so only construct with the exact expected count.
+
+## Files Worth Reading Before You Start
```
huaychou, line 206: **[T-2] (Nit)** Fusion guideline §5 Pitfall #5 labels reshape_gemm_fusion as non-conforming, but this PR fixes it

`docs/qnn_node_group_fusion_guidelines.md` §5 Pitfall #5 states:

```
"Skipping the validate gate in TryFusion. ... Validating *only* in IsSupported (e.g. reshape_gemm_fusion, reshape_einsum_reshape) is non-conforming"
```

However, this PR itself fixes the validate-before-claim issue in both `reshape_gemm_fusion` and `reshape_einsum_reshape`, so they are no longer non-conforming examples. The document and the code are now inconsistent.

**Suggested fix:**

Update Pitfall #5 to remove `reshape_gemm_fusion` and `reshape_einsum_reshape` as non-conforming examples, or replace them with positive examples (e.g., "dq_q_fusion, channel_shuffle_fusion").

docs/qnn_op_builder_guidelines.md
qti-mbadnara: Done!

docs/qnn_op_builder_guidelines.md
qti-mbadnara: Done!

docs/qnn_node_group_fusion_guidelines.md
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_gemm_fusion.cc
qti-mbadnara: Done!

docs/qnn_node_group_fusion_guidelines.md
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
```diff
@@ -91,7 +91,7 @@ Ort::Status CastOpBuilder::ProcessExtraInputForNotEqual(QnnModelWrapper& qnn_mod
   }
 
   // Build additional static input with value 0.
-  const std::string& input_name = utils::UniqueNameGenerator().New(node_unit, "_notequal_zero");
+  const std::string input_name = node_unit.Name() + "_notequal_zero";
```
huaychou: **[N-1] (Minor)** Deterministic naming drops the OpType+Index fallback for empty node_unit.Name(), risking name collisions for unnamed nodes

The removed `UniqueNameGeneratorImpl::New(const OrtNodeUnit&, suffix)` overload contained a fallback: when `node_unit.Name()` is empty it uses `OpType() + std::to_string(Index())` as the base (see qnn_utils.cc lines 43-49). Switching to a bare `node_unit.Name() + "_suffix"` drops that fallback.

Sites that now use `node_unit.Name()` directly as a tensor-name base: cast's `_notequal_zero`/`_signabs_sign`/`_signabs_abs` (lines 94, 196, 197), resize's `_tfhp_resize_out` (line 186), simple's `_fold_i` (line 350). The ONNX spec permits empty or non-unique node names; if two nodes of the same type both have empty names, the concatenation yields identical strings (e.g. both become `"_notequal_zero"`).

**Impact:** `AddTensorWrapper` then sees a duplicate name and returns true (dedup, no overwrite), silently aliasing two nodes' intermediate tensors onto the same wrapper and producing an incorrect QNN graph. Model nodes usually carry unique names in practice, hence Minor — but the old code did defend against this case.

**Suggested fix:**

Replace the bare node_unit.Name() with a small helper (e.g. utils::GetNodeName(node_unit) returning Name() when non-empty, else OpType()+std::to_string(Index())) to preserve the old overload's empty-name fallback; or at minimum document the "node name assumed non-empty" precondition at these sites.

docs/qnn_op_builder_guidelines.md
```diff
@@ -0,0 +1,204 @@
...
+- **Don't hand-roll outputs** — call the base `ProcessOutputs(...)`. It handles graph-output detection, int64/uint64→int32/uint32 cast insertion, the qparam-override callback, and node emission. Note: base `ProcessAttributesAndOutputs` **early-returns OK if `input_names` is empty** ([base_op_builder.cc](../onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc)) — a node that processes zero inputs silently emits no QNN node.
+
+## 8. Decomposition — One ONNX Op → Multiple QNN Nodes
+
+- **When needed:** QNN has no single equivalent op, or its equivalent has shape/axis/dtype constraints the ONNX op doesn't (e.g. `LayerNormalization` → LayerNorm → Mul → Add when scale/bias must be externalized: [layernormalization_op_builder.cc](../onnxruntime/core/providers/qnn/builder/opbuilder/layernormalization_op_builder.cc)). Note: "decomposition" here is the op-builder fan-*out* (1 ONNX op → many QNN nodes); the fusion doc uses "decomposition" for the opposite — the multi-node ONNX *pattern* a fusion collapses back into one QNN op. Don't confuse the two.
+- Name intermediates with `utils::UniqueNameGenerator().New(node_unit, "_suffix")` so they never collide.
```
huaychou: **[N-2] (Minor)** New doc §8 still recommends UniqueNameGenerator for decomposition intermediate tensors, conflicting with this PR's fix and the same doc's own footgun note

§8 Decomposition line 79 says "Name intermediates with `utils::UniqueNameGenerator().New(node_unit, "_suffix")` so they never collide.", yet this PR's core change is precisely to move op-builder intermediate tensors off UniqueNameGenerator onto deterministic naming; the same doc's footgun later (lines 143, 161) explicitly warns the generator is "stateful/global" and "emits a different name on the second (Compile) pass".

**Impact:** A reader following line 79 for decomposition intermediates walks straight into the cross-pass naming trap this PR fixes. Doc-internal consistency issue only; no runtime effect.

**Suggested fix:**

Reword line 79 to recommend deterministic naming (tensor_name/node_unit.Name() + suffix) for decomposition intermediates, or at minimum add a caveat cross-referencing the lines 143/161 footgun, clearly distinguishing node names (created unconditionally each pass; generator acceptable) from intermediate-tensor names (must be stable across passes; use deterministic naming).

huaychou (commented):
Just wondering, does using `UniqueNameGenerator` cause any issues, or does it just make the tensor names weird?

qti-mbadnara:
> Just wondering, does using `UniqueNameGenerator` cause any issues, or does it just make the tensor names weird?

Thanks for catching this, so when GetCapability and Compile run as separate sessions, the counter has advanced and Phase 2 generates a different name for the same tensor, silently corrupting the graph.

Replaced UniqueNameGenerator with deterministic node_unit.Name() + "_suffix" for intermediate tensor names, and added utils::NodeUnitBaseName() to preserve the empty-name fallback (OpType()+Index()) from the original overload

onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
qti-mbadnara: Added utils::NodeUnitBaseName() in qnn_utils.h that returns Name() when non-empty, else OpType()+std::to_string(Index()), preserving the original fallback.

docs/qnn_op_builder_guidelines.md
qti-mbadnara: Updated §8 to recommend utils::NodeUnitBaseName(node_unit) + "_suffix" for intermediate tensor names, with a note that UniqueNameGenerator is acceptable for QNN node names only. Also added a caveat that existing builders pre-dating this guideline still use UniqueNameGenerator for tensor names and are a known cleanup gap. Will raise a follow up PR to address it

PR #542, opened 2026-06-17
[QNN EP] Add MaxRoiPool support via StridedSlice/ReduceMax/Concat decomposition

qti-niscmami:
## Description

  Adds QNN EP support for the ONNX `MaxRoiPool` operator. Because QNN has no `MaxRoiPool`/`RoiPooling` kernel that runs on the NPU, the op is **decomposed** into primitives that are supported on all backends (CPU, HTP, GPU): `StridedSlice` + `ReduceMax` per pooled bin, assembled with `Concat`/`Reshape`.

  ### What it does

  For each ROI, the adaptive bin geometry is computed at build time from the (constant) `rois` tensor, following the ONNX spec (`hstart = y1 + floor(i·roi_h/ph)`, `hend = y1 + ceil((i+1)·roi_h/ph)`, and likewise for width). Each bin becomes:

  - `StridedSlice` of the exact bin region from the NHWC feature map → `ReduceMax` over H,W (keepdims) → `[1,1,1,C]`
  - Empty/degenerate bins emit a static zero tensor (matching ONNX, which fills them with 0)

  Per ROI the bin results are concatenated and reshaped to `[1, ph, pw, C]`; the per-ROI tiles are concatenated along the batch axis to form the `[num_rois, ph, pw, C]` output. The decomposition is exact for any constant ROI (handles overlapping / non-uniform bins and `spatial_scale`).

  ### Constraints

  - `rois` must be a constant initializer (bin geometry is resolved at build time); a non-constant `rois` is rejected in `IsOpSupported` and falls back to the CPU EP.
  - An op-count guard (`num_rois · pooled_h · pooled_w ≤ 4096`) bounds graph size; larger configs fall back to the CPU EP.


qti-chuteng (approved):

onnxruntime/test/providers/qnn/maxroipool_test.cc
```diff
@@ -0,0 +1,290 @@
...
+                                                        -9.0f, -10.0f, -11.0f, -12.0f, -13.0f, -14.0f, -15.0f, -16.0f}),
+                                   TestInputDef<float>({1, 5}, true, {0.0f, 0.0f, 3.0f, 1.0f, 6.0f}),
+                                   {test::MakeAttribute("pooled_shape", std::vector<int64_t>{2, 2}),
+                                    test::MakeAttribute("spatial_scale", 1.0f)},
+                                   ExpectedEPNodeAssignment::All,
+                                   /*opset=*/13,
+                                   /*use_contrib_qdq=*/true);
+}
+
+#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
```
qti-mbadnara, line 320: Can you add unit tests for the following QDQ Int16 Cases MultiRoi, MultiImage and spatial_scale similar to U8 tests?

onnxruntime/core/providers/qnn/builder/opbuilder/maxroipool_op_builder.cc
```diff
@@ -0,0 +1,452 @@
...
+  const uint32_t pooled_w = out_info.shape[2];
+
+  const Qnn_DataType_t dtype = out_info.qnn_data_type;
+  const float spatial_scale = node_helper.Get("spatial_scale", 1.0f);
+
+  // Read the constant rois corners (input-image space).
+  std::vector<float> rois_flat;
+  RETURN_IF_ERROR(ReadRoisAsFloat(qnn_model_wrapper, inputs[1], num_rois, rois_flat));
+
+  const std::string name_base = node_unit.Name().empty()
```
qti-mbadnara: I believe this might cause some overlap when two nodes share the same name? Something to think about

onnxruntime/core/providers/qnn/builder/opbuilder/maxroipool_op_builder.cc
```diff
@@ -0,0 +1,452 @@
...
+    const uint32_t bh = hend - hstart;
+    const uint32_t bw = wend - wstart;
+    std::vector<uint32_t> slice_shape{1u, bh, bw, channels};
+    QnnTensorWrapper slice_tensor(slice_out, QNN_TENSOR_TYPE_NATIVE, x_info.qnn_data_type,
+                                  x_info.quant_param.Copy(), std::vector<uint32_t>(slice_shape));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(slice_tensor)),
+                  "Failed to add MaxRoiPool slice tensor.");
+
+    std::vector<uint32_t> ranges_dims{4u, 3u};
+    std::vector<uint32_t> ranges_data{
```
qti-mbadnara, line 298: I think QNN uses int32 rather than unint32. Please check once. If that is the case, use static_cast to change it back to uint32

onnxruntime/core/providers/qnn/builder/opbuilder/maxroipool_op_builder.cc
```diff
@@ -0,0 +1,452 @@
...
+    int32_t offset = 0;
+    RETURN_IF_ERROR(rois_info.quant_param.GetPerTensorScaleOffset(scale, offset));
+    const bool is_signed = (rois_info.qnn_data_type == QNN_DATATYPE_SFIXED_POINT_8);
+    for (size_t i = 0; i < num_elems; ++i) {
+      const double q = is_signed ? static_cast<double>(static_cast<int8_t>(rois_bytes[i]))
+                                 : static_cast<double>(rois_bytes[i]);
+      rois_flat[i] = static_cast<float>(utils::Dequantize(offset, scale, q));
+    }
+  } else {
+    return MAKE_EP_FAIL("MaxRoiPool only supports float32 or 8-bit quantized rois.");
```
qti-mbadnara: This contradicts your tests?

Please correct me if I am wrong, doesnt the QDQ tests use FIXED_POINT_16 which will fall back to CPU EP

Can you add a 16 bit QDQ branch simillar to the 8 bit one? Or modify the tests

qti-niscmami: Added test cases.

onnxruntime/test/providers/qnn/maxroipool_test.cc
qti-niscmami: Added

onnxruntime/core/providers/qnn/builder/opbuilder/maxroipool_op_builder.cc
qti-niscmami: The ranges tensor is already emitted as signed int32 via the is_signed=true flag on QnnParamWrapper. No static_cast is needed.

PR #543, opened 2026-06-18
[QNN EP] Refactor ProcessAxisAttribute to be op-def-driven

huaychou:
### Description

- Replace ad-hoc ProcessAxisAttribute overloads with a general design
- Replace qnn_axis_scalar with AddQnnScalar

### Motivation and Context

- The follow up of https://github.com/onnxruntime/onnxruntime-qnn/pull/525
- Update ProcessAxisAttribute behavior to Replace qnn_axis_scalar with AddQnnScalar

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
@@ -403,30 +403,21 @@ Ort::Status BaseOpBuilder::SetOutputQParamEqualToInputIfNearlyEqual(QnnModelWrap
 
 Ort::Status BaseOpBuilder::ProcessAxisAttribute(const QnnModelWrapper& qnn_model_wrapper,
                                                 const OrtNodeUnit& node_unit,
-                                                Qnn_Scalar_t& axis_qnn_scalar,
-                                                int32_t& default_axis_value) const {
+                                                const std::string& attr_name,
+                                                int32_t default_axis,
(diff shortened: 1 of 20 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
```diff
@@ -403,30 +403,21 @@ Ort::Status BaseOpBuilder::SetOutputQParamEqualToInputIfNearlyEqual(QnnModelWrap
 
 Ort::Status BaseOpBuilder::ProcessAxisAttribute(const QnnModelWrapper& qnn_model_wrapper,
```
minfhong-qti: In my imagination, `ProcessXXX` prefix function should build up QNN object (`Qnn_Scalar_t` in this case). Just curious why not design like this.
If it is intentional, I prefer to rename to `GetCanonicalizedAxisAttribute` or others without `Process` prefix.

huaychou: We follow the original behavior of ProcessAxisAttribute and make it more general to reduce code change. In addition, decoupling it from AddQnnScalar can increase the flexibility.

huaychou: Done

minfhong-qti (approved):

qti-chuteng (approved):

PR #544, opened 2026-06-18
[QNN EP] Fix offload_graph_io_quantization tensor name binding on cached-context load path

qti-mbadnara:
### Description

Fixes a crash when running a QNN EP context binary (EPContext model) compiled with `offload_graph_io_quantization=1`. The EP now serializes the internal→external tensor-name rename map into the EPContext node as a new `io_name_overrides` attribute at generate time, and populates it on the cached-context load path so graph I/O is resolved by name rather than
by position. Building on top of #263 

## Motivation and Context

When `offload_graph_io_quantization=1`, the QNN EP moves boundary Q/DQ nodes outside the QNN partition. To keep names aligned with the original ONNX graph, it renames the QNN boundary tensor via an internal `tensor_name_overrides_` map built by the op builders during `Compile`.

This map existed only in memory and was never written into the EPContext model. On the JIT path the map is live and `SetGraphInputOutputInfo` resolves graph I/O by name correctly, which is why running the source `.onnx` directly works. On the cached-context load path the op builders never run, the map is `nullptr`, and name resolution is disabled.

PR #263 added a positional fallback to compensate, but it pairs QNN tensors with fused-node outputs by index. For multi-output models where QNN does not preserve output ordering this silently causes the size assert to fire

The fix closes the serialization gap at its source. The rename map is now written into the EPContext node at generate time and read back on load, making the cached-context path behaviorally equivalent to the JIT path. Backward compatibility is preserved: context binaries generated before this fix continue to load via the positional fallback.


onnxruntime/core/providers/qnn/builder/onnx_ctx_model_helper.cc
```diff
@@ -116,6 +116,38 @@ Ort::Status GetMainContextNode(const OrtGraph** graphs,
...
 }
 
+std::unordered_map<std::string, std::string> ParseIoNameOverrides(const OrtNode* ep_context_node) {
+  std::unordered_map<std::string, std::string> overrides;
+  if (ep_context_node == nullptr) {
+    return overrides;
+  }
+  OrtNodeAttrHelper node_helper(*ep_context_node);
+  const std::string encoded = node_helper.Get(IO_NAME_OVERRIDES, std::string{});
+  // Decode "internal=external;" pairs.
```
tirupath-qti, line 126: we can't depend on selected delimiters = and ; as there is a risk of having these special characters in the tensor names of any future model. We can address this risk in a follow-up.

tirupath-qti (approved):

PR #547, opened 2026-06-18
[QNN EP] Add per-partition DLC debug bundle

qti-niscmami:
### Description

  Adds a per-partition DLC debug bundle for sharing reproducible QNN graphs for further debugging.

  **New session options**

  | Option | Type | Description |
  |---|---|---|
  | `qnn.dump_partition_dlc_bundle` | bool (default `false`) | Master switch. |
  | `qnn.partition_dlc_bundle_dir`  | path | Output dir. Required when the flag is on. |

  **What gets written**

  ```
  <dir>/
  ├── manifest.json             # ordered partitions, I/O (name/dtype/shape), edges
  └── <partition>/              # one folder per partition (helper consolidates here)
      ├── <partition>.dlc
      ├── inputs/<name>.raw      # filled by the host helper
      └── goldens/<name>.raw     # filled by the host helper
  ```

  The C++ side (compile-time) emits the DLCs and `manifest.json`. A Python helper fills each partition's `inputs/` and `goldens/`:

  ```bash
  python onnxruntime/python/tools/qnn/generate_partition_dlc_bundle.py \
      --bundle-dir /tmp/bundle \
      --model model.onnx \
      --inputs input0=input0.raw
  ```

  The helper marks every boundary tensor as a graph output of a copy of the original ONNX, runs once on CPU EP, and writes raw little-endian `.raw` files in the format `qnn-net-run` consumes via `--input_list`.

  **Constraints**

  - Forces IR-backend serialization regardless of `backend_type`. Same as the existing `dump_qnn_ir_dlc`: a session with the flag on cannot execute the model — use a separate session for inference.
  - No-op for AOT Phase 2 (loading an EPContext model) and Hybrid AOT — `ComposeGraph()` is not invoked in those paths.

  ### Motivation and Context

  When an "ORT-on-HTP gives wrong numbers" report needs to be reproduced, a self-contained QNN graph plus inputs and reference outputs are needed. A model ORT splits into N partitions cannot be reproduced from the ONNX alone:

  - Only partition 0's inputs are model-level. Every later partition's inputs are activations from the previous partition.
  - Without per-partition boundary tensors, partitions 2..N are not independently runnable.
  - A single end-to-end golden cannot point at which partition introduced divergence.

  This bundle produces N runnable DLCs with their own inputs and goldens, captured automatically at compile time.


onnxruntime/python/tools/qnn/partition_dlc_bundle/generate_partition_dlc_bundle.py
```diff
@@ -0,0 +1,206 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Prefer to create its own directory to avoid this directory getting messy later.

onnxruntime/test/python/test_generate_partition_dlc_bundle.py
```diff
@@ -0,0 +1,111 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Test file can be placed under `test/python`.

onnxruntime/python/tools/qnn/partition_dlc_bundle/generate_partition_dlc_bundle.py
qti-niscmami: Done

onnxruntime/test/python/test_generate_partition_dlc_bundle.py
qti-niscmami: Done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2242,6 +2280,41 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
     }
 
     RETURN_IF_NOT_OK(qnn_model->ComposeGraph(context));
+
+    if (ep->dump_partition_dlc_bundle_) {
```
minfhong-qti: I think below logic worths defining a new .h/.cc for them instead of making QnnEp harder to maintain.

qti-niscmami: Done

qti-niscmami: Extracted to `onnxruntime/core/providers/qnn/partition_dlc_bundle.{h,cc}` (`RecordPartitionBundle` per-partition + `WritePartitionBundleManifest` at end of Compile). `QnnEp` now just calls these two and holds the records vector.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -258,6 +258,10 @@ class QnnEp : public OrtEp, public ApiPtrs {
   // from a single thread today, so plain size_t is sufficient.
   size_t dump_qnn_ep_input_graph_count_ = 0;
 
+  bool dump_partition_dlc_bundle_ = false;
+  std::string partition_dlc_bundle_dir_ = "";
+  std::vector<qnn::PartitionBundleRecord> partition_bundle_records_;
```
qti-kromero, line 263: nit: partition_bundle_dlc_records_ might be slightly more consistent with rest of naming in PR

onnxruntime/python/tools/qnn/partition_dlc_bundle/generate_partition_dlc_bundle.py
```diff
@@ -0,0 +1,206 @@
...
+# SPDX-License-Identifier: MIT
+
+"""Fill a per-partition DLC debug bundle with inputs and CPU-EP goldens.
```
qti-kromero, line 4: the naming of this script seems a little bit off cause you are saying it's generating a partition dlc bundle but I thought the partition dlc bundle was created by the EP based on the naming in the EP code

I think a better name for this script might be "populate_partition_dlc_bundle.py" because you are taking a partition DLC bundle created by ORT EP and adding the golden outputs to it

onnxruntime/python/tools/qnn/partition_dlc_bundle/generate_partition_dlc_bundle.py
```diff
@@ -0,0 +1,206 @@
...
+
+Usage:
+    python generate_partition_dlc_bundle.py \\
+        --bundle-dir /tmp/bundle \\
+        --model model.onnx \\
+        --inputs name1=path1.raw [name2=path2.raw ...]
```
qti-kromero, line 10: where do these inputs come from if not provided? is there a means of auto-generating some inputs of the correct dimensions if this is not provided?

onnxruntime/python/tools/qnn/partition_dlc_bundle/generate_partition_dlc_bundle.py
```diff
@@ -0,0 +1,206 @@
...
+
+Reads <bundle-dir>/manifest.json (produced at compile time by the QNN EP when
+qnn.dump_partition_dlc_bundle=1), marks every boundary tensor as a graph output
+of a copy of the ONNX model, runs once on the CPU EP with the user-supplied
+inputs, then consolidates each partition into <bundle-dir>/<partition>/ holding
+its <partition>.dlc, inputs/<name>.raw and goldens/<name>.raw. Files are raw
+little-endian tensor data (no header), which is the format qnn-net-run consumes
```
qti-kromero, line 17: what is the use case of this script with regards to accuracy comparisons? trying to think about how we could expand this or better serve the use case cause this script is still missing half the picture right? we'd want to have something to compare against, either from qnn-net-run or HTPExecutionProvider in ORT

onnxruntime/python/tools/qnn/partition_dlc_bundle/generate_partition_dlc_bundle.py
```diff
@@ -0,0 +1,206 @@
...
+            sys.exit(f"--inputs expects name=path, got {spec!r}")
+        name, path = spec.split("=", 1)
+        user_inputs[name] = Path(path)
+
+    boundary_names = collect_boundary_tensors(manifest)
+    free_dims = {}
+    for spec in args.free_dim:
+        name, _, value = spec.partition("=")
+        free_dims[name] = int(value)
+    modified_model, alias = add_boundary_outputs(args.model, boundary_names, free_dims)
```
qti-kromero, line 137: this modified_model might leak if something below throws an exception, may be well placed in a context manager or try/finally

onnxruntime/test/python/test_generate_partition_dlc_bundle.py
qti-kromero: does this contain any package namespacing today? or is it just a flat directory

onnxruntime/test/python/test_generate_partition_dlc_bundle.py
```diff
@@ -0,0 +1,114 @@
...
+        for p in ("qnn_0", "qnn_1"):
+            assert (bundle / p / f"{p}.dlc").is_file(), f"{p}.dlc not consolidated"
+        assert not (bundle / "partitions").exists(), "empty partitions/ should be removed"
+        manifest_after = json.loads((bundle / "manifest.json").read_text())
+        for p in manifest_after["partitions"]:
+            assert p["dlc_path"] == f"{p['name']}/{p['name']}.dlc", p["dlc_path"]
+        assert manifest_after["goldens_source"] == "cpu"
+
+
+if __name__ == "__main__":
```
qti-kromero, line 112: I would generally prefer to move to a model where we just rely on pytest to run tests in python files

PR #484, opened 2026-06-04
[QNN EP] Doc changes for 2.3.0 release

qti-mbadnara:
As per title

PR #381, opened 2026-05-14
[QNN EP] Doc changes for 2.2.0 Release

qti-mbadnara:
This PR has the documentation changes for the `2.2.0` release.

PR #549, opened 2026-06-19
[QNN EP] Minor CI pipeline improvements

qti-mbadnara:
This PR makes minor changes in the CI pipeline

- Remove dependency on JFROG_API_KEY, need to maintain two separate keys for essentially the same function. Could be performed by ARTIFACTORY_PASSWORD
- Add the title and content for the release notes when index_server_to is set as GitHub for Zip and Tgz
- Prevent creating multiple drafts of the release notes when Zip and Tgz jobs are triggered one after another
- Upgrade actions v8
- Fix naming convention of test_package.zip created on each nightly and RC

### DIFF
--- a/.github/workflows/qualcomm-internal-ci-trigger.yml
+++ b/.github/workflows/qualcomm-internal-ci-trigger.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ["self-hosted", "X64", "Linux", "Build"]
     steps:
       - name: Dispatch CI workflow on PR branch
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             // Fetch the PR to get its head branch and draft status.
--- a/qcom/scripts/upleveling/qnn_ep_uplevel.py
+++ b/qcom/scripts/upleveling/qnn_ep_uplevel.py
@@ -322,8 +322,9 @@ class ArtifactUpleveler(ABC):
 
     def _download_signed_libs(self, target_dir: str) -> str:
         """Download <format>.zip from artifactory into target_dir; return its path."""
-        api_key = os.environ.get("ARTIFACTORY_PASSWORD", "")
-        if not api_key:
+        artifactory_user = os.environ.get("ARTIFACTORY_USERNAME", "")
+        artifactory_password = os.environ.get("ARTIFACTORY_PASSWORD", "")
+        if not artifactory_password:
             raise RuntimeError("ARTIFACTORY_PASSWORD environment variable is required when --sign_artifact true")
 
         version_url = self.config_manager.get_repository_url(
@@ -334,7 +335,9 @@ class ArtifactUpleveler(ABC):
         target_path = os.path.join(target_dir, zip_filename)
 
         logging.info(f"Downloading signed libs ({zip_filename}) for version {self.args.version_from}")
-        response = requests.get(url, auth=("", api_key), verify=ARTIFACTORY_CERTS_FILE, timeout=60)
+        response = requests.get(
+            url, auth=(artifactory_user, artifactory_password), verify=ARTIFACTORY_CERTS_FILE, timeout=60
+        )
         if response.status_code != 200:
             raise RuntimeError(f"Failed to download signed libs: HTTP {response.status_code}")
 
--- a/qcom/scripts/upleveling/qnn_ep_uplevel.py
+++ b/qcom/scripts/upleveling/qnn_ep_uplevel.py
@@ -318,9 +318,9 @@ class ArtifactUpleveler(ABC):
 
     def _download_signed_libs(self, target_dir: str) -> str:
         """Download <format>.zip from artifactory into target_dir; return its path."""
-        api_key = os.environ.get("JFROG_API_KEY", "")
+        api_key = os.environ.get("ARTIFACTORY_PASSWORD", "")
         if not api_key:
-            raise RuntimeError("JFROG_API_KEY environment variable is required when --sign_artifact true")
+            raise RuntimeError("ARTIFACTORY_PASSWORD environment variable is required when --sign_artifact true")
 
         version_url = self.config_manager.get_repository_url(
             _SIGNED_LIBS_INDEX, self.args.product_name, self._signed_libs_version
@@ -423,8 +423,6 @@ class WheelUpleveler(ArtifactUpleveler):
       ARTIFACTORY_USERNAME / ARTIFACTORY_PASSWORD  — Artifactory basic auth (download + upload)
       PYPI_API_KEY                                 — PyPI upload token (when index_server_to=pypi)
       TEST_PYPI_API_KEY                            — TestPyPI upload token (when index_server_to=testpypi)
-      JFROG_API_KEY                                — Read-only token for the signed-libs bundle
-                                                     (only when --sign_artifact true)
     """
 
     @property
@@ -662,8 +660,6 @@ class NugetUpleveler(ArtifactUpleveler):
       ARTIFACTORY_USERNAME / ARTIFACTORY_PASSWORD  — Artifactory basic auth (download + upload)
       NUGET_API_KEY                                — nuget.org API key (when index_server_to=nuget)
       TEST_NUGET_API_KEY                           — int.nugettest.org API key (when index_server_to=testnuget)
-      JFROG_API_KEY                                — Read-only token for the signed-libs bundle
-                                                     (only when --sign_artifact true)
     """
 
     def __init__(self, args: argparse.Namespace):
@@ -938,8 +934,6 @@ class ZipUpleveler(ArtifactUpleveler):
 
     Credentials (never in argv):
       ARTIFACTORY_USERNAME / ARTIFACTORY_PASSWORD  — Artifactory basic auth (download + upload)
-      JFROG_API_KEY                                — Read-only token for the signed-libs bundle
-                                                     (only when --sign_artifact true)
     """
 
     @property
@@ -1077,6 +1071,37 @@ class ZipUpleveler(ArtifactUpleveler):
 
             logging.info(f"Version update completed for {zip_file}, updated to {updated_zip_path}")
 
+    @staticmethod
+    def _read_release_notes() -> tuple:
+        """Return (title, body) from the first section of docs/release-notes.md.
+
+        title — text of the first H1 line (without the leading '# ').
+        body  — all lines after that H1 up to (not including) the next H1.
+        Both are empty strings if the file is missing or has no H1.
+        """
+        notes_path = os.path.join(SCRIPT_DIR, "..", "..", "..", "docs", "release-notes.md")
+        try:
+            with open(notes_path, encoding="utf-8") as f:
+                lines = f.readlines()
+        except OSError:
+            return "", ""
+
+        title = ""
+        body_lines = []
+        in_section = False
+        for line in lines:
+            stripped = line.rstrip("\n")
+            if stripped.startswith("# "):
+                if not in_section:
+                    title = stripped[2:].strip()
+                    in_section = True
+                else:
+                    break
+            elif in_section:
+                body_lines.append(stripped)
+
+        return title, "\n".join(body_lines).strip()
+
     def _upload_to_github(self, distribution_dir: str) -> None:
         """Create a git tag + GitHub Release tagged v{version_to} and attach the artifacts."""
         tag = f"v{self.args.version_to}"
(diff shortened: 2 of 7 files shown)


.github/workflows/qualcomm-internal-ci-trigger.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     runs-on: ["self-hosted", "X64", "Linux", "Build"]
     steps:
       - name: Dispatch CI workflow on PR branch
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
```
kuanyul-qti, line 37: Just curious, why do we need to upgrade to v8?

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -1120,23 +1145,30 @@ def _upload_to_github(self, distribution_dir: str) -> None:
...
-        if create_result.returncode == 0:
-            logging.info(f"Created draft GitHub Release {tag}")
+        if release_exists:
+            logging.info(f"GitHub Release {tag} already exists, reusing it")
         else:
-            stderr = create_result.stderr.decode(errors="replace")
-            if "already exists" in stderr.lower():
-                logging.info(f"GitHub Release {tag} already exists, reusing it")
+            release_title, release_notes = self._read_release_notes()
+            release_title = release_title or tag
```
kuanyul-qti: The `tag` is "v{version}". But I believe the release title should be "ONNX Runtime QNN Execution Provider v{version}". Something like that:

```suggestion
            title_with_tag = "ONNX Runtime QNN Execution Provider" + tag
            release_title = release_title or title_with_tag
```

kuanyul-qti: **[M-1] (Major)** _read_release_notes() does not validate that release-notes.md's title version matches the tag — risks publishing a GitHub Release whose title and tag disagree

`_upload_to_github` uses the first H1 of `release-notes.md` (e.g. `# ONNX Runtime QNN Execution Provider v2.3.0`) as the GitHub Release title, but never checks that it belongs to the same version as `tag = f"v{self.args.version_to}"`:

```
release_title, release_notes = self._read_release_notes()
release_title = release_title or tag
create_result = subprocess.run(
    ["gh", "release", "create", tag,
     "--title", release_title,
     "--notes", release_notes, "--draft"],
    ...)
```

The `git log` for `docs/release-notes.md` (PR #484, #381, #373, …) shows the file is updated by hand alongside each release. If the maintainer forgets to bump it before cutting the next tag, this code will silently produce:

- tag `v2.4.0`
- title `ONNX Runtime QNN Execution Provider v2.3.0`
- body containing v2.3.0 content
And because the draft release was "successfully created", subsequent re-runs hit the `gh release view` early-exit and the mismatch is never re-checked.

**Impact**: a public-facing GitHub Release with the wrong version banner; downstream consumers (Maven, wheel index links, blog references) inherit the misleading title. Recovering requires manually deleting the release and re-tagging — harder to undo than "I forgot to update release-notes.md" alone. This is the single point in the release flow that is easiest for a human to miss.

**Suggested fix:**

After parsing the title in `_read_release_notes()`, add a simple version-consistency check:

```
release_title, release_notes = self._read_release_notes()
if release_title and self.args.version_to not in release_title:
    raise RuntimeError(
        f"docs/release-notes.md top H1 ({release_title!r}) does not mention "
        f"--version_to={self.args.version_to}; refusing to publish a tag/title-mismatch release. "
        f"Bump docs/release-notes.md before re-running.")
```

Keeps the empty-file fallback (drop back to `tag` when parse fails), and bakes the "don't forget to update release-notes.md" rule into the script. A stricter version can match `vMAJOR.MINOR.PATCH` with a regex.

.github/workflows/qualcomm-internal-ci-trigger.yml
kuanyul-qti, line 37: **[N-1] (Minor)** actions/github-script@v8 requires Actions Runner v2.327.1+ / Node 24 — self-hosted runner upgrade status should be confirmed in the PR description

All four workflows bump `actions/github-script` from `@v7` to `@v8`. The v8 release notes call out:

- script runtime upgraded from Node 20 to Node 24
- minimum Actions Runner version is now `v2.327.1`
This repo runs entirely on self-hosted runners (labels include `"self-hosted", "X64", "Linux", "Build"`, etc.) — unlike GitHub-hosted runners, they do not auto-upgrade. The PR description does not mention whether the runner pool has been bumped to ≥ v2.327.1.

**Impact**: any self-hosted runner still on an older release will hard-fail the `Dispatch CI workflow on PR branch` / `Verify admin permission` steps that depend on `github-script`, breaking the whole PR-trigger mechanism.

**Suggested fix:**

Add a line to the PR description naming the runner-pool version (or link to a CI run on the upgraded pool proving the new bits run cleanly). If the runner upgrade is staggered: (1) trial-run `qualcomm-internal-ci-trigger.yml` first (the earliest trigger point), verify the whole pool runs green, then merge the v7→v8 bump across all four workflows; (2) if the runner upgrade is not yet complete, revert the four v7→v8 changes and re-submit the bump in its own PR after the upgrade lands.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -1120,23 +1145,30 @@ def _upload_to_github(self, distribution_dir: str) -> None:
...
-        # `gh release create` exits non-zero both for "already exists" (expected on re-runs and
-        # cross-format runs) and for real failures (auth, repo not found, …). Inspect stderr to
-        # tell them apart so genuine errors don't surface as a confusing upload failure later.
-        create_result = subprocess.run(
-            ["gh", "release", "create", tag, "--title", tag, "--notes", "", "--draft"],
-            check=False,
-            capture_output=True,
+        # Create the draft GitHub Release only if it does not already exist.
+        # Check with `gh release view` first so that parallel zip/tgz jobs never
+        # race on `gh release create` and accidentally produce two draft releases.
```
kuanyul-qti: **[N-2] (Minor)** New _upload_to_github comment overstates the gh release view race guarantee — there's still a TOCTOU window; the actual safety comes from the 'already exists' fallback

The new comment reads:

```
# Create the draft GitHub Release only if it does not already exist.
# Check with `gh release view` first so that parallel zip/tgz jobs never
# race on `gh release create` and accidentally produce two draft releases.
```

However, `gh release view → gh release create` is still TOCTOU: two parallel jobs can both `view`, both miss, and both `create`. **What actually prevents duplicate creation is the existing fallback:**

```
if create_result.returncode == 0:
    logging.info(f"Created draft GitHub Release {tag}")
else:
    stderr = create_result.stderr.decode(errors="replace")
    if "already exists" in stderr.lower():
        logging.info(f"GitHub Release {tag} already exists, reusing it")
```

The race-loser's `create` will fail with "already exists", and the fallback rescues it. The comment's "never race" framing risks misleading a future maintainer into removing the fallback under the impression that the `view` probe is sufficient on its own.

**Impact**: documentation accuracy only — no runtime defect. But future readers may draw the wrong conclusion.

**Suggested fix:**

Soften the comment to acknowledge the fallback's role, e.g.:

```
# Create the draft GitHub Release only if it does not already exist.
# `gh release view` is a fast-path: when zip/tgz jobs run sequentially the second
# one short-circuits without invoking `gh release create`. The view+create pair
# is still TOCTOU under true parallelism — the `"already exists"` fallback below
# is what actually guarantees no duplicate drafts; keep it intact.
```

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -1077,6 +1071,37 @@ def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir:
 
             logging.info(f"Version update completed for {zip_file}, updated to {updated_zip_path}")
 
+    @staticmethod
+    def _read_release_notes() -> tuple:
```
kuanyul-qti: **[T-1] (Nit)** _read_release_notes() return annotation should be tuple[str, str] to match the rest of the file

The new helper annotates `-> tuple` (un-parameterised):

```
def _read_release_notes() -> tuple:
```

The existing helpers in the same file use parameterised annotations:

```
def get_credentials(repository_index: str) -> tuple[str, str]:
def _get_credentials(self, repository_index: str) -> tuple[str, str]:
```

**Suggested fix:**

Change to `-> tuple[str, str]`.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -1077,6 +1071,37 @@ def update_artifacts(self, artifact_list: list[str], input_dir: str, output_dir:
...
 
+    @staticmethod
+    def _read_release_notes() -> tuple:
+        """Return (title, body) from the first section of docs/release-notes.md.
+
+        title — text of the first H1 line (without the leading '# ').
+        body  — all lines after that H1 up to (not including) the next H1.
+        Both are empty strings if the file is missing or has no H1.
+        """
+        notes_path = os.path.join(SCRIPT_DIR, "..", "..", "..", "docs", "release-notes.md")
```
kuanyul-qti: **[T-2] (Nit)** Repo-root path is computed by '../../..' string walk-up — pathlib + parents[2] is harder to silently break

The new helper uses:

```
notes_path = os.path.join(SCRIPT_DIR, "..", "..", "..", "docs", "release-notes.md")
```

If `qcom/scripts/upleveling/` is moved one level deeper or shallower in the future (e.g. `qcom/scripts/release/upleveling/`), the `../../..` count silently breaks — no type-check or unit-test catches it. The release-notes parse will return empty, the title will quietly fall back to `tag`, and nobody notices until a release is cut.

**Suggested fix:**

Use `pathlib` or hoist the repo root into a module-level constant:

```
from pathlib import Path
notes_path = Path(SCRIPT_DIR).resolve().parents[2] / "docs" / "release-notes.md"
```

The file currently uses `os.path` consistently — you can keep that style but factor `parents[2]` into a single `_REPO_ROOT` constant so future directory moves touch one line.

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -318,9 +318,9 @@ def _setup_signing_dirs(self) -> None:
 
     def _download_signed_libs(self, target_dir: str) -> str:
         """Download <format>.zip from artifactory into target_dir; return its path."""
-        api_key = os.environ.get("JFROG_API_KEY", "")
+        api_key = os.environ.get("ARTIFACTORY_PASSWORD", "")
```
huaychou: Nit: Rename `api_key` to `artifactory_password` (or `password`) to match the updated error message and env name.

.github/workflows/qualcomm-internal-ci-trigger.yml
qti-mbadnara: Check if user is admin uses Node.js 20 which is deprecated. 

We need to upgrade to @v8 which uses Node.js 24

qti-mbadnara: I guess CI ran for this PR with the version bump from v7 to v8.

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Fixed with the latest commit

qti-mbadnara: Fixed with the latest commit

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Done!

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Done!

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Done!

kuanyul-qti (approved):

qcom/scripts/upleveling/qnn_ep_uplevel.py
huaychou: Do we need to specify the ARTIFACTORY_USERNAME after replacing JFROG_API_KEY with ARTIFACTORY_PASSWORD?

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Done!

qcom/scripts/upleveling/qnn_ep_uplevel.py
qti-mbadnara: Good catch, done!

qcom/scripts/upleveling/qnn_ep_uplevel.py
```diff
@@ -318,9 +322,10 @@ def _setup_signing_dirs(self) -> None:
 
     def _download_signed_libs(self, target_dir: str) -> str:
         """Download <format>.zip from artifactory into target_dir; return its path."""
-        api_key = os.environ.get("JFROG_API_KEY", "")
-        if not api_key:
-            raise RuntimeError("JFROG_API_KEY environment variable is required when --sign_artifact true")
+        artifactory_user = os.environ.get("ARTIFACTORY_USERNAME", "")
+        artifactory_password = os.environ.get("ARTIFACTORY_PASSWORD", "")
```
huaychou, line 326: nit: can it be replaced by self._get_credentials(self.args.index_server_from)?

huaychou (approved):
LGTM. Remember to test on release workflow internally.

qti-ashwshan (approved):
LGTM

PR #550, opened 2026-06-19
[QNN EP] Use prebuilt ORT version and skip source fetch

tirupath-qti:
Previously the build always derived ORT_CORE_VER from the pinned deps.txt
URL and always fetched+patched the ort_core source, even when a prebuilt
ORT was supplied via --ort-prebuilt (onnxruntime_ORT_HOME). This forced the
prebuilt to be the exact pinned version (1.26.0); any other version failed.

In prebuilt mode:
- Read ORT_CORE_VER from <onnxruntime_ORT_HOME>/VERSION_NUMBER instead of the
  deps.txt URL (FATAL_ERROR if the file is missing).
- Skip onnxruntime_fetchcontent_declare/FetchContent_Populate(ort_core) and
  the ort_core patches; point ExternalProject_Add at an empty stub SOURCE_DIR
  since its build/install steps are no-ops there.
- Skip the onnxruntime_provider_test testdata copy and the Java sigmoid.ort
  copy, which both read from the now-unfetched ort_core source.

Source-build mode (no --ort-prebuilt) is unchanged: version is parsed from
the URL and ort_core is fetched and patched as before.


tirupath-qti:
Validated below scenarios:
1) Prebuilt 1.24.4 (--ort-prebuilt)
2) Source build (no prebuilt)
3) Prebuilt 1.26.0 (--ort-prebuilt)

qti-mbadnara (approved):

cmake/onnxruntime_unittests.cmake
```diff
@@ -366,12 +366,19 @@ block()
...
 
-  add_custom_command(
-    TARGET onnxruntime_provider_test POST_BUILD
-    COMMAND ${CMAKE_COMMAND} -E copy_directory
-    ${ort_core_SOURCE_DIR}/onnxruntime/test/testdata
-    $<TARGET_FILE_DIR:onnxruntime_provider_test>/testdata
-  )
+  # Copy upstream ORT test data next to the test binary. The data lives in the fetched ort_core
+  # source, which is not available in prebuilt mode (onnxruntime_ORT_HOME set), so skip the copy
+  # there. Tests needing upstream models must have that data provided separately in prebuilt builds.
```
qti-mbadnara, line 370: I think this will cause silent test failures, if we migrate the CI to use the prebuilt as the default option. If not LGTM!

cmake/external/onnxruntime_external_deps.cmake
```diff
@@ -470,14 +470,33 @@ else()
...
-  message(STATUS "Extracted ORT_CORE_VER: ${ORT_CORE_VER}")
+# Determine the ORT Core version and (for source builds) fetch the ORT Core source.
+#
+# When a prebuilt ORT is supplied (onnxruntime_ORT_HOME set), the build must use ONLY that
+# prebuilt: its version comes from the prebuilt's VERSION_NUMBER file, and the upstream ORT
+# source is NOT fetched or patched. Otherwise (source build) the version is parsed from the
+# pinned deps.txt URL and the ort_core source is fetched and patched as before.
+if(onnxruntime_ORT_HOME)
+  # Pin 1: read ORT_CORE_VER from the prebuilt instead of from the deps.txt URL.
+  set(_ort_version_file "${onnxruntime_ORT_HOME}/VERSION_NUMBER")
```
kuanyul-qti: Since the `VERSION_NUMBER` file records the QNN EP version, how could it represent the ORT core version?

kuanyul-qti (commented):
Will CI use the prebuilt method, or will it still fetch and build?

cmake/onnxruntime_unittests.cmake
qti-chuteng: I agree with you.
If a user wants to build their own ORT, we shouldn’t assume that the existing test data will be placed in somewhere already.

cmake/external/onnxruntime_external_deps.cmake
```diff
@@ -470,14 +470,33 @@ else()
...
-  message(FATAL_ERROR "Could not extract version from DEP_URL_ort_core: ${DEP_URL_ort_core}")
-endif()
+  # Extract version from ort_core URL
+  # Expected URL format: https://github.com/microsoft/onnxruntime/archive/refs/tags/v1.24.1.zip
+  if(DEP_URL_ort_core MATCHES ".*refs/tags/v([0-9]+\\.[0-9]+\\.[0-9]+)\\.zip$")
+    set(ORT_CORE_VER ${CMAKE_MATCH_1})
+    message(STATUS "Extracted ORT_CORE_VER: ${ORT_CORE_VER}")
+  else()
+    message(FATAL_ERROR "Could not extract version from DEP_URL_ort_core: ${DEP_URL_ort_core}")
+  endif()
```
huaychou, line 499: Missing endif()

cmake/external/onnxruntime_prebuilt.cmake
```diff
@@ -31,6 +31,13 @@ if(onnxruntime_ORT_HOME)
     endif()
     message(STATUS "ORT_PREBUILT_SOURCE: ${ORT_PREBUILT_SOURCE}")
     set(ONNXRUNTIME_APPLICATION_INCLUDES "${onnxruntime_ORT_HOME}/include")
+    # In prebuilt mode the ort_core source is not fetched, so ort_core_SOURCE_DIR/ort_core_BINARY_DIR
+    # (and thus ORT_SOURCE_DIR/ORT_BUILD_DIR set above) are empty. ExternalProject_Add below still
+    # requires a valid, non-empty SOURCE_DIR/BINARY_DIR even though its build/configure/install steps
+    # are no-ops here, so point them at an empty stub directory.
+    set(ORT_SOURCE_DIR "${CMAKE_BINARY_DIR}/ort_core_prebuilt_stub")
+    set(ORT_BUILD_DIR "${CMAKE_BINARY_DIR}/ort_core_prebuilt_stub")
```
huaychou, line 39: Do we need to add a log to notice users that `ORT_SOURCE_DIR` `ORT_BUILD_DIR` are updated?

PR #557, opened 2026-06-22
[QNN EP] NPU detection on Linux

qti-niscmami:
### Description

Register a QNN NPU device on Linux ARM64. Checks `/dev/fastrpc-cdsp*` and advertises an NPU if found. Same fallback we already use for older WoS devices.

### Motivation and Context

After #500, Linux ARM64 users get `No QNN EP devices found`. ORT Core doesn't enumerate the Hexagon on Linux, and #500 removed the CPU placeholder that used to fill in. 

Verified on CI Linux ARM64 runner

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
+++ b/onnxruntime/core/providers/qnn/qnn_provider_factory.cc
@@ -11,6 +11,10 @@
 #include <iostream>
 #include <optional>
 
+#if !defined(_WIN32)
+#include <glob.h>
+#endif
+
 #include "onnxruntime_c_api.h"
 #include "onnxruntime_ep_device_ep_metadata_keys.h"
 #include "QnnCommon.h"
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -185,11 +190,25 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetSupportedDevicesImpl(OrtEpFactory* this
...
 
   if (!has_npu_hw_device && num_ep_devices < max_ep_devices) {
-    if (qnn::soc::GetSocId() != 0) {
-      // If ORT Core does not detect NPU hardware but we recognize the device as WoS (through qnn::soc::GetSocId),
-      // exploit virtual hardware device to create an NPU hardware device for user to select from.
-      // Such case happens for older WoS devices (e.g., Makena) that ORT Core's device discovery logic could not detect
-      // NPU through DXCore.
+    bool synthesize_npu = qnn::soc::GetSocId() != 0;
+
+#if !defined(_WIN32) && defined(__aarch64__)
```
minfhong-qti: Prefer to define a new function in `soc_utils.cc` for below logic and make factory as clean as possible.

minfhong-qti (commented):
I guess it is probably worth revisiting whether we should set that macro for all UTs.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-niscmami: Done

minfhong-qti (approved):

kuanyul-qti (approved):

qti-chuteng (approved):

PR #558, opened 2026-06-22
Fix softmax output enc not in [0, 1] range

yath1:
### Description
This PR honors non-natural encoding populated at the output of softmax layer. 

image

In the above network, the output of softmax is populated with an encoding of [Scale: 0.0078432, zp: 128] which translates to a min/max encodings of [-1.004, 0.996]. This is done by AIMET to satisfy older architectures where HTP only supports symmetric second input for Matmul. 

But Softmax output is mathematically between [0, 1]. When the above encodings generated by AIMET is directly populated at the output of softmax, HTP misbehaves because it expects the range to be [0, 1]. This PR fixes this issue by populating correct encodings at the output of softmax, and then by inserting a convert op to convert to the encodings generated by AIMET.

### Motivation and Context
YoloV10_det w8a8 precision shows an on-target accuracy of close to 0. With this change, the accuracy recovers to 34.257 (AIMET simulated is 34.8). PS: There is still a small gap which needs to be further investigated. YoloV11_det also has the same issue. 

Next step for this PR: Extend the solution to Sigmoid, HardSigmoid etc.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/softmax_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/softmax_op_builder.cc
@@ -39,6 +39,135 @@ constexpr int32_t GetDefaultAxisAttribute(int opset_version) {
   return opset_version < 13 ? 1 : -1;
 }
 
+// Returns true if the Softmax output uses a non-natural quantized encoding that should be
+// decoupled from the consumer's encoding via an explicit Convert.
+//
+// Softmax output is in [0, 1] (strictly non-negative), so its natural quantized encoding is
+// asymmetric with offset (zero-point) 0. However, when the output feeds an op that requires a
+// symmetric encoding on that input (e.g. an HTP MatMul RHS on older HTP archs), aimet assigns the softmax
+// output a symmetric encoding (zero-point != 0). At 8-bit HTP softmax kernel mishandles the symmetric
+// output encoding producing garbage and near-zero accuracy
+//
+// The fix emits Softmax with its natural (zero-point 0) encoding and then a QNN_OP_CONVERT to the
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/softmax_op_builder.cc
```diff
@@ -39,6 +39,135 @@ constexpr int32_t GetDefaultAxisAttribute(int opset_version) {
   return opset_version < 13 ? 1 : -1;
 }
```
qti-yuduo, line 41: We have a fusion pass: `ScaleSoftmaxFusion` that still creates `QNN_OP_SOFTMAX` directly from `softmax_node_unit.Outputs()[0]` in `onnxruntime/core/providers/qnn/builder/qnn_node_group/scale_softmax_fusion.cc:221-244`. That fusion is registered for Mul in qnn_node_group.cc:98, so Mul(scalar) -> Softmax can still feed HTP Softmax the problematic symmetric/non-natural output encoding. Best fix: extract/share the bounded-output split helper so both standalone Softmax and fused ScaleSoftmax use identical output construction.

qti-yuduo: Also make sure we `AssertOpInQnnGraph(dir, "Convert", 1)` to validate the behavior.

yath1: Forgot to mention. I will doing this as part of refactor as well. This needs to be extended to other ops like Sigmoid as well.

yath1: Added asserts for `convert` count

qti-yuduo (approved):

PR #561, opened 2026-06-23
Tanh gelu fusion and test case

nagesh-qti:
### Description

Added support for Gelu Tanh Fusion in ORT-QNNEP code. This merges many ops to one op increasing model performance. This is similar to QNN converter fusion.    

### Motivation and Context

QNN only has one Gelu op. The tanh-GELU approximation expands as:
x * 0.5 * (1 + Tanh(sqrt(2/π) * (x + 0.044715 * x³)))
which can be written in ONNX ops as:
x -> Pow(3) -> Mul(0.044715) -> Add(x) -> Mul(sqrt(2/π)) -> Tanh -> Add(1) -> Mul(0.5) -> Mul(x) ==>
The starting node for fusion discovery needs to be Tanh.
The total fused node list is 9 nodes: mul_x2, mul_x3, mul_0044715, add_inner, mul_coeff, tanh, add_one, mul_x, mul_half
Added the test case as well.

### DIFF
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -519,6 +519,7 @@ QNN EP recognizes the following multi-op patterns and fuses them into a single Q
 | Pattern | Fused QNN op | Notes |
 |---|---|---|
 | `[Div/Mul(√2)] → Erf → [Mul(0.5) →] Add(1) → Mul → [Mul(0.5)]` | `QNN_OP_GELU` | Matches two common Gelu decomposition variants (ErfAdd and ErfMul). Surrounding DQ nodes are also handled. |
+| `Mul(x,x) → Mul(x²,x) → Mul(0.044715) → Add(x) → Mul(√(2/π)) → Tanh → Add(1) → Mul(x) → Mul(0.5)` | `QNN_OP_GELU` | Tanh-based GELU approximation. SingleNode (non-QDQ) only. The QNN GELU op uses the exact-erf definition; max substitution error is ~4.7e-4 over [-10, 10]. |
 | `HardSigmoid(α=1/6, β=0.5) → Mul` (shared input) | `QNN_OP_ELEMENT_WISE_NEURON` (HardSwish) | Both inputs to Mul must originate from the same source tensor. |
 | `ReduceMean → Sub → Pow(2) → ReduceMean → Add(ε) → Sqrt → Div → Mul(γ) → Add(β)` | `QNN_OP_LAYER_NORM` | Matches the manual LayerNorm decomposition. Gamma and beta must be constants. |
 | `Mul(scalar constant) → Softmax` | `QNN_OP_SOFTMAX` | The scalar multiplier is folded into the beta parameter of QNN's Softmax. |
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc
@@ -29,6 +29,7 @@
 #include "core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.h"
 #include "core/providers/qnn/builder/qnn_node_group/reshape_transpose_rank5.h"
 #include "core/providers/qnn/builder/qnn_node_group/scale_softmax_fusion.h"
+#include "core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.h"
 #include "core/providers/qnn/builder/qnn_node_group/transpose_reshape_transpose_fusion.h"
 #include "core/providers/qnn/builder/qnn_node_group/udo_fusion.h"
 #include "core/providers/qnn/builder/qnn_utils.h"
@@ -98,6 +99,7 @@ static std::unordered_map<std::string, std::vector<FusionFunc>> fusions = {
     {"Mul", {ScaleSoftmaxFusion::TryFusion}},
     {"Cast", {CastLoneQFusion::TryFusion}},
     {"Erf", {GeluFusion::TryFusion}},
+    {"Tanh", {TanhGeluFusion::TryFusion}},
     {"ReduceMean", {LayerNormFusion::TryFusion}},
     {"Einsum", {ReshapeEinsumReshapeNodeGroup::TryFusion}},
     {"Reshape", {SpaceToDepthFusion::TryFusion, Rank6ToRank5Fusion::TryFusion}},
@@ -142,6 +144,7 @@ static std::unique_ptr<IQnnNodeGroup> TryQnnFusions(
       starting_node_unit.OpType() != "Gather" &&
       starting_node_unit.OpType() != "MatMul" &&
       starting_node_unit.OpType() != "Erf" &&
+      starting_node_unit.OpType() != "Tanh" &&
       starting_node_unit.OpType() != "Reshape") {
     return nullptr;
   }
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.cc
@@ -39,6 +39,10 @@ bool HasSingleOutputConsumer(const OrtNodeUnit& node_unit) {
 }
 
 // Returns true if `node_unit` is a standalone (non-QDQ) SingleNode with the given op type.
+// QDQ support is intentionally out of scope for this fusion: in a quantized tanh-GELU model
+// every Mul/Add/Tanh is a QDQGroup, and the DQ/Q unwrap logic needed to fuse across QDQ
+// boundaries (see GeluFusion::TryFusion) adds significant complexity. That work is left as a
+// follow-up; for now only float32/float16 non-QDQ graphs are fused.
 bool IsSingleNode(const OrtNodeUnit* node_unit, std::string_view op_type) {
   return node_unit != nullptr &&
          node_unit->UnitType() == OrtNodeUnit::Type::SingleNode &&
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.cc
@@ -1,11 +1,8 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
 
 #include "core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.h"
 
-#include <algorithm>
-#include <cassert>
-#include <cmath>
 #include <memory>
 #include <string>
 #include <string_view>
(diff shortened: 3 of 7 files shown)


qti-chuteng:
## [M-1] `onnxruntime/core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.cc:L172`
The square-node check uses `&&` instead of `||`:
```cpp
if (sqi[0].name != candidate_root && sqi[1].name != candidate_root) continue;
```
This passes when *either* input equals root, so `Mul(x, w)` (w ≠ x) is wrongly accepted as `x²` and a non-GELU subgraph mis-fuses into a QNN Gelu (numerically wrong). A genuine square needs *both* inputs to equal root.
**Fix:**
```cpp
if (sqi[0].name != candidate_root || sqi[1].name != candidate_root) continue;
```
Add a negative test (square operand not root) asserting `AssertOpInQnnGraph(dir, "Gelu", 0)`.

## [M-2] `onnxruntime/core/providers/qnn/builder/qnn_node_group/tanh_gelu_fusion.cc:L119`
The backward walk uses `GetParentOfInput` / `GetParentOfInputByName`, which (unlike the forward `GetOnlyChildOfOutput`) do **not** enforce that the matched producer's output has a single consumer. This fusion walks 5 backward hops; if any intermediate (`x²`, `x³`, `mul_0044715_out`, `add_inner_out`, `mul_coeff_out`) is also consumed outside the pattern, the fold swallows it and leaves the external consumer with a dangling input (silent CPU drop or graph-build failure). The sibling Erf GeluFusion uses the same helper but only one hop.
**Fix:** add a single-consumer check for each backward-matched intermediate (mirror `GetOnlyChildOfOutput`), and add a negative test where an intermediate feeds an extra op.

## [N-1] `tanh_gelu_fusion.cc:L180`
Root fan-in equality uses raw `name == name` (no alias canonicalization). Per [F02] rule 1, an `Identity(x)` / duplicated alias would silently miss the fusion. Reuse `gelu_fusion.cc`'s `CanonicalizeRootTensorName`/`HasInputWithEquivalentName`.

## [N-2] `tanh_gelu_fusion.cc:L1` (and .h / _test.cc)
New files use the Microsoft copyright header; per [B1]/[C10] new QNN-EP files should use the Qualcomm header + `SPDX-License-Identifier: MIT`. (Repo is currently inconsistent — confirm convention.)

qti-yuduo:
🔴 F1 — Backward‑walked intermediates are deleted without a single‑consumer guard — HIGH, blocking

  tanh_gelu_fusion.cc reaches mul_x2 (x²), mul_x3 (x³), mul_0044715, add_inner, mul_coeff via GetParentOfInput / GetParentOfInputByName. I verified in utils.cc that neither
  backward helper checks the parent's consumer count (only IsGraphOutput + already‑claimed + SingleNode), whereas the forward helper GetOnlyChildOfOutput (utils.cc:425‑430) does
  enforce consumers.size()==1. All 9 matched units are inserted into node_unit_to_qnn_node_group and skipped in the build loop (qnn_node_group.cc:217‑238) — i.e. removed from the
  graph; only the fused Gelu is emitted.

  Consequence: if any backward intermediate feeds a consumer outside the chain, its producer is deleted and that consumer dangles → silent wrong output (no crash, no CPU fallback).

  - Trigger is real, not hypothetical: x*x and x²*x are textbook common subexpressions. ORT's CommonSubexpressionElimination (confirmed present in ort_core-src and active in the
  run log) merges duplicate Mul(x,x) when two GELUs share the same x, producing a shared mul_x2 with fan‑out 2.
  - Mitigations (honest): the root x — the tensor that's legitimately multi‑consumer — is correctly not claimed, so the common case is safe. And the sibling gelu_fusion.cc:163 has
  the same gap (claims div_node_unit backward, unguarded) and shipped.
  - But: the codebase treats fan‑out guards as mandatory — layer_norm_fusion.cc:126‑127 explicitly checks consumers.size()==2 for a legitimately‑shared node. This PR's exposure (5
  backward nodes, two of them CSE‑magnet squares/cubes) is materially larger than gelu's single Div. A pre‑existing gap doesn't license a new, wider one.
  - Fix (~5‑8 lines): after matching each of the 5 backward nodes, verify its output has exactly one consumer before claiming it (mirror GetOnlyChildOfOutput). File a separate
  issue to backport the guard to gelu_fusion.cc.

  🟠 F2 — Square check accepts Mul(x, y) as x²; contradicts its own comment — MEDIUM, fix before merge

  tanh_gelu_fusion.cc:203‑206:
  // sq_candidate = Mul(x,x): both inputs must be candidate_root.
  if (sqi[0].name != candidate_root && sqi[1].name != candidate_root) continue;
  The comment says both inputs must be root; the code rejects only when neither is — so Mul(x, gate) passes as x². If the rest of the pattern (and forward chain) matches, the
  fusion emits Gelu(x) over math that is not GELU → silently wrong, even on a clean single‑consumer graph. Trigger is contrived, but the guard is simply weaker than documented.
  Fix: require both — if (sqi[0].name != candidate_root || sqi[1].name != candidate_root) continue;.

  🟡 F3 — Uses QNN_OP_GELU instead of the codebase's ElementWiseNeuron path — MEDIUM

  Verified: QNN_OP_GELU is a real, param‑less, HTP‑supported op (QnnOpDef.h:367; HtpOpDefSupplement lists fp32/fp16/int8/int16) — so this is not a correctness bug. But every other
  GELU in the provider — including sibling gelu_fusion.cc and simple_op_builder.cc — maps to QNN_OP_ELEMENT_WISE_NEURON + OPERATION_GELU. QNN_OP_GELU is used nowhere else. Real
  concerns: (a) a model with both erf‑GELU and tanh‑GELU emits two different QNN ops for the same math, which may quantize/schedule differently; (b) the standalone op's HTP
  perf/quant behavior is unproven vs the established path; (c) the tests bake in a "Gelu" op name no other compiled graph produces. Recommend: route through ElementWiseNeuron
  (ideally a shared EmitGeluNode helper used by both fusions), or justify QNN_OP_GELU with HTP perf/accuracy data. Update tests to assert "ElementWiseNeuron".

  🟡 F4 — Matcher is more complex than necessary (set‑then‑reset nested loops) — MEDIUM, maintainability

  TryFusion is ~210 lines (over the 50‑line guideline) with nested for i/for j and a manual "Reset — different root" block nulling four variables. It's currently correct but is a
  latent‑bug factory (the next forgotten reset becomes the next F1). Simplify following the sibling's own idiom: small bool TryMatch* helpers returning a result struct with
  single‑assignment‑on‑success, plus a tiny commutative‑input helper (OtherInput(node, const) / BothInputsAre(node, name)) to kill every ci[0]/ci[1] ternary. Roughly halves the
  function and removes the reset dance.

  🟢 F5 — Silent tanh‑approx → exact‑erf substitution — LOW, document only

  Measured max abs diff ≈ 4.7e‑4 over [‑10,10], well inside the test tolerances (2e‑3…6e‑3). This is the standard GELU approximation swap (ONNX Gelu(approximate="tanh") vs erf).
  Non‑blocking; add a one‑line comment at the QNN_OP_GELU call noting the deliberate swap and error magnitude.

  🟢 F6 — Header diagram shows Pow(3); code matches Mul(Mul(x,x),x) — LOW, doc fix

  tanh_gelu_fusion.h diagram is stale/misleading (the .cc inline comment is correct and explicitly notes "no Pow node"). The matcher does not handle a real Pow node. Replace the
  header diagram with the two‑Mul form already in the .cc.

  Minor / simplification

  - Duplicated scalar‑constant reader: IsScalarConstantApprox substantially duplicates ExtractScalarValueFromMul in scale_softmax_fusion.cc. Hoist a std::optional<float>
  GetScalarConstantValue(...) into qnn_node_group/utils.{h,cc} and layer the tolerance check on top. (Keep this PR's fp16 branch — scale_softmax lacks it.)
  - Test boilerplate: 3 positive tests are near‑identical; extract RunTanhGeluFusionTest(name, input_def, abs_err). Coverage gaps: no commutative‑input‑ordering test (the matcher's
  entire for i/for j complexity is untested), no fp16 test (the fp16 branch is dead in CI), only one negative.
  - Copyright header: files say // Copyright (c) Microsoft Corporation.; sibling Qualcomm‑authored fusions (e.g. lpbqmatmul_fusion.cc) use Qualcomm. Align with team convention.
  - Naming/constants: k0044715 → kGeluCubicCoeff; name the bare 1.0f/0.5f literals; document the single tol=1e-3f (≈2% relative for 0.044715). Prefer an anonymous namespace over
  file‑static for consistency with siblings.

  Note on the prompt's "better onnx‑ir/onnxscript API" question: N/A here — this is C++ EP graph‑matching code, not Python; the analogous simplification is reusing the shared
  utils.cc helpers and the ElementWiseNeuron lowering path (F3/F4).

qti-chuteng:
### 🔵 Completeness

**[D-2]** `🟠 Concern` `[Completeness]` — SingleNode-only design silently no-ops on quantized HTP models; the Erf sibling handles QDQ, and this scope cut is undocumented
`tanh_gelu_fusion.cc (IsSingleNode gate + entry point)`

Every pattern node must pass IsSingleNode() (which rejects QDQGroup). In a real quantized tanh-GELU model each Mul/Add/Tanh is a DQ->op->Q QDQGroup, so TryFusion bails on the very first hop and the pattern never fuses — it stays 9 separate quantized ops.System-level impact: the behavior is safe (no crash, no stranding, individual ops still lower), but the PR's headline benefit is not delivered for the model class that actually ships on HTP. The adjacent Erf GeluFusion went to real trouble here: it unwraps the DQ around root and the Q after output (gelu_fusion.cc:425-437, 451-463) to keep the fused Gelu float-in/float-out, and warns on mixed topology (WarnAndFailOnStandaloneQdq). Tanh does none of these. The header states the limitation ("All contained NodeUnits must be SingleNode") but never justifies why quantized is excluded.

**Suggested fix:** Either (a) mirror gelu_fusion's DQ/Q unwrap so quantized tanh-GELU fuses, or (b) explicitly document this as an out-of-scope cut in the PR description and the fusion doc, so it is a decision rather than an accident. If (b), add a one-line comment near the entry check explaining why the quantized path is unsupported for now, and consider the mixed-topology warning gelu_fusion uses so partially-quantized graphs are not a silent miss.

**[D-3]** `🔵 Improvement` `[Completeness]` — The 'Operator decomposition fusions' doc table was not updated with the tanh-GELU row, and the exact-erf substitution is not disclosed
`docs/execution_providers/QNN-ExecutionProvider.md:551`

The doc has a dedicated Operator decomposition fusions table (line 547) with an explicit row for the Erf-based Gelu decomposition at line 551. This PR adds a second Gelu-producing fusion (tanh approximation -> QNN_OP_GELU) and adds no row.System-level impact: the enumerated table is now silently incomplete; and — relevant to ⟦C8.4⟧ — there is no user-facing disclosure that a tanh-approx model's numeric output is being swapped for the exact-erf Gelu. A user cannot tell the tanh path is supported, nor that a function substitution happened.

**Suggested fix:** Add a sibling row, e.g. Mul(x,x) → Mul(x²,x) → Mul(0.044715) → Add(x) → Mul(√(2/π)) → Tanh → Add(1) → Mul(x) → Mul(0.5) -> QNN_OP_GELU, noting SingleNode-only and the exact-erf substitution tolerance already documented at tanh_gelu_fusion.cc:354.

**[D-4]** `🔵 Improvement` `[Completeness]` — Test matrix lacks a QDQ cell, the substitution error is untested in its peak region, and there is no forward-tail stranding test
`tanh_gelu_fusion_test.cc`

Coverage is HTP-float only; three gaps:No QDQ/quantized cell: the exact scope boundary D-2 leans on is never exercised, so a future change that partially enables QDQ has no guardrail.Substitution error untested at peak: every positive test uses input range [-1,1] or [-1.5,1.5] (:767/:775/:783), but the peak divergence between the tanh approximation and exact-erf is around |x|≈2. So the ~4.7e-4 claim at tanh_gelu_fusion.cc:354 and the 6e-3/2e-3 tolerances are never actually stressed.No forward-tail stranding test: the shared-intermediate negative test only stresses the backward side (x² feeds an extra Relu); there is no case where add_one_out or mul_x_out has a second consumer or is a graph output (the branch is provably safe — see design highlights — but unverified).

nagesh-qti:
@qti-chuteng  
Addressed all your comments. 
for [D-2] I selected option b, Will add QDQ support in the new patch along with the tests (both positive and negative).
This patch will allow customer to run there fp16 model.

I will raise a new patch with QDQ support. Creating a JIRA on my name.

qti-yuduo (approved):

PR #566, opened 2026-06-23
[Tests] Add GQA unit tests

qti-mattsinc:
### Description
* Add GQA unit tests

### Motivation and Context
* Test coverage for GQA op builder and backend support


qti-mattsinc:
~~cb7b07ca21270a575c3d00a3c5fd832a7250b05d is the main commit of the PR. This PR depends on #424 and #504~~ (edit: those two PRs are now merged), and also will require the EP upleveling to QAIRT 2.48 for the added unit tests to pass.

onnxruntime/test/providers/qnn/group_query_attention_test.cc
```diff
@@ -0,0 +1,1481 @@
...
+  const GetTestModelFn build_test_case = BuildGQATestCase<T, M>(query_def, key_def, value_def,
+                                                                past_key_def, past_value_def,
+                                                                seqlens_k_def, total_sequence_length_def,
+                                                                cos_cache_def, sin_cache_def,
+                                                                position_ids_def, attention_bias_def, head_sink_def,
+                                                                do_rotary, k_quant_type, kv_cache_bit_width, kv_num_heads,
+                                                                local_window_size, num_heads, qk_output, rotary_interleaved,
+                                                                scale, smooth_softmax, v_quant_type);
+  // The GPU backend only supports GQA with past/present buffer sharing on gpu-accessible shared memory. So, the GQA UTs do not
+  // use RunQnnModelTest and instead manually create/run the QNN inference session with the buffer sharing.
```
qti-mattsinc: This comment is stale now, as the GPU backend was updated to support the non-zero-copy path. It should be more like:

```suggestion
  // To test GQA with past/present KV cache buffer sharing, we can't use RunQnnModelTest
  // and so instead manually create/run the QNN inference session with the buffer sharing.
```

onnxruntime/test/providers/qnn/group_query_attention_test.cc
```diff
@@ -0,0 +1,1481 @@
...
+    bool use_shared_memory_allocator = false) {
+  const GetTestModelFn build_test_case = BuildGQATestCase<T, M>(query_def, key_def, value_def,
+                                                                past_key_def, past_value_def,
+                                                                seqlens_k_def, total_sequence_length_def,
+                                                                cos_cache_def, sin_cache_def,
+                                                                position_ids_def, attention_bias_def, head_sink_def,
+                                                                do_rotary, k_quant_type, kv_cache_bit_width, kv_num_heads,
+                                                                local_window_size, num_heads, qk_output, rotary_interleaved,
+                                                                scale, smooth_softmax, v_quant_type);
+  // To test GQA with past/present KV cache buffer sharing, we can't use RunQnnModelTest
```
chunghow-qti, line 172: Hi @qti-mattsinc,

I was wondering whether the GPU EP supports KV cache without buffer sharing.

f it does, do you happen to have test coverage for the non-buffer-sharing case? We found that ORT CPU EP explicitly clears `present_key` when buffer sharing is not enabled, not sure if gpu have the same clean up behavior.
Thanks!

Ref:
[onnxruntime/onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h at abbc2abc646e1f45e8534cec5991abd…](https://github.com/microsoft/onnxruntime/blob/abbc2abc646e1f45e8534cec5991abdcabf9e889/onnxruntime/contrib_ops/cpu/bert/gqa_attention_base.h#L324-L327)

qti-mattsinc: Hi @chunghow-qti,

GPU does support KV cache without buffer sharing. The non-`*SharedMemoryAllocator*`  tests added in this PR cover that case.

onnxruntime/test/providers/qnn/group_query_attention_test.cc
```diff
@@ -0,0 +1,1481 @@
...
+    VerifyOutput(output_names[i], cpu_outputs[i], *qnn_outputs[i], verification_params.tensor_verifier);
+  }
+}
+
+#if defined(_M_ARM64)
+//
+// GPU tests:
+//
+
+TEST_F(QnnGPUBackendTests, GroupQueryAttention_Basic_FP32) {
```
chunghow-qti, line 334: suggest same gaurd as op builder to maintain backward compatibility

onnxruntime/test/providers/qnn/group_query_attention_test.cc
```diff
@@ -0,0 +1,1481 @@
...
+  // Derived sizes
+  const int32_t packed_qkv_d = num_heads * head_size + 2 * kv_num_heads * head_size;
+
+  // === Inputs ===
+  auto query_def = TestInputDef<Ort::Float16_t>({batch_size, sequence_length, packed_qkv_d},
+                                                false, Ort::Float16_t(-1.0f), Ort::Float16_t(1.0f));
+  const std::optional<std::reference_wrapper<TestInputDef<Ort::Float16_t>>> key_def = std::nullopt;
+  const std::optional<std::reference_wrapper<TestInputDef<Ort::Float16_t>>> value_def = std::nullopt;
+
+  static auto pk_max = TestInputDef<Ort::Float16_t>({batch_size, kv_num_heads, total_seq_len, head_size},
```
chunghow-qti: Can we avoid using static TestInputDef?

qti-mattsinc: Done

onnxruntime/test/providers/qnn/group_query_attention_test.cc
qti-mattsinc: The plan has been to remove the guard from the op builder. Now that we've upleveled to 2.48, I've done so. None of the other ops include a guard for backwards compatibility when compiling against a newer EP against an older SDK, so I think it's best to conform to the current standard

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -533,8 +533,8 @@ ort.unregister_execution_provider_library(ep_registration_name)
 |com.microsoft:DequantizeLinear|Provides 16-bit integer dequantization support|
 |com.microsoft:FusedMatMul||
 |com.microsoft:Gelu||
-|com.microsoft:GatherBlockQuantized|GPU backend only; bits=4; block_size must be a power-of-2 ≥ 16; quantize_axis=1; symmetric quantization only (no zero points); requires QAIRT SDK ≥ 2.48|
-|com.microsoft:GroupQueryAttention|GPU backend only; requires QAIRT SDK ≥ 2.48 (QNN opset 2.12); rotary_interleaved=0; no k_quant_type/v_quant_type|
```
qti-mattsinc, line 621: I am modifying this because a QAIRT SDK requirement is not made explicit for any other ops in this table. If this is something we want to include, then we should add another column and try to document the version requirement for all ops. On top of that, I think the 2.48 requirement given for GatherBlockQuantized is inaccurate (GPU supports block quantized gather from 2.46 and I don't see anything 2.48-specific in the op-builder), and I'm worried that keeping it like this will encourage future additions to the docs to accumulate errors like these.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -13,8 +13,6 @@
 namespace onnxruntime {
 namespace qnn {
 
-// GQA not available until opset version 2.12.0 (QAIRT 2.48). TODO: Remove this check once the EP uplevels to 2.48.
```
qti-mbadnara, line 16: Instead of removing it, can we use this check

#if !(defined(QNN_SDK_VERSION_MAJOR) && QNN_SDK_VERSION_MAJOR == 2 && defined(QNN_SDK_VERSION_MINOR) && QNN_SDK_VERSION_MINOR >= 48)

Or if you prefer we can keep the existing check.

Users can compile with any version of QAIRT and we should fail gracefully, but if a decision has been taken to remove the QNN SDK Guards for Opsets then this should be fine

Maybe @tirupath-qti  can confirm

onnxruntime/test/providers/qnn/group_query_attention_test.cc
qti-mbadnara: Agree with @chunghow-qti, a guard maybe needed. @tirupath-qti can help us out here

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
tirupath-qti: agree with @qti-mbadnara

tirupath-qti (dismissed):

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
tirupath-qti: i mistakenly hit approval. @qti-mattsinc can you please make the suggested changes.

onnxruntime/test/providers/qnn/group_query_attention_test.cc
qti-mattsinc: Restored the version check, but this won't meaningfully help with backwards compatibility when building against any older SDK version. If we want to actually have that compatibility, then we need to address this in all other op builders as well as other locations in the EP that assume definitions provided by recent QNN headers. Otherwise we will just have a hodge-podge of guards present in a few locations and missing in others that still ultimately would lead to a build failure.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
qti-mattsinc: > Instead of removing it, can we use this check [...]

Same reply as https://github.com/onnxruntime/onnxruntime-qnn/pull/566#discussion_r3669452190. Additionally, I think using the Opset version check makes more sense here than the QAIRT SDK check though they are ultimately equivalent.

qti-mbadnara (approved):

qti-yuduo (approved):

PR #171, opened 2026-03-25
Reshape-Transpose-Reshape => SpaceToDepth fusion

qti-shubham:
### Description
Folds Reshape-Transpose-Reshape along with their Pre-Post Transpose ops into SpaceToDepth 
- ONNX S2D spec don't support CRD mode and unequal block size, so this pass can't implemented as Pre-optimizer and is done in the form of QNN Node Group fusion.

### Motivation and Context
Fixes https://github.com/qcom-ai-hub/tetracode/issues/17949 and https://github.com/qcom-ai-hub/tetracode/issues/16543
- Reshape-Transpose-Reshape contains 6D transpose and reshape op causing Qnn-Ctx-bin failure
- Perf improvements
- Sets parity with Qairt-converter behavior

PR #567, opened 2026-06-24
[QNN EP] Fix SpaceToDepth CRD fusion for dynamic batch models

yuhuchua-qti:
## Summary

- Fix `HasSpaceToDepthCoreSignature` to handle dynamic batch dimension (`-1`) in
Reshape shape initializers, enabling SpaceToDepth CRD fusion for dynamic-batch models.

## Problem

Models exported with dynamic batch size use `-1` in Reshape shape initializers
(e.g., `[-1, 3, 64, 2, 64, 2]`). `HasSpaceToDepthCoreSignature` checked `v <= 0`
which rejects `-1`, causing `MatchPattern` to return `nullopt`. The RTR nodes then
fell through to standalone QNN validation where 6-D tensors are rejected with
**error code 3110**, resulting in CPU fallback.

Regression introduced by PR #171 (Reshape-Transpose-Reshape → SpaceToDepth fusion),
which was only tested with static batch-size models.

## Fix

Three changes in `HasSpaceToDepthCoreSignature`:
1. **Positive-dimension guard**: skip check when `v == -1`
2. **Batch-dim match**: treat `r_n == -1` as matching any actual batch size
3. **Output shape comparison**: treat `shape_4d[i] == -1` as a wildcard

The `node_count == 3` guard in `MatchPattern` is intentionally preserved —
it avoids redundant transposes in the DLC during the 1st `GetCapability` call.
In the 2nd call (after Layout Transform), the pattern matches as 4-node or 5-node
and the guard is never triggered.

## Test plan

- [ ] Existing SpaceToDepth fusion unit tests pass (static batch, HTP backend)
- [ ] `ImageSuperResolutionV100001` model: SpaceToDepth CRD op appears in QNN graph
	(confirmed via `ComposeQnnGraph` log: `fused_spacetodepth QNN_op_type: SpaceToDepth`)
- [ ] No regression on models with static batch size

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
@@ -182,14 +182,18 @@ bool HasSpaceToDepthCoreSignature(
     return false;
   }
 
-  // check Reshape 'shape' dimensions to be positive.
-  for (int64_t v : *shape_6d) {
+  // check Reshape 'shape' dimensions to be positive (allow -1 only for the batch dim at index 0).
+  for (size_t i = 0; i < shape_6d->size(); ++i) {
+    const int64_t v = (*shape_6d)[i];
+    if (i == 0 && v == -1) continue;  // dynamic batch
     if (v <= 0) {
       return false;
     }
   }
 
-  for (int64_t v : *shape_4d) {
+  for (size_t i = 0; i < shape_4d->size(); ++i) {
+    const int64_t v = (*shape_4d)[i];
+    if (i == 0 && v == -1) continue;  // dynamic batch
     if (v <= 0) {
       return false;
     }
@@ -209,8 +213,8 @@ bool HasSpaceToDepthCoreSignature(
   const int64_t w_div = (*shape_6d)[4];
   const int64_t b1 = (*shape_6d)[5];
 
-  // r_ are expected to match input [N,C]
-  if (r_n != n || r_c != c || b0 < 1 || b1 < 1) {
+  // r_ are expected to match input [N,C]; allow r_n = -1 for dynamic batch.
+  if ((r_n != -1 && r_n != n) || r_c != c || b0 < 1 || b1 < 1) {
     return false;
   }
 
@@ -225,8 +229,12 @@ bool HasSpaceToDepthCoreSignature(
   // [N, C * b0 * b1, H / b0, W / b1]
   const int64_t expected_c = c * b0 * b1;
   const std::array<int64_t, 4> expected_shape_4d = {n, expected_c, h / b0, w / b1};
-  if (!std::equal(shape_4d->begin(), shape_4d->end(), expected_shape_4d.begin())) {
-    return false;
+  for (size_t i = 0; i < 4; ++i) {
+    // Allow -1 only for the batch dimension (index 0) in the output shape.
+    if (i == 0 && (*shape_4d)[i] == -1) continue;
+    if ((*shape_4d)[i] != expected_shape_4d[i]) {
+      return false;
+    }
   }
 
   // check transpose perm to be either {0,3,5,1,2,4} or {0,1,3,5,2,4}.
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
```diff
@@ -225,8 +225,11 @@ bool HasSpaceToDepthCoreSignature(
   // [N, C * b0 * b1, H / b0, W / b1]
   const int64_t expected_c = c * b0 * b1;
   const std::array<int64_t, 4> expected_shape_4d = {n, expected_c, h / b0, w / b1};
-  if (!std::equal(shape_4d->begin(), shape_4d->end(), expected_shape_4d.begin())) {
-    return false;
+  for (size_t i = 0; i < 4; ++i) {
+    // Allow -1 for dynamic batch dimension in the output shape.
+    if ((*shape_4d)[i] != -1 && (*shape_4d)[i] != expected_shape_4d[i]) {
```
minfhong-qti: The check here as well allows -1 in dim other than batch dim. Is it expected?
If so, then above logic should be modified as well. If not, probably add an index check.

minfhong-qti (commented):
Just curious. If this fusion is the cause of regression and the On1 model can pass in older version of QNN-EP, why it isn't fallback to original behavior if this fusion isn't matched?

onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
yuhuchua-qti: The "original behavior" in older QNN-EP (before PR #171) was handled by Rank6ToRank5Fusion, not by falling back to standalone 6-D ops. When Rank6ToRank5Fusion matched the RTR pattern (for batch=1, using the batch dim as the "unit dim" to squeeze), it reduced the intermediate 6-D tensors to 5-D ops that QNN supports — so the model ran without issues.

PR #171 explicitly excluded S2D CRD/DCR patterns from Rank6ToRank5Fusion:

```
// Added in PR #171
if (std::equal(perm...kPermS2dCrd) || std::equal(perm...kPermS2dDcr)) {
  return std::nullopt;  // ← removed the old accidental workaround
}
```

So after PR #171, when SpaceToDepthFusion fails to match (due to the -1 bug), there is no fallback —
Rank6ToRank5Fusion was intentionally disabled for S2D patterns. The RTR nodes fall through to standalone QNN
validation, where 6-D tensors are rejected with error 3110.

In short: the regression is not just "new fusion doesn't match", it's "new fusion doesn't match AND old workaround was removed simultaneously."

yuhuchua-qti: Thanks. Updated all three check sites to guard on i == 0 — only the batch dimension should be -1. Also added unit tests for the dynamic batch case.

onnxruntime/test/providers/qnn/qnn_node_group/spacetodepth_fusion_test.cc
```diff
@@ -55,7 +56,7 @@ GetTestModelFn BuildSpaceToDepthTestCase(const std::vector<int64_t>& input_shape
                                                  use_contrib_qdq);
     }
 
-    const int64_t n = input_shape[0];
+    const int64_t n = use_dynamic_batch ? -1 : input_shape[0];
```
minfhong-qti: Directly give the parameter `input_shape[0]` to `-1` instead of a new parameter.

minfhong-qti (commented):
Kindly remind for triggering CI.

onnxruntime/test/providers/qnn/qnn_node_group/spacetodepth_fusion_test.cc
yuhuchua-qti: Done. Removed use_dynamic_batch parameter — callers now pass {-1, 2, 4, 4} directly as input_shape. Added a concrete data_shape (replacing -1 with 1) for TestInputDef allocation to avoid std::vector size overflow when -1 is interpreted as size_t. All 20 SpaceToDepth tests pass.

minfhong-qti (approved):

qti-chuteng (approved):

PR #569, opened 2026-06-25
[QNN EP] Support FP16/BFP16 scale initializers in UnpackScales

chunghow-qti:
### Description

`QnnModelWrapper::UnpackScales` previously accepted only `FLOAT` scale initializers. This change adds `FLOAT16` and `BFLOAT16` decode branches: values are read as `Ort::Float16_t` / `Ort::BFloat16_t` and converted to `float` (the type QNN's quantization structs use) before being appended to the output. A dtype guard is placed before `UnpackInitializerData` so unsupported types return early without performing any initializer I/O, matching the existing guard pattern in the `uint8_t` branch.

`QLinearMatMulOpBuilder::ReadScaleAsFloat32`, which previously duplicated the same float/fp16/bf16 decode logic inline, is refactored to delegate to `UnpackScales`.

Unit tests are added (`onnxruntime_provider_test`, run under the `coverage-linux-x86_64` CI job) covering:
- `FLOAT` decode (regression)
- `FLOAT16` decode — per-axis and per-tensor scalar
- `BFLOAT16` decode — per-axis
- unsupported dtype → early error (no I/O performed)
- null tensor → error
- per-tensor `FLOAT16` scale flowing through `QnnQuantParamsWrapper::Init` to a `SCALE_OFFSET` encoding

`AddTensorFloat16` and `AddTensorBFloat16` helpers are added to the test-only `mock_init_registry.h`.

### Motivation and Context

In fp16 models, quantization tooling matches the scale initializer's dtype to the activation dtype, so the `scale` input of `QuantizeLinear` / `DequantizeLinear` (and the weight/bias scales consumed by Conv/Gemm/MatMul builders) is emitted as `FLOAT16` rather than `FLOAT`. The old code rejected these initializers, forcing the affected Q/DQ node onto the CPU EP. This had a cascading effect — downstream ops that depend on the dequantized tensor's shape then failed with `Cannot get shape`. QNN itself supports these scales (its quantization param structs store `scale` as `float`); the gap was purely a missing decode branch on the ONNX side.

`BFLOAT16` support is added at the same time for completeness, matching the dtype set already handled by `QLinearMatMulOpBuilder::ReadScaleAsFloat32`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
@@ -385,17 +385,26 @@ class QnnModelWrapper {
 
     // Handle float scales
     if constexpr (std::is_same_v<T, float>) {
-      // Verify data type for float scales
-      RETURN_IF_NOT(onnx_data_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT,
-                    "Expected scale initializer to be of type FLOAT");
-
       std::vector<uint8_t> initializer_bytes;
       RETURN_IF_ERROR(UnpackInitializerData(scale_tensor, initializer_bytes));
 
-      gsl::span<const float> src = gsl::make_span(reinterpret_cast<const float*>(initializer_bytes.data()),
-                                                  initializer_bytes.size() / sizeof(float));
-
-      scales.insert(scales.end(), src.begin(), src.end());
+      if (onnx_data_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT) {
+        gsl::span<const float> src = gsl::make_span(reinterpret_cast<const float*>(initializer_bytes.data()),
+                                                    initializer_bytes.size() / sizeof(float));
+        scales.insert(scales.end(), src.begin(), src.end());
+      } else if (onnx_data_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16) {
+        // fp16 scale initializers are produced by quantization tools that match the scale dtype to
+        // the activation dtype (e.g. fp16 models). QNN quantization structs use float, so decode here.
+        gsl::span<const Ort::Float16_t> src =
+            gsl::make_span(reinterpret_cast<const Ort::Float16_t*>(initializer_bytes.data()),
+                           initializer_bytes.size() / sizeof(Ort::Float16_t));
+        scales.reserve(scales.size() + src.size());
+        for (const auto& fp16_val : src) {
+          scales.push_back(fp16_val.ToFloat());
+        }
+      } else {
+        return MAKE_EP_FAIL("Expected scale initializer to be of type FLOAT or FLOAT16");
+      }
     }
     // Handle uint8_t scales (for block quantization)
     else if constexpr (std::is_same_v<T, uint8_t>) {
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -385,17 +385,26 @@ class QnnModelWrapper {
 
     // Handle float scales
     if constexpr (std::is_same_v<T, float>) {
-      // Verify data type for float scales
-      RETURN_IF_NOT(onnx_data_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT,
-                    "Expected scale initializer to be of type FLOAT");
-
       std::vector<uint8_t> initializer_bytes;
       RETURN_IF_ERROR(UnpackInitializerData(scale_tensor, initializer_bytes));
```
huaychou, line 396: **[N-1] (Minor)** UnpackInitializerData called before dtype dispatch — unnecessary I/O on error path

The original code validated dtype first with `RETURN_IF_NOT(onnx_data_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT, ...)` and returned early on a type mismatch, calling `UnpackInitializerData` only after the check. The new code moves `UnpackInitializerData` before the dtype dispatch, so an unsupported dtype (e.g. INT32) causes a full initializer unpack before the error is returned.

**Impact**: For an externally-stored initializer with an unsupported dtype, the new code performs a full external-data read before returning the error — unnecessary I/O. The practical cost is negligible for scale initializers (scalar or small 1D), but an early dtype guard is more defensive and more semantically symmetric with the `uint8_t` branch (which retains `RETURN_IF_NOT` for early validation), making the two branches inconsistent.

**Suggested fix:**

Either move `UnpackInitializerData` inside each dtype branch, or restore an up-front dtype guard before the unpack call:

```
if (onnx_data_type != ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT &&
    onnx_data_type != ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16) {
  return MAKE_EP_FAIL("Expected scale initializer to be of type FLOAT or FLOAT16");
}

std::vector<uint8_t> initializer_bytes;
RETURN_IF_ERROR(UnpackInitializerData(scale_tensor, initializer_bytes));

if (onnx_data_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT) { ... }
else { /* FLOAT16 path */ }
```

huaychou:
We noticed that QLinearMatMulOpBuilder also includes ReadScaleAsFloat32 to support float16 and bfloat16 scales, so it may be worth integrating them.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -385,17 +385,39 @@ class QnnModelWrapper {
...
+      // float (identity), Ort::Float16_t, and Ort::BFloat16_t (both have operator float()). fp16/bf16
+      // scale initializers are produced by quantization tools that match the scale dtype to the
+      // activation dtype (e.g. fp16 models); QNN quantization structs use float, so decode here.
+      auto append_scales = [&scales, &initializer_bytes](auto src_type_tag) {
+        using SrcT = decltype(src_type_tag);
+        gsl::span<const SrcT> src = gsl::make_span(reinterpret_cast<const SrcT*>(initializer_bytes.data()),
+                                                   initializer_bytes.size() / sizeof(SrcT));
+        scales.reserve(scales.size() + src.size());
+        for (const auto& val : src) {
+          scales.push_back(static_cast<float>(val));
```
huaychou, line 408: Just wondering, will `scales.push_back` be slower than `scales.insert` in float use case?

huaychou (approved):
LGTM

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
chunghow-qti: Yes, insert would be faster for float. But the lambda is shared with the fp16/bf16 branches, where static_cast<float>(val) is a real decode, different types, so insert can't take the bulk-copy path there anyway. Since this runs once per tensor at compile time on tiny arrays, I kept the uniform loop for readability.

qti-chuteng (approved):

PR #570, opened 2026-06-25
[CI] Run pytests in CI

qti-niscmami:
### Description
Adds a `pytest-unit-tests` job to the Qualcomm Internal PR Workflow that runs the
  pytest suite under `qcom/scripts/all/tests/` (69 tests, pure stdlib + pytest).
  The job installs pytest and runs `python3 -m pytest -q` in that directory on the
  existing Linux X64 Build runner. Gated on the same draft-PR condition as the other
  jobs

### Motivation and Context
The suite covering the test-archiving helper scripts (`archive_testdata`,
  `archive_tests`, `extract_testdata`) was not executed by any workflow. The tests
  passed locally but never gated a PR, so regressions in these scripts could merge
  undetected. This wires them into CI.


qti-yuduo (approved):

.github/workflows/qualcomm-internal-ci.yml
```diff
@@ -47,3 +47,19 @@ jobs:
     if: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
     uses: ./.github/workflows/qualcomm-internal-asan.yml
     secrets: inherit
+
+  pytest-unit-tests:
```
qti-kromero: is it intentional that this is living in the dispatcher job?

taking linting as an analogue it may fit better at the top-level build-and-test.yml

.github/workflows/qualcomm-internal-ci.yml
```diff
@@ -47,3 +47,19 @@ jobs:
...
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v6
+      - name: Run pytest
+        working-directory: qcom/scripts/all/tests
+        run: |
+          python3 -m venv .venv
+          .venv/bin/pip install pytest
```
qti-kromero: it would probably be good to pin this pytest install to a major version of pytest

.github/workflows/qualcomm-internal-ci.yml
```diff
@@ -47,3 +47,19 @@ jobs:
     if: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
     uses: ./.github/workflows/qualcomm-internal-asan.yml
     secrets: inherit
+
+  pytest-unit-tests:
+    name: PyTest Unit Tests
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.draft == false }}
+    runs-on: ["self-hosted", "X64", "Linux", "Build"]
```
qti-kromero: do we want these tests to run on other platforms aside from Linux? may also be worth integrating this into existing testing infrastructure if we envision the matrix being valuable

I think it may be worth it because the script itself is ran on multiple platforms when ORT Core is built right? e.g. Linux vs. Windows, as-is we are only really testing the Linux part of that

.github/workflows/qualcomm-internal-ci.yml
qti-niscmami: Yeah. Will move it.

.github/workflows/qualcomm-internal-ci.yml
qti-niscmami: Agreed.

.github/workflows/qualcomm-internal-ci.yml
qti-niscmami: Could leave it as-is.

We currently use pathlib + fake directory trees for testing the Windows specific folder codepaths. But otherwise, the scripts are platform agnostic. The only scenario where I might consider testing on Windows is if pathlib has some bugs in directory traversal, but that would mean the std library itself is broken.

We could add Windows test runner, but then we are likely not getting much benefit out of it, as now we are just trying to catch pathlib error cases.

.github/workflows/qualcomm-internal-ci.yml
qti-niscmami: Moved to a reusable workflow (qualcomm-internal-pytest.yml) . Dispatcher now just calls it, same pattern as asan/coverage.

.github/workflows/qualcomm-internal-ci.yml
qti-niscmami: Pinned against current stable release of pytest.

.github/workflows/qualcomm-internal-ci.yml
qti-kromero: I was thinking more in .github/workflows/qualcomm-internal-build-and-test.yml - I'm not sure I agree that asan/coverage should have been added to the dispatcher job to begin with

PR #571, opened 2026-06-25
[QNN EP] Fix if_test.cc to use new RunQnnModelTest EPVerificationParams signature

qti-niscmami:
### Description
`if_test.cc` failed to compile (linux-aarch64, android, windows-arm64ec) after #504 changed `RunQnnModelTest`'s 4th arg to a single `const EPVerificationParams&`, while #409 was authored against the old signature and merged in parallel. 

Fix wraps the args as
  `EPVerificationParams{expected_ep_assignment, ElementwiseAbsoluteVerifier(fp32_abs_err)}`, matching every other caller.


qti-ashwshan (approved):
LGTM

PR #573, opened 2026-06-25
[QNN EP] Fix naming convention of Linux x86_64 Wheels in Nightly

qti-mbadnara:
As per title.  Pin to the build's target platform tag. Without --plat, auditwheel auto-detects the lowest policy the binaries qualify for and emits a compressed tag set (e.g. manylinux_2_34_x86_64.manylinux_2_35_x86_64) for Linux x86_64 wheels.

RC Workflow: https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/28262631182

Nightly Workflow: https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/28262650899


setup.py
```diff
@@ -110,12 +110,32 @@ def run(self):
                 assert self.dist_dir is not None
                 file = glob(path.join(self.dist_dir, "*linux*.whl"))[0]
                 logger.info("repairing %s for manylinux1", file)
-                auditwheel_cmd = ["auditwheel", "-v", "repair", "-w", self.dist_dir, file]
+                # Pin to the build's target platform tag. Without --plat, auditwheel
+                # auto-detects the lowest policy the binaries qualify for and emits a
+                # compressed tag set (e.g. manylinux_2_34_x86_64.manylinux_2_35_x86_64).
+                auditwheel_plat = environ["AUDITWHEEL_PLAT"]
```
qti-kromero, line 116: what happens here if someone invokes this outside the docker container? maybe worth adding an error message but not sure if someone will hit issues before this line if they are outside of the container 

```suggestion
auditwheel_plat = environ["AUDITWHEEL_PLAT"] or sys.exit("AUDITWHEEL_PLAT not set, are you in docker container?")
```

qti-kromero: if they would hit other issues first then it might be overkill

qti-kromero:
LGTM although surprising to me that there isn't an easier way to do this, couldn't really find anything in my research that would make this cleaner

setup.py
qti-mbadnara: The code is guarded behind is_manylinux = environ.get("AUDITWHEEL_PLAT", None) in manylinux_tags

On a Linux build outside the docker, AUDITWHEEL_PLAT isn't set, so is_manylinux is False. The if is_manylinux and not disable_auditwheel_repair: block is skipped entirely — no auditwheel call, no wheel tags re-tag, no --plat.

The Windows wheel keeps the tag that bdist_wheel assigns natively, e.g. ...-cp311-cp311-win_amd64.whl. auditwheel is a Linux-only tool, so it was never part of the Windows path to begin with.

qti-kromero: oh I see, that was cut off by GitHub

qti-kromero (approved):

PR #574, opened 2026-06-25
Add convert for mixed-bitwidth quantized conv

qti-yuduo:
## Summary

QNN HTP `Conv2d` requires matched input/output bitwidth. A quantized Conv whose output Q has a different bitwidth than the input (e.g. uint16 activations feeding a uint8 KV-cache, as in Llama-3.2 v_proj heads) was rejected by `OrtConvNodeGroupSelector` and silently fell back to a Conv with float weight, producing UNDEFINED weight encoding and an HTP finalize failure.

Two coordinated changes mirror the QAIRT converter's lowering:

- **`qnn_ep_utils.cc OrtConvNodeGroupSelector::Check`** — allow input/output dtype mismatch only for the unsigned uint8 <-> uint16 case the op-builder can decompose; reject all other mismatches. Extend the 16-bit allow guard to cover the output dtype.
- **`conv_op_builder.cc ProcessAttributesAndOutputs`** — for the mixed-bitwidth case emit `Conv (output @ input bitwidth, output range re-encoded at input bitwidth) -> QNN_OP_CONVERT -> output (at the user-authored bitwidth/encoding)`. The matched-bitwidth path is unchanged. The 1D Conv path (no Convert insertion there) explicitly rejects mixed-bitwidth I/O.

Adds a static `RequantizePerTensorAtBitwidth` helper that preserves the floating-point range across bitwidths.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -3,6 +3,8 @@
 
 #include <gsl/gsl>
 
+#include <cmath>
+
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
 #include "core/providers/qnn/builder/qnn_model_wrapper.h"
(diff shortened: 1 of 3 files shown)


qti-yuduo:
Will do a proper redesigned mixed precision handling.

PR #580, opened 2026-06-29
Add unit tests for onnx_ctx_model_helper.cc

yuhuchua-qti:
## Summary

- Add 47 function-level unit tests for `onnx_ctx_model_helper.cc`
- Unit-only coverage: 0% → 66.0% (171/259); combined 97.7% (253/259)
- Extend `qnn_fake_ort_graph.h` with `FakeOpAttr` struct and
`Node_GetAttributeByName` / `OpAttr_GetType` / `ReadOpAttr` stubs
to support `OrtNodeAttrHelper` in unit test context

## Functions covered

| Function | Tests | Notes |
|---|---|---|
| `GraphHasEpContextNode` | 10 | All source/type branches |
| `GraphHasDlcContextNode` | 2 | |
| `IsOrtGraphHasCtxNode` / `IsOrtGraphHasDlcCtxNode` | 5 | |
| `GetEpContextDlcPath` | 5 | |
| `TryGetMaxSpillFillSize` | 5 | Including swap path |
| `ParseIoNameOverrides` | 7 | All edge cases |
| `GetMainContextNode` | 6 | All error paths + success |
| `GetEpContextFromMainNode` | 5 | Validation paths only (op_type, empty/absolute/dotdot/missing path); embed_mode=true and file I/O are integration territory |
| `CreateEPContextNodes` | 2 | IS_MULTI_SOC_EP_CONTEXT attribute written 1/0; DLC-vs-single-SoC binary divergence lives in QnnBackendManager |

## Test plan

- [x] `onnxruntime_provider_test --gtest_filter="QnnUnit_OnnxCtxModelHelperTest.*"` — 47/47 pass

### DIFF
--- a/onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
+++ b/onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
@@ -22,8 +22,31 @@
 #include "test/providers/qnn/unit/qnn_fake_ort_graph.h"
 #include "test/providers/qnn/unit/qnn_unit_test_utils.h"
 
-using namespace onnxruntime::qnn;
-using namespace onnxruntime::test;
+namespace onnxruntime {
+namespace test {
+
+// Specific using-declarations (not `using namespace`) so each helper/constant
+// pulled from onnxruntime::qnn is named explicitly. Fake* graph types and the
+// stub helpers already live in onnxruntime::test, this file's own namespace.
+using onnxruntime::qnn::EMBED_MODE;
+using onnxruntime::qnn::EP_CACHE_CONTEXT;
+using onnxruntime::qnn::EP_CONTEXT_TYPE;
+using onnxruntime::qnn::EP_CONTEXT_TYPE_BIN;
+using onnxruntime::qnn::EP_CONTEXT_TYPE_DLC;
+using onnxruntime::qnn::GetEpContextDlcPath;
+using onnxruntime::qnn::GetEpContextFromMainNode;
+using onnxruntime::qnn::GetMainContextNode;
+using onnxruntime::qnn::GraphHasDlcContextNode;
+using onnxruntime::qnn::GraphHasEpContextNode;
+using onnxruntime::qnn::IO_NAME_OVERRIDES;
+using onnxruntime::qnn::IsOrtGraphHasCtxNode;
+using onnxruntime::qnn::IsOrtGraphHasDlcCtxNode;
+using onnxruntime::qnn::MAIN_CONTEXT;
+using onnxruntime::qnn::MAX_SIZE;
+using onnxruntime::qnn::ParseIoNameOverrides;
+using onnxruntime::qnn::QnnModelLookupTable;
+using onnxruntime::qnn::SOURCE;
+using onnxruntime::qnn::TryGetMaxSpillFillSize;
 
 namespace {
 
@@ -326,7 +349,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, GetEpContextDlcPath_SecondGraphHasDlcNode_R
 // TryGetMaxSpillFillSize
 //
 // Iterates main_context_pos_list, reads MAX_SIZE from each EPContext node, and
-// swaps the largest to position 0 in main_context_pos_list (L270: swap path).
+// swaps the largest to position 0 in main_context_pos_list.
 // =============================================================================
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, TryGetMaxSpillFillSize_ZeroContexts_ReturnsOk) {
@@ -379,8 +402,8 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, TryGetMaxSpillFillSize_SecondContextLarger_
 }
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, TryGetMaxSpillFillSize_NonConsecutivePosList_IndexesThroughPosList) {
-  // Regression guard for the `graphs[main_context_pos_list[idx]]` indirection
-  // (source L261). pos_list is non-identity ({2, 0}) and graph 1 — which the
+  // Regression guard for the `graphs[main_context_pos_list[idx]]` indirection.
+  // pos_list is non-identity ({2, 0}) and graph 1 — which the
   // pos_list never references — carries the largest MAX_SIZE as a trap.
   // Correct code reads graphs 2 then 0 (max=500, already at front → no swap).
   // A mutation to `graphs[idx]` would instead read graphs 0 then 1, pick up the
@@ -412,26 +435,42 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, TryGetMaxSpillFillSize_NonConsecutivePosLis
   EXPECT_EQ(pos_list[1], 0);
 }
 
+TEST(QnnUnit_OnnxCtxModelHelperTest, TryGetMaxSpillFillSize_GraphWithMultipleNodes_ReturnsError) {
+  // Each main-context graph must contain exactly one EPContext node; a graph
+  // whose node count is not 1 is rejected before any MAX_SIZE read. Two nodes
+  // exercises that guard (ZeroContexts_ReturnsOk above instead skips the loop
+  // via count=0, so it never reaches this check).
+  CtxHelperTestContext ctx;
+  FakeNode n0{"n0", "EPContext", "", 1, {}, {}};
+  FakeNode n1{"n1", "EPContext", "", 1, {}, {}};
+  FakeGraph g{{n0, n1}, {}, {}, {}};
+  const OrtGraph* graphs[] = {g.AsGraph()};
+  std::vector<int> pos_list = {0};
+  int64_t max_size = 0;
+  auto status = TryGetMaxSpillFillSize(graphs, ctx.api, 1, max_size, pos_list);
+  EXPECT_FALSE(status.IsOK());
+}
+
 // =============================================================================
 // ParseIoNameOverrides
 //
 // Parses the IO_NAME_OVERRIDES attribute from an EPContext node into an
 // internal→external name map. Edge cases:
-//   - nullptr node           → L122: early return empty map
-//   - no trailing semicolon  → L131: sep = encoded.size()
-//   - consecutive semicolons → L136: empty pair → continue
-//   - pair with no '='       → L140: malformed pair → continue
+//   - nullptr node           → early return empty map
+//   - no trailing semicolon  → separator falls back to encoded.size()
+//   - consecutive semicolons → empty pair → continue
+//   - pair with no '='       → malformed pair → continue
 // =============================================================================
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_NullNode_ReturnsEmpty) {
-  // L122: ep_context_node == nullptr → returns empty map immediately.
+  // ep_context_node == nullptr → returns empty map immediately.
   auto overrides = ParseIoNameOverrides(nullptr);
   EXPECT_TRUE(overrides.empty());
 }
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_SinglePairNoTrailingSemicolon_Parsed) {
   CtxHelperTestContext ctx;
-  // "a=b" — no trailing semicolon; sep falls back to encoded.size() (L131).
+  // "a=b" — no trailing semicolon; separator falls back to encoded.size().
   FakeOpAttr attr = FakeOpAttr::MakeString(IO_NAME_OVERRIDES, "a=b");
   FakeNode node{"ep", "EPContext", "", 1, {}, {}};
   node.attrs[IO_NAME_OVERRIDES] = &attr;
@@ -442,7 +481,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_SinglePairNoTrailingSe
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_EmptyPair_Skipped) {
   CtxHelperTestContext ctx;
-  // ";a=b" — leading semicolon produces an empty first pair (L136: continue).
+  // ";a=b" — leading semicolon produces an empty first pair (skipped via continue).
   FakeOpAttr attr = FakeOpAttr::MakeString(IO_NAME_OVERRIDES, ";a=b");
   FakeNode node{"ep", "EPContext", "", 1, {}, {}};
   node.attrs[IO_NAME_OVERRIDES] = &attr;
@@ -453,7 +492,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_EmptyPair_Skipped) {
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_MalformedPairNoEquals_Skipped) {
   CtxHelperTestContext ctx;
-  // "noeq;a=b" — first pair has no '=' (L140: continue). Second is valid.
+  // "noeq;a=b" — first pair has no '=' (skipped via continue). Second is valid.
   FakeOpAttr attr = FakeOpAttr::MakeString(IO_NAME_OVERRIDES, "noeq;a=b");
   FakeNode node{"ep", "EPContext", "", 1, {}, {}};
   node.attrs[IO_NAME_OVERRIDES] = &attr;
@@ -477,7 +516,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_MultiplePairs_AllParse
 TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_EqualsInsideValue_SplitsOnFirst) {
   CtxHelperTestContext ctx;
   // "a=b=c" — pair.find('=') splits on the FIRST '=', so internal="a",
-  // external="b=c" (the remaining '=' stays in the value, L142-143).
+  // external="b=c" (the remaining '=' stays in the value).
   FakeOpAttr attr = FakeOpAttr::MakeString(IO_NAME_OVERRIDES, "a=b=c");
   FakeNode node{"ep", "EPContext", "", 1, {}, {}};
   node.attrs[IO_NAME_OVERRIDES] = &attr;
@@ -489,7 +528,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_EqualsInsideValue_Spli
 TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_EmptyInternalOrExternal_Skipped) {
   CtxHelperTestContext ctx;
   // "=b" has an empty internal, "a=" has an empty external; both fail the
-  // L144 !internal.empty() && !external.empty() guard and are skipped. Only the
+  // !internal.empty() && !external.empty() guard and are skipped. Only the
   // fully-populated "c=d" survives.
   FakeOpAttr attr = FakeOpAttr::MakeString(IO_NAME_OVERRIDES, "=b;a=;c=d");
   FakeNode node{"ep", "EPContext", "", 1, {}, {}};
@@ -507,7 +546,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, ParseIoNameOverrides_EmptyInternalOrExterna
 // =============================================================================
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_ZeroGraphs_ReturnsError) {
-  // count=0 → loop never executes → L115: pos empty → error.
+  // count=0 → loop never executes → pos empty → error.
   CtxHelperTestContext ctx;
   std::vector<int> pos;
   auto status = GetMainContextNode(nullptr, 0, ctx.api, pos);
@@ -515,7 +554,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_ZeroGraphs_ReturnsError)
 }
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_GraphWithTwoNodes_ReturnsError) {
-  // L97: num_nodes != 1 → error.
+  // num_nodes != 1 → error.
   CtxHelperTestContext ctx;
   FakeNode n0{"n0", "Relu", "", 13, {}, {}};
   FakeNode n1{"n1", "Relu", "", 13, {}, {}};
@@ -527,7 +566,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_GraphWithTwoNodes_Return
 }
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_WrongOpType_ReturnsError) {
-  // L106: op_type != EPCONTEXT_OP → error.
+  // op_type != EPCONTEXT_OP → error.
   CtxHelperTestContext ctx;
   FakeNode node{"relu", "Relu", "", 13, {}, {}};
   FakeGraph g{{node}, {}, {}, {}};
@@ -538,7 +577,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_WrongOpType_ReturnsError
 }
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_NoMainContextAttr_ReturnsError) {
-  // No MAIN_CONTEXT attr → defaults to 0 → not marked main → L115 error.
+  // No MAIN_CONTEXT attr → defaults to 0 → not marked main → empty pos → error.
   CtxHelperTestContext ctx;
   FakeNode node{"ep", "EPContext", "", 1, {}, {}};
   FakeGraph g{{node}, {}, {}, {}};
@@ -585,10 +624,10 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_TwoGraphsSecondIsMain_Re
 // GetEpContextFromMainNode — path validation (no QnnBackendManager needed)
 //
 // Unit-testable error paths that fire before any filesystem or backend access:
-//   L159: wrong op_type
-//   L178: embed_mode=false, ep_cache_context empty (default "")
-//   L194: embed_mode=false, absolute path (starts with '/')
-//   L198: embed_mode=false, path contains ".."
+//   wrong op_type
+//   embed_mode=false, ep_cache_context empty (default "")
+//   embed_mode=false, absolute path (starts with '/')
+//   embed_mode=false, path contains ".."
 //
 // Deliberately NOT unit-tested here (deferred to integration tests):
 //   - embed_mode=true: calls QnnBackendManager::LoadCachedQnnContextFromBuffer,
@@ -602,7 +641,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, GetMainContextNode_TwoGraphsSecondIsMain_Re
 // =============================================================================
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, GetEpContextFromMainNode_WrongOpType_ReturnsError) {
-  // L159: op_type != EPCONTEXT_OP → error before any attr or path access.
+  // op_type != EPCONTEXT_OP → error before any attr or path access.
   CtxHelperTestContext ctx;
   FakeNode node{"relu", "Relu", "", 13, {}, {}};
   QnnModelLookupTable models;
@@ -664,7 +703,7 @@ TEST(QnnUnit_OnnxCtxModelHelperTest, GetEpContextFromMainNode_NonEmbedDotDotPath
 }
 
 TEST(QnnUnit_OnnxCtxModelHelperTest, GetEpContextFromMainNode_NonEmbedFileNotFound_ReturnsError) {
-  // embed_mode=0, valid relative path but file does not exist → L206-208 error.
+  // embed_mode=0, valid relative path but file does not exist → not-found error.
   // No real file is needed; std::filesystem::is_regular_file returns false for
   // a non-existent path, triggering the "does not exist or is not accessible" guard.
   CtxHelperTestContext ctx;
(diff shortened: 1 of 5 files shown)


qti-chuteng:
### 🟪 Architecture Fit

**[D-3]** `🟠 Concern` `[Arch]` — `qnn_fake_ort_graph.h:725-731`
The `ReadOpAttr` type-mismatch branch returns `nullptr` (OK) + `*out_size=0`, with a comment claiming `CheckAttrType` makes it unreachable. That's true for STRING/array reads (`GetValueImpl` calls CheckAttrType first) but **false for the scalar numeric path**: `GetNumericValue` (onnxruntime_cxx_inline.h:2924-2929) calls `ReadOpAttr` directly with no preceding CheckAttrType. For a wrong-typed numeric Get, the real API returns an error (→ def_val), whereas this stub returns OK with `out` uninitialized. Latent contract infidelity in the shared harness; unexercised today, but the comment misleads future users.
*Better design:* Either return a non-null error status on type mismatch (faithful to the real API), or narrow the comment to state it only holds for string/array reads.

### 🔵 Completeness

**[D-1]** `🟠 Concern` `[Completeness]` — `onnx_ctx_model_helper_test.cc:361`
The source (onnx_ctx_model_helper.cc:257-261) deliberately indexes `graphs[main_context_pos_list[idx]]` and warns in a comment that `graphs[idx]` would read the wrong graph at non-consecutive positions. All three tests use an identity `pos_list` (`{0}`, `{0,1}`), so a mutation regressing to `graphs[idx]` still passes. This function's core invariant is left unguarded.
*Better design:* Add a non-consecutive `pos_list` case (e.g. `{1,0}`) where MAX_SIZE values make "use idx" vs "use pos_list[idx]" diverge.

**[D-2]** `🟠 Concern` `[Completeness]` — `onnx_ctx_model_helper_test.cc:291`
The source lowercases the DLC path via `GetLowercaseString` (onnx_ctx_model_helper.cc:79). Both happy-path tests use already-lowercase inputs, so `EXPECT_EQ` passes whether or not the lowercasing exists. A mixed-case regression goes undetected — and a green test entrenches the (suspect) practice of lowercasing a filesystem path.
*Better design:* Use a mixed-case value (e.g. `Model.DLC`) and assert the observed result; note the path-lowercasing as a source-side concern.

**[D-4]** `🟠 Concern` `[Completeness]` — `onnx_ctx_model_helper_test.cc:561`
The source has two path-validation impls: Linux tests raw `[0]=='/'` / `find("..")`; Windows (`_WIN32`) uses `is_absolute()` / `find(L"..")` on a `lexically_normal()`ed wstring. The whole unit tier is gated on `ENABLE_COVERAGE AND UNIX AND x86_64` (cmake:391-401), so the shipping `_WIN32` guards (incl. the directory-traversal check) get zero coverage. Inputs are POSIX-baked too: `/absolute/path.bin` is `is_absolute()`==false on Windows.
*Better design:* Extract a platform-agnostic path-policy helper tested per-platform, or add a `_WIN32`-gated test with `C:\absolute\path.bin`; at minimum note the platform asymmetry in the test file.

**[D-5]** `🔵 Improvement` `[Completeness]` — `onnx_ctx_model_helper.cc:288,320`
`LoadQnnCtxFromOnnxGraph` and `CreateEPContextNodes` are entirely untested with no deferral note, unlike the carefully-documented intra-function deferrals elsewhere. `LoadQnnCtxFromOnnxGraph`'s `num_nodes != 1` guard (L297) fires before any backend use — trivially unit-testable and reachable by this harness.
*Better design:* Add a deferral note for both, or a test for the early `num_nodes != 1` guard.

**[D-7]** `🔵 Improvement` `[Completeness]` — `README.md:23-31`
The maintained "Current test suites" table (7 rows) is not updated with the new 8th file/suite (`onnx_ctx_model_helper_test.cc` / `QnnUnit_OnnxCtxModelHelperTest`). The cmake glob auto-discovers it, so functionality is fine — pure doc-sync.
*Better design:* Add the row to keep the discoverability table complete.

**[D-8]** `🔵 Improvement` `[Completeness]` — `onnx_ctx_model_helper_test.cc:395`
`ParseIoNameOverrides` tests miss: multiple valid pairs (`a=b;c=d`), `=` inside a value (`a=b=c`, split on first `=`), and the empty-internal/external guard (false branch of L144).
*Better design:* Add those three cases to fully pin the decode loop.

### 🟩 Simplicity

**[D-6]** `🔵 Improvement` `[Simplicity]` — `onnx_ctx_model_helper_test.cc:43-45`
`CtxHelperTestContext` initializes via a comma operator: `global_guard((InstallFakeGraphApiStubs(api), &api))`, relying on `api` being declared before `global_guard`. Reordering the members silently breaks setup — a hidden precondition.
*Better design:* Use a plain ctor body (like sibling `EpUtilsTestContext`): `CtxHelperTestContext() { InstallFakeGraphApiStubs(api); }`. Safe because `OrtGlobalApiOverride` only stores `&api` (dereferenced at call time).

yuhuchua-qti:
> ### 🟪 Architecture Fit
> **[D-3]** `🟠 Concern` `[Arch]` — `qnn_fake_ort_graph.h:725-731` The `ReadOpAttr` type-mismatch branch returns `nullptr` (OK) + `*out_size=0`, with a comment claiming `CheckAttrType` makes it unreachable. That's true for STRING/array reads (`GetValueImpl` calls CheckAttrType first) but **false for the scalar numeric path**: `GetNumericValue` (onnxruntime_cxx_inline.h:2924-2929) calls `ReadOpAttr` directly with no preceding CheckAttrType. For a wrong-typed numeric Get, the real API returns an error (→ def_val), whereas this stub returns OK with `out` uninitialized. Latent contract infidelity in the shared harness; unexercised today, but the comment misleads future users. _Better design:_ Either return a non-null error status on type mismatch (faithful to the real API), or narrow the comment to state it only holds for string/array reads.
> 
> ### 🔵 Completeness
> **[D-1]** `🟠 Concern` `[Completeness]` — `onnx_ctx_model_helper_test.cc:361` The source (onnx_ctx_model_helper.cc:257-261) deliberately indexes `graphs[main_context_pos_list[idx]]` and warns in a comment that `graphs[idx]` would read the wrong graph at non-consecutive positions. All three tests use an identity `pos_list` (`{0}`, `{0,1}`), so a mutation regressing to `graphs[idx]` still passes. This function's core invariant is left unguarded. _Better design:_ Add a non-consecutive `pos_list` case (e.g. `{1,0}`) where MAX_SIZE values make "use idx" vs "use pos_list[idx]" diverge.
> 
> **[D-2]** `🟠 Concern` `[Completeness]` — `onnx_ctx_model_helper_test.cc:291` The source lowercases the DLC path via `GetLowercaseString` (onnx_ctx_model_helper.cc:79). Both happy-path tests use already-lowercase inputs, so `EXPECT_EQ` passes whether or not the lowercasing exists. A mixed-case regression goes undetected — and a green test entrenches the (suspect) practice of lowercasing a filesystem path. _Better design:_ Use a mixed-case value (e.g. `Model.DLC`) and assert the observed result; note the path-lowercasing as a source-side concern.
> 
> **[D-4]** `🟠 Concern` `[Completeness]` — `onnx_ctx_model_helper_test.cc:561` The source has two path-validation impls: Linux tests raw `[0]=='/'` / `find("..")`; Windows (`_WIN32`) uses `is_absolute()` / `find(L"..")` on a `lexically_normal()`ed wstring. The whole unit tier is gated on `ENABLE_COVERAGE AND UNIX AND x86_64` (cmake:391-401), so the shipping `_WIN32` guards (incl. the directory-traversal check) get zero coverage. Inputs are POSIX-baked too: `/absolute/path.bin` is `is_absolute()`==false on Windows. _Better design:_ Extract a platform-agnostic path-policy helper tested per-platform, or add a `_WIN32`-gated test with `C:\absolute\path.bin`; at minimum note the platform asymmetry in the test file.
> 
> **[D-5]** `🔵 Improvement` `[Completeness]` — `onnx_ctx_model_helper.cc:288,320` `LoadQnnCtxFromOnnxGraph` and `CreateEPContextNodes` are entirely untested with no deferral note, unlike the carefully-documented intra-function deferrals elsewhere. `LoadQnnCtxFromOnnxGraph`'s `num_nodes != 1` guard (L297) fires before any backend use — trivially unit-testable and reachable by this harness. _Better design:_ Add a deferral note for both, or a test for the early `num_nodes != 1` guard.
> 
> **[D-7]** `🔵 Improvement` `[Completeness]` — `README.md:23-31` The maintained "Current test suites" table (7 rows) is not updated with the new 8th file/suite (`onnx_ctx_model_helper_test.cc` / `QnnUnit_OnnxCtxModelHelperTest`). The cmake glob auto-discovers it, so functionality is fine — pure doc-sync. _Better design:_ Add the row to keep the discoverability table complete.
> 
> **[D-8]** `🔵 Improvement` `[Completeness]` — `onnx_ctx_model_helper_test.cc:395` `ParseIoNameOverrides` tests miss: multiple valid pairs (`a=b;c=d`), `=` inside a value (`a=b=c`, split on first `=`), and the empty-internal/external guard (false branch of L144). _Better design:_ Add those three cases to fully pin the decode loop.
> 
> ### 🟩 Simplicity
> **[D-6]** `🔵 Improvement` `[Simplicity]` — `onnx_ctx_model_helper_test.cc:43-45` `CtxHelperTestContext` initializes via a comma operator: `global_guard((InstallFakeGraphApiStubs(api), &api))`, relying on `api` being declared before `global_guard`. Reordering the members silently breaks setup — a hidden precondition. _Better design:_ Use a plain ctor body (like sibling `EpUtilsTestContext`): `CtxHelperTestContext() { InstallFakeGraphApiStubs(api); }`. Safe because `OrtGlobalApiOverride` only stores `&api` (dereferenced at call time).

**D-1 — `TryGetMaxSpillFillSize` invariant not guarded (identity `pos_list`)**
Fixed. Added `TryGetMaxSpillFillSize_NonConsecutivePosList_IndexesThroughPosList`:
`pos_list = {2, 0}`, graphs `g0(MAX_SIZE=100)`, `g1(999)`, `g2(500)`. `g1` is a trap — it is *not* referenced by `pos_list`, so a regression to `graphs[idx]` (instead of `graphs[pos_list[idx]]`) would read `g0, g1`, pick up `999`, and swap it to the front.
The test asserts `max_size == 500` and `pos_list == {2, 0}`, so the mutation turns it red.

**D-2 — DLC-path lowercasing not exercised (already-lowercase input)**
Fixed. `GetEpContextDlcPath_DlcNodeWithPath_ReturnsLowercasedPath` now feeds mixed-case `"Path/To/Model.DLC"` and asserts `"path/to/model.dlc"`; an already-lowercase input could not distinguish whether `GetLowercaseString` ran. Test comment kept minimal/neutral. Side note (not acted on in this PR): unconditionally lowercasing a filesystem path is questionable on case-sensitive filesystems (Linux). That is a source-side design question outside the scope of a test-only PR — flagging it here in case it's worth a separate issue.

**D-3 — `ReadOpAttr` fake diverges from real API on type mismatch**
Fixed (faithful version). The fake `ReadOpAttr` stub now returns a real `ORT_INVALID_ARGUMENT` status on attribute type mismatch instead of returning OK with `*out_size = 0`. This matches the real contract for the scalar numeric path (`Ort::ConstOpAttr::GetNumericValue` calls `ReadOpAttr` with no preceding `CheckAttrType`), where the caller (`OrtNodeAttrHelper::Get`) maps the error to its default value. The STRING/array path is unaffected — it goes through `CheckAttrType` (→ `OpAttr_GetType`) first, so it never reaches this branch.

**D-4 — `_WIN32` path-validation branches uncovered / POSIX-only inputs**
Not changed (documented no-op). The entire unit tier is gated on `ENABLE_COVERAGE AND UNIX AND x86_64`, so the `_WIN32` guards cannot execute in this tier regardless of input — a `_WIN32`-gated test would simply not run here. Extracting a platform-agnostic path-policy helper would modify EP source, which is scope creep for a test-only PR. Happy to file a follow-up if cross-platform coverage of these guards is wanted.

**D-5 — `LoadQnnCtxFromOnnxGraph` / `CreateEPContextNodes` untested, no deferral note**
Deferred to the integration tier. Both functions need a real backend / session to run past their entry guards, which is out of the unit tier's boundary (no session by definition). Their functional correctness is covered by the end-to-end EP-context integration suite, so we defer them there rather than add stub-level tests here.

**D-6 — `CtxHelperTestContext` member-ordering fragility**
Fixed. Added a comment clarifying that `OrtGlobalApiOverride global_guard{&api}` only records the *pointer* `&api` (the API is dereferenced at call time), so filling in the stubs in the constructor body afterwards is safe and does not depend on member-declaration order relative to the install call.

**D-7 — README suites table missing the new file**
Fixed. Added the `onnx_ctx_model_helper_test.cc → QnnUnit_OnnxCtxModelHelperTest → builder/onnx_ctx_model_helper.cc` row to `unit/README.md`.

**D-8 — `ParseIoNameOverrides` edge cases not covered**
Fixed. Added three tests:
- `ParseIoNameOverrides_MultiplePairs_AllParsed` (`"a=b;c=d"` → size 2)
- `ParseIoNameOverrides_EqualsInsideValue_SplitsOnFirst` (`"a=b=c"` → `a == "b=c"`)
- `ParseIoNameOverrides_EmptyInternalOrExternal_Skipped` (`"=b;a=;c=d"` → only `c=d` survives)

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
```diff
@@ -419,6 +461,48 @@ inline void InstallFakeGraphApiStubs(OrtApi& api) {
...
+      // numeric Get reaches here; the real API returns an error, which the caller
+      // (OrtNodeAttrHelper::Get) maps to its default value. Return a real error to
+      // match that contract. (STRING/array reads never hit this branch because
+      // ConstOpAttr::GetValue<T>() calls CheckAttrType via OpAttr_GetType first.)
+      *out_size = 0;
+      return reinterpret_cast<OrtStatus*>(
+          new FakeOrtStatus{ORT_INVALID_ARGUMENT, "ReadOpAttr: attribute type mismatch"});
+    }
+    if (expected_type == OrtOpAttrType::ORT_OP_ATTR_STRING) {
+      *out_size = fa->string_value.size();
```
huaychou, line 497: **[N-1] (Minor)** ReadOpAttr string stub does not document whether size includes the null terminator

The string probe path of the `ReadOpAttr` stub (`buf=nullptr, buf_size=0`) returns `fa->string_value.size()`, which excludes the null terminator.

The real ORT API's size semantics for string `ReadOpAttr` could be either "byte count with null" or "byte count without null". The stub chooses "without null", and the caller `GetValueImpl<std::string>::GetValue` (`onnxruntime_cxx_inline.h:2775`) calls `result.resize(size)` then memcpy-s size bytes — consistent in the current test context so **tests pass correctly**. However, the stub comment only describes the "probe-then-read pattern" without stating **whether size includes the null terminator**. A future contributor adding a new string-reading stub — or porting this one to another test binary — could assume it includes null and end up with a trailing '\0' in the string, causing comparison failures.

**Impact:** Current test behavior is correct; impact is limited to maintainability if the stub is ported or extended.

**Suggested fix:**

Add a one-line clarifying comment next to the string branch of the `ReadOpAttr` lambda:

`// size = byte count of string_value, NOT including a null terminator —
// consistent with how GetValueImpl<std::string> allocates: result.resize(size) + memcpy(size).`

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
```diff
@@ -0,0 +1,681 @@
...
+}
+
+TEST(QnnUnit_OnnxCtxModelHelperTest, TryGetMaxSpillFillSize_SingleContext_NoSwap) {
+  CtxHelperTestContext ctx;
+  FakeOpAttr max_size_attr = FakeOpAttr::MakeInt64(MAX_SIZE, 100);
+  FakeNode ep_ctx{"e", "EPContext", "", 1, {}, {}};
+  ep_ctx.attrs[MAX_SIZE] = &max_size_attr;
+  FakeGraph g{{ep_ctx}, {}, {}, {}};
+  const OrtGraph* graphs[] = {g.AsGraph()};
+  std::vector<int> pos_list = {0};
```
huaychou, line 385: **[T-2] (Nit)** TryGetMaxSpillFillSize missing test for num_nodes != 1 error path

The `TryGetMaxSpillFillSize` implementation (`onnx_ctx_model_helper.cc:264`) has a defensive `RETURN_IF(num_nodes != 1, ...)` check — returning an error when a subgraph does not contain exactly one node. All four existing test cases use subgraphs with exactly one EPContext node; the "subgraph with 0 nodes" error path is not covered.

**Suggested fix:**

Optionally add:

```
TEST(QnnUnit_OnnxCtxModelHelperTest, TryGetMaxSpillFillSize_GraphWithZeroNodes_ReturnsError) {
  CtxHelperTestContext ctx;
  FakeGraph g{{}, {}, {}, {}}; // 0 nodes
  const OrtGraph* graphs[] = {g.AsGraph()};
  std::vector<int> pos_list = {0};
  int64_t max_size = 0;
  auto status = TryGetMaxSpillFillSize(graphs, ctx.api, 1, max_size, pos_list);
  EXPECT_FALSE(status.IsOK());
}
```

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
```diff
@@ -0,0 +1,681 @@
...
+
+#include <vector>
+
+#include "core/providers/qnn/builder/onnx_ctx_model_helper.h"
+#include "core/providers/qnn/builder/qnn_model.h"
+#include "test/providers/qnn/unit/qnn_fake_ort_graph.h"
+#include "test/providers/qnn/unit/qnn_unit_test_utils.h"
+
+using namespace onnxruntime::qnn;
+using namespace onnxruntime::test;
```
huaychou: Just wondering, could we avoid "using namespace" to specify the namespace?

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
```diff
@@ -120,11 +122,46 @@ static_assert(offsetof(FakeOrtValue, elem_type) == offsetof(FakeValueInfo, elem_
 static_assert(offsetof(FakeOrtValue, shape) == offsetof(FakeValueInfo, shape),
               "FakeOrtValue/FakeValueInfo layout drift: shape offset mismatch");
 
+// ---------------------------------------------------------------------------
+// FakeOpAttr
```
huaychou, line 126: Just wondering, can we use a real OrtOpAttr for these unit tests?

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
yuhuchua-qti: Added in 26fc44baee: `TryGetMaxSpillFillSize_GraphWithMultipleNodes_ReturnsError`, using a two-node graph to hit the "exactly one EPContext node" guard. 
Thanks.

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
yuhuchua-qti: Clarified in 26fc44baee. 
For `ORT_OP_ATTR_STRING` the stub now sets `*out_size = string_value.size()` with a comment noting this excludes the null terminator, matching the real `ReadOpAttr` contract (ORT _custom_ops.cc_: STRING case does `*out = s.size()`, no `+1`) and how `Ort::ConstOpAttr::GetValue<std::string>` consumes it (`resize(size) + memcpy(size)`, no trailing `'\0'`). 
Thanks.

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
yuhuchua-qti: Done. 
Removed both `using namespace onnxruntime::qnn;` and `using namespace onnxruntime::test;`, wrapped the file in `namespace onnxruntime { namespace test {`, and added explicit `using onnxruntime::qnn::X;` declarations for each symbol used.

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
yuhuchua-qti: We considered it. 
A real `OrtOpAttr` is constructible via `CreateOpAttr`, but the node it hangs off (`FakeNode`) stays fake, so we'd end up with a real/fake hybrid plus extra lifetime management, for no added coverage — the EP only reads the attr back through `OrtApi` function pointers, which the fake already satisfies. We'd prefer to keep the harness uniformly fake. Thanks.

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
```diff
@@ -120,11 +122,46 @@ static_assert(offsetof(FakeOrtValue, elem_type) == offsetof(FakeValueInfo, elem_
...
+//
+// Acts as OrtOpAttr* in tests that exercise Node_GetAttributeByName /
+// OpAttr_GetType / ReadOpAttr stubs (e.g., OrtNodeAttrHelper::Get(...)).
+//
+// Holds a name + type + value. Currently supports ORT_OP_ATTR_STRING and
+// ORT_OP_ATTR_INT. Add more value fields if other attribute types are needed.
+// ---------------------------------------------------------------------------
+struct FakeOpAttr {
+  std::string name;
```
huaychou, line 137: **[T-3] (Nit)** FakeOpAttr::name field unused by stubs — needs a clarifying comment

`FakeOpAttr` (lines 134–136) declares a `name` field and both `MakeString` (line 178) and `MakeInt64` (line 185) populate it. However, the stub implementation (`Node_GetAttributeByName`, lines 826–830) never reads `FakeOpAttr::name` — attribute lookup is done via `fn.attrs.find(name)` keyed on the map entry, not on the `FakeOpAttr`'s own `name` field.

The field is harmless but the name/population pattern can mislead future maintainers into thinking stubs dispatch through it.

**Suggested fix:**

Add a one-line comment next to the `name` field, e.g.: `// For debugging only. Node_GetAttributeByName looks up attrs by map key, not this field.`

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
```diff
@@ -0,0 +1,723 @@
...
+// calls reach our stubs instead of the real ORT runtime.
+
+#include "gtest/gtest.h"
+
+#if !defined(ORT_MINIMAL_BUILD) && QNN_EP_INTERNAL_SYMBOL_ACCESS
+
+#include <vector>
+
+#include "core/providers/qnn/builder/onnx_ctx_model_helper.h"
+#include "core/providers/qnn/builder/qnn_model.h"
```
huaychou, line 27: **[T-1] (Nit)** Unnecessary #include of qnn_model.h

The test file includes `"core/providers/qnn/builder/qnn_model.h"` at line 22, but this include is redundant.

The test only needs `QnnModelLookupTable`, which is already defined in `onnx_ctx_model_helper.h` at line 19 (via a forward declaration of `class QnnModel` plus a using alias). Pulling in the full `qnn_model.h` drags in `QnnModelWrapper`, `QnnBackendManager`, `RpcmemLibrary`, and other heavy headers, adding unnecessary compile time.

**Suggested fix:**

Remove the line `#include "core/providers/qnn/builder/qnn_model.h"`; the other includes are sufficient.

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
huaychou: Acknowledge

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
huaychou: Acknowledge

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
huaychou: Acknowledge

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
```diff
@@ -0,0 +1,723 @@
...
+  EXPECT_FALSE(status.IsOK());
+}
+
+TEST(QnnUnit_OnnxCtxModelHelperTest, GetEpContextDlcPath_DlcNodeWithPath_ReturnsLowercasedPath) {
+  CtxHelperTestContext ctx;
+  FakeOpAttr source = FakeOpAttr::MakeString(SOURCE, "qnn");
+  FakeOpAttr ctx_type = FakeOpAttr::MakeString(EP_CONTEXT_TYPE, "dlc");
+  // Mixed-case input pins the source's GetLowercaseString call; an
+  // already-lowercase input could not distinguish it.
+  FakeOpAttr dlc_ctx = FakeOpAttr::MakeString("ep_dlc_context", "Path/To/Model.DLC");
```
huaychou: nit: we can add this in `onnx_ctx_model_helper.h`:
```C++
static const std::string EP_DLC_CONTEXT = "ep_dlc_context";
```

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
```diff
@@ -0,0 +1,723 @@
...
+using onnxruntime::qnn::GraphHasEpContextNode;
+using onnxruntime::qnn::IO_NAME_OVERRIDES;
+using onnxruntime::qnn::IsOrtGraphHasCtxNode;
+using onnxruntime::qnn::IsOrtGraphHasDlcCtxNode;
+using onnxruntime::qnn::MAIN_CONTEXT;
+using onnxruntime::qnn::MAX_SIZE;
+using onnxruntime::qnn::ParseIoNameOverrides;
+using onnxruntime::qnn::QnnModelLookupTable;
+using onnxruntime::qnn::SOURCE;
+using onnxruntime::qnn::TryGetMaxSpillFillSize;
```
huaychou: Would using a namespace alias be a better approach?
```C++
namespace qnn = onnxruntime::qnn;
qnn::EMBED_MODE
```

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
```diff
@@ -0,0 +1,723 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+//
+// Function-level unit tests for onnx_ctx_model_helper.cc.
+//
+// Tests the EP-context-node-detection helpers using FakeGraph / FakeNode /
```
huaychou, line 6: Do we need to add unit tests for multi_soc_ep_context?

onnxruntime/test/providers/qnn/unit/qnn_fake_ort_graph.h
yuhuchua-qti: Agreed — added a comment in b5e0129839 noting the field is debug-only: `Node_GetAttributeByName` dispatches via the `FakeNode::attrs` map key, not `FakeOpAttr::name`. 
Thanks.

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
yuhuchua-qti: Thanks for flagging — I checked, and this include is actually required, not redundant. 
Several tests declare `QnnModelLookupTable models;` locals (e.g. the `GetEpContextFromMainNode_*` cases). 
`QnnModelLookupTable` is `unordered_map<string, unique_ptr<QnnModel>>`, so destroying such a local instantiates `unique_ptr<QnnModel>::~unique_ptr`, which needs the complete `QnnModel` type. 
`onnx_ctx_model_helper.h` only forward-declares `QnnModel`, so dropping the include fails to compile: `invalid application of 'sizeof' to incomplete type 'onnxruntime::qnn::QnnModel'`. 
So I've kept the include as-is.

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
yuhuchua-qti: Done in b912bc77ed. Added EP_DLC_CONTEXT next to the other EPContext attribute-name constants; GetEpContextDlcPath (.cc:79) and the DLC-path tests now share it instead of each hardcoding "ep_dlc_context" — removes the drift risk where the attr name could change in one place and silently pass the test.

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
yuhuchua-qti: Done in b5e0129839. 
Added a `namespace qnn = onnxruntime::qnn;` alias and shortened the using-declarations to `using qnn::X;`. 
Thanks.

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
yuhuchua-qti: Done in b5e0129839. 
Added `EP_DLC_CONTEXT` next to the other EPContext attribute-name constants; `GetEpContextDlcPath` (`onnx_ctx_model_helper.cc:79`) and the DLC-path tests now share it instead of each hardcoding `"ep_dlc_context"` independently — removes the drift risk where the attr name could change in one place and the test would silently still pass.
Thanks.

onnxruntime/test/providers/qnn/unit/onnx_ctx_model_helper_test.cc
yuhuchua-qti: Added. 
The only multi-SoC behaviour observable in this file is the serialization of the flag, so I tested that: two `CreateEPContextNodes` tests assert the `IS_MULTI_SOC_EP_CONTEXT` attribute is written with value 1 when `enable_multi_soc_ep_context=true` and `0` when `false` (captured via a `CreateOpAttr` stub).

The read side (`GetEpContextFromMainNode`, `.cc:166`) only reads the same attribute back and forwards it unchanged to `LoadCachedQnnContextFromBuffer` — no branch in this file varies on it. 
The actual behavioural divergence — multi-SoC routes the context binary through the DLC path (`system_dlc_plugin_->GetDlcBinaryBuffer`) instead of a single-SoC QNN context blob — lives in `QnnBackendManager` (`qnn_backend_manager.cc:1590/1631`), so that coverage belongs in `qnn_backend_manager_test.cc` rather than here.

huaychou (approved):
LGTM

qti-chuteng (approved):

PR #581, opened 2026-06-29
[QNN EP] Remove Dummy Bias tensor creation in RMS Norm

qti-mbadnara:
## Description

Removes the dummy all-zeros beta tensor that was unconditionally synthesized for `QNN_OP_RMS_NORM` on the HTP backend. SDK 2.49+ accepts beta as an optional input, so the workaround is only compiled in when building against SDK < 2.49.

---

## Motivation & Context

`QNN_OP_RMS_NORM` on the HTP backend previously required an explicit beta/bias input even though the ONNX `RMSNormalization` op has no bias parameter. To satisfy QNN's input contract, the op builder synthesized a dummy all-zeros beta tensor for every NPU graph build.

QAIRT SDK 2.49 relaxed this requirement, beta is now a truly optional input. The dummy tensor is no longer needed and adds unnecessary graph weight.

**Why `QNN_SDK_VERSION_MINOR` instead of `QNN_API_VERSION_MINOR`:**

SDK 2.47 and SDK 2.48 both define `QNN_API_VERSION_MINOR = 36`. The QNN API version did not increment between these two releases. `QNN_API_VERSION_MINOR` cannot distinguish them at compile time. Instead, `QNN_SDK_VERSION_MINOR` is injected by CMake by parsing the `version` field from `sdk.yaml` at configure time, which correctly resolves to `47` vs `48`.

The `!defined(QNN_SDK_VERSION_MINOR)` fallback in the `#if` guard preserves the old behavior (dummy tensor always added) when the SDK version cannot be determined at configure time.

---

## Files Changed

| File | Change |
|------|--------|
| `builder/opbuilder/rmsnormalization_op_builder.cc` | Wrapped dummy beta synthesis in `#if !defined(QNN_SDK_VERSION_MINOR) \|\| QNN_SDK_VERSION_MINOR < 48`. Added VERBOSE log on both paths indicating whether the dummy tensor is added or skipped. |
| `cmake/onnxruntime_providers_qnn.cmake` | Parse `QNN_SDK_VERSION` string (from `sdk.yaml`) into `QNN_SDK_VERSION_MAJOR` / `QNN_SDK_VERSION_MINOR` compile definitions on the `onnxruntime_providers_qnn` target. |

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
@@ -87,15 +87,15 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
   RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[X_IDX], logger, input_names));
   RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[SCALE_IDX], logger, input_names));
 
-#if !defined(QNN_SDK_VERSION_MINOR) || QNN_SDK_VERSION_MINOR < 48
-  // QNN SDK < 2.48 requires an explicit beta/bias input for QNN_OP_RMS_NORM on NPU.
-  // SDK 2.48+ accepts beta as optional, so the dummy tensor is only needed for older SDKs.
+#if !defined(QNN_SDK_VERSION_MINOR) || (QNN_SDK_VERSION_MAJOR == 2 && QNN_SDK_VERSION_MINOR < 49)
+  // QNN SDK < 2.49 requires an explicit beta/bias input for QNN_OP_RMS_NORM on NPU.
+  // SDK 2.49+ accepts beta as optional, so the dummy tensor is only needed for older SDKs.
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
```diff
@@ -87,9 +87,15 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
   RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[X_IDX], logger, input_names));
   RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, inputs[SCALE_IDX], logger, input_names));
 
-  // Create dummy beta tensor for NPU backend
+#if !defined(QNN_SDK_VERSION_MINOR) || QNN_SDK_VERSION_MINOR < 48
```
qti-yuduo: Does it hurt adding a dummy one? Is it still compatible?

qti-mbadnara: When the SDK used is < 2.49 I think we need to add a dummy tensor for it to work with QNN

qti-yuduo (approved):

PR #582, opened 2026-06-30
Fix incorrectly reusing bias initializer in BN

yath1:
### Description
tldr: Fixes BN layer's bias and scale inputs being incorrectly computed using the the ONNX op's inputs.

ONNX BN op has the following inputs: Input Activation, Scale, Bias, Input_mean and Input_variance. 
HTP BN op has the following inputs: Input Activation, Scale_htp, Bias_htp 

Scale_htp is computed using Scale / sqrt(Input_variance+ eps)
Bias_htp is computed using Bias - Input_mean * Scale / sqrt(Input_variance+ eps)

BN Op builder skips computing bias (ie Bias_htp) above if a tensor already exists with a given name. In Sinet, two BN ops share the same bias initializer but their Input_mean and Input_variance are different. Once one of the BN is processed and its Bias_htp is computed, the second BN op would silently use the same Bias_htp as the first op because they share the same name leading to incorrect output. 

### Motivation and Context
Sinet w8a8 shows an on-target accuracy of 18.5, expected 92.8. After this fix, the accuracy drop is completely recovered.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
@@ -595,8 +595,15 @@ Ort::Status BatchNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_mode
   // QNN only accept 3 input. We need to first combine mean and variance into scale and bias.
   //
   {
-    const std::string& scale_name = inputs[1].name;
-    const std::string& bias_name = inputs[2].name;
+    // The fused scale/bias QNN tensors are specific to each bn node: PreprocessScale/PreprocessBias fold in
+    // this BN's mean/variance. Naming them after the source ONNX initializer (inputs[1]/inputs[2])
+    // causes a collision when two BN nodes share an initializer (e.g. AIMET-quantized models that
+    // share a saturated int32 bias init): the second node finds the name already registered, skips
+    // recompute, and silently reuses the first node's fused params. Use a per-node-unique name so
+    // each BN gets its own fused scale/bias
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -595,8 +595,15 @@ Ort::Status BatchNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_mode
...
-    const std::string& bias_name = inputs[2].name;
+    // The fused scale/bias QNN tensors are specific to each bn node: PreprocessScale/PreprocessBias fold in
+    // this BN's mean/variance. Naming them after the source ONNX initializer (inputs[1]/inputs[2])
+    // causes a collision when two BN nodes share an initializer (e.g. AIMET-quantized models that
+    // share a saturated int32 bias init): the second node finds the name already registered, skips
+    // recompute, and silently reuses the first node's fused params. Use a per-node-unique name so
+    // each BN gets its own fused scale/bias
+    const std::string node_prefix = std::to_string(node_unit.Index()) + "_" + node_unit.Name();
+    const std::string scale_name = node_prefix + "_bn_fused_scale";
+    const std::string bias_name = node_prefix + "_bn_fused_bias";
```
qti-yuduo, line 606: Should we utilize unique name generator here?

qti-yuduo (approved):
LGTM

onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
yath1: These are unique names, but it makes sense to keep it uniform and use UniqueNameGenerator. Lemme update this as part of next PR I am working on now.

PR #232, opened 2026-04-10
add reshape-gemm-reshape-reshape fusion

quic-muchhsu:
### Description

Add 2 QNN group node fusion to reshape_gemm_fusion:
1. Reshape - Gemm - Reshape
The ONNX pattern:
`Nd -> matmul  -> add -> Nd `
ORT-core optimize to
`Nd -> reshape_in -> 2d -> GEMM -> 2d -> reshape_out -> Nd`
optimize to QNN pattern:
`Nd -> FC  -> 2d -> reshape_out`

2. Reshape - Gemm - Reshape - Reshape
The ONNX pattern:
`Nd -> matmul  -> add -> Nd -> reshape_original`
ORT-core optimize to
`Nd -> reshape_in -> 2d -> GEMM  -2d -> reshape_out -> Nd -> reshape_original`
optimize to QNN pattern:
`Nd -> FC  -> 2d -> reshape_original`

The reshape - gemm - reshape is a very common pattern when ORT-core optimizes matmul-add, and since ort requires gemm to have 2d input and output, ort-core will add reshape before and after gemm. However fc in QNN is more flexable, it accepts any input rank/shape as long as it can be reshaped to [batch_size. N] where N in fc's weight input dim.

Adding reshape - gemm - reshape - reshape pattern as well because matmul - add - reshape is another popular pattern in attention models.

Another reason for this change, we observed HTP compiler has a long chain optimization for attention modules, and it relies on certain pattern matching to hit the optimization. I have belief from previous experiments that the long chain attention pattern expects a `Nd -> FC -> 2d` pattern. That's why we observed 44x perf imrpovement.

### Motivation and Context

With this change, this model's performance can be improved from 800 ms to 28ms
834 ms job w/o fusion: https://dev.aihub.qualcomm.com/jobs/jgjz34085 (w/ qairt2.43)
829 ms job w/o fusion: https://dev.aihub.qualcomm.com/jobs/jpyj37z0p/ (w/ qairt2.45)
28 ms job w/ with fusion: https://dev.aihub.qualcomm.com/jobs/jgokq134p (w/ qairt 2.45)

Other experimental jobs:
904ms job with `ReshapeGemmReshapeReshapeFusion` disabled: https://dev.aihub.qualcomm.com/jobs/jpr23n6k5/ (w/ qairt 2.45)

PR #583, opened 2026-06-30
[QNN EP] Fix ReshapeGemmFusion claiming shared Reshape with multiple consumers

yuhuchua-qti:
### Description

In ReshapeGemmFusionGroup::TryFusion2/3/4, add a guard to skip fusion when the input Reshape's output has more than one consumer.

When ORT's CSE (Common Subexpression Elimination) optimization runs before graph partitioning, it can merge two structurally-identical Reshape nodes
into one shared node. If such a shared Reshape feeds multiple Gemm nodes, the fusion logic would incorrectly claim it — removing the shared Reshape
from the graph and breaking every other consumer's input edge.

The fix conservatively skips fusion in this case. A future optimization could duplicate the Reshape per consumer to re-enable fusion.

**Files changed:**
- onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_gemm_fusion.cc — single-consumer guard in TryFusion2, TryFusion3, TryFusion4
- onnxruntime/test/providers/qnn/qnn_node_group/reshape_gemm_fusion_test.cc — negative test ReshapeGemmFusion_Negative_SharedReshape

**Kill-test fence:** 
ReshapeGemmFusion_Negative_SharedReshape was verified to fail when the InputReshapeIsShared guard is temporarily removed — the second Gemm's input tensor has already been claimed by the first fusion group, causing a Compile error. This confirms the test is not vacuously passing.

### Motivation and Context

This is an edge case missed in the initial implementation of ReshapeGemmFusionGroup (introduced in #232). The fusion logic did not account for the
possibility that ORT graph optimizations (e.g., CSE) could produce a shared Reshape consumed by multiple Gemms. Without this fix, such a model would
produce a broken QNN graph during Compile because the second Gemm's input tensor has already been claimed and removed by the first fusion group.

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_node_group/reshape_gemm_fusion_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_node_group/reshape_gemm_fusion_test.cc
@@ -358,6 +358,50 @@ TEST_F(QnnHTPBackendTests, ReshapeGemmReshapeFusion_Transformer) {
 // Negative Tests - Fusion should NOT happen
 // ============================================================================
 
+// Build a test case where input Reshape has two consumers (fusion should not happen).
+// Graph: Input -> Reshape -> Gemm1 (output1)
+//                        \-> Gemm2 (output2)
+// Claiming the shared Reshape for Gemm1 would break Gemm2, so fusion must not fire.
+GetTestModelFn BuildReshapeSharedByTwoGemmsTestCase() {
+  return [](ModelTestBuilder& builder) -> void {
+    builder.graph_->set_name("reshape_shared_two_gemms_graph");
+
+    auto input_def = TestInputDef<float>({1, 4, 8}, false, -1.0f, 1.0f);
+    MakeTestInput<float>(builder, "input", input_def);
+
+    // Shared Reshape: [1, 4, 8] -> [4, 8]
+    builder.Make1DInitializer<int64_t>("reshape_shape", {4, 8});
+    builder.AddNode("reshape", "Reshape", {"input", "reshape_shape"}, {"reshape_out"}, kOnnxDomain);
+
+    // Gemm1: [4, 8] x [8, 16] -> [4, 16]
+    builder.MakeInitializer<float>("weight1", {8, 16}, -0.5f, 0.5f);
+    builder.MakeInitializer<float>("bias1", {16}, -0.1f, 0.1f);
+    builder.AddNode("gemm1", "Gemm", {"reshape_out", "weight1", "bias1"}, {"output1"}, kOnnxDomain);
+
+    // Gemm2: [4, 8] x [8, 32] -> [4, 32]  — shares reshape_out
+    builder.MakeInitializer<float>("weight2", {8, 32}, -0.5f, 0.5f);
+    builder.MakeInitializer<float>("bias2", {32}, -0.1f, 0.1f);
+    builder.AddNode("gemm2", "Gemm", {"reshape_out", "weight2", "bias2"}, {"output2"}, kOnnxDomain);
+
+    builder.MakeOutput("output1");
+    builder.MakeOutput("output2");
+  };
+}
+
+// Test: Fusion should NOT happen when the input Reshape has two consumers.
+// If fusion incorrectly claimed the shared Reshape for one Gemm, the other Gemm
+// would lose its input and the model would fail. All nodes must still run on QNN EP.
+TEST_F(QnnHTPBackendTests, ReshapeGemmFusion_Negative_SharedReshape) {
+  SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V68);
+
+  ProviderOptions provider_options = GetProviderOptions();
+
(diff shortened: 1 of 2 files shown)


minfhong-qti (approved):

qti-chuteng (approved):

onnxruntime/test/providers/qnn/qnn_node_group/reshape_gemm_fusion_test.cc
```diff
@@ -358,6 +358,50 @@ TEST_F(QnnHTPBackendTests, ReshapeGemmReshapeFusion_Transformer) {
...
+
+    builder.MakeOutput("output1");
+    builder.MakeOutput("output2");
+  };
+}
+
+// Test: Fusion should NOT happen when the input Reshape has two consumers.
+// If fusion incorrectly claimed the shared Reshape for one Gemm, the other Gemm
+// would lose its input and the model would fail. All nodes must still run on QNN EP.
+TEST_F(QnnHTPBackendTests, ReshapeGemmFusion_Negative_SharedReshape) {
```
huaychou, line 394: **[M-1] (Major)** Negative test does not verify fusion was suppressed (missing AssertOpInQnnGraph)

The new `ReshapeGemmFusion_Negative_SharedReshape` test only uses `ExpectedEPNodeAssignment::All` to verify that all nodes run on QNN EP, but does NOT verify that fusion was actually suppressed (i.e., that no `FullyConnected` node was emitted).

By contrast, the existing `ReshapeGemmFusion_Negative_Rank5Input` test (lines 572–591 in the same file) establishes the expected pattern: set up a QNN graph dump directory, run `RunQnnModelTest`, then call `AssertOpInQnnGraph(dir, "FullyConnected", 0)` (or an equivalent count) to confirm no FullyConnected was emitted.

**Impact:** If fusion accidentally fires but produces numerically correct results, this test will pass vacuously (false pass). The test cannot confirm that the `InputReshapeIsShared` guard actually took effect — it only confirms the model runs correctly. The kill-test fence described in the PR description is only a manual verification and is not self-evident from the test structure.

**Suggested fix:**

Follow the pattern established by `ReshapeGemmFusion_Negative_Rank5Input` — add a QNN graph dump validation step:

1. Declare and clean up `json_qnn_graph_dir`
2. Set `dump_json_qnn_graph=1` and `json_qnn_graph_dir` in `provider_options`
3. After `RunQnnModelTest`, add:
   `AssertOpInQnnGraph(json_qnn_graph_dir, "FullyConnected", 0);`
   (confirm no FullyConnected node was emitted)
   or
   `AssertOpInQnnGraph(json_qnn_graph_dir, "Gemm", 2);`
   (confirm both Gemms run as standalone ops)

huaychou, line 394: **[N-1] (Minor)** Negative test only covers TryFusion2; TryFusion3 and TryFusion4 shared-Reshape scenarios are untested

The new test graph is `Reshape → Gemm1` and `Reshape → Gemm2` (2-node pattern), which only triggers `TryFusion2`. `TryFusion3` (Reshape→Gemm→Reshape) and `TryFusion4` (Reshape→Gemm→Reshape→Reshape) also received the `InputReshapeIsShared` guard, but have no corresponding negative test.

All three paths are logically equivalent (they call the same `InputReshapeIsShared` helper), but each has its own `GetInputReshapeNodeUnit` + guard sequence — code review cannot rule out a typo or omission in one of the paths.

**Impact:** If the guard in TryFusion3 or TryFusion4 is accidentally removed during a future refactor, no regression test will catch the regression.

**Suggested fix:**

Add at least one more negative test covering the 3-node shared-Reshape scenario (Reshape→Gemm→Reshape, where the input Reshape has two Gemm consumers). This can be added to the same test function or as a new `TEST_F(QnnHTPBackendTests, ReshapeGemmFusion_Negative_SharedReshape_Fusion3)`.

yuhuchua-qti: Agreed. Added `AssertOpInQnnGraph` validation (graph dump + op count assertion) to `ReshapeGemmFusion_Negative_SharedReshape`:

- `AssertOpInQnnGraph(dir, "Reshape", 1)` — shared Reshape remains standalone
- `AssertOpInQnnGraph(dir, "FullyConnected", 2)` — both Gemms run as independent FC nodes

Follows the same pattern as `ReshapeGemmFusion_Negative_Rank5Input`. See follow-up PR #592 .

yuhuchua-qti: Agreed. Added negative tests for both remaining paths:

| Test | Pattern | Reshape | FC |
|------|---------|---------|-----|
| `ReshapeGemmReshapeFusion_Negative_SharedReshape` | TryFusion3 | 3 | 2 |
| `ReshapeGemmReshapeReshapeFusion_Negative_SharedReshape` | TryFusion4 | 5 | 2 |

All three TryFusion variants now have dedicated shared-Reshape regression tests. See follow-up PR #592.

PR #476, opened 2026-06-03
[QNN EP] MatMul BQ Phase 1: support MatMul with block-quantized weight on HTP

huaychou:
### Description

Adds BW_FLOAT_BLOCK (block-quantized) weight support to the QNN EP MatMul op builder, mirroring the merged Conv BQ PR (#429). QNN HTP requires the BQ MatMul to run with 4-D tensors, so the implementation:
  - Dequantizes INT16 activation to FP16
  - Reshapes activation [..., M, K] → [batch, 1, M, K]
  - Adds weight as [1, 1, K, N] SFIXED_POINT_8 with BW_FLOAT_BLOCK quant (block_sizes = {1, 1, block_size, 1}, transpose_in1 = 0)
  - Transposes ONNX [num_blocks, N] scales to QNN [N, num_blocks] order
  - Reshapes 4-D FP16 MatMul output back to [..., M, N]
  - Re-quantizes FP16 output to INT16

### Motivation and Context

- Supports INT4/UINT4/INT8/UINT8 weights with HTP block_size constraints ({4-bit: ×8, 8-bit: ×4}). Requires QAIRT >= 2.47.

PR #584, opened 2026-06-30
[QNN EP] Consolidate duplicated BQ helpers across op builders

huaychou:
### Description

Add a shared qnn::bq utility (qnn_bq_utils.{h,cc}) and route the conv, matmul, and gemm builders through it:
- GetBQBitwidth and the {2:16,4:8,8:4} bits/block-size constraint map
- ValidateBQBitwidthAndBlockSize (HTP block_size divisor check)
- ResolveBlockSize: derives block_size from the scale shape (contraction_dim / num_blocks) and cross-checks against the DQ/Q "block_size" attribute introduced by PR307 — rejects a malformed model where the two disagree, is a no-op when the attribute is absent
- IsUnsignedBQType / MaybeShiftUnsignedWeightToSigned
- ComputeBQOffsets (unsigned_bias - onnx_zp convention)
- TransposeBlockMajorToChannelMajor ([num_blocks,N] -> [N,num_blocks])
- IsBlockedScale (shared tail of the per-op weight detection)
- AddInt16ToFp16DequantForActivation (INT16 -> FP16 activation dequant)

### Motivation and Context

The block-quantization PRs (Conv #429, MatMul #476, Gemm #477) each copy-pasted the same BW_FLOAT_BLOCK weight plumbing into their op builders, so a fix to a BQ constraint, dequant pattern, or scale layout had to be applied in three places.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
@@ -1,0 +1,190 @@
+// Copyright (c) Qualcomm. All rights reserved.
+// Licensed under the MIT License.
+
+#include "core/providers/qnn/builder/qnn_bq_utils.h"
+
+#include <cassert>
+#include <string>
+#include <vector>
+
+#include "QnnOpDef.h"
+
+#include "core/providers/qnn/builder/qnn_def.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/providers/qnn/common/inlined_containers.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+namespace {
+// HTP BQ: supported weight bitwidths (2/4/8) mapped to their block_size divisor
+// constraint — block_size must be a multiple of the corresponding value (same as
+// the MatMulNBits HTP constraints).
+const InlinedHashMap<uint32_t, int64_t> kHtpBQBitsAndBlockSizeMultipliers{
+    {2, 16}, {4, 8}, {8, 4}};
+}  // namespace
+
+namespace bq {
+uint32_t GetBQBitwidth(ONNXTensorElementDataType onnx_type) {
+  switch (onnx_type) {
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT2:
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT2:
+      return 2;
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT4:
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4:
+      return 4;
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT8:
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8:
+      return 8;
+    default:
+      return 0;
+  }
+}
+
+bool IsUnsignedBQType(ONNXTensorElementDataType onnx_type) {
+  return onnx_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT2 ||
+         onnx_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4 ||
+         onnx_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8;
+}
+
+Ort::Status ValidateBQBitwidthAndBlockSize(uint32_t bitwidth, int64_t block_size, const char* op_tag) {
+  auto bq_it = kHtpBQBitsAndBlockSizeMultipliers.find(bitwidth);
+  RETURN_IF(bq_it == kHtpBQBitsAndBlockSizeMultipliers.end(),
+            ("QNN HTP " + std::string(op_tag) + " BQ: unsupported weight bitwidth=" +
+             std::to_string(bitwidth))
+                .c_str());
+  RETURN_IF(block_size % bq_it->second != 0,
+            ("QNN HTP " + std::string(op_tag) + " BQ: block_size=" + std::to_string(block_size) +
+             " must be a multiple of " + std::to_string(bq_it->second) +
+             " for " + std::to_string(bitwidth) + "-bit weight")
+                .c_str());
+  return Ort::Status();
+}
+
+Ort::Status ResolveBlockSize(const OrtNodeUnitIODef& weight, int64_t contraction_dim,
+                             int64_t num_blocks, const char* op_tag,
+                             /*out*/ int64_t& block_size) {
+  RETURN_IF(num_blocks <= 0 || contraction_dim % num_blocks != 0,
+            ("QNN EP: BQ " + std::string(op_tag) +
+             ": contraction dim must be a positive multiple of num_blocks")
+                .c_str());
+  block_size = contraction_dim / num_blocks;
+
+  // Since PR307 the DQ/Q "block_size" attribute is surfaced on quant_param->block_size. When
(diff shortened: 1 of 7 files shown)


onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+
+bool IsUnsignedBQType(ONNXTensorElementDataType onnx_type) {
+  return onnx_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT2 ||
+         onnx_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4 ||
+         onnx_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8;
+}
+
+Ort::Status ValidateBQBitwidthAndBlockSize(uint32_t bitwidth, int64_t block_size, const char* op_tag) {
```
minfhong-qti:
```suggestion
Ort::Status ValidateBQBitwidthAndBlockSize(uint32_t bitwidth, int64_t block_size, const std::string& op_tag) {
```
Then no need to manually convert, and as well for others.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -0,0 +1,170 @@
...
+    RETURN_IF(attr_block_size != block_size,
+              ("QNN EP: BQ " + std::string(op_tag) + ": block_size attribute (" +
+               std::to_string(attr_block_size) + ") disagrees with scale-derived block_size (" +
+               std::to_string(block_size) + ")")
+                  .c_str());
+  }
+  return Ort::Status();
+}
+
+bool IsBlockedScale(gsl::span<const int64_t> scale_shape,
```
minfhong-qti: Just curious. Why `blocked` instead of `block`? As well for parameter `blocked_axis`.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -0,0 +1,170 @@
...
+    RETURN_IF_NOT(static_cast<int64_t>(zp_values.size()) == count,
+                  "QNN EP: BQ zero_point size mismatch");
+    for (size_t idx = 0; idx < zp_values.size(); ++idx) {
+      offsets[idx] = unsigned_bias - static_cast<float>(zp_values[idx]);
+    }
+  }
+  return Ort::Status();
+}
+
+Ort::Status TransposeBlockMajorToChannelMajor(gsl::span<const float> in, int64_t num_blocks, int64_t N,
```
minfhong-qti: I think there is an existing function for this which has an optimization for validation only. https://github.com/onnxruntime/onnxruntime-qnn/blob/main/onnxruntime/core/providers/qnn/builder/qnn_utils.h#L592

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -905,6 +905,49 @@ Ort::Status QnnModelWrapper::AddCastNode(const std::string& cast_node_name,
...
+                                             std::vector<uint32_t> output_shape,
+                                             bool do_op_validation) {
+  RETURN_IF(output_data_type != QNN_DATATYPE_SFIXED_POINT_4 &&
+                output_data_type != QNN_DATATYPE_UFIXED_POINT_4 &&
+                output_data_type != QNN_DATATYPE_SFIXED_POINT_8 &&
+                output_data_type != QNN_DATATYPE_UFIXED_POINT_8 &&
+                output_data_type != QNN_DATATYPE_SFIXED_POINT_16 &&
+                output_data_type != QNN_DATATYPE_UFIXED_POINT_16 &&
+                output_data_type != QNN_DATATYPE_SFIXED_POINT_32 &&
+                output_data_type != QNN_DATATYPE_UFIXED_POINT_32,
```
minfhong-qti: Can we collect these somewhere as `IsQnnFxpDataType` or a container `QnnFxpDataType`?

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
huaychou: Fixed

minfhong-qti (approved):

qti-chuteng:
### 🟠 Major

**[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:L871`
BQ MatMul output tail passes `op_output_shape` by `std::move` into `bq::AddFp16ToInt16QuantizeOutput()`, but the `if (reshape_output)` block at L884 reads `op_output_shape` again for `AddReshapeNode`. The original code passed a copy here, never a move — this is a use-after-move regression introduced by the refactor. It is reachable for a rank-1 activation × BQ-matrix MatMul that needs an output reshape (`reshape_input_0` true while `is_bq_matmul` true), producing a wrong/empty reshape shape and an incorrect graph or QNN compile failure.

---

### 🔵 Minor

**[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc:L928`
The shared `AddDequantizeNode`/`AddQuantizeNode` helpers changed the inserted node's unique-name suffix from `_int16_dequantize`/`_fp16_quantize` (based on `act_name`/`output_name`) to `_dequantize`/`_quantize` (based on the helper's `input_name`). Functionally harmless, but it changes generated QNN node names — for a "pure refactor" PR this shows up in QnnSaver/QnnIr dumps and name-keyed debugging. Either thread a suffix parameter through the helper to preserve the old names, or note the change as intentional in the PR description.

huaychou:
> ### 🟠 Major
> **[M-1]** `onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc:L871` BQ MatMul output tail passes `op_output_shape` by `std::move` into `bq::AddFp16ToInt16QuantizeOutput()`, but the `if (reshape_output)` block at L884 reads `op_output_shape` again for `AddReshapeNode`. The original code passed a copy here, never a move — this is a use-after-move regression introduced by the refactor. It is reachable for a rank-1 activation × BQ-matrix MatMul that needs an output reshape (`reshape_input_0` true while `is_bq_matmul` true), producing a wrong/empty reshape shape and an incorrect graph or QNN compile failure.

Fixed

> ### 🔵 Minor
> **[N-1]** `onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc:L928` The shared `AddDequantizeNode`/`AddQuantizeNode` helpers changed the inserted node's unique-name suffix from `_int16_dequantize`/`_fp16_quantize` (based on `act_name`/`output_name`) to `_dequantize`/`_quantize` (based on the helper's `input_name`). Functionally harmless, but it changes generated QNN node names — for a "pure refactor" PR this shows up in QnnSaver/QnnIr dumps and name-keyed debugging. Either thread a suffix parameter through the helper to preserve the old names, or note the change as intentional in the PR description.

This is a tradeoff

qti-chuteng (approved):

PR #585, opened 2026-06-30
Fix arm64ReproDir to use CACHE PATH in cmake/arm64x.cmake

kuanyul-qti:
### Description
Make arm64ReproDir a CMake cache variable so that passing `-Darm64ReproDir=<path>` at configure time is respected.

### Motivation and Context
`cmake/arm64x.cmake` declared `arm64ReproDir` with a plain `set()` call, which CMake does not expose as a cache variable. As a result, passing `-Darm64ReproDir=<path>` at configure time was silently ignored — CMake's `-D` flag only overrides `CACHE` variables.

Patch `0003` already applied the correct fix to the ort_core dependency's copy of the same file. This PR applies the matching fix to the fork's own `cmake/arm64x.cmake`, which is included directly when building with `BUILD_AS_ARM64X`.


kuanyul-qti:
Nightly workflow to test arm64x build: https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/28429486657

qti-chuteng (approved):

PR #586, opened 2026-06-30
Support Clip with DequantizeLinear-wrapped constant min/max

qti-yuduo:
## Summary
- ClipOpBuilder now accepts Clip nodes whose min/max inputs come from a standalone `DequantizeLinear` over a constant initializer (a common AIMET-export pattern where the data input is bare-float and so the Clip is not formed into a QDQ NodeUnit).
- The existing `FoldConstantDequantizeLinear` pass already folds those DQs into STATIC fp32 tensors before the Clip op builder runs, so the fix is: replace `IsConstantInput` with `IsEffectivelyConstantInput` in `ExplicitOpCheck`, and read the dequantised scalar from the folded `QnnTensorWrapper` in `ProcessClipMinMax`.
- Exports `GetEffectivelyConstantTensorBytes` from `qdq_constant_folding.h` for the new read path.
- Adds two HTP regression tests (uint16 zp=65535 and uint8 zp=128) that exercise asymmetric zero_point so the dequant-sign convention can't silently regress.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc
@@ -6,6 +6,7 @@
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
 #include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
+#include "core/providers/qnn/builder/opbuilder/qdq_constant_folding.h"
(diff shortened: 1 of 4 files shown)


yath1 (approved):
I am curious how did this issue show up? This change might also be needed for other ops like prelu for example

qti-yuduo:
>I am curious how did this issue show up? This change might also be needed for other ops like prelu for example

Discovered from client model.

PR #587, opened 2026-06-30
Fix QNN MatMul singleton weight lowering

qti-yuduo:
## Summary
Route static MatMul weights that carry only leading unit dimensions (for example `[1,1,K,N]`) to the existing FullyConnected lowering path, since `[1,...,1,K,N]` is exactly equivalent to `[K,N]` under ONNX broadcasting.

## Why
On QAIRT 2.35 / HTP V73, QNN `MatMul` **passes op validation but fails graph finalization** with `QNN_COMMON_ERROR_MEM_ALLOC` (Code 1002) whenever `input[1]` carries a per-axis (`QNN_QUANTIZATION_ENCODING_AXIS_SCALE_OFFSET`) encoding **and** the effective batch dim is > 1. A 9-variant factor screen showed this is independent of N (fails at N=42 and N=420), of weight bit-width (int8 and int16), and of whether the leading dims are 1 or a real batch. Per-tensor weights pass at the same shapes, and per-channel passes at batch 1.

The shape itself is supported; the trigger is the per-axis weight combined with a batch. `FullyConnected` flattens the batch into rows, so the failing pattern never reaches the backend.

This is a targeted workaround for the squeezable subset. A genuinely batched per-channel weight such as `[1,6,3,420]` still fails finalize and is not addressed here — squeezing is not semantics-preserving when the leading dims are not 1. See the discussion thread for the full screen.

## Changes
- `matmul_op_builder.cc`: one routing decision (`squeeze_static_weight_to_2d`) computed in `CheckInputs` and consumed downstream; the squeeze is gated on the quantization-axis remap actually succeeding, so a weight it cannot handle stays on the existing QNN MatMul path rather than failing the node.
- `qnn_quant_params_wrapper.h`: new `HandleSqueeze`, alongside the existing `HandleTranspose` / `HandleUnsqueeze`, remapping the quantization axis across the squeeze for `AXIS_SCALE_OFFSET`, `BW_AXIS_SCALE_OFFSET` and `BLOCKWISE_EXPANSION`.
- `qnn_utils.h`: shared `TrySqueezeLeadingOnesTo` helper; the duplicate local copy in `rmsnormalization_op_builder.cc` now uses it.
- Tests: HTP regression test for the AI Hub failure shape `[1,6,3,3] x [1,1,3,420]` (uint16 activation/output, static per-channel int16 weight) asserting `FullyConnected` is emitted and `MatMul` is not; a CPU test covering rank-2 `input_0` with a rank > 2 weight; and unit tests for `HandleSqueeze` including the K=1 axis-collision case.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -18,17 +18,6 @@ namespace onnxruntime {
 namespace qnn {
 
 namespace {
-template <typename Shape>
-bool LeadingDimsAllOne(const Shape& shape) {
-  for (size_t i = 0; i + 2 < shape.size(); ++i) {
-    if (shape[i] != 1) {
-      return false;
-    }
-  }
-
-  return true;
-}
-
 // Detects a block-quantized MatMul weight (ONNX MatMul input[1]).
 // Accepts weight rank 2–4: shape [..., K, N] where any leading dims beyond K/N must equal 1
 // (i.e. reshapeable to [1, 1, K, N]). Per ONNX opset 21 the scale has the same rank as the
@@ -50,13 +39,13 @@ bool IsBQWeight(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnitIODef
   if (rank < 2 || rank > 4) {
     return false;  // BQ supports weight rank 2–4 (reshapeable to [1,1,K,N]).
   }
-  if (!LeadingDimsAllOne(weight_shape)) {
+  if (!utils::TrySqueezeLeadingOnesTo(weight_shape, 2).has_value()) {
     return false;
   }
   if (scale_shape.size() != rank) {
     return false;  // Scale must have the same rank as the weight.
   }
-  if (!LeadingDimsAllOne(scale_shape)) {
+  if (!utils::TrySqueezeLeadingOnesTo(scale_shape, 2).has_value()) {
     return false;
   }
   // Blocked axis is rank-2 (K dimension). scale_shape[rank-2] < weight_shape[rank-2].
@@ -105,11 +94,14 @@ Ort::Status CheckInputs(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeU
   // Just use QNN MatMul for these older QNN SDK versions.
   use_fully_connected = false;
 #else
-  // Static [1, ..., 1, K, N] weights are ONNX-broadcast-equivalent to [K, N].
+  // Static [1, ..., 1, K, N] weights are ONNX-broadcast-equivalent to [K, N]. The rank > 2 term is
+  // required (not just implied by the squeeze check, which is vacuously true at rank 2): it keeps
+  // rank-2 weights on the existing path, where no output reshape is needed.
   const bool is_rank2_static_weight = input_info_1.shape.size() == 2 && input_info_1.is_initializer;
   const bool is_rank1_input_1 = input_info_1.shape.size() == 1;
   squeeze_static_weight_to_2d =
-      input_info_1.is_initializer && input_info_1.shape.size() > 2 && LeadingDimsAllOne(input_info_1.shape) &&
+      input_info_1.is_initializer && input_info_1.shape.size() > 2 &&
+      utils::TrySqueezeLeadingOnesTo(input_info_1.shape, 2).has_value() &&
       input_info_0.shape.size() >= 2;
   // FullyConnected cannot pass the Op validation if keep_dims is true, so if input_0 is per-channel quantized tensor
   // with rank > 2, it's not easy to set the quantization parameters for the output reshaped rank 2 tensor.
@@ -454,6 +446,9 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnFullyConnected(QnnModelWrapper& 
   const std::string& org_input_1_name = inputs[1].name;
   std::string input_1_name = org_input_1_name;
   std::vector<uint32_t> shape_2d;
+  // input_1's [k, n] shape, i.e. its trailing two dims with any leading unit dims squeezed off.
+  // Unused when input_1 is rank 1. Not const: TwoDimensionTranspose rewrites it in place below.
+  std::vector<uint32_t> weight_shape_kn;
   QnnQuantParamsWrapper quant_param_2d = input_info_1.quant_param.Copy();
   if (reshape_input_1) {
     // Input[1] is a rank 1 tensor that needs to be reshaped.
@@ -467,11 +462,11 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnFullyConnected(QnnModelWrapper& 
     input_1_name = utils::UniqueNameGenerator().New(org_input_1_name, "_transpose");
     const size_t weight_k_dim_index = input_info_1.shape.size() - 2;
     const size_t weight_n_dim_index = input_info_1.shape.size() - 1;
-    const std::vector<uint32_t> squeezed_weight_shape = {input_info_1.shape[weight_k_dim_index],
-                                                         input_info_1.shape[weight_n_dim_index]};
-    shape_2d = {input_info_1.shape[weight_n_dim_index], input_info_1.shape[weight_k_dim_index]};
+    weight_shape_kn = {input_info_1.shape[weight_k_dim_index], input_info_1.shape[weight_n_dim_index]};
+    // FullyConnected requires input_1's shape to be [n, k].
+    shape_2d = {weight_shape_kn[1], weight_shape_kn[0]};
     if (squeeze_static_weight_to_2d) {
-      RETURN_IF_ERROR(quant_param_2d.HandleSqueeze<uint32_t>(input_info_1.shape, squeezed_weight_shape));
+      RETURN_IF_ERROR(quant_param_2d.HandleSqueeze<uint32_t>(input_info_1.shape, weight_shape_kn));
     }
     RETURN_IF_ERROR(quant_param_2d.HandleTranspose<uint32_t>(std::vector<uint32_t>({1, 0})));
   }
@@ -481,13 +476,8 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnFullyConnected(QnnModelWrapper& 
   if (input_info_1.is_initializer) {
     std::vector<uint8_t> unpacked_tensor;
     if (!reshape_input_1) {
-      const size_t input_1_rank = input_info_1.shape.size();
-      std::vector<uint32_t> weight_shape_for_transpose =
-          squeeze_static_weight_to_2d
-              ? std::vector<uint32_t>{input_info_1.shape[input_1_rank - 2], input_info_1.shape[input_1_rank - 1]}
-              : input_info_1.shape;
       RETURN_IF_ERROR(utils::TwoDimensionTranspose(qnn_model_wrapper,
-                                                   weight_shape_for_transpose,  // Modified by TwoDimensionTranspose.
+                                                   weight_shape_kn,  // Modified by TwoDimensionTranspose.
                                                    input_info_1.initializer_tensor,
                                                    unpacked_tensor,
                                                    logger,
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -18,11 +18,8 @@ namespace onnxruntime {
 namespace qnn {
 
 namespace {
-bool HasOnlyLeadingUnitDimensions(const std::vector<uint32_t>& shape) {
-  if (shape.size() <= 2) {
-    return false;
-  }
-
+template <typename Shape>
+bool LeadingDimsAllOne(const Shape& shape) {
   for (size_t i = 0; i + 2 < shape.size(); ++i) {
     if (shape[i] != 1) {
       return false;
@@ -30,32 +27,6 @@ bool HasOnlyLeadingUnitDimensions(const std::vector<uint32_t>& shape) {
   }
 
   return true;
-}
-
-Ort::Status AdjustQuantAxisAfterSqueezingLeadingDimensions(QnnQuantParamsWrapper& quant_param,
-                                                           size_t num_squeezed_leading_dims) {
-  if (num_squeezed_leading_dims == 0 || (!quant_param.IsPerChannel() && !quant_param.IsLPBQ())) {
-    return Ort::Status();
-  }
-
-  Qnn_QuantizeParams_t& params = quant_param.Get();
-  int32_t* quant_axis = nullptr;
-  if (params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_AXIS_SCALE_OFFSET) {
-    quant_axis = &params.axisScaleOffsetEncoding.axis;
-  } else if (params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BW_AXIS_SCALE_OFFSET) {
-    quant_axis = &params.bwAxisScaleOffsetEncoding.axis;
-  } else if (params.quantizationEncoding == QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION &&
-             params.blockwiseExpansion != nullptr) {
-    quant_axis = &params.blockwiseExpansion->axis;
-  } else {
-    return MAKE_EP_FAIL(("Unhandled quantization encoding: " + std::to_string(params.quantizationEncoding)).c_str());
-  }
-
-  RETURN_IF_NOT(*quant_axis >= static_cast<int32_t>(num_squeezed_leading_dims),
-                "Cannot squeeze leading unit dimensions that contain the quantization axis.");
-  *quant_axis -= static_cast<int32_t>(num_squeezed_leading_dims);
-
-  return Ort::Status();
 }
 
 // Detects a block-quantized MatMul weight (ONNX MatMul input[1]).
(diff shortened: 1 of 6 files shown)


qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🟠 Concern` `[Arch]` — `matmul_op_builder.cc:L38` (belongs in `qnn_quant_params_wrapper.h`)
The quant-axis family — `HandleTranspose` (`qnn_quant_params_wrapper.h:130`) and `HandleUnsqueeze` (`:156`) — are member templates on `QnnQuantParamsWrapper`, each encapsulating the three-encoding dispatch over `AXIS_SCALE_OFFSET` / `BW_AXIS_SCALE_OFFSET` / `BLOCKWISE_EXPANSION`. The operation this PR adds is literally the inverse of `HandleUnsqueeze`, yet it is a free function in the op-builder TU that reaches into `quant_param.Get()` and mutates the raw union from outside. This trips three things at once: a Sibling-Consistency HOME-axis miss, a fourth copy of the encoding dispatch (change amplification), and leaked union-member dispatch. Since `CheckInputs` now excludes LPBQ from the FC path, the `BLOCKWISE_EXPANSION` branch here is even dead. A sibling `QLinearMatMul` (`qlinear_matmul_op_builder.cc:334-368`) has the same-shaped path and would reuse a wrapper member.
*Better design:* Add `QnnQuantParamsWrapper::HandleSqueeze<IntType>(orig_shape, new_shape)` mirroring `HandleUnsqueeze`, keeping the dispatch and the "axis falls inside a squeezed dim -> fail" guard inside the wrapper; the op-builder then calls one method and never touches `.Get()`.

### 🟩 Maintainability

**[D-2]** `🟠 Concern` `[Maint]` — `matmul_op_builder.cc:L511`
`CheckInputs` (the single routing owner) decides FC eligibility with `can_canonicalize_static_weight_to_2d = is_initializer && HasOnlyLeadingUnitDimensions(shape) && input_0.rank>2`. But `ProcessInputsForQnnFullyConnected` re-derives whether to squeeze with `squeeze_static_weight_to_2d = is_initializer && HasOnlyLeadingUnitDimensions(shape)` — dropping the `input_0.rank>2` term. It is correct today only by the non-obvious fact that no other FC-routing branch can make `HasOnlyLeadingUnitDimensions` true. If anyone later routes a leading-unit-dim weight with `input_0.rank<=2` to FC, the squeeze fires without the guard that justified it — an undocumented implicit precondition spanning two functions.
*Better design:* Make `CheckInputs` the sole owner and pass a single signal downstream (an out `bool squeeze_leading_dims` or a small routing enum) that `ProcessInputsForQnnFullyConnected` consumes rather than re-computing.

**[D-5]** `🔵 Improvement` `[Maint]` — `matmul_op_builder.cc:L25`
The scan `for (i; i+2<rank) if (shape[i]!=1) return false;` is byte-for-byte the loop `IsBQWeight` runs at `:47-51` (weight) and `:56-59` (scale). With this PR it becomes a third copy — the rule-of-three is met. The two also diverge at rank 2: `IsBQWeight` treats rank-2 as vacuously "all leading dims unit", while `HasOnlyLeadingUnitDimensions` returns false for rank<=2. A future edit to one scan won't propagate, and the rank-2 asymmetry is an easy trap.
*Better design:* Extract one file-local predicate (e.g. `LeadingDimsAllOne(shape)`) called by both `IsBQWeight` and the FC path.

### 🟩 Simplicity

*(All Simplicity findings — D-1 over-complexity/wheel-reinvention, D-5 duplicated scan, D-7 rename churn — are reported under their primary dimensions above/below. Simplicity bottom line: delete `AdjustQuantAxisAfterSqueezingLeadingDimensions` and add a `HandleSqueeze` sibling on the wrapper — see D-1.)*

**[D-7]** `🔵 Improvement` `[Simplicity]` — `matmul_op_builder.cc:L506`
The PR renames `shape_2d` -> `fc_weight_shape` and `quant_param_2d` -> `fc_weight_quant_param` throughout `ProcessInputsForQnnFullyConnected`, but the sibling `ProcessInputsForQnnMatMul` (`:317-318`) still uses `shape_2d`/`quant_param_2d`. One concept now has two naming schemes across sibling functions, and it inflates the diff.
*Better design:* If renaming, align the sibling too — or drop the rename to reduce churn.

### 🔵 Completeness

**[D-3]** `🟠 Concern` `[Completeness]` — `matmul_op_builder.cc:L177`
The canonicalization only fires when `input_0` has rank>2. For a legal ONNX MatMul with a 2D activation `[M,K]` and a rank-4 weight `[1,1,K,N]`, `can_canonicalize` is false and `is_rank2_static_weight` is false (weight rank 4), so the node goes to QNN MatMul. `ProcessInputsForQnnMatMul` registers the weight at its raw `[1,1,K,N]` shape (no squeeze), producing precisely the `[M,K] × [1,1,K,N] → [1,1,M,N]` batched MatMul whose HTP finalization the PR exists to avoid. FC would handle a 2D activation directly, so the rank>2 bound excludes a same-bug-family case, with no comment explaining why.
*Better design:* Either widen `can_canonicalize` to `input_0.rank>=2`, or add a one-line rationale plus a tripwire test asserting the 2D case's intended routing.

**[D-4]** `🟠 Concern` `[Completeness]` — `matmul_test.cc:L659`
The whole point of the PR is routing selection (FC instead of batched MatMul), but the added test checks only `TestQDQModelAccuracy` + `ExpectedEPNodeAssignment::All` — it never asserts the emitted QNN op is `FULLY_CONNECTED`. It has teeth only indirectly (reverting to MatMul would fail finalize -> assignment != All). If a future QAIRT SDK makes the batch-MatMul pattern finalize, the test silently stops guarding the routing and nobody notices the fix rotted.
*Better design:* Add `provider_options["dump_json_qnn_graph"]="1"` + `AssertOpInQnnGraph(dir, "FullyConnected", 1)` (with the `if (::testing::Test::IsSkipped()) return;` bail). The sibling `ReshapeGemmFusion_ProducesFullyConnected` is the canonical template.

**[D-6]** `🔵 Improvement` `[Completeness]` — `matmul_test.cc:L659`
The one added HTP case runs only rank-4 input, rank-4 weight, per-channel int16, axis=3 (on N). It does mutation-kill the axis-adjust bug, so it is real coverage for that shape. But missing: (a) a per-tensor leading-unit-dim weight (hits the AdjustQuantAxis early-return branch); (b) a rank-3 `[1,K,N]` weight (num_squeezed=1 vs 2); (c) a CPU-backend case (FC selection is not gated by `IsNpuBackend`, so the x86 CPU lane never exercises the new path); (d) the rejection path when the axis lands on a squeezed dim; (e) the combination with the uint16-weight Convert workaround (`:496-519`), which this int16 per-channel test cannot reach.
*Better design:* Add at minimum a per-tensor leading-unit-dim case and a CPU-backend case; the sibling BQ tests (`:546-583`) already model the Rank3/Rank4 enumeration discipline.

qti-yuduo:
Thanks @qti-chuteng, feedback addressed.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -22,6 +22,17 @@ inline bool IsQuant16bit(Qnn_DataType_t qnn_data_type) {
   return qnn_data_type == QNN_DATATYPE_UFIXED_POINT_16 || qnn_data_type == QNN_DATATYPE_SFIXED_POINT_16;
 }
 
+template <typename Shape>
+bool LeadingDimsAllOne(const Shape& shape) {
```
yath1: I think instead of hard coding `2` here for matmul. You could generalize and move to utils. I think this is similar to the `TrySqueezeLeadingOnesTo` you added for RMSNorm.

onnxruntime/test/providers/qnn/matmul_test.cc
```diff
@@ -362,6 +364,40 @@ TEST_F(QnnCPUBackendTests, MatMulOp) {
   // RunMatMulOpTest({2, 3, 3, 3}, {3, 2}, true, false);
 }
 
+TEST_F(QnnCPUBackendTests, MatMulOp_QDQ_StaticLeadingUnitDimsPerTensorWeight) {
```
yath1, line 367: Are these tests still needed? I think we are deprecating CPU option

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -452,10 +476,13 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnFullyConnected(QnnModelWrapper&
   if (input_info_1.is_initializer) {
     std::vector<uint8_t> unpacked_tensor;
     if (!reshape_input_1) {
-      // 2D initializer should be transposed to [n, k].
-      std::vector<uint32_t> original_shape_copy = input_info_1.shape;
+      const size_t input_1_rank = input_info_1.shape.size();
+      std::vector<uint32_t> weight_shape_for_transpose =
```
yath1: How is this shape different from `squeezed_weight_shape`. Is it possible to use a single var instead?

yath1 (commented):
Sorry for the late review. Could you please explain what are the advantages of squeezing unit dimensions and calling FC instead of a matmul layer? Any perf benefits or something else which is not directly apparent.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
qti-yuduo: Done in 47f97e5a. Both copies are gone: `LeadingDimsAllOne` here and the `TrySqueezeLeadingOnesTo` that landed in `rmsnormalization_op_builder.cc` after this branch's base. Single home is now `utils::TrySqueezeLeadingOnesTo` in `qnn_utils.h`, used by all four call sites (3 here, 1 in RMSNorm). Templated rather than `gsl::span<const uint32_t>` because `IsBQWeight` scans the weight shape as `vector<uint32_t>` and the scale shape as `vector<int64_t>`.

Two rank bounds are kept deliberately, since the helper treats an already-short shape as vacuously squeezable:
- `CheckInputs` keeps `shape.size() > 2` — without it a rank-2 weight sets `squeeze_static_weight_to_2d` and wrongly flips `reshape_output`.
- `IsBQWeight` keeps `rank > 4` — otherwise `[1,1,1,K,N]` would be accepted.

Both now have a comment saying why.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
qti-yuduo: Collapsed in 47f97e5a. `squeezed_weight_shape` and `weight_shape_for_transpose` were equal on every reachable path — rank-2 gives `{shape[0], shape[1]}` == `input_info_1.shape`, and the squeeze path gives `{shape[r-2], shape[r-1]}` for both — so they are now one `weight_shape_kn`. `HandleSqueeze` reads it before `TwoDimensionTranspose` rewrites it in place, and nothing reads it afterwards (`shape_2d` is what goes into the tensor wrapper); the in-place rewrite is called out in a comment on the declaration.

onnxruntime/test/providers/qnn/matmul_test.cc
qti-yuduo: I'd rather keep it, or convert it, than drop it — a few things point the other way:

- I can't find a deprecation marker for the CPU backend anywhere in the repo, and `QnnCPUBackendTests` has ~363 references today. Is the deprecation tracked somewhere (roadmap/Jira) I should be reading? And is the policy "no *new* CPU tests" or "remove existing ones too"?
- It's the only coverage of this path on `build-and-test-windows-x86_64` — the HTP tests in this file are all inside the `__aarch64__ || _M_ARM64 || __linux__` guard (matmul_test.cc:401-850), so on x86_64 deleting this leaves the new squeeze path with zero tests.
- It's also the only case that exercises `HandleSqueeze`'s per-tensor early-return; the HTP test covers the per-channel remap.
- @qti-chuteng's D-6(c) asked specifically for the per-tensor case.

If "no new CPU tests" is the rule, I'll move it into the HTP block as a per-tensor uint8 test instead of deleting it, so the per-tensor branch keeps coverage. Which do you prefer?

qti-yuduo:
@yath1 It's a compile/finalize fix rather than a perf change — the perf side is a secondary benefit.

**What it fixes.** The trigger was an AI Hub job with the batched pattern `[1, 6, 3, 3] x [1, 1, 3, 420]`, i.e. a static rank-4 weight whose leading dims are all 1. That failed HTP graph finalization when lowered to QNN `MatMul`. Under ONNX broadcasting `[1, ..., 1, K, N]` is exactly equivalent to `[K, N]` — the leading unit dims carry no information — so squeezing them is semantics-preserving and gets the graph to finalize. Without it the node is emitted, finalization fails, and the whole partition is lost.

**Why FC specifically.** Once the weight is 2D it's the same shape FC already requires, and FC's existing path folds the `[K, N] -> [N, K]` transpose into the initializer at build time (`utils::TwoDimensionTranspose`), so there is no runtime transpose node and no extra tensor. Routing to `MatMul` with a squeezed weight would need an in-graph reshape instead. So FC is the cheaper of the two lowerings here, not a new mechanism — this only widens the existing `is_rank2_static_weight` gate to cover the broadcast-equivalent case.

**What I haven't quantified:** I don't have a measured latency delta, only the finalize failure. If a number would help I can run a before/after A/B on HTP V73 locally and post it.

yath1:
>  Without it the node is emitted, finalization fails, and the whole partition is lost.

I think matmul does support  `[1, 6, 3, 3] x [1, 1, 3, 420]`? Could you explain why finalization failed here? Just trying to understand

onnxruntime/test/providers/qnn/matmul_test.cc
yath1: I am fine either way. Maybe we can err on the side of caution and leave the test. Removing this test along with other CPU tests should be easy later on

qti-yuduo: Leaving it. One more reason I found while re-reviewing: it's the only test covering the case where `input_0` is rank 2 and the weight is rank > 2, i.e. where the output rank is driven by the weight rather than by `input_0`. That's the `reshape_fc_output` branch with `batch = input_0.shape[0]`; the HTP test uses a rank-4 `input_0` and exercises the flattened-batch path instead. So when the CPU tests do get swept, this one is worth converting rather than deleting.

qti-yuduo:
> I think matmul does support `[1, 6, 3, 3] x [1, 1, 3, 420]`? Could you explain why finalization failed here? Just trying to understand

You're right that QNN MatMul supports that shape — my earlier explanation was wrong, and the shape isn't the trigger. I ran a factor screen on QAIRT 2.35 / HTP V73 to isolate it (9 variants, one factor at a time):

| # | input[0] | input[1] | input[1] quant | finalize |
|---|---|---|---|---|
| A | `[1,6,3,3]` | `[1,1,3,420]` | per-channel int16 | **FAIL** |
| B | `[1,6,3,3]` | `[1,1,3,420]` | per-tensor int16 | pass |
| C | `[1,6,3,3]` | `[1,1,3,420]` | per-channel int8 | **FAIL** |
| D | `[1,6,3,3]` | `[1,1,3,42]` | per-channel int16 | **FAIL** |
| E | `[1,1,3,3]` | `[1,1,3,420]` | per-channel int16 | pass |
| F | `[1,6,3,3]` | `[3,420]` (→ FC) | per-channel int16 | pass |
| G | `[1,6,3,3]` | `[1,6,3,420]` | per-channel int16 | **FAIL** |
| H | `[1,6,3,3]` | `[1,6,3,420]` | per-tensor int16 | pass |
| I | `[1,2,3,3]` | `[1,1,3,420]` | per-channel int16 | **FAIL** |

The rule: **QNN MatMul fails graph finalization iff `input[1]` carries a per-axis (`QNN_QUANTIZATION_ENCODING_AXIS_SCALE_OFFSET`) encoding and the effective batch dim is > 1.** Op validation passes (`Validation PASSED for 4 nodes in OrtNodeUnit (MatMul)`); the failure comes after Graph Optimizations, with `QNN_COMMON_ERROR_MEM_ALLOC: Memory allocation related error, Code: 1002`.

It's independent of N (D still fails at N=42), of weight bit-width (int8 and int16 both fail), and of whether the leading dims are 1 or a real batch (G). B and H show per-tensor passes at the same shapes; E shows per-channel passes at batch 1. D failing at N=42 is why I read this as a bug surfacing as `MEM_ALLOC` rather than genuine capacity exhaustion.

Why FullyConnected avoids it (F): FC flattens the batch into rows, so the per-axis-weight-with-batch pattern never reaches the backend.

Two things worth being straight about, since they follow from the table rather than from the PR description:

- **The fix is narrower than the bug.** G proves a genuinely batched per-channel weight `[1,6,3,420]` still dies, and this PR can't help there — squeezing isn't semantics-preserving when the leading dims aren't 1. Should I file a QAIRT ticket for the underlying finalize failure? I'd rather not add an SDK-version-blind rejection in the EP; my screen is 2.35 / V73 only and CI runs newer.
- **The reroute is broader than the failure.** Per-tensor squeezable weights get routed to FC too, even though B/H show they were never broken. I kept it that way because `[1,...,1,K,N]` is exactly equivalent to `[K,N]` under ONNX broadcasting, and rank-2 static weights already go to FC unconditionally, so gating on quant type would be arbitrary. Happy to narrow it to the per-channel case if you'd prefer.

Updating the PR description, which still frames this as a shape limitation.

qti-yuduo:
Self-review pass pushed in 043bec53, on top of the earlier review fixes:

- Reverted an unrelated behavior change I'd introduced: `use_fully_connected = ... && !use_conv2d;` is back to what it was. The extra `&& !input_info_1.quant_param.IsLPBQ()` term also re-routed pre-existing **rank-2** LPBQ weights from FullyConnected to MatMul whenever `use_conv2d` was false, which has nothing to do with this fix.
- The LPBQ case that does matter is now guarded where it belongs: the squeeze is gated on a trial `HandleSqueeze` succeeding on a `Copy()`, so a weight whose quantization axis sits in a leading 1-dim stays on the QNN MatMul path that handles it today instead of failing the node. Only LPBQ can hit that — its axis is constrained to 0 or 1 (`qnn_quant_params_wrapper.cc:656`, and the QNN axis is `1 - axis`), while a plain per-axis encoding always has >= 2 scales because the single-scale case returns per-tensor earlier (`:518`, `:525`), so its axis dim can never be 1. Deriving the guard from `HandleSqueeze` rather than restating the rule keeps the two from drifting.
- Extracted the duplicated `use_fully_connected && (rank > 2 || squeeze)` condition as `reshape_fc_output`, and dropped a redundant branch — `FlattenLeadingDims` already yields `shape[0]` at rank 2.
- Corrected the code comment to state the real trigger (per-axis `input[1]` + batch > 1) rather than framing it as a shape limitation.

Verified on QAIRT 2.35 / HTP V73: no new failures against a same-tree baseline. The 22 `MatMulBQ_*` / `GemmBQ_*` / `MatMulNBits_LPBQ_*` failures and the `MatMulNBits_LPBQ_M1_N32_K256_B8_BS32` segfault reproduce identically on plain `main` on this box.

1duo:
Revalidated on current main with the default QAIRT 2.50.40.260831. The original [1,6,3,3] x [1,1,3,420] per-channel int16 case now finalizes and runs as QNN MatMul, with no FullyConnected rewrite, on HTP V73. The QAIRT 2.35 MEM_ALLOC failure no longer reproduces, so this workaround is no longer required.

qti-yuduo:
Revalidated on current main with the default QAIRT 2.50.40.260831. The original [1,6,3,3] x [1,1,3,420] per-channel int16 case now finalizes and runs as QNN MatMul, with no FullyConnected rewrite, on HTP V73. The QAIRT 2.35 MEM_ALLOC failure no longer reproduces, so this workaround is no longer required.

PR #588, opened 2026-06-30
[QNN EP] Fix backend lib loaded check

qti-mattsinc:
### Description
* `LoadBackend` checks if `backend_lib_handle_` is null to determine if the backend has already been loaded. But `GetQnnInterfaceProvider` can fail after successfully loading the lib, if e.g. it fails in the API version check. This means that if a) a backend lib with a incomptabile version is used, and b) `LoadBackend` is called twice (as can happen in the `IsDx12SharedMemoryAllocatorSupported` path), then, the second call to `LoadBackend` will 'succeed', the fields in `qnn_interface_` will be null, and a crash will follow, instead of getting the expected graceful API version check error.
* To fix this, use a separate flag (`backend_lib_loaded_`) that is only set if `GetQnnInterfaceProvider` succeeded, similar to `system_lib_loaded_`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -302,7 +302,7 @@ void QnnBackendManager::SetQnnBackendType(uint32_t backend_id) {
 }
 
 Ort::Status QnnBackendManager::LoadBackend() {
-  if (backend_lib_handle_) {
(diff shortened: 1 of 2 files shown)


qti-chuteng (dismissed):

qti-mattsinc:
@qti-chuteng, tests are clean after re-triggering. Looks like it was a flaky CI issue.

qti-chuteng (approved):

PR #589, opened 2026-06-30
Add L2Norm Fusion pass

yath1:
### Description
This PR implements L2Norm fusion pass. L2Norm op implements the following math (with slight change): output = x / sqrt(max(sum(x**2), epsilon))

### Motivation and Context

This fixes CVT model issue for FP precision. Native path shows an accuracy of 0.376 but EP shows an accuracy of 0.319. Expected is 0.377.

This op involves the following sub operations: ReduceL2->Add->Div. ReduceL2 sums the squares of input activation at the given axis. In FP16 precision this operation overflows for sufficiently large input activations. But, HTP has a fused op QNN_OP_L2_NORM which maintains the intermediate tensor at high precision which avoids overflow issues. This PR generates this fused op which solves the accuracy issue now. EP now shows 0.3759 which is inline with what's expected.


qti-yuduo (approved):
LGTM

PR #595, opened 2026-07-01
[QNN EP] Fix null tensor dtype in GRU/GQA and add IR composition tests

qti-niscmami:
### Description

Use `QNN_DATATYPE_FLOAT_32` for null tensor placeholders in GRU and GQA op builders. Add IR backend composition tests for LSTM and GRU.

### Motivation and Context

QnnIr rejects `QNN_TENSOR_TYPE_NULL` tensors with `QNN_DATATYPE_UNDEFINED` (error 7004). LSTM was already fixed in #323; GRU and GQA had the same latent bug. IR backend tests catch future regressions at graph-composition time.

### DIFF
--- a/.github/workflows/qualcomm-internal-coverage.yml
+++ b/.github/workflows/qualcomm-internal-coverage.yml
@@ -6,21 +6,27 @@ name: Qualcomm Internal Coverage
 permissions:
   contents: read
   checks: write
-  pull-requests: write
+  # pull-requests: write is only needed by the "Post Diff Coverage to PR" step, which is
+  # currently disabled (if: false). Requesting it here breaks callers that don't grant it
+  # (e.g. the nightly release workflow). Re-add it when that step is re-enabled.
 
 on:
   workflow_call:
-    inputs:
+    inputs: &common_inputs
       target_py_vsn:
         description: "Python version for the coverage build wheel (does not affect C++ coverage results)"
         type: string
         default: '3.10'
+      nightly_coverage:
+        description: "Set to true to upload the per-file coverage Excel to Artifactory"
+        type: boolean
+        default: false
+      qa_tag:
+        description: "Pre-computed QA tag from the caller; leave blank to auto-generate from today's date + SHA"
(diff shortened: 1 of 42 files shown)


qti-yuduo (approved):

qti-mattsinc (approved):
It seems like QnnIr rejecting `QNN_TENSOR_TYPE_NULL` tensors on the basis of having `QNN_DATATYPE_UNDEFINED` is incorrect behavior by that backend. From QnnTypes.h:

```c
typedef enum {
  // ...
  
  /// Tensor type NULL which can be used to represent optional tensors. Other Qnn_Tensor_t metadata
  /// is ignored.
  QNN_TENSOR_TYPE_NULL = 5,

  // ...  
} Qnn_TensorType_t;
```

So QnnIr should really not be checking the data type at all for a null tensor. Has a ticket already been filed to fix the behavior in the backend? Working around it with FLOAT_32 makes sense, though. (Perhaps we could create a helper fn for creating null tensors, and leave a comment there explaining why FLOAT_32 is used).

qti-niscmami:
> It seems like QnnIr rejecting `QNN_TENSOR_TYPE_NULL` tensors on the basis of having `QNN_DATATYPE_UNDEFINED` is incorrect behavior by that backend. From QnnTypes.h:
> 
> ```c
> typedef enum {
>   // ...
>   
>   /// Tensor type NULL which can be used to represent optional tensors. Other Qnn_Tensor_t metadata
>   /// is ignored.
>   QNN_TENSOR_TYPE_NULL = 5,
> 
>   // ...  
> } Qnn_TensorType_t;
> ```
> 
> So QnnIr should really not be checking the data type at all for a null tensor. Has a ticket already been filed to fix the behavior in the backend? Working around it with FLOAT_32 makes sense, though. (Perhaps we could create a helper fn for creating null tensors, and leave a comment there explaining why FLOAT_32 is used).

Yes, we have a ticket tracking this issue. Refactored to use a helper function for creating null tensors.

PR #596, opened 2026-07-01
[QNN EP] Support quantized ReduceL2

qti-yuduo:
## Summary
- `ReduceL2` was missing from the QDQ node-unit selector map in `qnn_ep_utils.cc`, so ORT never fused the surrounding DQ/Q pair into a quantized node unit for this op.
- `reduce_op_builder.cc` also explicitly rejected quantized `ReduceL2` input and fell back to running the Mul->ReduceSum->Sqrt decomposition entirely in float32, leaving a stray float32 activation tensor in an otherwise fully-quantized QDQ graph. That silently breaks HTP targets without fp16 support (surfaces later as a confusing device-compatibility failure rather than a clear EP error).
- Register `ReduceL2` in the selector map, and rewrite the decomposition to Dequantize -> compute in float32 -> Quantize using the node unit's real output quant params, mirroring the existing pattern in `BatchNormalizationOpBuilder` for the same "intermediate activations have no well-defined quant params" problem.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/reduce_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/reduce_op_builder.cc
@@ -181,13 +181,19 @@ Ort::Status ReduceOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper, c
               "QNN EP: ReduceLogSumExp operator only supports float input dtypes.");
   }
 
-  if (reduce_op_type == ReduceOpType::REDUCE_OP_TYPE_L2 && node_unit.Inputs()[0].quant_param.has_value()) {
-    // QNN's Dequantize op does not accept per-channel quantized inputs; the float32 decomposition
-    // below assumes the inserted Dequantize can consume the input tensor's quant params as-is.
-    bool is_per_channel = false;
-    int64_t axis = 0;
-    RETURN_IF_ERROR(qnn_model_wrapper.IsPerChannelQuantized(node_unit.Inputs()[0], is_per_channel, axis));
-    RETURN_IF(is_per_channel, "QNN EP: ReduceL2 operator does not support per-channel quantized input.");
+  if (reduce_op_type == ReduceOpType::REDUCE_OP_TYPE_L2) {
+    // QNN's Dequantize/Quantize ops do not accept per-channel quantized tensors; the float32
+    // decomposition below assumes the inserted Dequantize/Quantize can consume the node unit's
+    // input/output quant params as-is.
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/reduce_op_builder.cc
```diff
@@ -186,6 +181,15 @@ Ort::Status ReduceOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper, c
               "QNN EP: ReduceLogSumExp operator only supports float input dtypes.");
   }
 
+  if (reduce_op_type == ReduceOpType::REDUCE_OP_TYPE_L2 && node_unit.Inputs()[0].quant_param.has_value()) {
+    // QNN's Dequantize op does not accept per-channel quantized inputs; the float32 decomposition
+    // below assumes the inserted Dequantize can consume the input tensor's quant params as-is.
+    bool is_per_channel = false;
+    int64_t axis = 0;
+    RETURN_IF_ERROR(qnn_model_wrapper.IsPerChannelQuantized(node_unit.Inputs()[0], is_per_channel, axis));
```
yath1: Inputs()[0] is the activation input right? We can assume its not going to be per-channel but good to error out. Our hardware doesn't support it.

onnxruntime/core/providers/qnn/builder/opbuilder/reduce_op_builder.cc
```diff
@@ -242,24 +246,51 @@ Ort::Status ReduceOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mo
...
+    // params -- mirroring how BatchNormalizationOpBuilder handles the same problem.
     const auto& input = node_unit.Inputs()[0];
     const auto& output = node_unit.Outputs()[0];
     std::vector<uint32_t> input_shape;
     RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(input.shape, input_shape), "Cannot get input shape.");
     std::vector<uint32_t> output_shape;
     RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(output.shape, output_shape), "Cannot get output shape.");
-    RETURN_IF(input.quant_param.has_value(), "Input tensor must not be quantized.");
-    ONNXTensorElementDataType output_type = output.type;
+    const bool is_quantized_op = input.quant_param.has_value();
```
yath1: Nit: I'd call this `is_quantized_input` instead

yath1 (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/reduce_op_builder.cc
qti-yuduo: Updated

PR #597, opened 2026-07-03
[QNN EP] Add component-level unit tests for QnnEpFactory

yuhuchua-qti:
## Summary

Adds component-level unit tests for `QnnEpFactory` in a new file
`onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc`
(**40 tests, 6 groups**). Groups 1–5 (36 tests) exercise the factory's logic
**without** a real QNN backend `.so` or a real `QnnEp` instance, using
captureless `noexcept` OrtApi stubs threaded through a `thread_local`
context. **Group 6** (4 tests, `QnnUnit_ProviderFactoryHtpTest` fixture,
added 2026-08-04) mirrors `QnnUnit_BackendManagerHtpTest` — stubbed OrtApi
plus a real `dlopen(libQnnHtp.so)`, with `GTEST_SKIP()` when the backend is
unavailable — to cover the factory paths that construct a real `QnnEp`.

Branch rebased onto the latest `origin/main` on 2026-08-04.

## Test groups

- **Group 1** — getters + ctor + trivial methods (name/vendor/version,
`IsStreamAware`, `CreateDataTransfer`, `ReleaseEp(nullptr)`).
- **Group 2** — `CreateEpFactories` version-gate error paths.
- **Group 3** — `GetSupportedDevices` filtering (NPU/GPU/CPU, vendor
mismatch, max-device truncation).
- **Group 4** — `Validate` / `GetHardwareDeviceIncompatibilityDetails`
error paths.
- **Group 5** — additional unit-coverable error/argument branches that the
model/integration suite skips (it always sets `backend_type`/`backend_path`,
bypassing the CreateEp autoep block).
- **Group 6 (new)** — real-HTP paths that reach `std::make_unique<QnnEp>`:
`CreateEp` autoep clone + `AddSessionConfigEntry`,
`ValidateCompiledModelCompatibilityInfo` temp-EP wiring, and
`GetHardwareDeviceIncompatibilityDetails` temp-EP wiring. Tests verify
factory *behaviour* via recorded stub side-effects (`added_config_entries`,
`clone_session_options_calls`) — the coverage-relevant lines execute
before the constructor call, so `make_unique<QnnEp>` success is not
required.

## Infrastructure additions (Group 6)

- `FactoryStubContext` gained `session_config`, `added_config_entries`,
`clone_session_options_calls` fields plus functional stubs for
`HasSessionConfigEntry` / `GetSessionConfigEntry` / `CloneSessionOptions` /
`AddSessionConfigEntry`.
- New `Logger_GetLoggingSeverityLevel` stub returning FATAL so
`Ort::Logger`'s C++ ctor short-circuits during `QnnEp` construction (no
deref on the fake `OrtLogger*` token).
- New `UseGlobalFactoryStubs` RAII combining `UseFactoryStubs` with
`OrtGlobalApiOverride`.

## Coverage

`qnn_provider_factory.cc`:

| Filter | Before Group 6 | After Group 6 | Δ |
|---|---|---|---|
| Unit-only (`QnnUnit_ProviderFactory*`) | 191/287 = 66.6% | **220/287 = 76.65%** | **+10.05 pp** |
| Combined (`*Qnn*`) | 228/287 = 79.4% | **243/287 = 84.67%** | **+5.27 pp** |

## Why coverage does not reach 90%

The 44 remaining uncovered lines fall into three buckets, all of which are
either environment-limited on x86_64 CI or already exercised by other test
tiers:

**1. NPU-synthesis fallback (~25 lines)** — lines `148–168`, `189–200`.
Requires `qnn::soc::GetSocId() != 0 || qnn::soc::HasFastRpcCdspDevice()`,
which is true only on real Qualcomm SoC platforms (arm64 Linux / Android /
Windows-on-Snapdragon). Deferred to on-device CI lanes.

**2. `CreateEpFactories` success tail (~15 lines)** — lines `577–615`
(`Ort::InitApi`, `GetEpApi`, `GetModelEditorApi`, factory construction and
return, `SetDefaultLogger`). Exercised end-to-end by every op-level test via
`RegisterExecutionProviderLibrary`, so the combined figure captures it
implicitly; an isolated unit test would essentially replicate what
ORT's plugin loader already does. Not worth pursuing.

**3. Defensive `catch(...)` and rare release branches (~4 lines)** — lines
`319–320` / `493–500` (require `std::make_unique<QnnEp>` to throw a
non-`runtime_error` exception; no deterministic trigger) and line `348`
(`ReleaseAllocator` HTP shared-memory branch; needs a real
`HtpSharedMemoryAllocator` instance which requires DSP / RpcMem).

76.65% unit-only / 84.67% combined is therefore a practical plateau for
this component, consistent with the device-dependent ceilings documented
for other QNN EP components (e.g. `qnn_backend_manager`).

## Test plan

- [x] `onnxruntime_provider_test --gtest_filter="QnnUnit_ProviderFactory*"` — 40/40 pass (36 stub + 4 real-HTP)
- [x] Coverage regenerated (unit-only 76.65%, combined 84.67%)
- [x] Full `*Qnn*` suite regression check via `generate_coverage.sh --test-filter="*Qnn*"` (exit 0)
- [x] `python qcom/build_and_test.py lint_and_fix` — clean

### DIFF
--- a/onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
+++ b/onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
@@ -23,11 +23,8 @@
 
 #if !defined(ORT_MINIMAL_BUILD) && QNN_EP_INTERNAL_SYMBOL_ACCESS
 
-#include <atomic>
 #include <cstdint>
-#include <cstring>
 #include <string>
-#include <utility>
 #include <vector>
 
 #include "gtest/gtest.h"
@@ -115,10 +112,6 @@ class FactoryStubContext {
   // instead of a fake pointer; the counter is decremented for each call.
   int fail_next_create_ep_device = 0;
 
-  // If true, HasSessionConfigEntry reports the corresponding key as present.
-  bool has_backend_type_entry = false;
-  bool has_backend_path_entry = false;
-
   // Version string reported by MakeFakeApiBase()'s GetVersionString. An empty
   // string is reported back as a nullptr (exercises the "(null)" branch of the
   // parse-error message); a non-empty string is passed through verbatim.
@@ -190,15 +183,12 @@ class FactoryStubContext {
     };
 
     stub_ort_api.HasSessionConfigEntry =
-        [](const OrtSessionOptions*, const char* key, int* out) noexcept -> OrtStatus* {
-      auto* self = current_;
+        [](const OrtSessionOptions*, const char*, int* out) noexcept -> OrtStatus* {
+      // No unit test configures backend_type/backend_path at the session-options
+      // level: setting either would make CreateEp skip the autoep block and reach
+      // std::make_unique<QnnEp>, which needs a real QnnEp (integration territory).
+      // Always reporting "not present" keeps CreateEp on the autoep branch.
       *out = 0;
-      if (self == nullptr) return nullptr;
-      const std::string k(key ? key : "");
-      if (k.find("backend_type") != std::string::npos)
-        *out = self->has_backend_type_entry ? 1 : 0;
-      else if (k.find("backend_path") != std::string::npos)
-        *out = self->has_backend_path_entry ? 1 : 0;
       return nullptr;
     };
     stub_ort_api.CreateSessionOptions = [](OrtSessionOptions** out) noexcept -> OrtStatus* {
@@ -602,7 +592,31 @@ TEST_F(QnnUnit_ProviderFactoryTest, GetSupportedDevices_MaxEpDevicesTruncates) {
   EXPECT_EQ(ctx.created_ep_devices[0], npu);
 }
 
-// ===========================================================================
+TEST_F(QnnUnit_ProviderFactoryTest, GetSupportedDevices_CreateEpDeviceFails_PropagatesError) {
+  FactoryStubContext ctx;
+  UseFactoryStubs use(ctx);
+  QnnEpFactory factory("ep", ctx.MakeApiPtrs());
+
+  OrtHardwareDevice* npu = MakeFakeHwDevice(17);
+  ctx.device_type_map[npu] = OrtHardwareDeviceType_NPU;
+  ctx.device_vendor_map[npu] = kQualcommVendorId;
+  ctx.fail_next_create_ep_device = 1;  // first CreateEpDevice call returns an error
+
+  const OrtHardwareDevice* devices[] = {npu};
+  OrtEpDevice* ep_devices[4] = {nullptr};
+  size_t num = 0;
+  // The supported NPU passes the filter, so CreateEpDevice is invoked and its
+  // error status must propagate out of GetSupportedDevices.
+  // Note: the create_ep_device lambda increments num_ep_devices before it
+  // returns, so the caller sees num == 1 with ep_devices[0] == nullptr.
+  OrtStatus* status = factory.GetSupportedDevices(&factory, devices, 1, ep_devices, 4, &num);
+  ASSERT_NE(status, nullptr);
+  EXPECT_EQ(StubStatusCode(ctx, status), ORT_FAIL);
+  EXPECT_EQ(num, 1u);
+  EXPECT_EQ(ep_devices[0], nullptr);
+  EXPECT_TRUE(ctx.created_ep_devices.empty());
+  ctx.stub_ort_api.ReleaseStatus(status);
+}
 // Group 4: GetHardwareDeviceIncompatibilityDetailsImpl /
 //          ValidateCompiledModelCompatibilityInfoImpl — error paths that stop
 //          before std::make_unique<QnnEp>. The fixture leaves the default
@@ -691,12 +705,13 @@ TEST_F(QnnUnit_ProviderFactoryTest, Ctor_CreateMemoryInfoFails_ReleasesMemoryInf
   FactoryStubContext ctx;
   UseFactoryStubs use(ctx);
   // CreateMemoryInfo_V2 returns a non-null status → ctor takes the
-  // ReleaseMemoryInfo cleanup branch. Must not crash.
+  // ReleaseMemoryInfo cleanup branch. Out-param is left null per the ORT
+  // convention that output params are null on failure. Must not crash.
   ctx.stub_ort_api.CreateMemoryInfo_V2 =
       [](const char*, OrtMemoryInfoDeviceType, uint32_t, int32_t,
          OrtDeviceMemoryType, size_t, OrtAllocatorType,
          OrtMemoryInfo** out) noexcept -> OrtStatus* {
-    *out = reinterpret_cast<OrtMemoryInfo*>(kFakeToken);
+    *out = nullptr;
     return reinterpret_cast<OrtStatus*>(new StatusRecord{ORT_FAIL, "stub CreateMemoryInfo_V2 failure"});
   };
   QnnEpFactory factory("ep", ctx.MakeApiPtrs());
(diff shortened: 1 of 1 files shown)


onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
```diff
@@ -0,0 +1,1184 @@
...
+// construction failures are tolerated with a diagnostic).
+// ===========================================================================
+
+class QnnUnit_ProviderFactoryHtpTest : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    OrtLoggingManager::SetDefaultLogger(nullptr);
+    QnnRealHtpBackendContext htp_check;
+    if (!htp_check.IsValid()) {
+      GTEST_SKIP() << "libQnnHtp.so not available";
```
huaychou, line 1009: **[N-1] (Minor)** Group 6 HTP tests can never execute under the current build gating

The whole file is wrapped in `#if !defined(ORT_MINIMAL_BUILD) && QNN_EP_INTERNAL_SYMBOL_ACCESS`, and `QNN_EP_INTERNAL_SYMBOL_ACCESS` is only defined when `ENABLE_COVERAGE AND UNIX AND CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64"` (see `cmake/onnxruntime_unittests.cmake:392-402`). On an x86_64 coverage host libQnnHtp.so is not loadable, so `QnnUnit_ProviderFactoryHtpTest::SetUp` always hits `GTEST_SKIP()`.

**Impact:** The four Group 6 tests (autoep clone + backend_path, multi-device NPU pick, Validate/Incompatibility temp-EP construction) always skip in the only configuration where the file compiles, so their contribution to coverage of those paths is currently zero. Not a correctness bug, but the autoep / temp-EP coverage the PR claims is not realized today.

**Suggested fix:**

Confirm this is intentional (Group 6 will run once the gate widens to ARM64 / on-device CI) and note explicitly in the file header or PR description that Group 6 is always skipped under the current x86_64 coverage gate, so it is not mistaken for realized coverage. If Group 6 should be effective now, QNN_EP_INTERNAL_SYMBOL_ACCESS must also be enabled in an HTP-capable configuration.

onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
```diff
@@ -0,0 +1,1184 @@
...
+  UseFactoryStubs use(ctx);
+  // CreateMemoryInfo_V2 returns a non-null status → ctor takes the
+  // ReleaseMemoryInfo cleanup branch. Out-param is left null per the ORT
+  // convention that output params are null on failure. Must not crash.
+  ctx.stub_ort_api.CreateMemoryInfo_V2 =
+      [](const char*, OrtMemoryInfoDeviceType, uint32_t, int32_t,
+         OrtDeviceMemoryType, size_t, OrtAllocatorType,
+         OrtMemoryInfo** out) noexcept -> OrtStatus* {
+    *out = nullptr;
+    return reinterpret_cast<OrtStatus*>(new StatusRecord{ORT_FAIL, "stub CreateMemoryInfo_V2 failure"});
```
huaychou, line 811: **[N-2] (Minor)** Ctor_CreateMemoryInfoFails leaks the StatusRecord (sanitizer builds would flag it)

This test's overridden `CreateMemoryInfo_V2` returns `new StatusRecord{...}`. The production ctor (`qnn_provider_factory.cc:99-102`) only calls `ReleaseMemoryInfo(mem_info)` when `status != nullptr` and never calls `ReleaseStatus` on that status, so the heap StatusRecord leaks.

**Impact:** Harmless under a plain coverage build, but if the coverage/CI config also enables ASan/LSan this test reports a memory leak. It also incidentally surfaces a latent production issue (the ctor drops the CreateMemoryInfo_V2 error status) — but the production file is out of scope for this test-only PR, noted for reference only.

**Suggested fix:**

Have this stub return a test-owned static/singleton status (no new), or track the new'd StatusRecord in a list and free it in TearDown, so the test does not depend on the production side to release it.

onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
```diff
@@ -0,0 +1,1184 @@
...
+  // returns, so the caller sees num == 1 with ep_devices[0] == nullptr.
+  OrtStatus* status = factory.GetSupportedDevices(&factory, devices, 1, ep_devices, 4, &num);
+  ASSERT_NE(status, nullptr);
+  EXPECT_EQ(StubStatusCode(ctx, status), ORT_FAIL);
+  EXPECT_EQ(num, 1u);
+  EXPECT_EQ(ep_devices[0], nullptr);
+  EXPECT_TRUE(ctx.created_ep_devices.empty());
+  ctx.stub_ort_api.ReleaseStatus(status);
+}
+// Group 4: GetHardwareDeviceIncompatibilityDetailsImpl /
```
huaychou, line 716: **[T-1] (Nit)** Group 4 section banner is malformed (missing opening rule line and blank separator)

Groups 1/2/3/5/6 each have a full top-and-bottom `// ====...====` banner and a preceding blank line; Group 4's comment starts right after the previous test's `}` (`// Group 4: ...`) and is missing the opening `// ====...====` rule line and the blank separator, keeping only the trailing rule. Cosmetic consistency only.

**Suggested fix:**

Add a blank line plus an opening `// ===========================================================================` rule before the Group 4 comment to match the other groups.

qti-chuteng (approved):

huaychou (commented):
Just wondering, should we define the class in .h?

onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
yuhuchua-qti: Verified on the x86_64 coverage host — all 4 `QnnUnit_ProviderFactoryHtpTest` tests run and pass, 0 skipped. `libQnnHtp.so` is loadable on x86_64: the build copies it from the QAIRT SDK into the build folder and the binary resolves it via RPATH. 
The sibling `BackendManagerHtpTest` (24) and `ExecutionProviderHtpTest` (2) suites in the same binary also all pass. So Group 6's coverage is realized today, not deferred. 
GTEST_SKIP() only fires if the SDK is absent from the run env — which would skip those sibling suites too.
Thanks.

yuhuchua-qti:
> Just wondering, should we define the class in .h?

Good point to raise. 
The only consumer of `FactoryStubContext` / `StatusRecord` / `UseFactoryStubs` is this one `.cc` — they're factory-specific stubs, not reusable infra. 
Extracting to `qnn_provider_factory_test.h` would be single-consumer/organizational only, so I'd prefer to keep them co-located with the tests that use them (matches the "class-specific fixtures live next to the test file that owns them" note in `qnn_unit_test_utils.h`).
Thanks.

onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
yuhuchua-qti: Good catch, fixed in PR #713 — added ReleaseStatus on the CreateMemoryInfo_V2 failure path.

onnxruntime/test/providers/qnn/unit/qnn_provider_factory_test.cc
yuhuchua-qti: Fixed in PR #713 — Group 4 banner now matches the other five groups.
Thanks.

PR #598, opened 2026-07-03
Fix GPU test flakiness

kuanyul-qti:
### Summary

Fix GPU test flakiness in CI.

#### Missing teardown on test failure

The five AutoEp tests all call `RegisterExecutionProviderLibrary` directly on the process-global `ort_env` and rely on a matching `UnregisterExecutionProviderLibrary` call at the end of the test body. That final call is written as an `ASSERT_*` statement — meaning if any earlier assertion in the same test fires, `ASSERT_*` expands to `return` and the unregister is silently skipped. The EP library stays registered in `ort_env` for the lifetime of the process, so the next test that tries to register the same name gets a duplicate-registration error.

The fix replaces the trailing unregister call with a `ScopedEpLibraryGuard` RAII struct whose destructor calls `UnregisterExecutionProviderLibrary` unconditionally. The EP library is now always released on scope exit regardless of how the test body terminates.

#### Python model tests: InferenceSession not released between tests

`ModelTestCase` creates an `InferenceSession` in `__init__` and holds it until the object is garbage collected. With many parametrized test cases, CPython's GC does not collect sessions promptly — they accumulate and can exhaust GPU memory or DX12 command queue capacity, causing non-deterministic failures in later tests. Fixed by wrapping `run()` in `try/finally` and explicitly deleting the session before the method returns.

### DIFF
--- a/qcom/model_test/model_test.py
+++ b/qcom/model_test/model_test.py
@@ -15,6 +15,7 @@ import jsonc
 import numpy as np
 import onnx
 import onnxruntime_qnn
+import pytest
 
 import onnxruntime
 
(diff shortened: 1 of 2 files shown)


qcom/model_test/model_test.py
```diff
@@ -114,36 +115,48 @@ def __tensors_from_files(cls, data_sets: Iterable[Iterable[Path]]) -> list[dict[
...
+                    logging.info(f"Comparing actual outputs for {name} to reference.")
+                    if cosine_similarity is not None:
+                        am.assert_cosine_similar(actual[name], expected[ds_idx][name], cosine_similarity)
+                        logging.info(f"{name} is cosine-similar enough (threshold: {cosine_similarity})")
+                    else:
+                        np.testing.assert_allclose(actual[name], expected[ds_idx][name], atol=atol, rtol=rtol)
+                        logging.info(f"{name} is close enough (rtol: {rtol}; atol: {atol})")
+        finally:
+            # Explicitly release the InferenceSession so GPU resources (DX12 command
+            # queues, QNN context memory) are freed before the next test begins rather
```
qti-mattsinc: The GPU backend doesn't use DX12 command queues, and that would be an internal implementation detail, anyway. We can just refer to the context in the comment.

```suggestion
            # Explicitly release the InferenceSession so GPU resources in the QNN
            # context memory are freed before the next test begins rather
```

qti-mattsinc:
>  Since GPU device enumeration can succeed or fail non-deterministically between runs, the failures appeared intermittent.

So the GPU EP device sometimes doesn't appear at all? Is it a case of [DeviceDiscovery::GetDevices](https://github.com/microsoft/onnxruntime/blob/dd32f3598a62d56a434439aa134176c96005ad69/onnxruntime/core/platform/device_discovery_common.cc#L15) in ORT not being deterministic?

qcom/model_test/model_test.py
kuanyul-qti: Fixed.

kuanyul-qti:
> So the GPU EP device sometimes doesn't appear at all? Is it a case of [DeviceDiscovery::GetDevices](https://github.com/microsoft/onnxruntime/blob/dd32f3598a62d56a434439aa134176c96005ad69/onnxruntime/core/platform/device_discovery_common.cc#L15) in ORT not being deterministic?

Not `DeviceDiscovery::GetDevices` -- that memoizes into a function-local `static` (enumerated once per process, with the "machine must be shutdown to change devices" assumption), so device membership is stable within a run. The flakiness isn't enumeration; it's that when that assertion fired, the old trailing `Unregister` (an `ASSERT_*`) was skipped, leaving the EP library registered in the shared `ort_env` and breaking the next test's registration. The RAII guard is really fixing. I've removed the wrong sentence in the PR description.

qti-mattsinc:
> Not `DeviceDiscovery::GetDevices` -- that memoizes into a function-local `static` (enumerated once per process, with the "machine must be shutdown to change devices" assumption), so device membership is stable within a run. The flakiness isn't enumeration; it's that when that assertion fired, the old trailing `Unregister` (an `ASSERT_*`) was skipped, leaving the EP library registered in the shared `ort_env` and breaking the next test's registration. The RAII guard is really fixing. I've removed the wrong sentence in the PR description.

But you are saying that e.g. QnnGPUBackendTests.AutoEp_GpuOnly is still flaky, right? And that the enumeration sometimes fails, which asserts and returns before the library get unregistered.

```cpp
ASSERT_TRUE(selected_devices.size() > 0) << "No QNN GPU device was found.";  // sometimes fails?
// ...
ASSERT_ORTSTATUS_OK(Ort::GetApi().UnregisterExecutionProviderLibrary(*ort_env, kQnnExecutionProvider));  // skipped if above fails
```

So I get the need for the RAII guard (and this PR LGTM), but I'm wondering why does the first assertion sometimes fail in the first place? i.e. is it something we need to investigate?

qti-mattsinc (approved):

qti-chuteng (approved):

kuanyul-qti:
> But you are saying that e.g. QnnGPUBackendTests.AutoEp_GpuOnly is still flaky, right? And that the enumeration sometimes fails, which asserts and returns before the library get unregistered.
> So I get the need for the RAII guard (and this PR LGTM), but I'm wondering why does the first assertion sometimes fail in the first place? i.e. is it something we need to investigate?

The `selected_devices.size() > 0` assertion is not actually the problem — `selected_devices` is populated from a function-local static enumerated once per process, so it's deterministically non-empty on any machine with a GPU. That assertion cannot be the flaky one.

The likely first trigger is `Ort::Session` construction. In the old `AutoEp_PreferGpu` test:

```c++
ASSERT_ORTSTATUS_OK(Ort::GetApi().RegisterExecutionProviderLibrary(...));

{
  Ort::Session session(*ort_env, model_path, so);   // throws on GPU memory exhaustion
  EXPECT_TRUE(SessionHasEp(...));
}

ASSERT_ORTSTATUS_OK(Ort::GetApi().UnregisterExecutionProviderLibrary(...));  // skipped
``` 

`Ort::Session` is a C++ wrapper that throws on failure. If session creation fails — e.g. GPU memory exhaustion (as described in the PR, which is also the motivation for the Python try/finally fix) — the exception unwinds past the trailing Unregister, leaving the EP library registered in `ort_env`. The next `AutoEp` test then hits a duplicate-registration error on its very first
`ASSERT_ORTSTATUS_OK(Register...)` and exits without unregistering either, cascading from there.

So, no GPU device enumeration issue — it's a resource exhaustion / exception-in-test-body issue. The RAII guard and the Python session teardown fix are addressing the same underlying cause from two different sides.

PR #599, opened 2026-07-03
[QNN EP][Affinity] Phase 1 - Add op_affinity provider option with backend scope

chunghow-qti:
## Summary

Adds an `op_affinity` QNN provider option controlling which ONNX op types the QNN EP claims in `GetSupportedNodes`. Filtered-off ops fall back to another EP (typically the CPU EP).

**Modes:**
- `exclude` (default): all ops go to QNN *except* the listed types.
- `include`: *only* the listed types are claimed; every other op type falls back.

**Backend scope (optional):** prefix with a backend name (`cpu`/`gpu`/`htp`/`dsp`/`ir`, same strings as `backend_type`) to scope the whole filter to one backend, e.g. `"htp:exclude:GroupQueryAttention"`. A QNN EP session runs one backend, so scope is a single filter-wide constant. On a session running a different backend the filter is inert (logged once at INFO). `htp` and `htp_fp16` are aliases for the same physical backend, so either scope name matches a session running the other.

## Option format

- **Inline**: `"[backend:][mode:]<comma-separated op types>"`, e.g. `"htp:exclude:Softmax"`, `"include:Conv,MatMul"`, `"Softmax"` (≡ `"exclude:Softmax"`). No spaces; exact, case-sensitive match. Only exact `backend:`/`exclude:`/`include:` words are treated as prefixes, so a UDO op type may itself contain a colon (`custom:MyOp`).
- **Config file**: `"@<path>"` to JSON `{ "backend": "htp", "mode": "exclude"|"include", "op_types": [...] }` (`backend` optional, JSONC comments allowed).

## Error handling

Errors are handled uniformly and loudly, then degrade to "no filtering" (a WARNING) rather than failing session creation:

- Unopenable / malformed config file.
- Config file with a wrong-typed or invalid field: non-string `backend`, `mode` that is neither `"exclude"` nor `"include"` (e.g. `"Include"`, `"excludes"`), non-array `op_types`, or a non-string element inside `op_types`.
- Unknown backend scope name (the filter can never match a real session).
- The reserved `OpType[name=...]` per-node qualifier (see follow-ups) — supplying it now degrades the whole filter to inactive.

An op type in the list that matches no node in the graph is reported per-graph with a WARNING (typically a typo). If the filter is scoped to a backend the session isn't running, those per-entry warnings are suppressed in favor of one INFO line explaining the filter was skipped.

If `op_affinity` is active and `session.disable_cpu_ep_fallback` is set, a WARNING is logged at construction (session creation will fail later if a filtered-off op has no other EP to run on).

## Changes

- **`qnn_onnx_op_affinity.{h,cc}`** (new): `qnn::OnnxOpAffinity` encapsulates parsing, the exclude/include policy, the optional backend scope, and diagnostics. `FromOptionValue()` is the single entry point (degrade-not-throw); `AppliesToBackend()` gates the whole filter for the session backend; `ShouldFilterOff(OrtNodeUnit, QnnBackendType)` applies the decision per node group; `WarnUnmatchedEntries()` is the typo guard. Backend scope is stored as the backend-name string and compared against `QnnBackendTypeToString` (the single source of truth for recognized names, so parsing and matching can't drift).
- **`qnn_execution_provider.{cc,h}`**: build one `OnnxOpAffinity` from the option and call `ShouldFilterOff` per node group in `GetSupportedNodes`, passing the session backend. Filtered-off nodes are tagged in the framework op trace as intentional (via a shared `AppendUnsupportedNodesForGroup` helper). Adds the `disable_cpu_ep_fallback` warning above.
- **Tests**:
  - `unit/qnn_onnx_op_affinity_test.cc` — host-side function tests (no device): `IsActive`, config-file constructor validation (missing/malformed/bad-field), `ShouldFilterOff` across exclude/include × backend match/mismatch, case-sensitive op and prefix matching, UDO colon preservation, `FromOptionValue` degrade paths, phase-2 syntax rejection, unknown backend, `WarnUnmatchedEntries` across multiple `GetSupportedNodes` calls, and a **device-independent tripwire** (`AppliesToBackend_EveryBackendMatchesItsOwnName`) that iterates every `QnnBackendType` enumerator to catch backend-name drift.
  - `qnn_basic_test.cc` — `QnnHTPBackendTests` end-to-end EP-assignment over `Conv → Softmax`: exclude/include, backend-scope match/mismatch, config-file exclude/include, unknown-op-type ignored, case-sensitive, config-not-found fallback, include-empty-forces-all-off, and `disable_cpu_ep_fallback` interaction. One `QnnGPUBackendTests` exclude test guards the non-HTP path.
- **`QNN-ExecutionProvider.md`**: documents the option in the provider-option table and an `OP Affinity - Details` section (matching examples, backend-scope mismatch, empty-list/default, `disable_cpu_ep_fallback` interaction, reserved syntax).

## Notes / follow-ups
- Phase 2 (`OpType[name=...]` per-node filtering) and an EP-built-in default filter are separate, later PRs. The `[` syntax is frozen now so phase-1 specs stay forward-compatible; users will be able to dump the EP-side graph to discover the post-transform node names to target.

### DIFF
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -59,7 +59,7 @@ For build instructions, please see the [BUILD page](./build.md).
     - Python 3.11.x
     - Numpy 1.25.2 or >= 1.26.4
   - Install: `pip install onnxruntime-qnn`
-- [NuGet package - `Qualcomm.ML.OnnxRuntime.QNN`](https://www.nuget.org/packages/Qualcomm.ML.OnnxRuntime.QNN) 
+- [NuGet package - `Qualcomm.ML.OnnxRuntime.QNN`](https://www.nuget.org/packages/Qualcomm.ML.OnnxRuntime.QNN)
   - A single package covering Windows ARM64 (ARM64X), ARM64EC, and x64 (`win-arm64`, `win-arm64ec`, `win-arm64x`, `win-x64` RuntimeIdentifiers) with appropriate fallbacks.
 - Archives (`.zip` and `.tgz`)
   - **Note**: Ships the QNN EP shared library and headers for use outside of Python on Windows and Linux hosts and targets.
@@ -247,6 +247,10 @@ Warning: Enabling HTP Monolithic LSTM may improve session creation time, but thi
 |`"op_packages"`|Description|
 |---|---|
 |Op package config (string)|Register custom op packages. Format: `<OpType>:<PackagePath>:<InterfaceSymbolName>[:<Target>]`. Multiple packages can be separated by commas.|
+
+|`"op_affinity"`|Description|
+|---|---|
+|Filter spec (string) — inline `"[backend:][mode:]<op types>"` or `"@<path to json file>"`|Controls which ONNX op types the QNN EP claims. **Two modes:** `exclude` (default) keeps every op on QNN *except* the listed op types, which fall back to another EP (typically the CPU EP); `include` claims *only* the listed op types and forces every other op type to fall back.<br><br>**Inline form:** `"exclude:Softmax,LayerNormalization"` or `"include:Conv,MatMul"`. `mode:` is optional and defaults to `exclude`, so `"Softmax"` ≡ `"exclude:Softmax"`. Op types are comma-separated, matched **case-sensitively**, and must not contain spaces.<br><br>**Backend scope (optional):** prefix with a backend name to scope the filter to one backend, e.g. `"htp:exclude:Softmax"`. Accepts the same QNN backend names as `backend_type`: `cpu`/`gpu`/`htp`/`dsp`/`ir` (`htp` also matches `htp_fp16` sessions). Without this prefix, the filter applies no matter what backend the session is actually running (via `backend_type`/`backend_path`).<br><br>**Config-file form:** `"@C:\path\to\filter.json"` — a path (prefixed with `@`) to a JSON file, for long lists. Schema: `{ "backend": "htp", "mode": "exclude" \| "include", "op_types": ["Softmax", "Conv"] }` (`"backend"` optional; JSONC comments allowed).<br><br>See [OP Affinity - Details](#op-affinity---details) for matching granularity, defaults, and the `disable_cpu_ep_fallback` interaction.|
 
 |`"dump_json_qnn_graph"`|Description|
 |---|---|
@@ -1667,3 +1671,24 @@ Currently, only pre-compiled models with EPContext nodes are supported. The  exa
 ```
 <graph name>;<adapter binary section path>
 ```
+
+
+## OP Affinity - Details
+
+#### Matching examples
+1. For a graph `Conv → Softmax` (two separate node groups), `"exclude:Softmax"` keeps `Conv` on QNN and falls `Softmax` back to
+another EP (typically CPU).
+2. For a fused QDQ group `DequantizeLinear → Conv → QuantizeLinear`, filter on the group's target op type: `"exclude:Conv"` removes
+  the whole group from QNN; `"exclude:DequantizeLinear"` has no effect.
+
+#### Backend scope mismatch
+On a session running a backend other than the one the filter is scoped to, the filter is skipped entirely (logged at INFO) rather
+than applied.
+
+#### Empty list and default
+- Default (option unset) is `exclude` with an empty list — all ops go to QNN.
+- `include` with an empty list forces **all** ops off QNN (logged at WARNING).
+- An op type in the list that matches no node is ignored with a WARNING (typically a typo).
+
+#### Interaction with `session.disable_cpu_ep_fallback`
+Setting both is logged at WARNING; if a filtered-off op has no other EP to run on, session creation fails.
--- a/onnxruntime/core/providers/qnn/builder/qnn_utils.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_utils.h
@@ -37,6 +37,21 @@ class QnnQuantParamsWrapper;
 
 namespace utils {
 /**
+ * Trims surrounding ASCII whitespace (space, tab, CR, LF) and returns a copy.
+ * Returns an empty string if the input is empty or all-whitespace.
+ * @param s The string to trim.
+ * @return The trimmed copy.
+ */
+inline std::string TrimWhitespace(std::string_view s) {
+  const auto begin = s.find_first_not_of(" \t\r\n");
+  if (begin == std::string_view::npos) {
+    return std::string();
+  }
+  const auto end = s.find_last_not_of(" \t\r\n");
+  return std::string(s.substr(begin, end - begin + 1));
+}
+
+/**
  * Returns a lowercase version of the input string.
  * /param str The string to lowercase.
  * /return The lowercased string.
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -889,6 +889,19 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                               logger_);
   }
 
+  // Op affinity: keep certain ONNX op types/op names on/off QNN, optionally scoped to a backend.
+  std::string op_affinity_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                 FormatEPConfigKey("op_affinity"), "", op_affinity_str);
+  op_affinity_ = qnn::OnnxOpAffinity::FromOptionValue(op_affinity_str, logger_);
+
+  if (op_affinity_.IsActive() && disable_cpu_ep_fallback_) {
+    ORT_CXX_LOG(logger_,
+                ORT_LOGGING_LEVEL_WARNING,
+                "op_affinity is configured and fallback to the CPU EP is disabled (disable_cpu_ep_fallback). "
+                "If op_affinity keeps a node off QNN and no other EP can run it, session creation will fail.");
+  }
+
   // HTP FP16 precision mode
   htp_graph_configs_.enable_htp_fp16_precision = ParseBoolOption(ort_api,
                                                                  session_options_,
@@ -1528,7 +1541,20 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
   }
 
   for (const std::unique_ptr<qnn::IQnnNodeGroup>& qnn_node_group : qnn_node_groups) {
-    Ort::Status support_status = qnn_node_group->IsSupported(qnn_model_wrapper, logger_);
+    // Apply the user's op affinity. Inactive affinity always returns false, so this is a no-op when
+    // the option is unset.
+    Ort::Status support_status;
+
+    if (op_affinity_.ShouldFilterOff(*qnn_node_group->GetTargetNodeUnit(),
+                                     qnn_backend_manager_->GetQnnBackendType())) {
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE,
+                  ("op_affinity: keeping op type '" + qnn_node_group->GetTargetNodeUnit()->OpType() +
+                   "' off QNN; it will fall back to another EP.")
+                      .c_str());
+      support_status = MAKE_EP_FAIL("Filtered off QNN by op_affinity provider option");
+    } else {
+      support_status = qnn_node_group->IsSupported(qnn_model_wrapper, logger_);
+    }
     const bool supported = support_status.IsOK();
 
     LogNodeSupport(logger_, *qnn_node_group, support_status);
@@ -1558,6 +1584,8 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
       }
     }
   }
+
+  op_affinity_.WarnUnmatchedEntries(qnn_backend_manager_->GetQnnBackendType(), logger_);
   return nullptr;
 }
 
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.h
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.h
@@ -24,6 +24,7 @@
 #include "core/providers/qnn/builder/onnx_ctx_model_helper.h"
 #include "core/providers/qnn/cache_compatibility/qnn_cache_compatibility_info.h"
 #include "core/providers/qnn/cache_compatibility/qnn_cache_compatibility_manager.h"
+#include "core/providers/qnn/qnn_onnx_op_affinity.h"
 #include "core/providers/qnn/qnn_telemetry.h"
 #include "core/providers/qnn/rpcmem_library.h"
 #include "core/providers/qnn/qnn_node_compute_info_base.h"
@@ -246,6 +247,10 @@ class QnnEp : public OrtEp, public ApiPtrs {
   uint32_t default_rpc_polling_time_ = 0;
   qnn::ModelSettings model_settings_ = {};
   qnn::HtpGraphConfigs_t htp_graph_configs_;
+
+  // Op Affinity controlling which ONNX op types the QNN EP claims in GetCapability, configured via
+  // the "op_affinity" provider option.
+  qnn::OnnxOpAffinity op_affinity_;
 
   bool dump_json_qnn_graph_ = false;
   std::string json_qnn_graph_dir_ = "";
--- a/onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.cc
+++ b/onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.cc
@@ -1,0 +1,261 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "core/providers/qnn/qnn_onnx_op_affinity.h"
+
+#include <exception>
+#include <fstream>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <utility>
+
+#include "nlohmann/json.hpp"
+
+#include "core/providers/qnn/builder/qnn_def.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+namespace {
+
+// Only the exact "include" spelling selects include mode; anything else (the default, or a typo)
+// resolves to the safer exclude mode.
+OnnxOpAffinity::Mode ParseMode(const std::string& mode_str) {
+  return (mode_str == "include") ? OnnxOpAffinity::Mode::kInclude : OnnxOpAffinity::Mode::kExclude;
+}
+
+// Recognized backend scope names, derived from QnnBackendTypeToString() (the single source of truth)
+// rather than a hand-maintained list, so the two can't drift. "htp" also covers HTP_FP16 -- see
+// AppliesToBackend().
+bool IsKnownBackendName(std::string_view name) {
+  for (uint8_t i = 0; i <= static_cast<uint8_t>(QnnBackendType::SERIALIZER); ++i) {
+    if (name == QnnBackendTypeToString(static_cast<QnnBackendType>(i))) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// Phase-2 syntax guard. '[' is reserved for a future op-name qualifier, "OpType[name=...]". Phase 1
+// doesn't implement name matching, so reject any token carrying it (rather than treat it as a literal
+// op type that would silently never match) -- this keeps a phase-1 spec a strict subset of the future
+// grammar. Colons are NOT reserved (UDO op types use them). Throws so FromOptionValue() degrades the
+// whole filter to inactive.
+void RejectReservedOpTypeChars(const std::string& op_type) {
+  if (op_type.find('[') != std::string::npos) {
+    throw std::runtime_error("op_affinity op type '" + op_type +
+                             "' uses the reserved '[' character (op-name qualifier); "
+                             "op-name filtering is not supported in this version.");
+  }
+}
+
+}  // namespace
+
+OnnxOpAffinity::OnnxOpAffinity(const std::string& inline_spec) {
+  // Inline spec: "[backend:][mode:]<comma-separated op types>". Only exact reserved words are treated
+  // as prefixes -- an op type may itself contain a colon (e.g. UDO "custom:MyOp"), so an unrecognized
+  // prefix stays part of the literal op-type list.
+  std::string remainder = inline_spec;
+
+  // Optional backend scope prefix (checked before mode: it is the outermost segment).
+  if (const auto colon = remainder.find(':'); colon != std::string::npos) {
+    const std::string maybe_backend = remainder.substr(0, colon);
+    if (IsKnownBackendName(maybe_backend)) {
+      backend_scope_ = maybe_backend;
+      remainder = remainder.substr(colon + 1);
+    }
+  }
+
+  // Optional mode prefix.
+  std::string mode_str = "exclude";
+  if (const auto colon = remainder.find(':'); colon != std::string::npos) {
+    const std::string maybe_mode = remainder.substr(0, colon);
+    if (maybe_mode == "exclude" || maybe_mode == "include") {
+      mode_str = maybe_mode;
+      remainder = remainder.substr(colon + 1);
+    }
+  }
+
+  mode_ = ParseMode(mode_str);
+  for (const auto& token : utils::SplitString(remainder, ",", /*keep_empty*/ false)) {
+    if (!token.empty()) {
+      std::string op_type(token);
+      RejectReservedOpTypeChars(op_type);  // phase-2 "[name=...]" syntax -> degrade to inactive
+      selected_op_types_.emplace(std::move(op_type));
+    }
+  }
+}
+
+OnnxOpAffinity::OnnxOpAffinity(const std::filesystem::path& config_file) {
+  // JSON config file: { "backend": "htp", "mode": "exclude"|"include", "op_types": [...] }
+  // ("backend" optional). Anything malformed -- unopenable file, bad JSON, or a field of the wrong
+  // type/value -- throws; FromOptionValue() catches and degrades to an inactive filter + WARNING. The
+  // per-field type/value checks below keep this path as loud as the inline path rather than silently
+  // coercing a typo'd config to a surprising default.
+  std::ifstream ifs(config_file);
+  if (!ifs) {
+    throw std::runtime_error("op_affinity config file could not be opened: " +
+                             config_file.string());
+  }
+
+  const nlohmann::json j = nlohmann::json::parse(ifs, /*cb*/ nullptr, /*allow_exceptions*/ true,
+                                                 /*ignore_comments*/ true);
+
+  if (j.contains("backend")) {
+    if (!j.at("backend").is_string()) {
+      throw std::runtime_error("op_affinity config: \"backend\" must be a string.");
+    }
+    std::string backend = utils::TrimWhitespace(j.at("backend").get<std::string>());
+    if (!backend.empty()) {
+      backend_scope_ = std::move(backend);
+    }
+  }
+
+  std::string mode_str = "exclude";  // default mode when not specified
+  if (j.contains("mode")) {
+    if (!j.at("mode").is_string()) {
+      throw std::runtime_error("op_affinity config: \"mode\" must be a string.");
+    }
+    mode_str = j.at("mode").get<std::string>();
+    // Unlike the inline path (where an unrecognized token falls into the op-type list), a config file
+    // names the mode explicitly, so a value that is neither "exclude" nor "include" is a typo -- reject
+    // it loudly rather than let ParseMode coerce it to exclude.
+    if (mode_str != "exclude" && mode_str != "include") {
+      throw std::runtime_error("op_affinity config: \"mode\" must be \"exclude\" or \"include\", got \"" +
+                               mode_str + "\".");
+    }
+  }
+  mode_ = ParseMode(mode_str);
+
+  if (j.contains("op_types")) {
+    if (!j.at("op_types").is_array()) {
+      throw std::runtime_error("op_affinity config: \"op_types\" must be an array of strings.");
+    }
+    for (const auto& item : j.at("op_types")) {
+      if (!item.is_string()) {
+        throw std::runtime_error("op_affinity config: every entry in \"op_types\" must be a string.");
+      }
+      // File values may be whitespace-padded; trim so entries match node op types.
+      std::string op_type = utils::TrimWhitespace(item.get<std::string>());
+      if (!op_type.empty()) {
+        RejectReservedOpTypeChars(op_type);  // phase-2 "[name=...]" syntax -> degrade to inactive
+        selected_op_types_.emplace(std::move(op_type));
+      }
+    }
+  }
+}
+
+OnnxOpAffinity OnnxOpAffinity::FromOptionValue(const std::string& option_value, const Ort::Logger& logger) {
+  const std::string trimmed = utils::TrimWhitespace(option_value);
+  if (trimmed.empty()) {
+    return OnnxOpAffinity{};  // No filter configured; keep default (exclude + empty list).
+  }
+
+  OnnxOpAffinity filter;
+  try {
+    if (trimmed[0] == '@') {
+      // Value is a path to a JSON config file.
+      const std::string path = utils::TrimWhitespace(std::string_view(trimmed).substr(1));
+      filter = OnnxOpAffinity(std::filesystem::path(path));
+    } else {
+      filter = OnnxOpAffinity(trimmed);
+    }
+  } catch (const std::exception& ex) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                (std::string("Failed to parse op_affinity value '") + trimmed +
+                 "': " + ex.what() + ". Falling back to no filtering.")
+                    .c_str());
+    return OnnxOpAffinity{};  // Bad value degrades to "no filtering" rather than failing the session.
+  }
+
+  if (filter.mode_ == Mode::kInclude && filter.selected_op_types_.empty()) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                "op_affinity mode is 'include' but the op-type list is empty; "
+                "ALL op types will fall back off QNN.");
+  }
+
+  // A backend scope that is not a recognized backend name can never match a real session; warn so a
+  // typo (e.g. "hpt:exclude:...") is visible rather than a silent no-op.
+  if (filter.backend_scope_.has_value() && !IsKnownBackendName(*filter.backend_scope_)) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                ("op_affinity backend scope '" + *filter.backend_scope_ +
+                 "' is not a recognized backend name; the filter will not apply to any session.")
+                    .c_str());
+  }
+
+  if (filter.IsActive()) {
+    std::string joined;
+    for (const auto& op_type : filter.selected_op_types_) {
+      joined += (joined.empty() ? "" : ", ") + op_type;
+    }
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
+                ("op_affinity backend=" +
+                 (filter.backend_scope_.has_value() ? *filter.backend_scope_ : std::string("<any>")) +
+                 " mode=" +
+                 std::string(filter.mode_ == Mode::kInclude ? "include" : "exclude") +
+                 " op_types=[" + joined + "]")
+                    .c_str());
+  }
+
+  return filter;
+}
+
+bool OnnxOpAffinity::AppliesToBackend(QnnBackendType backend_type) const {
+  if (!backend_scope_.has_value()) {
+    return true;  // Unscoped filter applies to any backend.
+  }
+  // "htp" and "htp_fp16" are the same physical backend at different precision settings, so a scope
+  // written as either name matches a session running the other.
+  if (backend_type == QnnBackendType::HTP || backend_type == QnnBackendType::HTP_FP16) {
+    return *backend_scope_ == "htp" || *backend_scope_ == "htp_fp16";
+  }
+  return *backend_scope_ == QnnBackendTypeToString(backend_type);
+}
+
+bool OnnxOpAffinity::ShouldFilterOff(const OrtNodeUnit& target_node_unit, QnnBackendType backend_type) const {
+  if (!IsActive() || !AppliesToBackend(backend_type)) {
+    return false;  // Inactive, or scoped to a different backend: never keeps anything off QNN.
+  }
+  const std::string target_op_type = target_node_unit.OpType();
+  const bool in_list = selected_op_types_.count(target_op_type) != 0;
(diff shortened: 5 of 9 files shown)


minfhong-qti:
Are we going to support skipping by node names?

chunghow-qti:
> Are we going to support skipping by node names?

good question, will study to check if we can also include this.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1527,6 +1527,77 @@ TEST_F(QnnHTPBackendTests, EPRejectsDynamicShapesF32) {
                   /*verify_output*/ true);
 }
 
+// Test the "op_types_to_skip" provider option: a fully-static, QNN-supported op listed in
+// op_types_to_skip must be kept off QNN and fall back to the CPU EP, while the rest of the
+// graph stays on QNN. Without the option both ops would be assigned to QNN.
+TEST_F(QnnHTPBackendTests, EPSkipsOpTypesInOpTypesToSkip) {
```
huaychou: **[N-2] (Minor)** No test for the documented <code class="c">disable_cpu_ep_fallback</code> conflict

The docs explicitly call out that this option conflicts with `session.disable_cpu_ep_fallback`: if a skipped op has no other EP to run on, session creation fails. The added test only covers the happy path (Softmax falls back to CPU, Conv stays on QNN); the documented failure interaction is untested.

**Impact:** A documented error-handling path has no regression coverage and could be silently broken by a future refactor.

**Suggested fix:**

Add a test that sets both `op_types_to_skip` and `session.disable_cpu_ep_fallback=1` where the skipped op is the sole/terminal node, and assert session creation throws. Optional hardening.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1389,7 +1412,37 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+  // can warn about entries that never matched (typically a misspelled op type, e.g. "Softmaxx").
+  std::unordered_set<std::string> matched_skip_op_types;
+
   for (const std::unique_ptr<qnn::IQnnNodeGroup>& qnn_node_group : qnn_node_groups) {
+    // User opt-out: keep this op type off QNN so ORT falls it back to another EP.
+    if (!op_types_to_skip_.empty() &&
+        op_types_to_skip_.count(qnn_node_group->GetTargetNodeUnit()->OpType()) != 0) {
+      matched_skip_op_types.emplace(qnn_node_group->GetTargetNodeUnit()->OpType());
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_INFO,
+                  ("Skipping op type '" + qnn_node_group->GetTargetNodeUnit()->OpType() +
```
huaychou: **[T-1] (Nit)** Per-node-group INFO log line is noisy when many nodes are skipped

The constructor already logs the full op_types_to_skip set once at VERBOSE. The `GetSupportedNodes` loop additionally logs "Skipping op type ..." at INFO for every skipped node group, producing many repetitive lines when a lot of nodes are skipped.

**Suggested fix:**

Consider lowering the per-node-group message to VERBOSE, or aggregating to one count line per op type. Log-noise nit only.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -772,6 +772,29 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     ParseOpPackages(op_packages_str, op_packages, logger_);
   }
 
+  // Op types to keep off QNN (comma-separated ONNX op types). Matching nodes are excluded from
+  // the QNN partition in GetSupportedNodes so ORT falls them back to another EP (e.g. CPU EP).
+  std::string op_types_to_skip_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                 FormatEPConfigKey("op_types_to_skip"), "", op_types_to_skip_str);
+  for (const auto& token : qnn::utils::SplitString(op_types_to_skip_str, ",", /*keep_empty*/ false)) {
+    // Trim surrounding whitespace so "GroupQueryAttention, LayerNormalization" matches both.
```
huaychou: Could we have a utility function to strip the whitespace?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1389,7 +1412,37 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
   }
 
+  // Track which entries in op_types_to_skip_ actually matched a node group in this graph so we
+  // can warn about entries that never matched (typically a misspelled op type, e.g. "Softmaxx").
+  std::unordered_set<std::string> matched_skip_op_types;
+
   for (const std::unique_ptr<qnn::IQnnNodeGroup>& qnn_node_group : qnn_node_groups) {
+    // User opt-out: keep this op type off QNN so ORT falls it back to another EP.
+    if (!op_types_to_skip_.empty() &&
+        op_types_to_skip_.count(qnn_node_group->GetTargetNodeUnit()->OpType()) != 0) {
```
huaychou: Why do we target at QNN op_type? I think the user cannot predict the QNN graph after graph transform, so we should target at ONNX op_type.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -241,6 +237,10 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
 |---|---|
 |Op package config (string)|Register custom op packages. Format: `<OpType>:<PackagePath>:<InterfaceSymbolName>[:<Target>]`. Multiple packages can be separated by commas.|
 
+|`"op_types_to_skip"`|Description|
+|---|---|
+|Comma-separated ONNX op types (string)|Keep the listed op types off QNN regardless of backend. Matching nodes are excluded from the QNN partition so ORT falls them back to another EP (typically the CPU EP). Example: `"Softmax,LayerNormalization"`. Default is empty (no opt-out). Note: this conflicts with `session.disable_cpu_ep_fallback` — if a skipped op has no other EP to run on, session creation fails.|
```
huaychou: Where is the code of conflict handling?

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
```diff
@@ -2245,7 +2245,7 @@ index 71f9050730..1e879bf5df 100644
 +                             "qnn_saver_path", "htp_graph_finalization_optimization_mode", "qnn_context_priority",
 +                             "htp_arch", "enable_htp_fp16_precision", "offload_graph_io_quantization",
 +                             "enable_htp_spill_fill_buffer", "enable_htp_shared_memory_allocator", "dump_json_qnn_graph",
-+                             "json_qnn_graph_dir", "disable_file_mapped_weights", "htp_bf16_enable", "enable_vtcm_backup_buffer_sharing", "extended_udma", "htp_share_resource_optimization"});
++                             "json_qnn_graph_dir", "disable_file_mapped_weights", "htp_bf16_enable", "enable_vtcm_backup_buffer_sharing", "extended_udma", "htp_share_resource_optimization", "op_types_to_skip"});
```
minfhong-qti: Does this still need to update?
Take `perf_test` for example. Any new options are not required to update the `perf_test` itself as it consumes any options given to plugin EP.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: Why even allow whitespace?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -772,6 +772,29 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     ParseOpPackages(op_packages_str, op_packages, logger_);
   }
 
+  // Op types to keep off QNN (comma-separated ONNX op types). Matching nodes are excluded from
+  // the QNN partition in GetSupportedNodes so ORT falls them back to another EP (e.g. CPU EP).
+  std::string op_types_to_skip_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                 FormatEPConfigKey("op_types_to_skip"), "", op_types_to_skip_str);
+  for (const auto& token : qnn::utils::SplitString(op_types_to_skip_str, ",", /*keep_empty*/ false)) {
```
minfhong-qti: Prefer to define a separate static function for parsing.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: I think the `OpType` here is ONNX one?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1389,7 +1412,37 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
     return qnn_status.release();
   }
 
+  // Track which entries in op_types_to_skip_ actually matched a node group in this graph so we
+  // can warn about entries that never matched (typically a misspelled op type, e.g. "Softmaxx").
+  std::unordered_set<std::string> matched_skip_op_types;
+
   for (const std::unique_ptr<qnn::IQnnNodeGroup>& qnn_node_group : qnn_node_groups) {
+    // User opt-out: keep this op type off QNN so ORT falls it back to another EP.
+    if (!op_types_to_skip_.empty() &&
```
minfhong-qti: Is it redundant to check the empty first?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1389,7 +1412,37 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+        op_types_to_skip_.count(qnn_node_group->GetTargetNodeUnit()->OpType()) != 0) {
+      matched_skip_op_types.emplace(qnn_node_group->GetTargetNodeUnit()->OpType());
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_INFO,
+                  ("Skipping op type '" + qnn_node_group->GetTargetNodeUnit()->OpType() +
+                   "' per op_types_to_skip; it will fall back to another EP.")
+                      .c_str());
+      // Record in the framework op trace (when enabled) so the diagnostic report shows these
+      // nodes were intentionally skipped by the user, not silently dropped or truly unsupported.
+      if (enable_framework_op_trace_) {
```
minfhong-qti: Not sure whether it is how op trace should be used. Let @kuanyul-qti comment.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: yes, it's onnx one.

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
chunghow-qti: I think this is for test_runner. but no need for our plugin_ep_test now. will remove in next commit.

chunghow-qti:
Update the `op_types_to_skip` options to `op_types_to_filter` for some certain UC provided by @tirupath-qti 

Usage
```
# inline command + skip op
--op_types_to_filter "exclude:Softmax,Conv"
# inline command + include op (other op fallback to cpu ep)
--op_types_to_filter "include:Softmax,Conv"
# config file
--op_types_to_filter "@<path to config file>"
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: It will be look like below when enabling together.
```
.\onnxruntime_plugin_ep_onnx_test.exe `
>>     --plugin_ep_libs "QNNExecutionProvider|onnxruntime_providers_qnn.dll" `
>>     --plugin_eps "QNNExecutionProvider" `
>>     -i "backend_path|QnnHtp.dll op_type_to_filter|exclude:GroupQueryAttention enable_framework_op_trace|1" `
>>     .\models\gqa_single_layer\ -a 0.01

PS C:\Users\HCKTest\Desktop\chunghow\AISW-191770-enable-fallback-options> cat .\qnn_op_trace.json
{
  "backend_type": "htp",
  "compilation_target": {
    "device_id": 0,
    "htp_arch": "V73",
    "soc_model": 0
  },
  "model_name": "model.onnx",
  "subgraph_traces": [],
  "summary": {
    "fusion_count": null,
    "qnn_subgraphs": 0,
    "supported_nodes": 0,
    "total_onnx_nodes": 1,
    "total_qnn_ops": 0,
    "unsupported_nodes": 1
  },
  "unsupported_nodes": [
    {
      "node_index": 0,
      "node_name": "GQA_0",
      "op_type": "GroupQueryAttention",
      "reason": "Filtered off QNN by op_type_to_filter provider option"
    }
  ]
}
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Done

onnxruntime/test/providers/qnn/qnn_basic_test.cc
chunghow-qti: UT added

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Now that we support using a config file in addition to inline commands, I added the TrimWhitespace function to strip whitespace from values read fonnrom the JSON file. For inline commands, whitespace was already not allowed.
https://github.com/microsoft/onnxruntime/blob/8a6eec2f7eae875b4bbfb944f89d8c2462f12425/onnxruntime/test/perftest/strings_helper.cc#L30-L33

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Add a member function `ParseOpTypeFilter` and a static function `ParseOpTypeFilterValue` for parsing.

minfhong-qti (commented):
Since `qnn_execution_provider.cc` is getting bigger and hard to maintain, how about define all the new logic in new file(s), probably with header and encapsulate logic in a class.
Remember to leave some space for future extension for node name.

chunghow-qti:
> Since `qnn_execution_provider.cc` is getting bigger and hard to maintain, how about define all the new logic in new file(s), probably with header and encapsulate logic in a class. Remember to leave some space for future extension for node name.

Thanks, updated in lasted commit

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou: If it's ONNX op_type, do we need to identify the domain (ex: contrib ops) and filter the node before ort graph transform?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -772,6 +772,13 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     ParseOpPackages(op_packages_str, op_packages, logger_);
   }
 
+  // Op-type filter: keep certain ONNX op types on/off QNN. Value is either inline "[mode:]list"
+  // or "@<path>" to a JSON config file. See qnn_op_filter.h / QNN-ExecutionProvider.md.
+  std::string op_type_to_filter_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                 FormatEPConfigKey("op_type_to_filter"), "", op_type_to_filter_str);
+  op_filter_ = qnn::QnnOpFilter::FromOptionValue(op_type_to_filter_str, logger_);
```
minfhong-qti: How about explicit constructors?
```C++
if (op_type_to_filter_str.empty()) {
  op_filter_ = qnn::QnnOpFilter();
} else if (op_type_to_filter_str[0] == '@') {
  op_filter_ = qnn::QnnOpFilter(std::filesystem::path(op_type_to_filter_str.substr(1)));
} else {
  op_filter_ = qnn::QnnOpFilter(op_type_to_filter_str);
}
```

onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.h
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Not sure but would the naming `qnn` in the file be confusing that it is filtering based on QNN op?

onnxruntime/core/providers/qnn/qnn_op_filter.cc
```diff
@@ -0,0 +1,157 @@
...
+    return filter;  // Nothing configured or parse failed; keep default (exclude + empty list).
+  }
+
+  if (filter.mode_ == Mode::kInclude && filter.op_types_.empty()) {
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
+                "op_type_to_filter mode is 'include' but the op-type list is empty; "
+                "ALL op types will fall back off QNN.");
+  }
+
+  if (filter.IsActive()) {
```
minfhong-qti: `IsActive` as well returns `true` if "include" but no op types specified. Is it expected?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1389,7 +1396,33 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
     return qnn_status.release();
   }
 
+  const bool op_type_filter_active = op_filter_.IsActive();
+
   for (const std::unique_ptr<qnn::IQnnNodeGroup>& qnn_node_group : qnn_node_groups) {
+    // Apply the user's op-type filter: decide whether this node group is kept off QNN.
+    if (op_type_filter_active && op_filter_.ShouldFilterOff(*qnn_node_group->GetTargetNodeUnit())) {
```
minfhong-qti: I think the behavior of `IsActive` should be inside `ShouldFilterOff`.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1510,9 +1504,127 @@ TEST_F(QnnHTPBackendTests, EPRejectsDynamicShapesF32) {
...
+  // Softmax was filtered off via op_type_to_filter (exclude) -> must be on CPU EP; Conv stays on QNN.
+  std::function<void(const Ort::Session&)> ep_graph_checker =
+      MakeOpAssignmentChecker({{"Softmax", kCpuExecutionProvider}, {"Conv", kQnnExecutionProvider}});
+
+  ProviderOptions provider_options;
 #if defined(_WIN32)
+  provider_options["backend_path"] = "QnnHtp.dll";
+#else
+  provider_options["backend_path"] = "libQnnHtp.so";
+#endif
```
minfhong-qti: Use `backend_type` and as well for others.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1510,9 +1504,127 @@ TEST_F(QnnHTPBackendTests, EPRejectsDynamicShapesF32) {
...
+}
+
+// Shared model + checker helpers for the "op_type_to_filter" provider-option tests below. Every
+// one of these tests uses the same tiny Conv -> Softmax graph (both ops are static-shape and
+// QNN-supported, so without any filter both land on QNN) and only differs in the filter value and
+// the expected per-op EP assignment. Factoring these out keeps each test to just its option value
+// and expectation.
+
+// Builds a Conv -> Softmax graph. Only the graph name varies between tests.
+static std::function<void(ModelTestBuilder&)> MakeConvSoftmaxBuilder(const std::string& graph_name) {
```
minfhong-qti, line 1537: Since there are new files, should the testcases adopt Yu-Hong's UT refactor?

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -35,6 +35,21 @@ class QnnOpConfigWrapper;
 class QnnModelWrapper;
 
 namespace utils {
+/**
+ * Trims surrounding ASCII whitespace (space, tab, CR, LF) and returns a copy.
+ * Returns an empty string if the input is empty or all-whitespace.
+ * @param s The string to trim.
+ * @return The trimmed copy.
+ */
+inline std::string TrimWhitespace(std::string_view s) {
```
minfhong-qti, line 45: Remember to add corresponding testcase in UT file.

onnxruntime/core/providers/qnn/qnn_op_filter.cc
chunghow-qti: Yes, that's the scenario where I think user specify all op_type fallback to cpu ep.
And there's a warning message above to notice user if they are intended to do this
```
  if (filter.mode_ == Mode::kInclude && filter.op_types_.empty()) {
                "op_type_to_filter mode is 'include' but the op-type list is empty; "
                "ALL op types will fall back off QNN.");
  }
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: I prefer we put the explicit ctor into `QnnOpFilter::FromOptionValue`.
This allow us to handle throw in `QnnOpFitler`.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
chunghow-qti: I prefer add `unit/qnn_onnx_op_filter_test.cc` (QnnUnit_OnnxOpFilterTest) covering the new OnnxOpFilter logic. And exsiting tests stay in qnn_basic_test.cc now. We can have a seperate  PR for migration since the change may be large.

onnxruntime/core/providers/qnn/builder/qnn_utils.h
chunghow-qti: added at `onnxruntime\test\providers\qnn\unit\qnn_utils_test.cc`

onnxruntime/test/providers/qnn/qnn_basic_test.cc
chunghow-qti: Done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Done

onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.h
chunghow-qti: update file name to `qnn_onnx_op_filter.cc` and class name to `OnnxOpFilter`

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -241,6 +241,10 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
...
...pace, **do not put spaces in the inline value**.<br><br>**Config-file form:** `"@C:\path\to\filter.json"` — the value is a path (prefixed with `@`) to a JSON file, which avoids an over-long command line. File schema: `{ "mode": "exclude" \| "include", "op_types": ["Softmax", "Conv"] }` (JSONC comments allowed). Whitespace is unrestricted inside the file.<br><br>**Matching granularity:** the filter is applied per QNN *node group* (a single op, or a fusion of several ops that QNN maps to one operator, such as a QDQ group `DQ → Conv → Q`), keyed on the group's **target** (main compute) op type. So `exclude:DequantizeLinear` has no effect (DQ/Q wrap a group, they are not its target), and an op that only appears inside a fusion won't match by itself — filter on the group's target op type instead.<br><br>An op type in the list that matches no node in the graph is ignored with a WARNING (typically a typo). `include` with an empty list forces **all** ops off QNN (logged at WARNING). Default (option unset) is `exclude` with an empty list — all ops go to QNN. Note: this conflicts with `session.disable_cpu_ep_fallback` — if a filtered-off op has no other EP to run on, session creation fails.|
```
kuanyul-qti: Could you add an example that shows how to filter a qnn node group?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1390,6 +1397,31 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+          for (const OrtNode* node : node_unit->GetAllNodesInGroup()) {
+            Ort::ConstNode const_node(node);
+            unsupported_nodes.push_back({
+                std::string(const_node.GetName()),
+                std::string(const_node.GetOperatorType()),
+                const_node.GetId(),
+                "Filtered off QNN by op_type_to_filter provider option",
+            });
+          }
+        }
```
kuanyul-qti: Can the common part be extracted as a function which also appears in line 1437?

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: Done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: I don't think MSFT would let us modify ORT's core graph transform to fit our use case, and more fundamentally, QNN EP's `GetCapability/GetSupportedNodes` only runs after ORT's Level 1 transforms are already applied, so there's no earlier hook for the EP to intercept before that point anyway.

So in practice we can only exclude/include ops on the graph as the EP sees it at that stage. This also matches the granularity our existing op-dispatch logic already uses: 
* `op builder` lookup (OpBuilderRegistrations::op_builder_map_) keys purely on ONNX op_type too, with no domain distinction, so `op_affinity` isn't introducing a new blind spot here, just following the same granularity the rest of the EP already uses.

This ties into the `op_name` qualifier we're planning for phase 2 as well: since the `op name` on the EP side may not match the original graph's name after transforms, our plan is to let users dump the graph as the EP actually sees it (dump_qnn_ep_input_graph) so they can find the real name to target.

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: add a warning message

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -871,88 +772,68 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     ParseOpPackages(op_packages_str, op_packages, logger_);
   }
 
-  // HTP graph finalization optimization mode
-  std::string htp_graph_finalization_opt_mode_str;
+
```
huaychou, line 891: nit: remove

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1583,3 +1532,24 @@ Currently, only pre-compiled models with EPContext nodes are supported. The  exa
...
 <graph name>;<adapter binary section path>
 ```
+
+
+## OP Affinity - Details
+
+#### Matching examples
+1. For a graph `Conv → Softmax` (two separate node groups), `"exclude:Softmax"` keeps `Conv` on QNN and falls `Softmax` back to
+another EP (typically CPU).
+2. For a fused QDQ group `DequantizeLinear → Conv → QuantizeLinear`, filter on the group's target op type: `"exclude:Conv"` removes
```
huaychou, line 1681: How do users know which op type will be a fused QDQ groups?

onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.h
```diff
@@ -0,0 +1,103 @@
...
+  // misspelled op type, which would otherwise be a silent no-op. Clears the match bookkeeping on every
+  // exit path, so repeated calls (GetSupportedNodes runs multiple times per session) each start clean.
+  // If the filter is scoped to a backend the session isn't running, no entry can match by design, so
+  // the per-entry warnings are suppressed in favor of a single INFO line. Takes the session backend to
+  // tell the two cases apart.
+  void WarnUnmatchedEntries(QnnBackendType backend_type, const Ort::Logger& logger) const;
+
+ private:
+  Mode mode_ = Mode::kExclude;
+  std::unordered_set<std::string> op_types_;
```
huaychou: Can we rename this variable? op_types_ is not very clear.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -248,6 +241,10 @@ Warning: Enabling HTP Monolithic LSTM may improve session creation time, but thi
...
...ed op types, which fall back to another EP (typically the CPU EP); `include` claims *only* the listed op types and forces every other op type to fall back.<br><br>**Inline form:** `"exclude:Softmax,LayerNormalization"` or `"include:Conv,MatMul"`. `mode:` is optional and defaults to `exclude`, so `"Softmax"` ≡ `"exclude:Softmax"`. Op types are comma-separated, matched **case-sensitively**, and must not contain spaces.<br><br>**Backend scope (optional):** prefix with a backend name to scope the filter to one backend, e.g. `"htp:exclude:Softmax"`. Accepts the same names as `backend_type`: `cpu`/`gpu`/`htp`/`dsp`/`ir` (`htp` also matches `htp_fp16` sessions). Without this prefix, the filter applies no matter what backend the session is actually running (via `backend_type`/`backend_path`).<br><br>**Config-file form:** `"@C:\path\to\filter.json"` — a path (prefixed with `@`) to a JSON file, for long lists. Schema: `{ "backend": "htp", "mode": "exclude" \| "include", "op_types": ["Softmax", "Conv"] }` (`"backend"` optional; JSONC comments allowed).<br><br>See [OP Affinity - Details](#op-affinity---details) for matching granularity, defaults, and the `disable_cpu_ep_fallback` interaction.|
```
huaychou: Do we need to add cpu backend to document since we don't support CPU backend by default?

huaychou:
In addition, there is a conflict with mainline. Please also fix the conflict.

onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.h
chunghow-qti: do you have any recommendation?

huaychou: selected_op_types_?

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: I think we can leave it for now. Since in `Qnn-ExecutionProviders.md` the backend_type cpu is still there

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Done

onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.h
chunghow-qti: Done

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: I think we can leverage the `framework_op_trace` to help dump the target onnx_op_type of a qnn node group.
I think there is a feasible change to make this happened?

chunghow-qti: We can have the framework op trace dump updated to below
```
 "sources": [
    {"name": "dq_1",   "type": "OP"},
    {"name": "conv_1", "type": "OP", "is_target": true},  // point out the target onnx type
    {"name": "q_1",    "type": "OP"}
 ]
```

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -20,6 +20,7 @@
 #include "core/providers/qnn/builder/qnn_model.h"
 #include "core/providers/qnn/builder/qnn_model_wrapper.h"
 #include "core/providers/qnn/builder/op_tracing/qnn_op_tracing.h"
+#include "core/providers/qnn/qnn_onnx_op_affinity.h"
```
minfhong-qti: Reorder it.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -247,6 +248,11 @@ class QnnEp : public OrtEp, public ApiPtrs {
   qnn::ModelSettings model_settings_ = {};
   qnn::HtpGraphConfigs_t htp_graph_configs_;
 
+  // Op Affinity controlling which ONNX op types the QNN EP claims in GetSupportedNodes, configured via
+  // the "op_affinity" provider option. See qnn_onnx_op_affinity.h. Default-constructed =
+  // inactive (all ops go to QNN).
```
minfhong-qti:
```suggestion
  // Op Affinity controlling which ONNX op types the QNN EP claims in GetCapability, configured via
  // the "op_affinity" provider option.
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -889,6 +889,21 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
   }
 
+  // Op affinity: keep certain ONNX op types/op names on/off QNN, optionally scoped to a backend.
+  std::string op_affinity_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                 FormatEPConfigKey("op_affinity"), "", op_affinity_str);
+  op_affinity_ = qnn::OnnxOpAffinity::FromOptionValue(op_affinity_str, logger_);
+  // Heads-up only (not a hard error): if op_affinity keeps a node off QNN and CPU EP fallback is also
+  // disabled, session creation will fail. Can't tell "will definitely conflict" from "possibly inert"
+  // here -- qnn_backend_manager_ (needed to know the running backend) doesn't exist yet.
```
minfhong-qti: Removed this comment which can be inferred from below message.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1528,6 +1561,23 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
...
+      // Record in the framework op trace (when enabled) so the report shows these nodes were
+      // intentionally filtered by the user, not truly unsupported.
+      if (enable_framework_op_trace_) {
+        AppendUnsupportedNodesForGroup(*qnn_node_group, "Filtered off QNN by op_affinity provider option",
+                                       unsupported_nodes);
+      }
+      continue;
+    }
+
     Ort::Status support_status = qnn_node_group->IsSupported(qnn_model_wrapper, logger_);
```
```suggestion
    Ort::Status support_status;

    // Apply the user's op affinity. Inactive affinity always returns false, so this is a no-op when
    // the option is unset.
    if (op_affinity_.ShouldFilterOff(*qnn_node_group->GetTargetNodeUnit(),
                                     qnn_backend_manager_->GetQnnBackendType())) {
      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE,
                  ("op_affinity: keeping op type '" + qnn_node_group->GetTargetNodeUnit()->OpType() +
                   "' off QNN; it will fall back to another EP.")
                      .c_str());
                      
      support_status = MAKE_EP_FAIL("Filtered off QNN by op_affinity provider option");
    } else {
      support_status = qnn_node_group->IsSupported(qnn_model_wrapper, logger_);
    }
```
And no need to define another `AppendUnsupportedNodesForGroup`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc

onnxruntime/core/providers/qnn/qnn_execution_provider.h
chunghow-qti: done

onnxruntime/core/providers/qnn/qnn_execution_provider.h
chunghow-qti: done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Done

minfhong-qti (approved):

docs/execution_providers/QNN-ExecutionProvider.md
huaychou: Should we add documentation saying we need to dump the framework op trace and then use op affinity?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1527,6 +1527,435 @@ TEST_F(QnnHTPBackendTests, EPRejectsDynamicShapesF32) {
...
+  auto model_build_fn = MakeConvSoftmaxBuilder("ep_filter_include_empty_graph");
+
+  // include + empty list -> everything falls back to CPU.
+  std::function<void(const Ort::Session&)> ep_graph_checker =
+      MakeOpAssignmentChecker({{"Softmax", kCpuExecutionProvider}, {"Conv", kCpuExecutionProvider}});
+
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = "htp";
+  provider_options["offload_graph_io_quantization"] = "0";
+  provider_options["op_affinity"] = "include:";  // empty list -> all ops off QNN + WARNING
```
huaychou, line 1884: This provider option looks pretty weird 🤔 How can specifying op_affinity as include make all ops off QNN?

onnxruntime/core/providers/qnn/qnn_onnx_op_affinity.cc
```diff
@@ -0,0 +1,261 @@
...
+
+OnnxOpAffinity OnnxOpAffinity::FromOptionValue(const std::string& option_value, const Ort::Logger& logger) {
+  const std::string trimmed = utils::TrimWhitespace(option_value);
+  if (trimmed.empty()) {
+    return OnnxOpAffinity{};  // No filter configured; keep default (exclude + empty list).
+  }
+
+  OnnxOpAffinity filter;
+  try {
+    if (trimmed[0] == '@') {
```
huaychou, line 159: Just wondering, can we just specify the filename without "@"? It is seldom to use "@" to specify filename.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -1527,6 +1527,435 @@ TEST_F(QnnHTPBackendTests, EPRejectsDynamicShapesF32) {
...
+
+  // Softmax was filtered off via op_affinity (exclude) -> must be on CPU EP; Conv stays on QNN.
+  std::function<void(const Ort::Session&)> ep_graph_checker =
+      MakeOpAssignmentChecker({{"Softmax", kCpuExecutionProvider}, {"Conv", kQnnExecutionProvider}});
+
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = "htp";
+  provider_options["offload_graph_io_quantization"] = "0";
+  // "exclude:" prefix omitted on purpose -> exercises the default-mode path.
+  provider_options["op_affinity"] = "Softmax";  // keep Softmax off QNN
```
huaychou, line 1597: The provider option looks weird. Setting Softmax in op_affinity makes it off QNN 🤔

PR #602, opened 2026-07-06
[QNN EP] Add component-level unit tests for QnnEp

yuhuchua-qti:
## Summary

Component-level unit tests for `QnnEp`
(`onnxruntime/core/providers/qnn/qnn_execution_provider.cc`), exercising the
constructor's option parsing, static impl methods, and early-return error paths
without loading a real QNN backend or constructing a full ORT session.

- **84 tests total, all unit-tier**, in `unit/qnn_execution_provider_test.cc`:
  - `QnnUnit_ExecutionProviderTest` — 82 tests (63 `TEST_F` + 19 `TEST_P`
    across three `INSTANTIATE_TEST_SUITE_P` for the valid htp_performance_mode /
    qnn_context_priority / htp_arch enum values). No backend, no session.
  - `QnnUnit_ExecutionProviderHtpTest` — 2 real-backend tests that
    `dlopen(libQnnHtp.so)` + `SetupBackend` (mirrors `QnnUnit_BackendManagerHtpTest`),
    `GTEST_SKIP()` when the SDK is unavailable. Still no ORT session, so unit-tier.
- Unit tests use `EpStubContext` (stubbed OrtApi + `session_config` map) and
  `UseGlobalEpStubs` RAII so the QnnEp ctor runs with a fake logger/backend path.
- The `_Logs{Error,Warning,Info,Verbose}` ctor tests assert the exact log
  severity + message substring via a `Logger_LogMessage` recording sink
  (`ExpectLogged`), not just no-throw.

## Coverage (qnn_execution_provider.cc)

Measured from `onnxruntime_provider_test` only (unit-only =
`QnnUnit_ExecutionProviderTest.*`; combined = `*Qnn*`). LF=1603 on the current
rebased source.

| Metric | Before (no unit suite) | After |
|---|---|---|
| Unit-only | 0% (0/1603) | **39.7% (637/1603)** |
| Combined  | ~80% | **82.0% (1314/1603)** |

Combined stays at 82.0% because the covered ctor branches are already reached by
the HTP op suites in the combined run — the unit tests make them reachable
**unit-only** (fast, no session/backend), which is the point.

## Why coverage is below 90%

The remaining ~18% is not unit-testable without either coverage theater or real
hardware/session, and is functionally validated by the E2E/op suites outside
this line-coverage measurement:

| Lines | Category | Reason |
|---|---|---|
| GetCapabilityImpl / CompileImpl (~640) | session | require a full ORT `InferenceSession` |
| GetSupportedNodes, InitQnnHtpGraphConfigs, EPContext partition/compile tails | backend/graph | need real backend `SetupBackend` + `OrtGraph*` |
| HTP/DX12 shared-memory allocator setup | env-limited | require real NPU/GPU hardware |
| file-mapped weights (RPCMem) | env-limited | forced off on Linux x86_64 |
| ProbeDumpDirectoryWritable secondary failure branches | defensive | no deterministic writable-but-probe-fails path |
| Genie/DLC (WoS-only), DX12 (Windows) paths | platform-gated | dead code on Linux x86_64 |

## Test plan

- [x] `onnxruntime_provider_test --gtest_filter="QnnUnit_ExecutionProviderTest.*"` — pass
- [x] `--gtest_filter="QnnUnit_*"` — 745/745 pass (incl. QnnUnit_ExecutionProviderHtpTest)
- [x] `--gtest_filter="*Qnn*"` — no failures in the unit suite (pre-existing
      op/backend-suite failures are x86_64-emulator/environment, unrelated to this change)
- [x] `lint_and_fix` clean

### DIFF
--- a/onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
+++ b/onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
@@ -482,6 +482,55 @@ TEST_F(QnnUnit_ExecutionProviderTest, Ctor_ContextPriorityInvalid_SetsUndefined)
   EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
 }
 
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpPerformanceModeBalanced_Succeeds) {
+  EpStubContext ctx;
+  ctx.session_config[EPKey("htp_performance_mode")] = "balanced";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpPerformanceModeDefault_Succeeds) {
+  EpStubContext ctx;
+  ctx.session_config[EPKey("htp_performance_mode")] = "default";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpPerformanceModeHighPerformance_Succeeds) {
+  EpStubContext ctx;
+  ctx.session_config[EPKey("htp_performance_mode")] = "high_performance";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpPerformanceModeHighPowerSaver_Succeeds) {
+  EpStubContext ctx;
+  ctx.session_config[EPKey("htp_performance_mode")] = "high_power_saver";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpPerformanceModeLowBalanced_Succeeds) {
+  EpStubContext ctx;
+  ctx.session_config[EPKey("htp_performance_mode")] = "low_balanced";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpPerformanceModeLowPowerSaver_Succeeds) {
+  EpStubContext ctx;
+  ctx.session_config[EPKey("htp_performance_mode")] = "low_power_saver";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpPerformanceModePowerSaver_Succeeds) {
+  EpStubContext ctx;
+  ctx.session_config[EPKey("htp_performance_mode")] = "power_saver";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+}
+
 // ===========================================================================
 // Group 6: Constructor — HTP graph finalization opt mode, HTP architecture
 // ===========================================================================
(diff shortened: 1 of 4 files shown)


onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
```diff
@@ -0,0 +1,1102 @@
...
+  auto ep = MakeEp(*factory, ctx);
+
+  // Only 3 colon-separated fields; function expects 6.
+  OrtCompiledModelCompatibility compat = OrtCompiledModelCompatibility_EP_SUPPORTED_OPTIMAL;
+  OrtStatus* s = ep->ValidateCompiledModelCompatibilityInfo(nullptr, 0, "1:2:3", &compat);
+  EXPECT_EQ(s, nullptr);
+  EXPECT_EQ(compat, OrtCompiledModelCompatibility_EP_NOT_APPLICABLE);
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, ValidateCompatibilityInfo_BadVersionFormat_NotApplicable) {
```
huaychou, line 920: **[N-1] (Minor)** Group 10 compatibility tests are vacuous passes on the x86_64 test host; the parsing logic the comments describe is never reached

`ValidateCompatibilityInfo_TooFewFields_NotApplicable` (line 908) and `ValidateCompatibilityInfo_BadVersionFormat_NotApplicable` (line 920) carry comments claiming they exercise the "function expects 6 fields" and "version field wrong format (2 parts, not 3)" parsing branches. But `ValidateCompiledModelCompatibilityInfo` in `qnn_execution_provider.cc` has an early guard `#if !defined(__aarch64__) && !defined(_M_ARM64) && !defined(_M_ARM64EC)` (around source line 3209) that logs "Skip compatibility validation on x86 platforms." and returns `EP_NOT_APPLICABLE` before any field parsing.

**Impact**: these unit tests are only compiled/run on Linux x86_64 (where `QNN_EP_INTERNAL_SYMBOL_ACCESS` is set — the coverage build). On that platform, any non-empty info string returns early at the x86 guard, so the field-count and version-format parsing code is **never reached**. All three Group 10 tests (including `EmptyString_NotApplicable`) pass, but for a reason unrelated to the branch their comments describe — they all actually validate the single x86-skip path. Not a bug, but it is misleading coverage: if someone later changes the field-parsing logic, these tests will not catch the regression.

**Suggested fix:**

Update the comments on these two tests to state that on the x86_64 test host they actually hit the x86 platform-skip branch (optionally assert the "Skip compatibility validation on x86 platforms." log). To genuinely cover the field/version parsing, run on aarch64 or extract the testable parsing logic into a platform-guard-free free function and test it directly. At minimum, note this limitation in the test or PR description so the parsing path is not assumed to be covered.

onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
```diff
@@ -0,0 +1,1102 @@
...
+}
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpFP16PrecisionInvalid_LogsVerbose) {
+  EpStubContext ctx;
+  ctx.log_severity = ORT_LOGGING_LEVEL_VERBOSE;
+  ctx.session_config[EPKey("enable_htp_fp16_precision")] = "invalid";
+  // Invalid value leaves enable_HTP_FP16_precision_ at its default true.
+  // On Linux x86_64, FP16+no-soc_model throws; provide a soc_model so the
+  // constructor can proceed past the FP16 validation. The invalid value is
+  // parsed by ParseBoolOption, which logs VERBOSE (not ERROR) at line 301.
```
huaychou, line 645: **[T-1] (Nit)** Comments cite absolute source line numbers (line 301 / 547 / 1084) that drift as the file changes

Several test comments cite absolute line numbers in `qnn_execution_provider.cc`, e.g. line 645 "logs VERBOSE (not ERROR) at line 301", line 750 "the else-branch at line 301", line 758 "(line 547)", lines 770/774 "current_path() (line 1084)".

**Impact**: these line numbers drift with any insertion/deletion in the source file and can later point readers to the wrong location. A readability nit only; does not affect test correctness.

**Suggested fix:**

Reference the function name or log message (e.g. "the else-branch of ParseBoolOption", "the IR backend path info log in InitQnnSerializerConfig") instead of absolute line numbers, so the comments stay accurate as the source evolves.

qti-chuteng (approved):

huaychou (commented):
Two questions:
1. Do we need to add a test for `enable_htp_fp16_clamp_overflow`?
2. Is it possible to test arm64ec artifacts?

onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
yuhuchua-qti: Fixed in PR #719 (commit 4d5e1602f2) — replaced the 5 absolute references (line 301, line 547, line 1084 ×2) with function/branch names (in its else-branch, in InitQnnSerializerConfig, in the QnnEp ctor). Reader can just grep for the anchor.

yuhuchua-qti:
> 1. Do we need to add a test for `enable_htp_fp16_clamp_overflow`?

Added Ctor_HtpFp16ClampOverflowTrueOnUnsupportedSdk_LogsWarning in PR #719 (commit 625494614b). 
It sets enable_htp_fp16_clamp_overflow=1 + soc_model=60 (to bypass the FP16+no-soc_model early throw) and asserts the ctor logs the "requested but the QNN HTP SDK in use does not support it" WARNING.

This branch is reachable on the current coverage host because QAIRT 2.48 = QNN API 2.37 < 2.38, so QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE is not defined (see qnn_def.h:40-45 for the version gate). 
The test itself is wrapped in #ifndef QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE, mirroring the source's own #ifndef around the warning branch — a future QAIRT ≥ 2.49 uplevel will remove both the source branch and this test at preprocessor time, so this won't become a landmine for the uplevel PR.

Context: the existing Conv_Fp16ClampOverflow_Smoke integration test in conv_test.cc:1199 (added in #686 alongside the feature) exercises the E2E path but 
- doesn't assert the WARNING message/severity and 
- SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(V75) on x86_64. So this is complementary coverage, not duplication.

onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
yuhuchua-qti: Thanks for catching this. 

Fixed in follow-up PR #719 (commit 4d5e1602f2):
- Consolidated TooFewFields + BadVersionFormat into one honestly-named NonEmptyOnX86Host_LogsSkipAndReturnsNotApplicable, and renamed EmptyString_NotApplicable → EmptyString_LogsNoInfoAndReturnsNotApplicable.
- Both tests now assert the actual WARNING string via ExpectLogged, so which branch fires is verified rather than assumed.
- Group 10 header notes that the field-parsing branches downstream of the x86-skip guard are unreachable from this UT tier and left for a future arm64 UT expansion.

yuhuchua-qti:
> 2. Is it possible to test arm64ec artifacts?

Not from this UT tier today — the whole unit/ directory is gated on QNN_EP_INTERNAL_SYMBOL_ACCESS, which is only enabled by the Linux x86_64 coverage build.
The plan is to stabilise this infra on Linux first and then evaluate extending to other platforms. 
arm64ec's real testing needs (ABI, x64 interop, linker behaviour) are best covered by integration tests running on arm64ec CI anyway — UT mocks can't reach that layer.

PR #603, opened 2026-07-06
[QNN EP] Fix enable_htp_fp16_precision default value in documentation

kuanyul-qti:
### Description
Update the document so the option's default behavior matches the actual code.

### Motivation and Context
The doc incorrectly labeled '1' as the default value for enable_htp_fp16_precision. The actual code default is '0' (disabled).


qti-chuteng:
I think that we only support htp_fp16 right now (there should be no htp_fp32 anymore)

chunghow-qti:
Just wondering, are we planning to remove this option? It appears to be a no-op now.

PR #604, opened 2026-07-06
[BQ] Extend HTP native BQ to MatMulNBits and refine Conv path

minfhong-qti:
### Description

Route block-quantized weights to the HTP native BQ kernel (QNN_QUANTIZATION_ENCODING_BLOCK) when supported, avoiding the extra INT16->FP16 activation dequant and FP16->INT16 output quantize that the non-native BW_FLOAT_BLOCK kernel requires.

- conv_op_builder: select native BLOCK vs BW_FLOAT_BLOCK via bq::IsHTPSupportedNativeBQ; only insert activation dequant / output quantize on the non-native fallback path.
- matmulnbits_op_builder: accept uint16/int16 activation on HTP, apply the same native-vs-fallback selection.
- qnn_quant_params_wrapper: clarify BQ fallback log wording.
- tests: add native BQ Conv and MatMulNBits QDQ coverage plus fallback edges.

### Motivation and Context

Support HTP native BQ for Conv2d and MatMulNBits (which essentially built into Conv2d).

**\*\*Note\*\*** Current implementation is guarded to adopt native BQ path only if QAIRT >= 2.51 to avoid breaking any test until QAIRT is upleveled, and so do the testcases.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
@@ -401,7 +401,7 @@ Ort::Status ConvOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
                                          std::vector<std::string>& input_names,
                                          bool do_op_validation) const {
   const auto& inputs = node_unit.Inputs();
-  assert(inputs.size() >= 2);
+  RETURN_IF_NOT(inputs.size() >= 2, "Expecting at least two inputs.");
 
   std::vector<uint32_t> input0_shape;
   RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[0].shape, input0_shape),
@@ -426,7 +426,7 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
   OnnxConvType conv_type = {};
   RETURN_IF_ERROR(GetOnnxConvType(node_unit.OpType(), conv_type));
 
-  assert(num_inputs >= 2);  // Checked by IsOpSupported.
+  RETURN_IF_NOT(num_inputs >= 2, "Expecting at least two inputs.");  // Checked by IsOpSupported.
 
   //
   // Input 0
@@ -486,8 +486,8 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
                                ? "weight quant params not LPBQ (enable_block_quant_weight_optimization=0, non-INT4, or asymmetric ZP)"
                                : "unknown";
       ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
-                  ("Conv weight encoding: BW_FLOAT_BLOCK for " + input1_name +
-                   " [LPBQ skipped: " + reason + "]")
+                  ("Conv weight encoding: non-LPBQ block-quant path for " + input1_name +
+                   " [LPBQ skipped: " + reason + "] — native BQ (BLOCK) vs BW_FLOAT_BLOCK is decided below")
                       .c_str());
     }
   }
@@ -495,15 +495,10 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
   if (is_bq_weight && !use_lpbq_path) {
     RETURN_IF_NOT(input_info.is_initializer, "QNN EP: BQ Conv weight must be a constant initializer");
 
-    // Activation handling: BQ kernel (BW_FLOAT_BLOCK) requires FP16, so INT16 → FP16 via Dequantize.
-    if (utils::IsQuant16bit(act_dtype)) {
-      // Reuse the original DequantizeLinear output name for the FP16 tensor so the QNN graph
-      // stays aligned with the ONNX graph naming.
-      const std::string fp16_act_name = Ort::ConstNode(&node_unit.GetNode()).GetInputs()[0].GetName();
-      RETURN_IF_ERROR(bq::AddInt16ToFp16DequantForActivation(qnn_model_wrapper, act_name,
-                                                             fp16_act_name, do_op_validation, "Conv"));
-      input_names[0] = fp16_act_name;
-    }
+    // Activation handling: the non-native BQ kernel (BW_FLOAT_BLOCK) requires FP16, so INT16 → FP16 via
+    // Dequantize. The native BQ kernel (ENCODING_BLOCK) consumes INT16 directly and needs no Dequantize.
+    // The insertion is deferred until the offsets are known, since offset symmetry is one of the
+    // native-BQ preconditions checked by bq::IsHTPSupportedNativeBQ below.
 
     // Common: transpose weight data from OIHW→HWIO (or IOHW→HWIO for ConvTranspose).
     // TransposeFromNchwToHwcn unpacks INT4 to INT8 internally (1 byte per element).
@@ -557,10 +552,51 @@ Ort::Status ConvOpBuilder::ProcessConv2D3DInputs(QnnModelWrapper& qnn_model_wrap
     RETURN_IF_ERROR(bq::ComputeBQOffsets(qnn_model_wrapper, inputs[1].quant_param->zero_point,
                                          is_unsigned_weight, bitwidth, OC * nb, offsets_qnn));
 
-    QnnQuantParamsWrapper bq_quant_params = QnnQuantParamsWrapper::BwFloatBlock(gsl::span<const float>(scales_qnn),
-                                                                                gsl::span<const float>(offsets_qnn),
-                                                                                bitwidth,
-                                                                                gsl::span<const uint32_t>(block_size_arr));
+    // Prefer the HTP native BQ kernel (QNN_QUANTIZATION_ENCODING_BLOCK) when the BQ parameters satisfy its
+    // constraints: it consumes the INT16 activation and produces an INT16 output directly, avoiding the
+    // INT16→FP16 activation Dequantize and FP16→INT16 output Quantize that BW_FLOAT_BLOCK requires.
+    // Restricted to the bias-less case (num_inputs < 3): with a bias present the BW_FLOAT_BLOCK path also
+    // dequantizes the INT32 bias to FP16 (see ProcessConv2D3DBias), which the native kernel does not expect.
+    const bool is_supported_native_bq = num_inputs < 3 &&
+                                        bq::IsHTPSupportedNativeBQ(qnn_model_wrapper.GetHtpArch(),
+                                                                   act_dtype,
+                                                                   bitwidth,
+                                                                   static_cast<uint32_t>(block_size),
+                                                                   static_cast<uint32_t>(OC),
+                                                                   gsl::span<const float>(offsets_qnn));
+
+    QnnQuantParamsWrapper bq_quant_params;
+    if (is_supported_native_bq) {
+      // Native BQ requires symmetric quantization; IsHTPSupportedNativeBQ has verified all offsets are zero.
+      const std::vector<int32_t> offsets_sym(offsets_qnn.size(), 0);
+      bq_quant_params = QnnQuantParamsWrapper::Block(gsl::span<const float>(scales_qnn),
+                                                     gsl::span<const int32_t>(offsets_sym),
+                                                     gsl::span<const uint32_t>(block_size_arr));
+      ORT_CXX_LOG(logger,
+                  ORT_LOGGING_LEVEL_VERBOSE,
+                  ("Conv weight encoding: native BQ (BLOCK) for " + input1_name).c_str());
+    } else {
+      bq_quant_params = QnnQuantParamsWrapper::BwFloatBlock(gsl::span<const float>(scales_qnn),
+                                                            gsl::span<const float>(offsets_qnn),
+                                                            bitwidth,
+                                                            gsl::span<const uint32_t>(block_size_arr));
+      ORT_CXX_LOG(logger,
+                  ORT_LOGGING_LEVEL_VERBOSE,
+                  ("Conv weight encoding: BW_FLOAT_BLOCK for " + input1_name).c_str());
+
+      // Non-native BQ kernel computes in FP16, so dequantize the INT16 activation first.
+      if (utils::IsQuant16bit(act_dtype)) {
+        // Reuse the original DequantizeLinear output name for the FP16 tensor so the QNN graph
+        // stays aligned with the ONNX graph naming.
+        const std::string fp16_act_name = Ort::ConstNode(&node_unit.GetNode()).GetInputs()[0].GetName();
+        RETURN_IF_ERROR(bq::AddInt16ToFp16DequantForActivation(qnn_model_wrapper,
+                                                               act_name,
+                                                               fp16_act_name,
+                                                               do_op_validation,
+                                                               "Conv"));
+        input_names[0] = fp16_act_name;
+      }
+    }
 
     // Always use SFIXED_POINT_8: unsigned types are pre-converted by TransformUnsignedToSignedFixedPoint.
     QnnTensorWrapper bq_weight_wrapper(input1_name, tensor_type,
@@ -749,7 +785,7 @@ Ort::Status ConvOpBuilder::ProcessConv1DInputs(QnnModelWrapper& qnn_model_wrappe
   OnnxConvType conv_type = {};
   RETURN_IF_ERROR(GetOnnxConvType(node_unit.OpType(), conv_type));
 
-  assert(num_inputs >= 2);  // Checked by IsOpSupported.
+  RETURN_IF_NOT(num_inputs >= 2, "Expecting at least two inputs.");  // Checked by IsOpSupported.
 
   //
   // Input 0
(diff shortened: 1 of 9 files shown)


onnxruntime/core/providers/qnn/builder/qnn_bq_utils.h
```diff
@@ -92,6 +92,14 @@ Ort::Status AddFp16ToInt16QuantizeOutput(QnnModelWrapper& qnn_model_wrapper,
                                          std::vector<uint32_t> output_shape,
                                          bool do_op_validation);
 
+// Determine whether given BQ parameters are natively supported by HTP. If true, activation data type can be kept in
+// fixed point; otherwise, insert Dequantize and Quantize nodes around to make activation in FP16.
+bool IsHTPSupportedNativeBQ(Qnn_DataType_t act_data_type,
```
huaychou: **[M-1] (Major)** IsHTPSupportedNativeBQ undocumented bias precondition and INT16 exclusion rationale; fallback for biased Conv emits no log

`IsHTPSupportedNativeBQ` returns false when `act_data_type != QNN_DATATYPE_UFIXED_POINT_16`: INT16 (SFIXED_POINT_16) activations always fall back to BW_FLOAT_BLOCK. Two documentation gaps exist:

- The comment only lists "activation data type: UFIXED_POINT_16" without explaining **why SFIXED_POINT_16 is excluded** — readers cannot tell whether this is a hard SDK constraint or an unvalidated scope.
- conv_op_builder gates on `num_inputs < 3` (bias-less) before calling this function, but this **bias precondition does not appear in the function signature or doc-comment**. If a future caller omits this pre-screen, a biased BQ Conv will be incorrectly classified as native-BQ-eligible. Additionally, the BW_FLOAT_BLOCK fallback for biased Conv emits no VERBOSE log explaining the reason.
**Impact:** (1) INT16 activations silently fall back to BW_FLOAT_BLOCK (correct behavior, undocumented). (2) Bias precondition absent from contract — future callers may misuse the function. (3) Biased-Conv fallback has no log, making it harder to debug.

**Suggested fix:**

(a) Update the doc-comment of `IsHTPSupportedNativeBQ`: state that only uint16 activation is supported (cite SDK constraint if known) and explicitly note the function does not check for bias — callers must pre-screen. (b) Add a VERBOSE log in conv_op_builder's biased-Conv BW_FLOAT_BLOCK fallback path: `"Conv weight encoding: BW_FLOAT_BLOCK for " + input1_name + " [native BQ skipped: bias present]"`

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -502,19 +517,44 @@ Ort::Status MatMulNBitsOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapp
...
-                                                                   input_names[0],
-                                                                   fp16_act_name,
-                                                                   do_op_validation,
-                                                                   "MatMulNBits"));
-            input_names[0] = fp16_act_name;
+
+          // Prefer the HTP native BQ kernel (QNN_QUANTIZATION_ENCODING_BLOCK) when the BQ parameters satisfy
+          // its constraints: it consumes the INT16 activation and produces an INT16 output directly, avoiding
+          // the INT16→FP16 activation Dequantize and FP16→INT16 output Quantize that BW_FLOAT_BLOCK requires.
+          const Qnn_DataType_t act_dtype = qnn_model_wrapper.GetQnnTensorWrapper(input_names[0]).GetTensorDataType();
```
huaychou, line 525: **[M-2] (Major)** matmulnbits_op_builder calls GetQnnTensorWrapper(input_names[0]) in the non-LPBQ path without an IsQnnTensorWrapperExist guard

At line 524 of `matmulnbits_op_builder.cc`, the non-LPBQ BQ path calls:

```
const Qnn_DataType_t act_dtype = qnn_model_wrapper.GetQnnTensorWrapper(input_names[0]).GetTensorDataType();
```

without first checking that the wrapper exists. On the current code paths the activation tensor is guaranteed to be registered (added during ProcessInputs section 1 Reshape/Cast), so the call is safe — but the assumption is implicit. The adjacent conv_op_builder (~line 1257) and matmulnbits ProcessAttributesAndOutputs (~line 688) both use the defensive `IsQnnTensorWrapperExist` guard; this site lacks equivalent protection. A future refactor that reaches this code before section 1 completes would cause a null-deref abort.

**Impact:** Latent null-deref abort; missing defensive guard is inconsistent with the established IsQnnTensorWrapperExist usage pattern throughout the file.

**Suggested fix:**

Add an existence check and return an observable error instead of aborting: `RETURN_IF_NOT(qnn_model_wrapper.IsQnnTensorWrapperExist(input_names[0]), "QNN EP: MatMulNBits activation tensor not found in wrapper map");` before the GetQnnTensorWrapper call.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -150,6 +159,48 @@ Ort::Status AddFp16ToInt16QuantizeOutput(QnnModelWrapper& qnn_model_wrapper,
...
+  }
+
+  // Weight bitwidth.
+  if (kHtpNativeBQBits.find(bitwidth) == kHtpNativeBQBits.end()) {
+    return false;
+  }
+
+  // Symmetric quantization.
+  for (float offset : offsets) {
+    if (offset != 0.0f) {
```
huaychou: **[N-1] (Minor)** IsHTPSupportedNativeBQ offset symmetry check uses != 0.0f, which may false-reject symmetric ZP with tiny FP rounding artifact

The symmetry check in `IsHTPSupportedNativeBQ`:

```
for (float offset : offsets) { if (offset != 0.0f) { return false; } }
```

The offsets come from `per_block_float_zp` computed as `-(zp - offset_shift)` from integer ZP values. When the original integer ZP is 0, the computed offset should be exactly 0.0f; asymmetric cases are already gated by `!zp_is_symmetric` before this point. However, if FP arithmetic introduces a tiny non-zero value (e.g. -0.0f or a denormal), the strict `!= 0.0f` check will false-negative reject native BQ, causing a model that qualifies to fall back unnecessarily to BW_FLOAT_BLOCK (performance regression; output correctness unaffected).

**Impact:** Minor performance regression in edge cases; output correctness is not affected.

**Suggested fix:**

Replace `!= 0.0f` with `std::abs(offset) > std::numeric_limits<float>::epsilon()`; or determine symmetry from the raw integer ZP values at the call site (before float conversion) and pass a pre-verified zero vector to IsHTPSupportedNativeBQ.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.h
huaychou: **[T-1] (Nit)** IsHTPSupportedNativeBQ doc-comment does not mention the minimum HTP arch requirement

Tests use `SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V68)`, implying native BQ requires HTP arch V69+, but the doc-comment in the header does not state this.

**Suggested fix:**

Add to the doc-comment: `// Note: HTP native BQ requires HTP arch >= V69.`

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -640,12 +680,17 @@ Ort::Status MatMulNBitsOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& q
     assert(output_info.shape.size() == 3);
```
huaychou: **[T-2] (Nit)** matmulnbits_op_builder.cc still contains pre-existing C-style asserts that should be replaced with RETURN_IF_NOT

The file has `#include <cassert>` and uses `assert(...)` at lines 300, 322, and 680 (all pre-existing). Per QNN EP coding convention, `assert` must not appear in op builders — it is silently compiled out in Release builds. This PR did not introduce new asserts but also did not clean up the existing ones. Cleanup is recommended.

**Suggested fix:**

Replace asserts with RETURN_IF_NOT and remove `#include <cassert>`.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
```diff
@@ -1212,12 +1247,17 @@ Ort::Status ConvOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
...
-  // Detect BQ Conv from the weight tensor's quant encoding.
-  // BQ  (BW_FLOAT_BLOCK):       Conv outputs FP16 → need FP16 intermediate + Convert(FP16→INT16).
+  // Detect the non-native BQ Conv from the weight tensor's quant encoding.
+  // native BQ (BLOCK)             : Conv outputs INT16 → standard quantized output path.
+  // non-native BQ (BW_FLOAT_BLOCK): Conv outputs FP16 → need FP16 intermediate + Quantize(FP16→INT16).
+  // LPBQ (BLOCKWISE_EXPANSION)    : Conv outputs INT16 → standard quantized output path.
+  // NOTE: IsBlockQuantized() is true for both BLOCK and BW_FLOAT_BLOCK, so match the encoding directly.
   // input_names[1] is the weight — IsOpSupported guarantees Conv has >= 2 inputs.
-  bool is_bq_conv = false;
+  bool is_non_native_bq_conv = false;
```
huaychou: Should we rename the variable to `is_float_bq_conv` or `is_bw_float_block`

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -150,6 +159,48 @@ Ort::Status AddFp16ToInt16QuantizeOutput(QnnModelWrapper& qnn_model_wrapper,
                                            std::move(output_shape), do_op_validation);
 }
 
+bool IsHTPSupportedNativeBQ(Qnn_DataType_t act_data_type,
+                            uint32_t bitwidth,
+                            uint32_t block_size,
+                            uint32_t output_channel,
+                            gsl::span<const float> offsets) {
+  // HTP native BQ constraints (subject to change):
+  // - activation data type: QNN_DATATYPE_UFIXED_POINT_16
```
huaychou, line 173: Just wondering, should we support QNN_DATATYPE_SFIXED_POINT_16 with symmetric quantization by adding a ConvertOp?

huaychou (commented):
I think we didn't register any MatMulNBits QDQ node group selector. Please update the description. In addition, do you think using QNN API version to filter unsupported unit tests is a good idea to avoid adding DISABLE unit tests?

huaychou:
I think this PR cannot be merged until QAIRT is upleveled to 2.50, otherwise the model with BQ can be broken in the nightly build.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Add explicit check.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.h
minfhong-qti: Added document.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
minfhong-qti: Replaced with `RETURN_IF_NOT`.

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
minfhong-qti: Renamed.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
minfhong-qti: From what I heard from HTP team, the reason only supporting UFIXED_POINT_16 is due to this is the only case currently. Therefore, I guess if there is really a use case requiring SFIXED_POINT_16, both HTP and QNN EP should have corresponding changes.

minfhong-qti:
> I think this PR cannot be merged until QAIRT is upleveled to 2.50, otherwise the model with BQ can be broken in the nightly build.

I added macro to guard only adopting native BQ path if QAIRT >= 2.51, and so do the testcases.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
huaychou: Acknowledge

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
huaychou: Nit: Could we also remove the assert in conv_op_builder.cc?

huaychou: ditto

huaychou: ditto

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -3372,6 +3376,45 @@ ProviderOptions GetLPBQConvProviderOptions() {
...
+  auto cleanup = gsl::finally([&json_qnn_graph_dir]() { std::filesystem::remove_all(json_qnn_graph_dir); });
+  provider_options["dump_json_qnn_graph"] = "1";
+  provider_options["json_qnn_graph_dir"] = json_qnn_graph_dir.string();
+
+  RunQnnModelTest(build_fn,
+                  provider_options,
+                  /*opset=*/21,
+                  EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(fp32_abs_err)});
+
+#if !(QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40))
```
huaychou: One alternative solution is to define symbole in `qnn_test_utils.h` like https://github.com/onnxruntime/onnxruntime-qnn/pull/718/changes#diff-c1248ecda018d877075eca8a053e8e5bd90b45ba8687a7736b186b98218825c2

huaychou (approved):
LGTM

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -150,6 +160,52 @@ Ort::Status AddFp16ToInt16QuantizeOutput(QnnModelWrapper& qnn_model_wrapper,
...
+  // - weight bitwidth: 4 bit
+  // - symmetric quantization: offset=0
+  // - block size: 32, 64, 128
+  // - output channel: multiplier of 32
+
+  // Activation data type.
```
huaychou, line 184:
```suggestion
  // Activation data type:
```

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -150,6 +160,52 @@ Ort::Status AddFp16ToInt16QuantizeOutput(QnnModelWrapper& qnn_model_wrapper,
...
+  if (act_data_type != QNN_DATATYPE_UFIXED_POINT_16) {
+    return false;
+  }
+
+  // Weight bitwidth.
```
huaychou, line 189:
```suggestion
  // Weight bitwidth:
```

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
```diff
@@ -150,6 +160,52 @@ Ort::Status AddFp16ToInt16QuantizeOutput(QnnModelWrapper& qnn_model_wrapper,
...
+  if (kHtpNativeBQBits.find(bitwidth) == kHtpNativeBQBits.end()) {
+    return false;
+  }
+
+  // Symmetric quantization.
```
huaychou, line 194:
```suggestion
  // Symmetric quantization:
```

onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc
minfhong-qti: Removed.

minfhong-qti: Removed.

minfhong-qti: Removed.

onnxruntime/test/providers/qnn/conv_test.cc
minfhong-qti: Revised to define in `qnn_test_utils.h` and reused.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
minfhong-qti: Since I prefer `.` instead of `:`, align in all places.

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
minfhong-qti: Ditto

onnxruntime/core/providers/qnn/builder/qnn_bq_utils.cc
minfhong-qti: Ditto

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -25,6 +25,7 @@
 
 // QNN SDK headers for platform attribute queries.
 #include "HTP/QnnHtpDevice.h"
+#include "QnnCommon.h"
```
huaychou: nit: can be removed.

huaychou (approved):
LGTM. Have the native BQ unit tests been run on the V81 device?

onnxruntime/test/providers/qnn/qnn_test_utils.h
minfhong-qti: Removed.

minfhong-qti:
> LGTM. Have the native BQ unit tests been run on the V81 device?

Yes, I've locally run on Glymur and make sure adopting native BQ.

qti-chuteng (approved):

PR #605, opened 2026-07-06
Fix the int4 weights issue in GatherBlockQuantize

sachjang-qti:
### Description
Adding fix for gather block quantize int4 weights

### Motivation and Context
Fixed the error for not picking qnn gpu for gather block quantize when weights are in int4 format


onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
```diff
@@ -184,6 +311,35 @@ TEST_F(QnnGPUBackendTests, GatherBlockQuantized_LargerIndices) {
...
+  params.hidden_size = 64;
+  params.batch_size = 1;
+  params.seq_len = 1;
+  params.block_size = 32;
+  params.gather_axis = 0;
+  params.quantize_axis = 1;
+  RunGatherBlockQuantizedUnpackedInt4Test(params);
+}
+
+// Same shapes as the reported failure: hidden_size=3072, block_size=32 (96 blocks/row),
```
qti-mattsinc, line 250: nit. Mention the previously-failing use-case instead of calling it "the reported failure" (which isn't specific)

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
```diff
@@ -41,6 +41,28 @@ void QuantizeBlockwiseOnly(
...
+    int32_t cols,
+    int32_t block_size) {
+  TestInputDef<float> weight_def({rows, cols}, true, raw_vals);
+  std::vector<Int4x2> zero_points;  // symmetric quant: zero-points resolve to 0
+  GetTestInputQuantParamsBlockQuant<Int4x2>(weight_def, scales, zero_points, block_size,
+                                            /*axis=*/1, /*symmetric=*/true);
+  QuantizeValuesBlockQuant<float, Int4x2>(raw_vals, quant_vals, {rows, cols}, scales, zero_points,
+                                          block_size, /*axis=*/1);
+}
+
```
qti-mattsinc, line 65: This only supports axis==1, but `RunGatherBlockQuantizedUnpackedInt4Test` accepts any axis value and doesn't pass axis to this function. So, should `RunGatherBlockQuantizedUnpackedInt4Test` be asserting that axis==1, or alternatively should `QuantizeBlockwiseInt4Unpacked` accept an axis param?

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
sachjang-qti: Done

onnxruntime/test/providers/qnn/gatherblockquantized_test.cc
sachjang-qti: Looking at how QuantizeBlockwiseOnly handles this, it actually accepts a quantize_axis parameter and forwards it as a boolean columnwise flag to QuantizeBlockwise, so QuantizeBlockwiseInt4Unpacked should follow the same pattern by accepting an axis parameter and passing it through, since GetTestInputQuantParamsBlockQuant and QuantizeValuesBlockQuant already support axis arguments. This way RunGatherBlockQuantizedUnpackedInt4Test can pass params.quantize_axis directly instead of hardcoding. Also the quantize_axis == 1 check already present in gatherblockquantized_op_builder.cc file.

qti-mattsinc (approved):

vjatoth-qti:
Requesting code-owners to review/approve this change.

qti-chuteng:
### 🟩 Simplicity

**[D-3]** `🔵 Improvement` `[Simplicity]` — `gatherblockquantized_test.cc:181`
`RunGatherBlockQuantizedUnpackedInt4Test` is ~85% byte-identical to the existing `RunGatherBlockQuantizedTest` (indices loop, scales_def, all four attributes, AddNode, RunQnnModelTest are line-for-line copies). The real divergence is ~15 lines (Int4x2 values + `MakeInitializer<Int4x2>` instead of `MakeTestInput<uint8_t>`). The compile-time constraint the author documents (Int4x2 has no `Ort::TypeToTensorType`, so `MakeTestInput<Int4x2>` won't compile) is real but only blocks the weight-input line, which already sits in an if/else-able block. Also, the helper `QuantizeBlockwiseInt4Unpacked` is a thin wrapper over two library calls that `matmul_test.cc:254-266` invokes inline; it has one caller.
*Better design:* delete the fork and add a `bool weight_is_unpacked_int4` (default false) to the existing runner, branching only the weight-construction block (~90 lines removed). Keep both test cases, including `_LargeHidden` (the shape that reproduces 6144!=3072).

sachjang-qti:
> ### 🟩 Simplicity
> **[D-3]** `🔵 Improvement` `[Simplicity]` — `gatherblockquantized_test.cc:181` `RunGatherBlockQuantizedUnpackedInt4Test` is ~85% byte-identical to the existing `RunGatherBlockQuantizedTest` (indices loop, scales_def, all four attributes, AddNode, RunQnnModelTest are line-for-line copies). The real divergence is ~15 lines (Int4x2 values + `MakeInitializer<Int4x2>` instead of `MakeTestInput<uint8_t>`). The compile-time constraint the author documents (Int4x2 has no `Ort::TypeToTensorType`, so `MakeTestInput<Int4x2>` won't compile) is real but only blocks the weight-input line, which already sits in an if/else-able block. Also, the helper `QuantizeBlockwiseInt4Unpacked` is a thin wrapper over two library calls that `matmul_test.cc:254-266` invokes inline; it has one caller. _Better design:_ delete the fork and add a `bool weight_is_unpacked_int4` (default false) to the existing runner, branching only the weight-construction block (~90 lines removed). Keep both test cases, including `_LargeHidden` (the shape that reproduces 6144!=3072).

It is addressed and resolved in new patch

tirupath-qti (approved):

PR #606, opened 2026-07-06
[QNN EP] Add graph switching support for large multi-graph contexts

qti-ashwshan:
### Description
- Add context_memory_limit_hint_mb provider option to enable QNN graph switching for contexts too large to fully reside in the PD aperture
- Set MEMORY_LIMIT_HINT + PERSISTENT_BINARY configs and hold context binaries alive for the context lifetime so QNN can reload graphs on demand; reclaim them after ReleaseContext
- Bypass file-mapping when graph switching is enabled (persistent binary needs a contiguous in-memory buffer)
- Disable graph switching when share-resource optimization is set (incompatible) and validate the option value


qti-kromero:
@qti-ashwshan needs mainline merged for CI to run

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1773,15 +1773,61 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
                     ORT_LOGGING_LEVEL_VERBOSE,
                     ("Max spill fill buffer size: " + std::to_string(max_spill_fill_size)).c_str());
 
-    const QnnContext_Config_t* context_configs[] = {&qnn_context_config, spill_fill_config_pointer, nullptr};
+    QnnContext_Config_t memory_limit_config = QNN_CONTEXT_CONFIG_INIT;
+    QnnContext_Config_t persistent_binary_config = QNN_CONTEXT_CONFIG_INIT;
+    bool enable_memory_limit = context_memory_limit_hint_mb_ > 0;
```
qti-kromero, line 1842: nit: could just use context_memory_limit_hint_mb_ in the if statements/if-else directly

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1773,15 +1773,61 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
     RETURN_IF(nullptr == qnn_interface_.contextCreateFromBinary,
               "Invalid function pointer for contextCreateFromBinary.");
 
     Qnn_ErrorHandle_t rt = QNN_SUCCESS;
+
+    // Graph switching is incompatible with file-mapped weights (CFBWithCallback).
+    // When enabled, skip the callback path entirely and use a persistent buffer
+    // that QNN can reload graphs from during execution. QNN requires this buffer
+    // to stay alive for the whole context lifetime (see persistent_context_buffers_).
+    bool skip_file_mapping = enable_memory_limit;
```
qti-kromero: do we anticipate this if-statement growing in size with more conditions? if not, it bothers me slightly that it is set to the same as enable_memory_limit

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -130,6 +130,7 @@ struct QnnBackendManagerConfig {
   // remains constant for the manager's lifetime.
   bool enable_framework_op_trace = false;
   bool skip_backend_op_validation = false;
+  uint64_t context_memory_limit_hint_mb = 0;
```
qti-kromero, line 140: nit: I feel like we could get away with this being uint32_t

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -661,6 +661,27 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
   }
 
+  // Context memory limit hint (MB) — triggers graph-switching for large multi-graph contexts
+  std::string context_memory_limit_hint_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                 FormatEPConfigKey("context_memory_limit_hint_mb"), "0",
+                                 context_memory_limit_hint_str);
+  if (!context_memory_limit_hint_str.empty() && context_memory_limit_hint_str != "0") {
+    try {
+      context_memory_limit_hint_mb_ = std::stoull(context_memory_limit_hint_str);
```
qti-kromero: idk how much of a real concern this is but Claude claims the following:

td::stoull without a pos argument silently accepts
  partial parses: "512xyz" produces 512 rather than falling into the catch block. Pass a
  pos argument and verify pos == str.size() to reject malformed input

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1773,15 +1773,61 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
+      if (!context_bin_filepath.empty()) {
+        ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE,
+                        "Graph switching enabled — bypassing file mapping, using persistent buffer.");
+        persistent_context_buffers_.emplace_back();
+        RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, persistent_context_buffers_.back()));
+        bin_buffer = static_cast<void*>(persistent_context_buffers_.back().data());
+        buffer_length = persistent_context_buffers_.back().size();
+      } else {
+        // No file to read from — the caller-provided bin_buffer is used as the
+        // persistent binary. It MUST outlive this context or graph reloads are UB.
```
qti-kromero, line 1884: this seems risky but I can't really think of a better way to handle this with the APIs available

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -661,6 +661,27 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     }
   }
 
+  // Context memory limit hint (MB) — triggers graph-switching for large multi-graph contexts
+  std::string context_memory_limit_hint_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
```
qti-kromero, line 859: this new option needs some accompanying documentation

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -156,6 +157,7 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
         profiling_file_path_(config.profiling_file_path),
         enable_framework_op_trace_(config.enable_framework_op_trace),
         context_priority_(config.context_priority),
+        context_memory_limit_hint_mb_(config.context_memory_limit_hint_mb),
```
qti-kromero, line 169: if we still have an opportunity to give feedback to HTP team on this API I think we should - e.g. why do we need to provide this hint? why can they not figure out eviction/enablement strategy? etc.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-ashwshan: fixed. added pos check

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -854,6 +854,43 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     ParseQnnContextPriority(context_priority_str, context_priority, logger_);
   }
 
+  // VTCM MB
```
minfhong-qti: I guess this VTCM block is incorrectly kept during rebase.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1836,15 +1836,61 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
     RETURN_IF(nullptr == qnn_interface_.contextCreateFromBinary,
               "Invalid function pointer for contextCreateFromBinary.");
 
     Qnn_ErrorHandle_t rt = QNN_SUCCESS;
+
+    // Graph switching is incompatible with file-mapped weights (CFBWithCallback).
+    // When enabled, skip the callback path entirely and use a persistent buffer
+    // that QNN can reload graphs from during execution. QNN requires this buffer
+    // to stay alive for the whole context lifetime (see persistent_context_buffers_).
+    bool skip_file_mapping = enable_memory_limit;
```
minfhong-qti: How about setting `enable_file_mapped_weights_` in `QnnEp` to false and thus no need to handle in here?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2422,6 +2472,11 @@ void QnnBackendManager::ReleaseResources() {
     ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_ERROR, ("Failed to ReleaseContext: " + result.GetErrorMessage()).c_str());
   }
 
+  // Safe to free persistent binaries only after the contexts referencing them are
+  // freed (QnnContext_free). ReleaseContext above did that; reclaim the memory now.
+  persistent_context_buffers_.clear();
```
minfhong-qti: Should these be moved inside `ReleaseContext`?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1358,10 +1357,11 @@ Ort::Status QnnBackendManager::CreateContextFromListAsync(const QnnContext_Confi
     RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, buffer, io_dispatch));
 
     size_t buffer_size = buffer.size();
-    buffer_list.push_back(std::move(buffer));
+    // Keep buffer alive for graph-switching (persistent binary) support.
+    persistent_context_buffers_.push_back(std::move(buffer));
```
minfhong-qti: Why modify this function? I think this function is only called when `htp_share_resource_optimization_=1`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-ashwshan: We could do that, but I think the feature owner had kept it default to True for a reason. We can check on that and make the respective change.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -881,6 +907,17 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     htp_share_resource_optimization_ = 1;
   }
 
+  // Graph switching cannot be combined with SHARE_RESOURCES: that path reuses a
+  // shared context handle and never applies the memory-limit config, so graph
+  // switching would silently no-op. Disable it and warn.
+  if (context_memory_limit_hint_mb_ > 0 && htp_share_resource_optimization_ == 1) {
```
huaychou, line 913: **[C-1] (Critical)** share_ep_contexts_ incompatibility not enforced — documented as incompatible but code does not guard it

Problem Code

```
// Current guard (line 913)
if (context_memory_limit_hint_mb_ > 0 && htp_share_resource_optimization_ == 1) {
    context_memory_limit_hint_mb_ = 0;
    // guards htp_share_resource_optimization, but misses share_ep_contexts_
}

// Documentation (QNN-ExecutionProvider.md)
// "Incompatible with shared EP contexts (share_ep_contexts). Only effective on HTP backend."
```

Root Cause

The documentation explicitly states this option is incompatible with `share_ep_contexts`, but the code only guards `htp_share_resource_optimization_ == 1` and does not guard `share_ep_contexts_ == true`. When a user sets only `share_ep_contexts=1` (without `htp_share_resource_optimization`), graph switching will still be activated and passed into `LoadCachedQnnContextFromBuffer`.

In embed mode (`is_embed_mode=true`), `onnx_ctx_model_helper.cc:172` passes the `.c_str()` of a local `const std::string& context_binary` as `bin_buffer` to `LoadCachedQnnContextFromBuffer`. When `enable_memory_limit` is true and `context_bin_filepath.empty()`, the code enters the "caller-provided buffer" warning branch, relying on the caller's buffer remaining valid. However, `context_binary` is a local string that becomes invalid when the caller's `if` block exits, causing QNN to read freed memory during graph reloads (UB).

**Impact:** When `share_ep_contexts=1` and `context_memory_limit_hint_mb > 0` are both set, behavior is undefined (use-after-free). QNN will read a dangling pointer when attempting to reload graphs, potentially causing crashes or silent incorrect results. The documentation/code mismatch also gives users false confidence that the combination is guarded.

**Suggested fix:**

Add a `share_ep_contexts_` guard to the compatibility check in `qnn_execution_provider.cc` and update the log message:

```
if (context_memory_limit_hint_mb_ > 0 &&
    (htp_share_resource_optimization_ == 1 || share_ep_contexts_)) {
  ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
              "context_memory_limit_hint_mb (graph switching) cannot be combined with "
              "htp_share_resource_optimization, enable_vtcm_backup_buffer_sharing, or "
              "share_ep_contexts. Graph switching disabled.");
  context_memory_limit_hint_mb_ = 0;
}
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -854,6 +854,32 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     ParseQnnContextPriority(context_priority_str, context_priority, logger_);
   }
 
+  // Context memory limit hint (MB) — triggers graph-switching for large multi-graph contexts
```
huaychou, line 857: **[M-1] (Major)** New session config key context_memory_limit_hint_mb lacks a parsing test

The PR introduces session config key `context_memory_limit_hint_mb` with string parsing (`std::stoull`), trailing-character validation, overflow catching, and two compatibility guards (against `htp_share_resource_optimization_` and `enable_file_mapped_weights_`), but includes no tests.

Every new session config key in this codebase ships with at least one parsing test (see the `QnnEpSessionOptionTest` family) to verify: (1) valid values are correctly parsed to `uint64_t`, (2) invalid values (non-numeric, trailing characters) correctly fall back to 0, (3) combining with `htp_share_resource_optimization=1` disables graph switching.

**Impact:** If the key string or parsing logic is ever refactored, tests won't catch typos or behavioral regressions. The compatibility guards (two if-guards) especially need tests since they perform silent-disable behavior and are easily removed accidentally without test coverage.

**Suggested fix:**

Add tests in the session option test file under `onnxruntime/test/providers/qnn/`, covering at minimum:

```
// Test 1: valid positive value
SessionOptions opts;
opts.AddConfigEntry("ep.qnn.context_memory_limit_hint_mb", "512");
// Verify QnnEp sets context_memory_limit_hint_mb_ = 512

// Test 2: invalid string → fallback to 0
opts.AddConfigEntry("ep.qnn.context_memory_limit_hint_mb", "abc");
// Verify context_memory_limit_hint_mb_ = 0

// Test 3: incompatibility with htp_share_resource_optimization
opts.AddConfigEntry("ep.qnn.context_memory_limit_hint_mb", "512");
opts.AddConfigEntry("ep.qnn.htp_share_resource_optimization", "1");
// Verify context_memory_limit_hint_mb_ = 0 (disabled)
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1836,12 +1837,57 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
+      ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE,
+                      ("Context memory limit hint: " + std::to_string(context_memory_limit_hint_mb_) + " MB").c_str());
+    }
+    QnnContext_Config_t* memory_limit_config_pointer =
+        enable_memory_limit ? &memory_limit_config : nullptr;
+    QnnContext_Config_t* persistent_binary_config_pointer =
+        enable_memory_limit ? &persistent_binary_config : nullptr;
+
+    // Build config array without nullptr gaps — QNN treats nullptr as terminator
+    std::vector<const QnnContext_Config_t*> context_configs_vec;
```
huaychou, line 1857: **[M-2] (Major)** context_configs_vec and persistent_context_buffers_ use std::vector instead of InlinedVector [B3]

Per the mandatory QNN EP convention [B3], small containers under `onnxruntime/core/providers/qnn/` must use `onnxruntime::InlinedVector<T, N>` instead of `std::vector`.

This PR has two violations:

- `qnn_backend_manager.cc` line 1857: `std::vector<const QnnContext_Config_t*> context_configs_vec`. This vector holds at most 4 elements (`qnn_context_config`, `spill_fill_config`, `memory_limit_config`, `persistent_binary_config`). Should be `InlinedVector<const QnnContext_Config_t*, 4>`.
- `qnn_backend_manager.h` line 807: member `std::vector<std::vector<char>> persistent_context_buffers_`. In normal usage, one context binary is loaded per session, so the count is 1. Should be `InlinedVector<std::vector<char>, 1>`.
**Impact:** Violates mandatory convention [B3]; each `std::vector` construction incurs a heap allocation on the context-loading hot path.

**Suggested fix:**

```
// qnn_backend_manager.cc — local config pointer array
#include "core/common/inlined_containers.h"
InlinedVector<const QnnContext_Config_t*, 4> context_configs_vec;

// qnn_backend_manager.h — member declaration
InlinedVector<std::vector<char>, 1> persistent_context_buffers_;
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -854,6 +854,32 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+  std::string context_memory_limit_hint_str;
```
huaychou, line 858: **[M-3] (Major)** Session config key uses raw string literal instead of constexpr const char* constant [coding-conv]

The PR uses the raw string literal `"context_memory_limit_hint_mb"` directly in the `GetSessionConfigEntryOrDefault` call, instead of defining a `constexpr const char*` constant first. Per the Session Config Key String Literals convention, all session config keys must be named constants so the compiler can catch typos at the definition site.

Additionally, the verbose log at line 877 prints the raw input string `context_memory_limit_hint_str` rather than the effective parsed value `context_memory_limit_hint_mb_`. If parsing fails and falls back to 0, the log still shows the original string, making it impossible to determine whether graph switching was actually activated.

**Impact:** Typos in key strings or log statements cannot be caught by the compiler, leading to silent feature misfires. The misleading log makes debugging difficult.

**Suggested fix:**

```
// At top of qnn_execution_provider.cc or in qnn_provider_factory.h
constexpr const char* kContextMemoryLimitHintMb = "context_memory_limit_hint_mb";

// Use the constant in parsing
GetSessionConfigEntryOrDefault(ort_api, session_options_,
    FormatEPConfigKey(kContextMemoryLimitHintMb), "0",
    context_memory_limit_hint_str);

// Log the effective value
    ("context_memory_limit_hint_mb: " +
     std::to_string(context_memory_limit_hint_mb_) +
     " MB (graph switching: " +
     (context_memory_limit_hint_mb_ > 0 ? "enabled" : "disabled") + ")").c_str());
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1836,12 +1837,57 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
+      if (!context_bin_filepath.empty()) {
+        ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE,
+                        "Graph switching enabled — bypassing file mapping, using persistent buffer.");
+        persistent_context_buffers_.emplace_back();
+        RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, persistent_context_buffers_.back(), io_dispatch));
+        bin_buffer = static_cast<void*>(persistent_context_buffers_.back().data());
+        buffer_length = persistent_context_buffers_.back().size();
+      } else {
+        // No file to read from — the caller-provided bin_buffer is used as the
+        // persistent binary. It MUST outlive this context or graph reloads are UB.
```
huaychou, line 1884: **[N-1] (Minor)** "Caller-provided buffer" path lacks API contract documentation

When `enable_memory_limit` is true and `context_bin_filepath` is empty, the code enters the `else` branch with a warning: "relying on the caller-provided buffer remaining valid for the context lifetime." This else branch represents a lifetime contract (the buffer must outlive the context), but this contract is only described inside the implementation — callers (e.g., `onnx_ctx_model_helper.cc:172`) have no visibility into this requirement.

**Impact:** Future maintainers or callers may not know about this lifetime requirement and pass short-lived buffers unwittingly (exactly the embed-mode scenario in C-1).

**Suggested fix:**

Add contract documentation to `LoadCachedQnnContextFromBuffer`'s declaration in `qnn_backend_manager.h` and strengthen the else-branch comment:

```
// In qnn_backend_manager.h declaration:
// REQUIRES (when context_memory_limit_hint_mb_ > 0 and context_bin_filepath is empty):
//   bin_buffer must remain valid for the lifetime of the created context.
//   Violating this requirement causes use-after-free when QNN reloads graphs.

// In qnn_backend_manager.cc else branch:
// LIFETIME CONTRACT: bin_buffer must outlive this context.
// Graph switching (QNN_CONTEXT_CONFIG_PERSISTENT_BINARY) causes QNN to
// re-read this pointer during execution; if the buffer is freed, the result is UB.
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou, line 858: **[N-2] (Minor)** HTP-only option lacks htp_ prefix, reducing discoverability [naming-conv]

Per the `enable_htp_*` naming convention, HTP-only session config options should carry an `htp_` prefix so users browsing the API can quickly identify their scope. The current key name `context_memory_limit_hint_mb` provides no hint that it is HTP-only.

Note: Since this is an MB numeric value (not a boolean toggle), the `enable_htp_` prefix is not applicable; the recommended alternative is `htp_context_memory_limit_hint_mb`, or keep the current name but more explicitly label it as "HTP-only" in docs and log messages.

**Impact:** Users setting this option on non-HTP backends (e.g., CPU/GPU) will get silent no-ops with no warning.

**Suggested fix:**

Choose one of:
(A) Rename the key to `htp_context_memory_limit_hint_mb` (update docs and logs accordingly).
(B) Keep the current name but add a backend-type check on enable: `if (backend_type_ != HTP) { /* warn and disable */ }`, and update the doc to more clearly state "HTP backend only — no effect on CPU/GPU backends."

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1836,12 +1837,57 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
+    }
+    QnnContext_Config_t* memory_limit_config_pointer =
+        enable_memory_limit ? &memory_limit_config : nullptr;
+    QnnContext_Config_t* persistent_binary_config_pointer =
+        enable_memory_limit ? &persistent_binary_config : nullptr;
+
+    // Build config array without nullptr gaps — QNN treats nullptr as terminator
+    std::vector<const QnnContext_Config_t*> context_configs_vec;
+    context_configs_vec.push_back(&qnn_context_config);
+    if (spill_fill_config_pointer) context_configs_vec.push_back(spill_fill_config_pointer);
```
huaychou, line 1859: **[T-1] (Nit)** Conditional if statements missing braces (coding style)

Lines 1859–1861 have three single-statement `if` bodies without braces, inconsistent with the codebase convention of always using `{ }`:

```
if (spill_fill_config_pointer) context_configs_vec.push_back(spill_fill_config_pointer);
if (memory_limit_config_pointer) context_configs_vec.push_back(memory_limit_config_pointer);
if (persistent_binary_config_pointer) context_configs_vec.push_back(persistent_binary_config_pointer);
```

**Suggested fix:**

```
if (spill_fill_config_pointer) {
  context_configs_vec.push_back(spill_fill_config_pointer);
}
if (memory_limit_config_pointer) {
  context_configs_vec.push_back(memory_limit_config_pointer);
}
if (persistent_binary_config_pointer) {
  context_configs_vec.push_back(persistent_binary_config_pointer);
}
```

huaychou (commented):
We need to add the corresponding unit tests for this feature.

qti-ashwshan:
Thanks for the review @huaychou. Please let me know if everything looks good.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -912,6 +912,42 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
   }
 
+  // Context memory limit hint (MB) — triggers graph-switching for large multi-graph contexts
+  static constexpr const char* kContextMemoryLimitHintMb = "context_memory_limit_hint_mb";
+  std::string context_memory_limit_hint_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                 FormatEPConfigKey(kContextMemoryLimitHintMb), "0",
+                                 context_memory_limit_hint_str);
+  if (!context_memory_limit_hint_str.empty() && context_memory_limit_hint_str != "0") {
+    try {
```
minfhong-qti: Leverage `ParseIntegerOption`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -839,6 +846,11 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   std::optional<bool> dx12_shared_memory_allocator_supported_ = std::nullopt;
   Qnn_ProfileHandle_t profile_backend_handle_ = nullptr;
   ContextPriority context_priority_;
+  uint64_t context_memory_limit_hint_mb_ = 0;
+  // Keeps context binary buffers alive for graph-switching (persistent binary).
+  // QNN requires the binary buffer to remain valid for the lifetime of the context
+  // so it can reload graphs on demand.
+  onnxruntime::InlinedVector<std::vector<char>, 1> persistent_context_buffers_;
```
minfhong-qti: Prefer a separate section for these two members while commenting `context_memory_limit_hint_mb_ > 0` will trigger the use of `persistent_context_buffers_`.
In addition, not sure the benefit of `InlinedVector` comparing to normal one.

minfhong-qti (commented):
Please add at least one E2E test to verify the new logic in `QnnBackendManager`.

PR #607, opened 2026-07-07
[QNN EP] Add int32 Max/Min unit tests for CPU and HTP backends

huaychou:
### Description

- Add int32 coverage for the existing Max/Min simple-op builders
- Add Max_1Input_NotSupported_Int32 test

### Motivation and Context

- Support Int32 ElementWiseMaximum and Minimum Ops in QNN for HTP

### DIFF
--- a/onnxruntime/test/providers/qnn/max_min_test.cc
+++ b/onnxruntime/test/providers/qnn/max_min_test.cc
@@ -48,6 +48,25 @@ static void RunQDQMinOrMaxOpTest(const std::string& op_type,
                        expected_ep_assignment);
 }
 
+// Runs an int32 Max/Min model on the given QNN backend ("cpu" or "htp"). Checks the graph node
+// assignment, and that inference outputs for QNN EP and ORT CPU EP match exactly (int32 has no
+// quantization tolerance to account for).
+static void RunInt32MinOrMaxOpTest(const std::string& op_type,
+                                   const std::vector<TestInputDef<int32_t>>& input_defs,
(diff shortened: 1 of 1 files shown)


onnxruntime/test/providers/qnn/max_min_test.cc
```diff
@@ -84,6 +103,26 @@ TEST_F(QnnCPUBackendTests, Max_2Inputs) {
                        ExpectedEPNodeAssignment::All, 13);
 }
 
+// Test int32 Min with 2 inputs on CPU backend.
```
minfhong-qti, line 106: Just curious. Do we have policy whether we should add new CPU tests considering we stop advertising CPU?

huaychou: @qti-niscmami to comment

minfhong-qti (approved):
LGTM. Wait for the confirmation whether new CPU testcases are still needed.

qti-chuteng (approved):

onnxruntime/test/providers/qnn/max_min_test.cc
qti-niscmami: We currently don't have a policy in place to stop adding new CPU tests. 

For the time being we can continue adding them, ideally with equivalent HTP and GPU coverage. In this PR the positive int32 min/max cases are covered on both CPU and HTP. The 1-input negative cases are HTP only, which is fine.

PR #608, opened 2026-07-07
[QNN EP] Add support for Buffer, StatefulLstm, StatefulGru block ops

kuanyul-qti:
Add QNN EP builder support for three qti_aisw custom block ops — Buffer, StatefulLstm, and StatefulGru — mapping to QNN_OP_BUFFER, QNN_OP_LSTM, and QNN_OP_GRU respectively.

Builders:
- buffer_op_builder.cc: QNN_OP_BUFFER mapping; optional reset input at ONNX in[1]; rejects BLOCKING mode (mode=0); supports FP16/INT16/INT8; mirrors quant encoding from activation input to output.
- stateful_lstm_op_builder.cc: QNN_OP_LSTM with reset input wired at in[24]; bidirectional supported for all dtypes (HtpOpDefSupplement places no direction constraint on the Lstm op).
- stateful_gru_op_builder.cc: QNN_OP_GRU with reset input wired at in[14]; bidirectional supported for float, rejected for quantized (INT8/INT16), per HtpOpDefSupplement.

Infrastructure:
- Register qti_aisw custom op domain and widen the QDQ selector domain filter.
- Extract shared helpers into rnn_op_utils.h/.cc, removing duplication across the four RNN builders.
- Fix activations[5]->[4] copy-paste bug in the bidirectional LSTM activation check; use QnnTensorWrapper::MakeNull with node-scoped unique names.

Model loadability (placeholder custom ops):
- The qti_aisw ops have no registered ONNX schema, so ORT's Graph::Resolve rejected any model containing them before the QNN EP could claim the nodes. Register schema-only placeholder custom ops from the QNN EP factory via GetCustomOpDomains so models pass validation and are then fused/compiled by QNN.

Tests:
- Unit tests for all three ops (FP16/QDQ positive cases and IsOpSupported=false negative cases) on a QNN-only assignment-check path, with HTP-hardware paths gated off the x86 simulator.
- IR-backend composition tests for StatefulLstm/StatefulGru (run on all platforms; no NPU required).

Reference:
ONNX op definition for Buffer, StatefulGru, StatefulLstm: https://docs.qualcomm.com/doc/80-63442-10/topic/blockop_definitions.html


onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -94,7 +95,20 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
   GetNumCustomOpDomains = GetNumCustomOpDomainsImpl;
   GetCustomOpDomains = GetCustomOpDomainsImpl;
 
-  // Build custom-op domains from ORT_QNN_CUSTOM_OP_DOMAINS env var.
+  // Register schema-only placeholder ops for the qti_aisw block ops so models that use them pass
```
minfhong-qti: Help move this custom op block below `CreateExternalResourceImporterForDevice`.

onnxruntime/core/providers/qnn/qnn_provider_factory.h
```diff
@@ -83,11 +84,15 @@ class QnnEpFactory : public OrtEpFactory, public ApiPtrs {
...
-  // Both vectors must outlive any session that uses this factory (factory is a per-library singleton).
-  // domain.Add(op*) does NOT transfer ownership; op objects must be kept alive here.
+  // Custom op domains reported to ORT via GetCustomOpDomains. Holds both the qti_aisw block-op
+  // domain (always registered) and any domains built from ORT_QNN_CUSTOM_OP_DOMAINS.
+  // The domains and the op objects below must outlive any session that uses this factory
+  // (factory is a per-library singleton); domain.Add(op*) does NOT transfer ownership.
   std::vector<Ort::CustomOpDomain> custom_op_domains_;
+  // Placeholder ops built from ORT_QNN_CUSTOM_OP_DOMAINS (env-var UDO ops).
   std::vector<std::unique_ptr<qnn::QnnUdoPlaceholderOp>> custom_op_objects_;
+  // Placeholder ops for the qti_aisw block ops (Buffer, StatefulLstm, StatefulGru).
```
minfhong-qti:
```suggestion
  // Placeholder ops for the qti_aisw block ops.
```
No need to explicitly write op types as they eventually will extend.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1904,7 +1904,9 @@ std::vector<OrtNodeGroup> OrtSelectorManager::GetOrtQDQSelections(const OrtGraph
 
     // Check domain (similar to the GraphViewer version)
     std::string domain_str(domain);
-    if (domain_str != kOnnxDomain && domain_str != kMSInternalNHWCDomain && domain_str != kMSDomain && domain_str != kMLOnnxDomain) {
+    if (domain_str != kOnnxDomain && domain_str != kMSInternalNHWCDomain && domain_str != kMSDomain &&
```
minfhong-qti:
```suggestion
    if (domain_str != kOnnxDomain &&
        domain_str != kMSInternalNHWCDomain &&
        domain_str != kMSDomain &&
```
Nit.

onnxruntime/core/providers/qnn/common/qnn_graph_utils.h
```diff
@@ -22,6 +22,9 @@ inline constexpr const char* kOnnxDomain = "";
 inline constexpr const char* kMSDomain = "com.microsoft";
 inline constexpr const char* kMSInternalNHWCDomain = "com.ms.internal.nhwc";
 inline constexpr const char* kMLOnnxDomain = "ai.onnx.ml";
+// Custom Qualcomm block-op domain (qti_aisw). Ops such as Buffer, StatefulLstm and StatefulGru
+// live here; QNN EP recognizes them via their op-builder registrations.
+inline constexpr const char* kQtiAiswDomain = "qti_aisw";
```
minfhong-qti: Prefer define in `qnn_qti_aisw_custom_op.h`.

onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h
```diff
@@ -0,0 +1,99 @@
...
+// for QNN-assigned nodes).
+
+struct QtiAiswPlaceholderOp
+    : Ort::CustomOpBase<QtiAiswPlaceholderOp, QnnUdoPlaceholderKernel, /*WithStatus*/ true> {
+  explicit QtiAiswPlaceholderOp(const char* op_name) : name_(op_name) {}
+
+  const char* GetName() const { return name_; }
+  // nullptr = CPU provider. Rejected nodes fall back here so session creation doesn't fail;
+  // Compute errors explicitly if such a node is ever executed (see QnnUdoPlaceholderKernel).
+  const char* GetExecutionProviderType() const { return nullptr; }
```
minfhong-qti: When is this function called? What's the difference in returning QNN EP name like `qnn_custom_op.h`?

onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h
```diff
@@ -0,0 +1,99 @@
...
+
+  // OPTIONAL inputs to accept empty interior slots (e.g. sequence_lens, B, initial_h).
+  // Most are UNDEFINED (any type); reset slot is BOOL to prevent InferOutputTypes from
+  // propagating bool to outputs (see GetInputType for slot indices).
+  static constexpr size_t kMaxInputs = 24;  // generous upper bound; covers every block op's arity
+  static constexpr size_t kMaxOutputs = 3;  // Y, Y_h, Y_c
+
+  size_t GetInputTypeCount() const { return kMaxInputs; }
```
minfhong-qti, line 53: Are these really matters? `QnnUdoPlaceholderOp` seems simply returning default value.

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
```diff
@@ -0,0 +1,37 @@
...
+--- a/onnxruntime/core/session/utils.cc
++++ b/onnxruntime/core/session/utils.cc
+@@ -318,7 +318,10 @@ static OrtStatus* CreateSessionAndLoadSingleModelImpl(_In_ const OrtSessionOptio
+       ORT_API_RETURN_IF_STATUS_NOT_OK(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
+ 
+       for (auto domain : domains) {
+-        if (ShouldAddDomain(domain, options->custom_op_domains_)) {
++        // Multiple EP devices can be backed by the same factory and expose the same
++        // custom-op domain. Keep one copy per domain in the session registry.
++        if (ShouldAddDomain(domain, options->custom_op_domains_) &&
```
minfhong-qti, line 26: It seems `ShouldAddDomain` already checks for duplicate domain [ref](https://github.com/microsoft/onnxruntime/blob/8d85527a010e294a26b274749f74294b2a32cec5/onnxruntime/core/session/utils.cc#L154). Is this patch mandatory?

yuhuchua-qti:
> Register qti_aisw custom op domain and widen the QDQ selector domain filter to enable INT8/INT16 quantized node-unit formation.

My understanding is that adding kQtiAiswDomain to the QDQ selector domain allow-list only lets qti_aisw nodes reach selector lookup. Without an op-type selector registration for "StatefulGru", DQ -> qti_aisw::StatefulGru -> Q would not form a QDQGroup NodeUnit. Is that expected here, or should StatefulGru be registered with a QDQ selector as well?

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
kuanyul-qti: Yes. Not for a single-device AutoEP session. `ShouldAddDomain` only checks `options->custom_op_domains_`; it does not check domains already appended to the local `all_ep_custom_op_domains` collection. When multiple AutoEP devices backed by the same QNN factory expose `qti_aisw`, both copies are appended and later registered together, causing duplicate-domain registration. The patch aims to avoid this.

onnxruntime/core/providers/qnn/common/qnn_graph_utils.h
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h
kuanyul-qti: ORT calls `GetExecutionProviderType()` while registering the custom kernel. A non-null value sets the kernel provider; `nullptr` explicitly registers it for CPU. Since `qti_aisw` block ops are QNN-only, make it return the QNN EP name.

onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h
kuanyul-qti: Yes. `QnnUdoPlaceholderOp` has one heterogeneous `VARIADIC` formal input/output. ORT type inference permits an absent input only when its formal parameter is `OPTIONAL`; `VARIADIC` is not sufficient. Stateful RNN models intentionally preserve empty interior ONNX slots such as `B`, `sequence_lens`, and `initial_h`, so inheriting the UDO schema fails model loading/type inference. The qti placeholder therefore provides fixed `OPTIONAL` slots and declares the reset slots as `BOOL`.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_provider_factory.h
kuanyul-qti: Done.

kuanyul-qti:
> My understanding is that adding kQtiAiswDomain to the QDQ selector domain allow-list only lets qti_aisw nodes reach selector lookup. Without an op-type selector registration for "StatefulGru", DQ -> qti_aisw::StatefulGru -> Q would not form a QDQGroup NodeUnit. Is that expected here, or should StatefulGru be registered with a QDQ selector as well?

`StatefulGru` is now registered with `OrtGRUNodeGroupSelector`, so `DQ -> qti_aisw::StatefulGru -> Q` can form a `QDQGroup`. The generic GRU selector is appropriate because it performs structural QDQ validation; native-versus-FP fallback decisions remain in the builder.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_lstm_op_builder.cc
```diff
@@ -0,0 +1,213 @@
...
+                                  const OrtNodeUnit& node_unit,
+                                  const std::string& direction,
+                                  const std::vector<std::string>& input_names,
+                                  const Ort::Logger& logger,
+                                  const bool& do_op_validation,
+                                  const bool& is_bidirection,
+                                  std::vector<std::string>& uni_lstm_output_names) const;
+};
+
+Ort::Status StatefulLstmOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
```
yuhuchua-qti: I see the regular LSTM path fixes the bidirectional activation validation bug by checking `activations[4]` instead of checking `activations[5]` twice. However, StatefulLstm has its own `IsOpSupported()` and only shares the lowering helper, so that validation does not seem to apply to qti_aisw::StatefulLstm.

Should StatefulLstm apply the same default-activation validation, or explicitly reject/document the activations attribute? Otherwise QNN EP may claim a StatefulLstm node with non-default activations, but lower it to QNN LSTM behavior that does not represent those semantics.

yuhuchua-qti: The regular LSTM builder explicitly rejects `layout=1`, but StatefulLstm does not seem to perform the same validation. Since the shared lowering path still appears to derive seq_length/batch_size and emit output reshapes assuming ONNX `layout=0` tensor order, should StatefulLstm also reject` layout=1` until batch-major mapping is supported?

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_gru_op_builder.cc
```diff
@@ -0,0 +1,290 @@
...
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[0], input_info));
+    RETURN_IF(input_info.quant_param.IsQuantized() && node_unit.UnitType() != OrtNodeUnit::Type::QDQGroup,
+              "QNN EP: bidirectional StatefulGru is only supported for float (FP16/FP32); "
+              "quantized (INT8/INT16) StatefulGru is forward-only (per HtpOpDefSupplement).");
+  }
+
+  if (node_unit.Inputs().size() > kOnnxResetInputIndex && node_unit.Inputs()[kOnnxResetInputIndex].Exists()) {
+    TensorInfo reset_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[kOnnxResetInputIndex], reset_info));
+    RETURN_IF_NOT(reset_info.qnn_data_type == QNN_DATATYPE_BOOL_8,
```
yuhuchua-qti: StatefulGru validates that the reset input is `QNN_DATATYPE_BOOL_8`, but StatefulLstm does not seem to perform the same check before wiring `input[8]` to QNN LSTM `input[24]`. Should StatefulLstm validate the reset input consistently with StatefulGru?

Also, should both StatefulLstm and StatefulGru validate the reset shape explicitly, e.g. scalar/0D if that is the QNN contract? The placeholder schema is intentionally permissive, so catching an invalid reset input in `IsOpSupported()` would give a clearer error than failing later during QNN graph validation/finalization.

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
kuanyul-qti: As discussed offline, I’ll use this patch and submit a PR to ORT core. If the ORT core PR is rejected, I’ll deprecate this patch and add `qti_aisw` domain to `session_option` in the AutoEP tests to address the issue.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_gru_op_builder.cc
```diff
@@ -0,0 +1,290 @@
...
+    std::vector<int32_t> sequence_lens(num_elems);
+    if (num_elems > 0) {
+      std::memcpy(sequence_lens.data(), sequence_lens_bytes.data(), num_elems * sizeof(int32_t));
+    }
+    RETURN_IF(std::any_of(sequence_lens.begin(), sequence_lens.end(),
+                          [&sequence_lens](int32_t i) { return i != sequence_lens[0]; }),
+              "QNN EP: Only support StatefulGru with same sequence length.");
+  }
+
+  OrtNodeAttrHelper node_helper(node_unit);
```
huaychou: **[M-1] (Major)** StatefulGru / StatefulLstm IsOpSupported drops the activations and layout validation the standard builders enforce — non-default values are accepted and silently produce wrong results

The standard `GRUOpBuilder::IsOpSupported` rejects non-default `activations` (requires sigmoid/tanh) and `layout=1`; `LSTMOpBuilder::IsOpSupported` likewise rejects non-default activations and requires `layout==0` (gru_op_builder.cc L82-85, lstm_op_builder.cc L183-190).

`StatefulGruOpBuilder::IsOpSupported` (L75-122) and `StatefulLstmOpBuilder::IsOpSupported` (L68-107) contain neither check — they only validate sequence_lens, clip, direction/quant, reset dtype, and input_forget.

**Impact:** The shared `rnn_details::AddUnidirection{GRU,LSTM}` lowering hardcodes the gate structure to sigmoid/tanh and never reads the activations attribute. A StatefulGru/StatefulLstm with non-default activations (e.g. `['relu','relu']`) therefore passes IsOpSupported, gets compiled into a QNN GRU/LSTM that uses fixed sigmoid/tanh gates, and silently produces numerically wrong output. The same applies to layout=1: the standard builders explicitly reject it, but the Stateful variants accept it. This is a validation regression relative to the standard builders.

**Suggested fix:**

Add the same activations and layout guards used by the standard GRU/LSTM builders to StatefulGru/StatefulLstm IsOpSupported (ideally factor this validation into a shared rnn_details helper so all four builders share one source of truth).

onnxruntime/core/providers/qnn/builder/opbuilder/rnn_op_utils.cc
```diff
@@ -0,0 +1,1206 @@
...
+                              QnnQuantParamsWrapper(), std::vector<uint32_t>(h_shape),
+                              std::vector<uint8_t>(utils::GetElementSizeByType(input_tensor_infos[0].qnn_data_type) * batch_size * hidden_size, 0));
+      RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(zero_h)), "Failed to add initial hidden state.");
+    }
+  }
+
+  // QAIRT lowers StatefulGru with reset as one native multi-time-step QNN GRU.
+  // QNN owns the state across inferences, so reset must be connected once to
+  // in[14], not copied to each of ORT's per-timestep GRU cells.
+  if (!reset.onnx_name.empty()) {
```
huaychou: **[M-2] (Major)** GRU stateful-reset path lacks the backend guard LSTM has — on non-HTP/IR backends reset is not cleanly declined

`AddUnidirectionLSTM` only takes the multi-time-step reset path when `is_npu_backend || is_ir_backend`; if reset appears on any other backend it is explicitly rejected with `RETURN_IF_NOT(false, "...the selected backend does not support the 3D reset input.")` (L692-695, L1040-1048).

By contrast `AddUnidirectionGRU` (L390) gates only on `if (!reset.onnx_name.empty())` and unconditionally emits a native multi-time-step QNN GRU node on any backend, including CPU/GPU.

**Impact:** Running a StatefulGru with reset on a QNN CPU/GPU backend that has no native in[14]-reset GRU will fail during ComposeGraph / do_op_validation with a QNN validation error and abort session init, rather than being cleanly declined in IsOpSupported (which the QNN EP prefers over reaching QNN validation). The two ops are handled inconsistently and the GRU side is missing the explicit guard.

**Suggested fix:**

Add the same backend guard the LSTM path uses: take the native multi-time-step reset GRU only on HTP/IR backends and explicitly report unsupported elsewhere (or decline in StatefulGru IsOpSupported for non-HTP/IR backend + reset). If QNN CPU/GPU genuinely supports this reset GRU, add a comment documenting why GRU needs no guard.

onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h
```diff
@@ -0,0 +1,87 @@
...
+  // Reset slot is BOOL (single-type) so InferOutputTypes doesn't propagate bool to outputs.
+  //   StatefulGru: in[6], StatefulLstm: in[8], Buffer: in[1]
+  ONNXTensorElementDataType GetInputType(size_t index) const {
+    if ((op_name_ == "StatefulGru" && index == 6) ||
```
huaychou: **[N-1] (Minor)** Reset-slot indices are hardcoded in two places (placeholder op and each op builder) with no single source of truth

`QtiAiswPlaceholderOp::GetInputType` (L48-55) uses magic numbers to mark each op's reset slot as BOOL: StatefulGru in[6], StatefulLstm in[8], Buffer in[1]. Those same three indices are independently defined again in each op builder as `kOnnxResetInputIndex` (stateful_gru=6, stateful_lstm=8) or inline (buffer in[1]).

**Impact:** If the two sites drift (e.g. a future input-layout change to one op), the slot the placeholder op types as BOOL and the slot the builder actually wires reset into diverge silently, causing wrong type inference or reset landing on the wrong slot.

**Suggested fix:**

Centralize each block op's reset ONNX index into a single constant table (e.g. in qnn_qti_aisw_custom_op.h) shared by the placeholder op and the op builders.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_lstm_op_builder.cc
```diff
@@ -0,0 +1,213 @@
...
+// initial_h, initial_c, P) plus a trailing "reset" input appended at ONNX index 8. It maps to the same
+// QNN_OP_LSTM ("Lstm") op as the standard LSTM builder, with one difference: the reset input is wired
+// into QNN LSTM in[24] (which the standard LSTM builder leaves as null_tensor).
+//
+// The ONNX->QNN decomposition (per-direction StridedSlice gate slicing, ifoc gate reordering, bias
+// summation, zero-bias / zero-initial-state stubs, bidirectional Concat, and the QNN LSTM input
+// vector of size 25) is shared with the standard LSTM builder via rnn_details::AddUnidirectionLSTM
+// in rnn_op_utils.cc. This builder passes a reset ResetInput; the standard LSTM passes kNoReset().
+//
+// Bidirectional is supported for all dtypes (float and quantized) via a forward + reverse unroll
```
huaychou: **[T-1] (Nit)** StatefulLstm comment claims quantized bidirectional support, but LSTM has no QDQ selector so the quantized path really runs fp via explicit DQ/Q

The class comment (L27-29) says bidirectional works 'for both float and quantized inputs'. But qnn_ep_utils.cc registers a QDQ selector only for GRU/StatefulGru (L1809-1812); there is no LSTM/StatefulLstm selector, and the rnn_op_utils LSTM lowering has no use_fp_fallback. In practice the QDQ StatefulLstm tests (RunHtpQDQStatefulLSTMOpTest) run DQ->StatefulLstm->Q as standalone nodes, i.e. effectively fp execution.

**Impact:** Wording only — could mislead a future maintainer into thinking a native quantized LSTM path exists. Suggest clarifying the comment.

**Suggested fix:**

Reword the comment to state that quantized StatefulLstm models run fp via explicit DQ/Q (no native quantized LSTM selector), to avoid misleading readers.

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
```diff
@@ -0,0 +1,42 @@
...
++++ b/onnxruntime/core/session/utils.cc
+@@ -318,7 +318,10 @@ static OrtStatus* CreateSessionAndLoadSingleModelImpl(_In_ const OrtSessionOptio
+       ORT_API_RETURN_IF_STATUS_NOT_OK(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
+ 
+       for (auto domain : domains) {
+-        if (ShouldAddDomain(domain, options->custom_op_domains_)) {
++        // Multiple EP devices can expose the same custom-op domain. Keep one copy
++        // in this session's collection before its custom registry is created.
++        if (ShouldAddDomain(domain, options->custom_op_domains_) &&
++            !DoesDomainWithNameExist(domain->domain_, all_ep_custom_op_domains)) {
```
huaychou:
```suggestion
+            ShouldAddDomain(domain, all_ep_custom_op_domains)) {
```

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
```diff
@@ -0,0 +1,42 @@
...
+       }
+@@ -431,7 +434,10 @@ static OrtStatus* CreateSessionAndLoadModelImpl(_In_ const OrtSessionOptions* op
+       ORT_API_RETURN_IF_STATUS_NOT_OK(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
+ 
+       for (auto domain : domains) {
+-        if (ShouldAddDomain(domain, options->custom_op_domains_)) {
++        // Multiple EP devices can expose the same custom-op domain. Keep one copy
++        // in this session's collection before its custom registry is created.
++        if (ShouldAddDomain(domain, options->custom_op_domains_) &&
++            !DoesDomainWithNameExist(domain->domain_, all_ep_custom_op_domains)) {
```
huaychou:
```suggestion
+            !ShouldAddDomain(domain, all_ep_custom_op_domains)) {
```

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
```diff
@@ -17,13 +17,25 @@ namespace qnn {
 // ORT's model-load schema validation; it must never actually execute. Returning
 // an explicit error (rather than a silent no-op) prevents wrong results if a UDO
 // node is accidentally not fused.
+//
+// Used by both env-var UDO ops and the built-in qti_aisw block ops. The optional op name is
+// folded into the error message for diagnostics; the message always contains "fused" for the
+// QNN path.
 struct QnnUdoPlaceholderKernel {
+  QnnUdoPlaceholderKernel() = default;
+  explicit QnnUdoPlaceholderKernel(std::string op_name) : op_name_(std::move(op_name)) {}
```
huaychou: I think we pass the op_type when QnnUdoPlaceholderOp but op_type when QtiAiswPlaceholderOp. This is inconsistent.

huaychou (commented):
Could we have the Buffer, StatefulGru, StatefulLstm op definition for reference?

onnxruntime/core/providers/qnn/builder/opbuilder/rnn_op_utils.cc
kuanyul-qti: Fixed.

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
kuanyul-qti: Fixed.

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_lstm_op_builder.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_gru_op_builder.cc
```diff
@@ -0,0 +1,294 @@
...
+                  "QNN EP: StatefulGru reset input must have QNN BOOL_8 data type.");
+  }
+
+  return Ort::Status();
+}
+
+Ort::Status StatefulGruOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
+                                                const OrtNodeUnit& node_unit,
+                                                const Ort::Logger& logger,
+                                                std::vector<std::string>& input_names,
```
huaychou: **[M-1] (Major)** StatefulGru / StatefulLstm IsOpSupported missing activations and layout validation — non-default values are silently accepted and produce wrong results

The standard `GRUOpBuilder::IsOpSupported` (gru_op_builder.cc L75-84) rejects non-default `activations` (requires sigmoid/tanh) and `layout=1`. The standard `LSTMOpBuilder::IsOpSupported` (lstm_op_builder.cc L183-190) similarly rejects non-default activations and requires `layout==0`.

`StatefulGruOpBuilder::IsOpSupported` (lines 80-131) and `StatefulLstmOpBuilder::IsOpSupported` (lines 73-111) contain neither check — they only validate sequence_lens, clip, direction/quant, reset dtype, and input_forget. Two silent-error scenarios:

**(1) activations**: `rnn_details::AddUnidirectionGRU/LSTM` hardcodes the gate structure to sigmoid/tanh and never reads the activations attribute. A model with non-default activations passes IsOpSupported, is compiled into a fixed sigmoid/tanh QNN op, and silently produces wrong numerical output.

**(2) layout=1**: The lowering code correctly extracts seq_length/batch_size for layout=1, but always emits output tensors in layout=0 format (Y=[seq, num_dir, batch, hidden], Y_h=[num_dir, batch, hidden]). ONNX layout=1 requires Y=[batch, seq, num_dir, hidden] and Y_h=[batch, num_dir, hidden] — a shape mismatch.

**Impact**: Models with `layout=1` or non-default activations are silently accepted and produce wrong results. This was flagged in the v1 review and remains unresolved after the v2 force-push.

**Suggested fix:**

Add the same two guards to StatefulGruOpBuilder::IsOpSupported (before return Ort::Status()), mirroring GRUOpBuilder:
(1) activations check (see gru_op_builder.cc L81-84)
(2) layout check: RETURN_IF(layout != 0, "QNN EP: layout=1 not supported for StatefulGru.")

Apply the same fix to StatefulLstmOpBuilder::IsOpSupported (see lstm_op_builder.cc L183-190).
Consider factoring this shared validation into an rnn_details::ValidateRnnBaseAttributes() helper so all four builders share one source of truth.

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -39,6 +39,8 @@ using Microsoft::WRL::ComPtr;
 namespace onnxruntime {
 namespace test {
 constexpr const char* kOnnxDomain = "";
+// ONNX domain used by the QNN block-op test models.
+constexpr const char* kQtiAiswDomain = "qti_aisw";
```
huaychou: **[N-1] (Minor)** kQtiAiswDomain defined independently in the production header and test utils — risk of desync

`kQtiAiswDomain` is declared as `inline constexpr const char*` in `onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h` (inside the `onnxruntime` namespace).

`qnn_test_utils.h` line 43 declares an independent `constexpr const char*` with the same string without including the production header. If the domain string ever changes and only one definition is updated, production and test will use different domains and all node-assignment tests will silently pass with the wrong domain.

**Impact**: A domain mismatch silently corrupts `VerifyQnnEpModelAssignment`/`RunQnnModelTest` assertions at runtime with no compile-time signal.

**Suggested fix:**

Include the production header in qnn_test_utils.h and reuse the single constant:
  #include "core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h"
Remove the independent definition at line 43, and update usages to onnxruntime::kQtiAiswDomain.

onnxruntime/core/providers/qnn/builder/opbuilder/buffer_op_builder.cc
```diff
@@ -0,0 +1,229 @@
...
+//   mode values: BLOCKING=0, NON_BLOCKING_LEFT=1, NON_BLOCKING_RIGHT=2
+//
+// The ONNX Buffer node attributes are passed straight through to the QNN op, so the ONNX
+// attribute names match the QNN param names verbatim.
+class BufferOpBuilder : public BaseOpBuilder {
+ public:
+  BufferOpBuilder() : BaseOpBuilder("BufferOpBuilder") {}
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(BufferOpBuilder);
+
+  Ort::Status IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
```
huaychou: **[N-2] (Minor)** BufferOpBuilder declares IsOpSupported / ProcessInputs / ProcessAttributesAndOutputs as public, deviating from the established protected pattern in the base class

`BaseOpBuilder` (base_op_builder.h) declares `ProcessInputs` and `ProcessAttributesAndOutputs` as `protected virtual` — the pattern followed by every other op builder in the codebase, including StatefulGruOpBuilder and StatefulLstmOpBuilder.

`BufferOpBuilder` (lines 26–44) puts all three methods (IsOpSupported, ProcessInputs, ProcessAttributesAndOutputs) under `public:`, making ProcessInputs and ProcessAttributesAndOutputs directly callable from outside the class hierarchy. Only `OverrideOutputQuantParam` is correctly placed under `protected`.

**Impact**: No functional impact (virtual dispatch still works correctly), but breaks encapsulation and is inconsistent with the codebase convention.

**Suggested fix:**

Move IsOpSupported, ProcessInputs, and ProcessAttributesAndOutputs to the protected section of the BufferOpBuilder class definition, matching all other op builders in the codebase.

onnxruntime/core/providers/qnn/builder/opbuilder/rnn_op_utils.cc
```diff
@@ -0,0 +1,1212 @@
...
+      RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(onnx_outputs[i], output_tensor_infos[i]));
+    } else {
+      output_tensor_infos[i].qnn_data_type = input_tensor_infos[0].qnn_data_type;
+      // TODO: an absent output slot keeps the default UNDEFINED quant_param, emitted verbatim below.
+      // Dormant only because LSTM has no QDQ selector yet (always fp, where UNDEFINED is valid); before
+      // enabling a u8 LSTM, either fp-degrade a missing-output group in this builder (as the GRU op
+      // builder does) or backfill absent slots' quant encodings here.
+    }
+  }
+
```
huaychou: **[T-1] (Nit)** TODO about u8 LSTM QDQ selector in rnn_op_utils.cc lacks an issue tracking reference

The TODO comment near line 675 of `AddUnidirectionLSTM` notes that absent output slot quant encodings need to be fp-degraded or backfilled before a u8 LSTM QDQ selector can be enabled. The TODO has no JIRA or GitHub issue reference, making it hard to track.

**Suggested fix:**

Add an issue reference to the TODO, e.g.: // TODO(AISW-XXXXX): before enabling u8 LSTM QDQ selector...

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_gru_op_builder.cc
huaychou: **[M-1] (Major)** StatefulGru / StatefulLstm IsOpSupported missing activations and layout validation — non-default values are silently accepted and produce wrong results

The standard `GRUOpBuilder::IsOpSupported` (gru_op_builder.cc L75-84) rejects non-default `activations` (requires sigmoid/tanh) and `layout=1`. The standard `LSTMOpBuilder::IsOpSupported` (lstm_op_builder.cc L183-190) similarly rejects non-default activations and requires `layout==0`.

`StatefulGruOpBuilder::IsOpSupported` (lines 80-131) and `StatefulLstmOpBuilder::IsOpSupported` (lines 73-111) contain neither check — they only validate sequence_lens, clip, direction/quant, reset dtype, and input_forget. Two silent-error scenarios:

**(1) activations**: `rnn_details::AddUnidirectionGRU/LSTM` hardcodes the gate structure to sigmoid/tanh and never reads the activations attribute. A model with non-default activations passes IsOpSupported, is compiled into a fixed sigmoid/tanh QNN op, and silently produces wrong numerical output.

**(2) layout=1**: The lowering code correctly extracts seq_length/batch_size for layout=1, but always emits output tensors in layout=0 format (Y=[seq, num_dir, batch, hidden], Y_h=[num_dir, batch, hidden]). ONNX layout=1 requires Y=[batch, seq, num_dir, hidden] and Y_h=[batch, num_dir, hidden] — a shape mismatch.

**Impact**: Models with `layout=1` or non-default activations are silently accepted and produce wrong results. This was flagged in the v1 review and remains unresolved after the v2 force-push.

**Suggested fix:**

Add the same two guards to StatefulGruOpBuilder::IsOpSupported (before return Ort::Status()), mirroring GRUOpBuilder:
(1) activations check (see gru_op_builder.cc L81-84)
(2) layout check: RETURN_IF(layout != 0, "QNN EP: layout=1 not supported for StatefulGru.")

Apply the same fix to StatefulLstmOpBuilder::IsOpSupported (see lstm_op_builder.cc L183-190).
Consider factoring this shared validation into an rnn_details::ValidateRnnBaseAttributes() helper so all four builders share one source of truth.

onnxruntime/test/providers/qnn/qnn_test_utils.h
huaychou: **[N-1] (Minor)** kQtiAiswDomain defined independently in the production header and test utils — risk of desync

`kQtiAiswDomain` is declared as `inline constexpr const char*` in `onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h` (inside the `onnxruntime` namespace).

`qnn_test_utils.h` line 43 declares an independent `constexpr const char*` with the same string without including the production header. If the domain string ever changes and only one definition is updated, production and test will use different domains and all node-assignment tests will silently pass with the wrong domain.

**Impact**: A domain mismatch silently corrupts `VerifyQnnEpModelAssignment`/`RunQnnModelTest` assertions at runtime with no compile-time signal.

**Suggested fix:**

Include the production header in qnn_test_utils.h and reuse the single constant:
  #include "core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h"
Remove the independent definition at line 43, and update usages to onnxruntime::kQtiAiswDomain.

onnxruntime/core/providers/qnn/builder/opbuilder/buffer_op_builder.cc
huaychou: **[N-2] (Minor)** BufferOpBuilder declares IsOpSupported / ProcessInputs / ProcessAttributesAndOutputs as public, deviating from the established protected pattern in the base class

`BaseOpBuilder` (base_op_builder.h) declares `ProcessInputs` and `ProcessAttributesAndOutputs` as `protected virtual` — the pattern followed by every other op builder in the codebase, including StatefulGruOpBuilder and StatefulLstmOpBuilder.

`BufferOpBuilder` (lines 26–44) puts all three methods (IsOpSupported, ProcessInputs, ProcessAttributesAndOutputs) under `public:`, making ProcessInputs and ProcessAttributesAndOutputs directly callable from outside the class hierarchy. Only `OverrideOutputQuantParam` is correctly placed under `protected`.

**Impact**: No functional impact (virtual dispatch still works correctly), but breaks encapsulation and is inconsistent with the codebase convention.

**Suggested fix:**

Move IsOpSupported, ProcessInputs, and ProcessAttributesAndOutputs to the protected section of the BufferOpBuilder class definition, matching all other op builders in the codebase.

onnxruntime/core/providers/qnn/builder/opbuilder/rnn_op_utils.cc
huaychou: **[T-1] (Nit)** TODO about u8 LSTM QDQ selector in rnn_op_utils.cc lacks an issue tracking reference

The TODO comment near line 675 of `AddUnidirectionLSTM` notes that absent output slot quant encodings need to be fp-degraded or backfilled before a u8 LSTM QDQ selector can be enabled. The TODO has no JIRA or GitHub issue reference, making it hard to track.

**Suggested fix:**

Add an issue reference to the TODO, e.g.: // TODO(AISW-XXXXX): before enabling u8 LSTM QDQ selector...

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_gru_op_builder.cc
kuanyul-qti: Per [op def](https://docs.qualcomm.com/doc/80-63442-10/topic/blockop_definitions.html), StatefulGru and StatefulLstm have neither activation nor layout. So, I don't think we need to add the check.

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
kuanyul-qti: Revert this based on the suggestion in the onnxruntime PR. `ShouldAddDomain` always describes a duplicate as already existing in session options (`utils.cc:154-156`). Passing `all_ep_custom_op_domains` here therefore emits that misleading warning for the expected duplicate from another AutoEP device, even when the user supplied no domain. Keep `ShouldAddDomain` for the session-options check, but use `!DoesDomainWithNameExist(domain->domain_, all_ep_custom_op_domains)` for this collection-only check. Apply the same correction to the mirrored `OrtModel` path at lines 439-440.

onnxruntime/core/providers/qnn/builder/opbuilder/buffer_op_builder.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_gru_op_builder.cc
huaychou: Could we have the Buffer, StatefulGru, StatefulLstm op definition in codebase or document for reference?

kuanyul-qti: Added in the document.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_lstm_op_builder.cc
```diff
@@ -0,0 +1,212 @@
...
+                                  const bool& is_bidirection,
+                                  std::vector<std::string>& uni_lstm_output_names) const;
+};
+
+Ort::Status StatefulLstmOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                                 const OrtNodeUnit& node_unit,
+                                                 const Ort::Logger& logger) const {
+  ORT_UNUSED_PARAMETER(logger);
+  if (node_unit.Inputs().size() > 4 && node_unit.Inputs()[4].Exists()) {
+    TensorInfo tensor_info = {};
```
huaychou: **[M-1] (Major)** StatefulLstmOpBuilder::IsOpSupported missing backend guard and BOOL_8 type validation for reset input

`StatefulLstmOpBuilder::IsOpSupported` does not validate the backend type (NPU/IR) or the `QNN_DATATYPE_BOOL_8` data type of the reset input (ONNX in[8]) when it is present.

Compare with `StatefulGruOpBuilder::IsOpSupported` (stateful_gru_op_builder.cc:112–132), which correctly: (1) guards with `IsNpuBackend(backend_type) || IsIrBackend(backend_type)` (lines 115–119), and (2) validates that the reset's data type is `QNN_DATATYPE_BOOL_8` (lines 121–132).

If a StatefulLstm model with reset is submitted to a CPU/GPU backend, `IsOpSupported` silently accepts it; the failure is deferred to `AddToModelBuilder` via `RETURN_IF_NOT(false, ...)` at rnn_op_utils.cc:1050–1053, producing a harder-to-diagnose error.

**Impact**: StatefulLstm+reset on CPU/GPU backends produces a late-stage error rather than an early rejection; non-BOOL_8 reset tensors also cannot be caught at session initialization. This is a new finding in v3 — StatefulLstm is not symmetric with StatefulGru's guard logic.

**Suggested fix:**

Add the same backend guard and BOOL_8 type check to StatefulLstmOpBuilder::IsOpSupported when ONNX in[8] is present, mirroring StatefulGruOpBuilder::IsOpSupported lines 112–132.

onnxruntime/test/providers/qnn/qnn_test_utils.h
huaychou: **[N-1] (Minor)** kQtiAiswDomain defined independently in the production header and test utils — risk of desync

`kQtiAiswDomain` is declared as `inline constexpr const char*` in the production header `qnn_qti_aisw_custom_op.h` (line 18) and also independently as `constexpr const char*` in `qnn_test_utils.h` (line 43) without including the production header. If the domain string changes and only one definition is updated, production and test silently use different domains.

**Impact**: A domain mismatch silently corrupts `VerifyQnnEpModelAssignment`/`RunQnnModelTest` assertions with no compile-time signal. Still unresolved in v3.

**Suggested fix:**

Include the production header in qnn_test_utils.h and reuse the single constant:
  #include "core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h"
Remove the independent definition at line 43, and update usages to onnxruntime::kQtiAiswDomain.

onnxruntime/core/providers/qnn/builder/opbuilder/buffer_op_builder.cc
```diff
@@ -0,0 +1,229 @@
...
+//   mode values: BLOCKING=0, NON_BLOCKING_LEFT=1, NON_BLOCKING_RIGHT=2
+//
+// The ONNX Buffer node attributes are passed straight through to the QNN op, so the ONNX
+// attribute names match the QNN param names verbatim.
+class BufferOpBuilder : public BaseOpBuilder {
+ public:
+  BufferOpBuilder() : BaseOpBuilder("BufferOpBuilder") {}
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(BufferOpBuilder);
+
+ protected:
```
huaychou: **[N-2] (Minor)** BufferOpBuilder declares IsOpSupported / ProcessInputs / ProcessAttributesAndOutputs as public, deviating from the established protected pattern

`BufferOpBuilder` previously declared these three methods under `public:`, inconsistent with `BaseOpBuilder`'s `protected virtual` pattern.

**Suggested fix:**

Fixed in v3: IsOpSupported, ProcessInputs, and ProcessAttributesAndOutputs are now in the protected section.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -740,9 +741,22 @@ ort.unregister_execution_provider_library(ep_registration_name)
...
 
 Supported data types vary by operator and QNN backend. Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10/topic/operations.html) for more information.
 
+### QTI AISW block operators
+
+QNN EP also supports the following Qualcomm block operators in the `qti_aisw` domain. Their definitions are distinct from standard ONNX operators. Please refer to [this page](https://docs.qualcomm.com/doc/80-63442-10/topic/blockop_definitions.html) for Qualcomm's Block-op definitions.
+
+|Operator|Inputs|Outputs|Attributes|
+|---|---|---|---|
+|`qti_aisw:Buffer`|`data` (rank N); optional scalar-BOOL `reset`|`out`: same rank and type as `data`, with dimension `buffer_dim` set to `buffer_size`|Required: `buffer_size`, `buffer_dim`. Optional: `buffer_padding=0`, `stride=1`, `mode=0` (`BLOCKING`).|
```
huaychou: **[N-3] (Minor)** QNN-ExecutionProvider.md Buffer documentation does not note that mode=0 (BLOCKING) is unsupported on HTP

The documentation (line 756) lists `mode=0 (BLOCKING)` as the default for the Buffer mode attribute, but `BufferOpBuilder::IsOpSupported` (buffer_op_builder.cc:65–68) explicitly rejects mode=0 with: "Buffer mode 0 (BLOCKING) is not supported on HTP. Use mode 1 or 2."

**Impact**: A user following the documented default will receive an unexpected error at session creation time with no warning from the docs.

**Suggested fix:**

Update the mode attribute description in the Buffer row to note that mode=0 is unsupported on HTP and mode 1 or 2 is required, e.g.: "mode (uint32): 1=NON_BLOCKING_LEFT, 2=NON_BLOCKING_RIGHT. Note: mode=0 (BLOCKING) is not supported on HTP."

onnxruntime/core/providers/qnn/builder/opbuilder/rnn_op_utils.cc
```diff
@@ -0,0 +1,1212 @@
...
+    }
+  }
+  // because QNN LSTM three outputs are mandatory, we should provide them tensor info
+  std::vector<TensorInfo> output_tensor_infos(3);
+  for (size_t i = 0; i < 3; i++) {
+    if (onnx_outputs.size() > i && onnx_outputs[i].Exists()) {
+      RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(onnx_outputs[i], output_tensor_infos[i]));
+    } else {
+      output_tensor_infos[i].qnn_data_type = input_tensor_infos[0].qnn_data_type;
+      // TODO: an absent output slot keeps the default UNDEFINED quant_param, emitted verbatim below.
```
huaychou: **[T-1] (Nit)** TODO about u8 LSTM QDQ selector in rnn_op_utils.cc lacks an issue tracking reference

The TODO comment near line 669 of `AddUnidirectionLSTM` notes that absent output slot quant encodings need to be fp-degraded or backfilled before a u8 LSTM QDQ selector can be enabled. The TODO has no JIRA or GitHub issue reference, making it hard to track.

**Suggested fix:**

Add an issue reference to the TODO, e.g.: // TODO(AISW-XXXXX): before enabling u8 LSTM QDQ selector...

huaychou (commented):
Could you link the ORT PR in the comment?

docs/execution_providers/QNN-ExecutionProvider.md
kuanyul-qti: Added.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_lstm_op_builder.cc
kuanyul-qti: Added.

onnxruntime/core/providers/qnn/builder/opbuilder/buffer_op_builder.cc
```diff
@@ -0,0 +1,229 @@
...
+                                           const OrtNodeUnit& node_unit,
+                                           const Ort::Logger& logger) const {
+  ORT_UNUSED_PARAMETER(logger);
+
+  OrtNodeAttrHelper node_helper(node_unit);
+
+  // HTP constraint: mode 0 (BLOCKING) is not supported on HTP; only NON_BLOCKING_LEFT (1) and
+  // NON_BLOCKING_RIGHT (2) are accepted.
+  const int64_t mode = node_helper.Get("mode", static_cast<int64_t>(QNN_OP_BUFFER_MODE_BLOCKING));
+  RETURN_IF(mode == QNN_OP_BUFFER_MODE_BLOCKING,
```
huaychou: **[M-1] (Major)** Buffer applies an HTP-only mode restriction to every backend

Both the comment and public documentation describe `mode=0 (BLOCKING)` as an HTP restriction, but the current `RETURN_IF` is unconditional. CPU, GPU, or another backend that supports BLOCKING mode is therefore rejected during capability checking.

**Impact:** A backend-specific restriction leaks into every backend, reducing model support and contradicting the documented contract.

**Suggested fix:**

Following [C11], compute `is_npu_backend` and reject mode 0 only for HTP/NPU, with a non-HTP mode-0 test. If the op is intended to be HTP-only, explicitly restrict the entire op in both the builder and documentation.

onnxruntime/test/providers/qnn/qnn_test_utils.h
huaychou: **[N-1] (Minor)** Tests duplicate the qti_aisw domain string

The production header already defines `onnxruntime::kQtiAiswDomain`, while the test utility independently declares the same string. A future rename can leave tests constructing models for a different domain than production registers.

**Suggested fix:**

Reference the production constant from tests, or move the shared constant into a small dependency-light header.

onnxruntime/core/providers/qnn/builder/opbuilder/stateful_lstm_op_builder.cc
```diff
@@ -0,0 +1,224 @@
...
+            "QNN EP doesn't support input_forget=1 for StatefulLstm.");
+
+  if (node_unit.Inputs().size() > kQtiAiswStatefulLstmResetInputIndex &&
+      node_unit.Inputs()[kQtiAiswStatefulLstmResetInputIndex].Exists()) {
+    const QnnBackendType backend_type = qnn_model_wrapper.GetQnnBackendType();
+    RETURN_IF_NOT(IsNpuBackend(backend_type) || IsIrBackend(backend_type),
+                  "QNN EP: StatefulLstm reset requires the multi-time-step LSTM path; "
+                  "the selected backend does not support the 3D reset input.");
+    TensorInfo reset_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[kQtiAiswStatefulLstmResetInputIndex], reset_info));
```
huaychou: **[N-2] (Minor)** Reset inputs validate BOOL_8 but not the required scalar shape

All three builders define reset as a 0-D BOOL scalar, but capability validation checks only `QNN_DATATYPE_BOOL_8`. Because the `qti_aisw` placeholder schema does not constrain rank, a BOOL vector can pass model resolution and fail only during later QNN validation.

**Impact:** Invalid models are not rejected clearly in `IsOpSupported`; since this custom domain has no CPU fallback, session initialization ultimately fails.

**Suggested fix:**

Require `shape.empty()` in all three reset validations and add rank-1 BOOL reset negative tests. Op Spec Guard was unavailable, so this finding is downgraded to Minor.

onnxruntime/core/providers/qnn/builder/opbuilder/rnn_op_utils.cc
huaychou: **[T-1] (Nit)** Quantized LSTM TODO has no tracking issue

This TODO describes a missing-output quantization problem that must be fixed before enabling u8 LSTM, but it has no issue or owner and can become indefinite dormant correctness debt.

**Suggested fix:**

Link a tracking issue or work item and require a regression test when enabling the quantized LSTM selector.

cmake/patches/ort_core/0009-deduplicate-autoep-custom-op-domains.patch
```diff
@@ -0,0 +1,42 @@
...
+within that collection before registering the session's custom registry.
+
+---
+diff --git a/onnxruntime/core/session/utils.cc b/onnxruntime/core/session/utils.cc
+index 8fc6c1d8c0..8d3a79740d 100644
+--- a/onnxruntime/core/session/utils.cc
++++ b/onnxruntime/core/session/utils.cc
+@@ -318,7 +318,10 @@ static OrtStatus* CreateSessionAndLoadSingleModelImpl(_In_ const OrtSessionOptio
+       ORT_API_RETURN_IF_STATUS_NOT_OK(GetCustomOpDomainsFromEpDevice(*ep_device, domains));
+
```
huaychou: **[T-2] (Nit)** The ort_core patch contains trailing whitespace

`git diff --check` reports trailing whitespace at lines 21 and 33 of the patch.

**Suggested fix:**

Remove the trailing spaces from the two blank added lines and rerun `git diff --check`.

onnxruntime/core/providers/qnn/builder/opbuilder/buffer_op_builder.cc
```diff
@@ -0,0 +1,229 @@
...
+            "QNN EP: Buffer mode 0 (BLOCKING) is not supported on HTP. Use mode 1 or 2.");
```
huaychou: Just wondering, do we support Buffer on other backend?

huaychou (commented):
_Draft review generated by Claude Code (`/code-reviewer`). Please review the inline comments and submit when ready._

**Verdict:** `not_ready`

## Summary

The review found **1 Major, 2 Minor, and 2 Nit issues**. Fix M-1 before merge. N-2 should be addressed in the same update so the custom block-op reset contract is fully validated during capability checking.

**Suggested Fix Priority**: M-1 → N-2 → N-1 → T-1 → T-2.

Op Spec Guard was unavailable: the environment has no onnx Python package and QNN_SDK_ROOT is unset.

## Good parts

- **The earlier reset guard issue is fixed**: StatefulLstm now checks the backend and BOOL_8 dtype consistently with StatefulGru.
- **Shared lowering reduces duplication**: GRU/LSTM decomposition is centralized in rnn_op_utils.
- **Tests exercise real state reset**: on-device tests use a true/false/true sequence within one session.
- **Clean rebase**: the head rebases cleanly onto origin/main.

*Reviewed at commit `5bdf1843f36e`*

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -740,9 +741,22 @@ ort.unregister_execution_provider_library(ep_registration_name)
...
 
 Supported data types vary by operator and QNN backend. Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10/topic/operations.html) for more information.
 
+### QTI AISW block operators
+
+QNN EP also supports the following Qualcomm block operators in the `qti_aisw` domain. Their definitions are distinct from standard ONNX operators. Please refer to [this page](https://docs.qualcomm.com/doc/80-63442-10/topic/blockop_definitions.html) for Qualcomm's Block-op definitions.
+
+|Operator|Inputs|Outputs|Attributes|
+|---|---|---|---|
+|`qti_aisw:Buffer`|`data` (rank N); optional scalar-BOOL `reset`|`out`: same rank and type as `data`, with dimension `buffer_dim` set to `buffer_size`|Required: `buffer_size`, `buffer_dim`. Optional: `buffer_padding=0`, `stride=1`, `mode=0` (`BLOCKING`). Note: QNN EP's HTP backend requires `mode=1` (`NON_BLOCKING_LEFT`) or `mode=2` (`NON_BLOCKING_RIGHT`).|
```
qti-chuteng: Could we use the bulletpoint to display the inputs and outputs?

onnxruntime/core/providers/qnn/custom_op/qnn_qti_aisw_custom_op.h
```diff
@@ -0,0 +1,94 @@
...
+  QtiAiswPlaceholderOp(std::string op_type, std::string ep_type)
+      : op_type_(std::move(op_type)), ep_type_(std::move(ep_type)) {}
+
+  const char* GetName() const { return op_type_.c_str(); }
+  const char* GetExecutionProviderType() const { return ep_type_.c_str(); }
+
+  // OPTIONAL inputs to accept empty interior slots (e.g. sequence_lens, B, initial_h).
+  // Most are UNDEFINED (any type); reset slot is BOOL to prevent InferOutputTypes from
+  // propagating bool to outputs (see GetInputType for slot indices).
+  static constexpr size_t kMaxInputs = 24;  // generous upper bound; covers every block op's arity
```
qti-chuteng: How do we get 24?

onnxruntime/core/providers/qnn/qnn_provider_factory.h
```diff
@@ -83,11 +84,15 @@ class QnnEpFactory : public OrtEpFactory, public ApiPtrs {
...
-  // Both vectors must outlive any session that uses this factory (factory is a per-library singleton).
-  // domain.Add(op*) does NOT transfer ownership; op objects must be kept alive here.
+  // Custom op domains reported to ORT via GetCustomOpDomains. Holds both the qti_aisw block-op
+  // domain (always registered) and any domains built from ORT_QNN_CUSTOM_OP_DOMAINS.
+  // The domains and the op objects below must outlive any session that uses this factory
+  // (factory is a per-library singleton); domain.Add(op*) does NOT transfer ownership.
   std::vector<Ort::CustomOpDomain> custom_op_domains_;
+  // Placeholder ops built from ORT_QNN_CUSTOM_OP_DOMAINS (env-var UDO ops).
   std::vector<std::unique_ptr<qnn::QnnUdoPlaceholderOp>> custom_op_objects_;
+  // Placeholder ops for the qti_aisw block ops.
```
qti-chuteng: for the qti_aisw ops. (ex: BlockOp)

PR #613, opened 2026-07-08
[QNN EP] Fix 2 unit test failures on QAIRT 2.49

kuanyul-qti:
### Description

PartitionAddedInput tests: Replace `RandomNormalLike` with `EyeLike`, which has no QNN op builder and is absent from every QAIRT backend, making it unconditionally CPU-only regardless of QAIRT version.

### Motivation and Context

Three unit tests fail when run against QAIRT 2.49 artifacts while passing on QAIRT 2.47:

- `QnnCPUBackendTests.PartitionAddedInputRegisteredAsGraphInput`
- `QnnCPUBackendTests.PartitionAddedInputRegisteredAsGraphInputOffloadGraphIoQuantization`

Root cause — PartitionAddedInput tests: QAIRT 2.49's QnnCpu.dll added a native implementation of `RandomNormalLike`, so `backendValidateOpConfig` now accepts the op. QNN EP claims the entire graph and `rnl_output` becomes a NATIVE tensor instead of a partition-added APP_WRITE input. The test was relying on the QAIRT CPU backend rejecting `RandomNormalLike` to drive the partition-boundary scenario.

### DIFF
--- a/onnxruntime/test/providers/qnn/qnn_basic_test.cc
+++ b/onnxruntime/test/providers/qnn/qnn_basic_test.cc
@@ -2399,7 +2399,7 @@ TEST(QnnSaverBackendTests, DISABLED_QnnSaver_OutputFiles) {
   EXPECT_TRUE(std::filesystem::exists(qnn_saver_output_dir / "params.bin"));
 }
 
-// Returns a function that builds a model with RandomNormalLike (CPU-only) + Add
+// Returns a function that builds a model with EyeLike (CPU-only) + Add
 // to test partition-added inputs.
 static GetTestModelFn BuildPartitionAddedInputModel() {
   return [](ModelTestBuilder& builder) {
@@ -2408,14 +2408,14 @@ static GetTestModelFn BuildPartitionAddedInputModel() {
     // Create input
     MakeTestInput<float>(builder, "input", TestInputDef<float>({1, 3}, false, {1.0f, 2.0f, 3.0f}));
 
-    // Create constant initializer for RandomNormalLike
-    builder.MakeInitializer<float>("constant", {1, 3}, {0.0f, 0.0f, 0.0f});
+    // "constant" is a graph input (not an initializer) to prevent ORT constant-folding EyeLike.
+    MakeTestInput<float>(builder, "constant", TestInputDef<float>({1, 3}, false, {0.0f, 0.0f, 0.0f}));
 
-    // RandomNormalLike: CPU-only op that creates a partition-added input
-    builder.AddNode("rnl", "RandomNormalLike", {"constant"}, {"rnl_output"}, kOnnxDomain);
+    // EyeLike: CPU-only op (no QNN builder) that creates a partition-added input
+    builder.AddNode("el", "EyeLike", {"constant"}, {"el_output"}, kOnnxDomain);
 
     // Add: combines graph input with partition-added input
-    builder.AddNode("add", "Add", {"input", "rnl_output"}, {"add_output"}, kOnnxDomain);
+    builder.AddNode("add", "Add", {"input", "el_output"}, {"add_output"}, kOnnxDomain);
 
     builder.MakeOutput("add_output");
   };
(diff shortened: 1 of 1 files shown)


huaychou:
Just wondering, if https://github.com/onnxruntime/onnxruntime-qnn/pull/599 is merged, could we use it to avoid offloading RandomNormalLike to QNN?

kuanyul-qti:
> Just wondering, if #599 is merged, could we use it to avoid offloading RandomNormalLike to QNN?

Using `op_type_to_filter = "exclude:RandomNormalLike"` in the test would work mechanically. It would force the op back to CPU. But that makes the test depend on a provider option to manufacture the partition boundary, which is a test-code-level workaround that obscures the intent. The test is supposed to verify what happens when the backend rejects an op organically; using a filter is a different (and less representative) trigger.

huaychou:
> > Just wondering, if #599 is merged, could we use it to avoid offloading RandomNormalLike to QNN?
> 
> Using `op_type_to_filter = "exclude:RandomNormalLike"` in the test would work mechanically. It would force the op back to CPU. But that makes the test depend on a provider option to manufacture the partition boundary, which is a test-code-level workaround that obscures the intent. The test is supposed to verify what happens when the backend rejects an op organically; using a filter is a different (and less representative) trigger.

But it will depend on the QNN backend support and need to update once QNN support eyelike op. Do we have any plan to avoid this issue?

kuanyul-qti:
> But it will depend on the QNN backend support and need to update once QNN support eyelike op. Do we have any plan to avoid this issue?

A 100% guarantee that this test will never need updating is not achievable — any op we pick can gain QNN support in a future QAIRT release, as `RandomNormalLike` did in 2.49.

`EyeLike` is a more stable choice than `RandomNormalLike` for a concrete reason: it has no entry in `QnnOpDef.h` currently (not a defined QNN API op). For `EyeLike` to break this test, two things must happen in sequence: Qualcomm adds it to `QnnOpDef.h`, and someone in this repo writes a new op builder for it — both are visible, reviewable changes. That's a lower-probability path than `RandomNormalLike`, which only needed Qualcomm to flip the backend implementation.

If it does break in the future, the fix is the same: swap to another CPU-only op. The comment in the test already documents why `EyeLike` was chosen, so the update path is clear.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -2408,11 +2408,11 @@ static GetTestModelFn BuildPartitionAddedInputModel() {
...
 
-    // Create constant initializer for RandomNormalLike
-    builder.MakeInitializer<float>("constant", {1, 3}, {0.0f, 0.0f, 0.0f});
+    // "constant" is a graph input (not an initializer) to prevent ORT constant-folding EyeLike.
+    MakeTestInput<float>(builder, "constant", TestInputDef<float>({1, 3}, false, {0.0f, 0.0f, 0.0f}));
 
-    // RandomNormalLike: CPU-only op that creates a partition-added input
-    builder.AddNode("rnl", "RandomNormalLike", {"constant"}, {"rnl_output"}, kOnnxDomain);
+    // EyeLike: CPU-only op (no QNN builder) that creates a partition-added input
+    builder.AddNode("rnl", "EyeLike", {"constant"}, {"rnl_output"}, kOnnxDomain);
```
huaychou: **[T-1] (Nit)** Node name "rnl" is a stale abbreviation for RandomNormalLike — now the node is EyeLike

`builder.AddNode("rnl", "EyeLike", ...)`'s first argument `"rnl"` is an abbreviation for `RandomNormalLike`, which is now stale after switching to `EyeLike`. The same occurs at line 2507.

The output tensor name `"rnl_output"` is hard-coded in test assertions (lines 2483, 2578); renaming it requires updating those assertions as well, which is beyond this PR's scope — but worth doing the next time these helpers are touched.

**Suggested fix:**

Rename `AddNode("rnl", "EyeLike", ...)` to `AddNode("el", "EyeLike", ...)` at both occurrences. When convenient, also rename `"rnl_output"` to `"el_output"` (requires updating lines 2418, 2483, 2507, 2510, 2578).

kuanyul-qti: Done

huaychou (approved):
Remember to update the PR description.

qti-chuteng (approved):

PR #614, opened 2026-07-08
[QNN EP] Extract neuron activations into a standalone op builder

huaychou:
### Description

- New neuron_op_builder.cc with NeuronOpBuilder for the 8 neuron ops, including the moved ProcessNodeAttribute helper, OverrideQuantParams, and the OverrideOutputQuantParam override.
- Removed those blocks + the Softplus check + the dead ProcessSigmoidOrTanhOutput declaration from simple_op_builder.cc.
- Removed the stale CreateHardSigmoidOpBuilder declaration (never defined or called) from the factory header.
- Registered the 8 ops via CreateNeuronOpBuilder in the factory.

### Motivation and Context

- The neuron activations (Elu, Gelu, Relu, Sigmoid, Tanh, HardSigmoid, HardSwish, Softplus) all lower to QNN_OP_ELEMENT_WISE_NEURON but carried per-op param construction, a Softplus rank guard, a Tanh uint16 QNN validation workaround, and 16-bit output quant-param overrides inside SimpleOpBuilder — again contradicting its "simple" contract. Move them into a new NeuronOpBuilder.


onnxruntime/core/providers/qnn/builder/opbuilder/neuron_op_builder.cc
```diff
@@ -0,0 +1,230 @@
...
+    // The QNN documentation states that the output scale and offset for ufixed_point_16 should be
+    // (1/32768) and -32768, respectively. However, the QNN validator incorrectly rejects these values.
+    if (op_type == "Tanh") {
+      TensorInfo output_info = {};
+      RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Outputs()[0], output_info));
+      if (output_info.qnn_data_type == QNN_DATATYPE_UFIXED_POINT_16) {
+        ORT_CXX_LOG(logger,
+                    ORT_LOGGING_LEVEL_INFO,
+                    ("Skipping QNN validation for Tanh node '" + node_unit.Name() +
+                     "' with quantized unit16 output.")
```
chunghow-qti: uint16

onnxruntime/core/providers/qnn/builder/opbuilder/neuron_op_builder.cc
```diff
@@ -0,0 +1,230 @@
...
+    return Ort::Status();
+  }
+
+  const std::string& op_type = node_unit.OpType();
+
+  if (do_op_validation) {
+    const auto qnn_backend_type = qnn_model_wrapper.GetQnnBackendType();
+
+    if (op_type == "Softplus" && qnn_backend_type != QnnBackendType::CPU) {
+      TensorInfo input_info = {};
```
chunghow-qti, line 62: would you like helper function like this?

https://github.com/onnxruntime/onnxruntime-qnn/blob/9ffa9f3fc33f20ed293b3b7342ad00993e12241a/onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc#L159-L183

huaychou: No, I don't think so.

onnxruntime/core/providers/qnn/builder/opbuilder/neuron_op_builder.cc
huaychou: Fixed

PR #615, opened 2026-07-08
[QNN EP] Extract Quantize/Dequantize into a standalone op builder

huaychou:
### Description

- New quantize_op_builder.cc with QuantizeOpBuilder for both Q and DQ.
- Removed the DQ/Q blocks from SimpleOpBuilder::ExplicitOpCheck and the constant-fold short-circuit from ProcessAttributesAndOutputs.
- Dropped the now-dead qdq_constant_folding.h include from simple_op_builder.cc (kept qnn_graph_utils.h for kMSInternalNHWCDomain).
- Registered both ops via CreateQuantizeOpBuilder in the factory.

### Motivation and Context

- SimpleOpBuilder had accreted large Q/DQ-specific logic (per-channel-on- non-const rejection, offload_graph_io_quantization external tensor-name overrides, and constant folding), contradicting its "no attributes / no special handling" contract. Move QuantizeLinear and DequantizeLinear into a new QuantizeOpBuilder, matching the existing standalone-builder pattern (selu, quickgelu, etc.).

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/quantize_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/quantize_op_builder.cc
@@ -1,0 +1,123 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include "core/providers/qnn/builder/op_builder_factory.h"
+#include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
+#include "core/providers/qnn/builder/opbuilder/qdq_constant_folding.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+// Handles ONNX QuantizeLinear and DequantizeLinear. Q/DQ have no QNN params, so
+// only inputs & outputs are processed (plus a compile-time constant-fold short-circuit).
+class QuantizeOpBuilder : public BaseOpBuilder {
+ public:
+  QuantizeOpBuilder() : BaseOpBuilder("QuantizeOpBuilder") {}
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(QuantizeOpBuilder);
+
+ protected:
+  Ort::Status ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                          const OrtNodeUnit& node_unit,
+                                          std::vector<std::string>&& input_names,
+                                          const Ort::Logger& logger,
+                                          bool do_op_validation) const override ORT_MUST_USE_RESULT;
+
+ private:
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/quantize_op_builder.cc
```diff
@@ -0,0 +1,121 @@
...
+
+  return Ort::Status();
+}
+
+Ort::Status QuantizeOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
+                                                           const OrtNodeUnit& node_unit,
+                                                           std::vector<std::string>&& input_names,
+                                                           const Ort::Logger& logger,
+                                                           bool do_op_validation) const {
+  if (input_names.empty()) {
```
minfhong-qti, line 94: Just curious. When will `input_names` be empty?

onnxruntime/core/providers/qnn/builder/opbuilder/quantize_op_builder.cc
```diff
@@ -0,0 +1,121 @@
...
+    return Ort::Status();
+  }
+
+  if (do_op_validation) {
+    RETURN_IF_ERROR(ValidateQdqNode(qnn_model_wrapper, node_unit));
+  }
+
+  // Emit a STATIC tensor instead of an APP_WRITE input for standalone Q/DQ on constant inputs.
+  if (CanFoldConstantQdq(qnn_model_wrapper, node_unit)) {
+    Ort::Status fold_status = TryFoldConstantQDQ(qnn_model_wrapper, node_unit);
```
minfhong-qti, line 104: Is `TryFoldConstantQDQ` guaranteed to fold if `CanFoldConstantQdq` passed?
The logic here is a bit weird.

huaychou: No, `CanFoldConstantQdq` only guarantee the single Q/DQ node with constant input. `TryFoldConstantQDQ` can fail due to the unsupported datatype or missing quantization parameters.

onnxruntime/core/providers/qnn/builder/opbuilder/quantize_op_builder.cc
huaychou: No, it won't be empty. This is copied from `simple_op_builder.cc`.

onnxruntime/core/providers/qnn/builder/opbuilder/quantize_op_builder.cc
minfhong-qti: Then the question will become why not combine them as one function?

huaychou: It is designed in https://github.com/onnxruntime/onnxruntime-qnn/pull/339. Hi @qti-yuduo, could you help answer this question?

qti-chuteng (approved):

huaychou:
Because there are still issues with `TryFoldConstantQDQ`, we close this, so it doesn’t get merged into mainline.

PR #618, opened 2026-07-08
[QNN EP] Extend LayerNormFusion support

qti-mattsinc:
### Description
* Support LayerNormFusion in FP16 graphs.
* Check for broadcastable gamma/beta shapes.
* Partially apply LayerNormFusion when gamma/beta can't fold into the QNN layernorm op, leaving them as standalone ElementWiseBinary ops.
* Add LayerNormFusion tests for the GPU.

### Motivation and Context
* Enable layernorm fusion in more models, particularly for GPU.


qti-mattsinc:
@yath1, I see that you are working on a similar fusion in #589. Maybe you can help review this PR? And I can review your PR when it's out of a draft 🙂

nagesh-qti:
Hi  @qti-mattsinc Is the case where models using NCHW layout [N, C, H, W] where normalization may be over the channel
axis (axes={1}) covered? Is not such models would silently fall back to executing all 9 decomposed ops individually on HTP instead of being fused into a single LayerNorm, losing the performance and
precision benefits of the hardware op.

My patch introduces "Pre transpose"-> Layernorm -> "Post transpose" which changes the axis to trailing axis acceptable by current layernorm.

qti-mattsinc:
> Hi @qti-mattsinc Is the case where models using NCHW layout [N, C, H, W] where normalization may be over the channel axis (axes={1}) covered? Is not such models would silently fall back to executing all 9 decomposed ops individually on HTP instead of being fused into a single LayerNorm, losing the performance and precision benefits of the hardware op.
> 
> My patch introduces "Pre transpose"-> Layernorm -> "Post transpose" which changes the axis to trailing axis acceptable by current layernorm.


@nagesh-qti, your changes in #649 are orthogonal to this PR, and my expectation would be that both that PR and this one will be merged.

nagesh-qti:
@qti-mattsinc Correct I will merge the patch after you.  By this I can rebase and test before merging.

nagesh-qti (approved):

qti-mattsinc:
Bump for review from code-owner.

cc. @tirupath-qti / @qti-chuteng / @yath1 / any others

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -53,20 +53,98 @@ static std::optional<float> GetConstantFloatScalar(const QnnModelWrapper& qmw,
...
+    gsl::span<const uint32_t> axes,
+    /*out*/ std::vector<uint32_t>& squeezed_shape) {
+  // Normalize axes: sort and deduplicate.
+  std::vector<uint32_t> normalized_axes(axes.begin(), axes.end());
+  std::sort(normalized_axes.begin(), normalized_axes.end());
+  normalized_axes.erase(std::unique(normalized_axes.begin(), normalized_axes.end()),
+                        normalized_axes.end());
+
+  // If already the right rank, verify each dim matches the input at the corresponding axis.
+  if (scale_bias_shape.size() == normalized_axes.size()) {
```
qti-yuduo: This equal-rank shortcut maps `scale_bias_shape[i]` positionally onto `normalized_axes[i]`, but ONNX broadcasting **right-aligns** gamma/beta against the input. The general path just below (L100-128) does this correctly by left-padding to the input rank, so the two disagree whenever the axes are not a trailing suffix.

That case is newly reachable in this PR because the trailing-suffix axes check was narrowed to `if (is_npu_backend)` (L279), so CPU/GPU now accept non-trailing axes.

Example - input `[1,16,64,16]`, `axes=[1]`, gamma `[16]`:
- ONNX right-aligns gamma to `[1,1,1,16]`, so `Mul` scales along **axis 3**.
- The shortcut checks `input_shape[1] == 16` -> passes, and emits QNN LayerNorm with `axes=[1]` and gamma `[16]`, i.e. scaling along **axis 1**. Silently wrong results, no error.
- The general path would left-pad to `[1,1,1,16]`, see `dim[1]=1 != 16`, and correctly reject.

When axes *are* a trailing suffix both paths agree, which is why this is invisible on NPU. `LayerNormFusion_4D_3D_Axis1_GammaBeta` uses rank-3 gamma `{16,1,1}` so it takes the general path and does not cover this branch.

Suggest just deleting the shortcut - the broadcast-aware path below already handles the equal-rank case correctly.

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -549,20 +619,28 @@ static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qmw,
   if (!qmw.IsQnnTensorWrapperExist(root_input.name)) {
     RETURN_IF_NOT(qmw.AddTensorWrapper(std::move(input_tensor)), "Failed to add input tensor.");
   }
-  if (!qmw.IsQnnTensorWrapperExist(gamma_input.name)) {
+  if (gamma_input != nullptr && !qmw.IsQnnTensorWrapperExist(gamma_input->name)) {
```
qti-yuduo, line 611: Validate/compose asymmetry can drop the squeezed shape:

- `ValidateQnnNode` (L605) receives `Qnn_Tensor_t` **by value** from the local `gamma_tensor`, so validation always sees the squeezed rank.
- Composition passes gamma **by name** (L634) and only registers the squeezed wrapper when none exists. `AddTensorWrapper` early-returns on an existing name (`qnn_model_wrapper.cc:132`), so a pre-existing wrapper wins and the squeezed shape is silently discarded.

Trigger: gamma/beta produced by an **upstream in-graph node** (e.g. a DQ or Reshape) rather than a graph input. Node groups compose in topological order (`qnn_node_group.cc:193`), so the producer registers gamma's output wrapper at its original rank first. A `[1,1,16]` gamma for `axes=[-1]` then reaches QNN LayerNorm as rank 3 while validation saw rank 1 -> passes validation, fails at graph finalization.

Note a plain graph input does *not* hit this: `RegisterGraphInputOutputInOrder` runs inside `ComposeQnnGraph` (L650), i.e. after all `AddToModelBuilder` calls, and skips names absent from `model_tensors_map_` (L614). So the new `DynamicGammaBeta`/`DynamicBeta` tests pass legitimately and this stays latent rather than showing up as a failure.

Fix options: register the squeezed wrapper unconditionally, or bail out of the fusion when an existing wrapper's rank differs from the squeezed rank.

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
qti-mattsinc: Good catch. Removed the shortcut.

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
qti-mattsinc: To fix this, I have gone back to preventing the folding when gamma/beta are not constants.

qti-mattsinc:
gentle bump for review @qti-yuduo

PR #620, opened 2026-07-08
Fix WinError 5 in archive step from Windows symlinks under _deps/

qti-niscmami:
### Description
  `_iter_release_files` in `qcom/scripts/all/archive_tests.py` intended to skip Windows symlinks, but evaluated `is_file()` first.  
Which in-turn follows the link and raises `WinError 5` before the `is_symlink()` filter runs. Swap the order.

### Motivation and Context
  A recent flatbuffers update added a symlink under 
`_deps/`:`flatbuffers-src/java/src/test/java/DictionaryLookup -> ../../../../tests/DictionaryLookup`. 

The target exists, but Windows' `stat` (follow-symlinks) denies access on symlinks created without `SeSymbolicLinkPrivilege` and non-privileged dev builds hit it every time. 

The existing `is_symlink()` check proves the case was meant to be filtered but it just never runs due to short-circuit order.


qti-chuteng (approved):

qcom/scripts/all/archive_tests.py
```diff
@@ -115,7 +115,10 @@ def accept_repo_relative(self, rel: Path) -> bool:
 
 def _iter_release_files(release_dir: Path):
     for p in release_dir.glob("**/*"):
-        if not p.is_file() or (platform.system() == "Windows" and p.is_symlink()):
+        # Check is_symlink() first: on Windows is_file() raises on some symlinks.
+        if platform.system() == "Windows" and p.is_symlink():
```
qti-kromero: nit: platform.system() == "Windows" could be placed and evaluated outside of loop

qti-kromero (approved):

qcom/scripts/all/archive_tests.py
qti-niscmami: Done

PR #621, opened 2026-07-08
Remove Not_Equal op work-around for FP->Bool Cast op

qti-shubham:
### Description
Not_Equal op has accuracy issue on v79 architecture: https://jira-dc.qualcomm.com/jira/browse/AISW-192595

All the Attention Mask Softmax block were affected.

FP->Bool Cast op is now natively supported by QNN (tested with v2.47 on device)

Fixes = https://github.com/qcom-ai-hub/tetracode/issues/19570

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
@@ -31,64 +31,7 @@ class CastOpBuilder : public BaseOpBuilder {
                                           std::vector<std::string>&& input_names,
                                           const Ort::Logger& logger,
                                           bool do_op_validation) const override ORT_MUST_USE_RESULT;
-
- private:
-  // QNN HTP currently does not support casting FP16/FP32 to Bool, and thus such Cast will be replaced by NotEqual with
-  // an additional static input 0.f to achieve the idential functional.
-  bool IsFpToBoolCast(const OrtNodeUnit& node_unit) const;
-  Ort::Status ProcessExtraInputForNotEqual(QnnModelWrapper& qnn_model_wrapper,
-                                           const OrtNodeUnit& node_unit,
-                                           std::vector<std::string>& input_names,
-                                           const Ort::Logger& logger) const;
 };
-
-bool CastOpBuilder::IsFpToBoolCast(const OrtNodeUnit& node_unit) const {
-  ONNXTensorElementDataType input_type = node_unit.Inputs()[0].type;
-  ONNXTensorElementDataType output_type = node_unit.Outputs()[0].type;
-
-  Qnn_DataType_t input_qnn_dtype = QNN_DATATYPE_UNDEFINED;
-  Qnn_DataType_t output_qnn_dtype = QNN_DATATYPE_UNDEFINED;
-
-  if (!utils::GetQnnDataType(false, input_type, input_qnn_dtype).IsOK() ||
-      !utils::GetQnnDataType(false, output_type, output_qnn_dtype).IsOK()) {
-    return false;
-  }
-
-  return ((input_qnn_dtype == QNN_DATATYPE_FLOAT_16 || input_qnn_dtype == QNN_DATATYPE_FLOAT_32) &&
-          output_qnn_dtype == QNN_DATATYPE_BOOL_8);
-}
-
-Ort::Status CastOpBuilder::ProcessExtraInputForNotEqual(QnnModelWrapper& qnn_model_wrapper,
-                                                        const OrtNodeUnit& node_unit,
-                                                        std::vector<std::string>& input_names,
-                                                        const Ort::Logger& logger) const {
-  const auto& input = node_unit.Inputs()[0];
-  if (input.quant_param.has_value()) {
-    return Ort::Status();
-  }
-
-  // Build additional static input with value 0.
-  const std::string& input_name = utils::UniqueNameGenerator().New(node_unit, "_notequal_zero");
-
-  Qnn_DataType_t qnn_data_type = QNN_DATATYPE_UNDEFINED;
-  ONNXTensorElementDataType input_type = input.type;
-  RETURN_IF_ERROR(utils::GetQnnDataType(false, input_type, qnn_data_type));
-
-  QnnTensorWrapper input_tensor_wrapper(input_name,
-                                        QNN_TENSOR_TYPE_STATIC,
-                                        qnn_data_type,
-                                        QnnQuantParamsWrapper(),
-                                        std::vector<uint32_t>{1},
-                                        std::vector<uint8_t>(utils::GetElementSizeByType(qnn_data_type), 0));
-  RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(input_tensor_wrapper)),
-                "Failed to add additional input tensor for QNN Cast node that will be replaced by NotEqual.");
-  input_names.push_back(input_name);
-
-  ORT_CXX_LOG(logger,
-              ORT_LOGGING_LEVEL_VERBOSE,
-              ("FP-to-Bool Cast node " + node_unit.Name() + " is replaced by NotEqual.").c_str());
-  return Ort::Status();
-}
 
 Ort::Status CastOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
                                          const OrtNodeUnit& node_unit,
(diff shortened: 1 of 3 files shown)


onnxruntime/test/providers/qnn/cast_test.cc
```diff
@@ -228,16 +228,22 @@ TEST_F(QnnHTPBackendTests, TestCastInt32ToInt64HTP) {
                          ExpectedEPNodeAssignment::All, "htp");
 }
 
-// Cast float to bool on HTP.
+// Cast float to bool on HTP using native QNN Cast op (JIRA: AISW-192595).
+// Skipped: native Cast FP->Bool not supported by the x86_64 HTP simulator: needs atleast v73 arch
 TEST_F(QnnHTPBackendTests, TestCastFloatToBoolHTP) {
+  QNN_SKIP_TEST_ON_LINUX_X86_64("Cast FP->Bool requires HTP arch >= v73; not supported by the x86_64 simulator.");
```
yath1: What would happen when we are targeting devices which are <73. For example: v68?

onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
```diff
@@ -31,65 +31,8 @@ class CastOpBuilder : public BaseOpBuilder {
                                           std::vector<std::string>&& input_names,
                                           const Ort::Logger& logger,
                                           bool do_op_validation) const override ORT_MUST_USE_RESULT;
-
- private:
-  // QNN HTP currently does not support casting FP16/FP32 to Bool, and thus such Cast will be replaced by NotEqual with
```
yath1, line 36: From here: https://docs.qualcomm.com/doc/80-63442-10/topic/HtpOpDefSupplement.html Cast still does not support casting FP16 to Bool. Are the docs outdated?

yath1:
Adding original author @minfhong-qti

onnxruntime/test/providers/qnn/cast_test.cc
qti-shubham: v68 don't have floating point support. The user should submit a Quantized model.

qti-shubham:
Another alternative to `NotEqual` op is `Greater` op. It doesn't have accuracy issue on v79.
I will check if CI fails again than I will switch to `Greater` but QAIRT converter uses `CAST` only.

yath1:
> Another alternative to `NotEqual` op is `Greater` op. It doesn't have accuracy issue on v79. I will check if CI fails again than I will switch to `Greater` but QAIRT converter uses `CAST` only.

What do you think of architecture specific switching then? <v73, keep the same logic as now. >v73 could be either direct `cast`/`greater` ?

qti-shubham:
> > Another alternative to `NotEqual` op is `Greater` op. It doesn't have accuracy issue on v79. I will check if CI fails again than I will switch to `Greater` but QAIRT converter uses `CAST` only.
> 
> What do you think of architecture specific switching then? <v73, keep the same logic as now. >v73 could be either direct `cast`/`greater` ?

I don't think EP receives HTP arch as option input. I think it may not be good aesthetics to check for HTP arch in code.

onnxruntime/test/providers/qnn/cast_test.cc
```diff
@@ -228,16 +228,22 @@ TEST_F(QnnHTPBackendTests, TestCastInt32ToInt64HTP) {
...
+  SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V69);
   RunCastOpTest<float>({3, 3},
                        ONNX_NAMESPACE::TensorProto_DataType::TensorProto_DataType_BOOL,
                        ExpectedEPNodeAssignment::All,
                        "htp");
 }
 
-// Cast float16 to bool on HTP.
+// Cast float16 to bool on HTP using native QNN Cast op (JIRA: AISW-192595).
```
minfhong-qti: We should not have JIRA number in code, and as well for below and PR description.

onnxruntime/test/providers/qnn/cast_test.cc
minfhong-qti: Why skip it instead of specifying HTP arch for Linux x86 environment?

onnxruntime/test/util/test_utils.cc
```diff
@@ -154,10 +154,12 @@ void VerifyOutput(const std::string& output_name,
       break;
     }
     case ONNX_TENSOR_ELEMENT_DATA_TYPE_BOOL: {
-      const bool* expected_data = expected_value.GetTensorData<bool>();
-      const bool* actual_data = actual_value.GetTensorData<bool>();
+      const uint8_t* expected_data = expected_value.GetTensorData<uint8_t>();
+      const uint8_t* actual_data = actual_value.GetTensorData<uint8_t>();
       for (size_t i = 0; i < element_count; ++i) {
-        EXPECT_EQ(expected_data[i], actual_data[i]) << "Element " << i << " mismatch for " << output_name;
+        // Compare as bool (nonzero = true): QNN's Bool output may return 0xFF instead of 0x01,
```
minfhong-qti: Is the statement "QNN's Bool output may return 0xFF" confirmed with HTP team? It seems a bit weird.
In addition, modify the base comparison logic here could have great impact, and I do not prefer it.

onnxruntime/test/providers/qnn/cast_test.cc
qti-shubham: will remove JIRA number.

onnxruntime/test/providers/qnn/cast_test.cc
qti-shubham: the reason I didn't added is that I have to replicate test for all HTP arch on Linux x86 which is not fruitful. Just like `QNN NotEqual` case, behavior of CAST(fp->bool) may vary on different HTP arch.

onnxruntime/test/util/test_utils.cc
qti-shubham: it was just a hypothesis. CI stilling failing for Win Arm64 (v73). The FP32→Bool test is running but comparing booleans incorrectly (elements 1-8 both say true yet fail).
I don't know the root cause but the logs are:

```
/local/mnt/workspace/repo/onnxruntime-qnn/onnxruntime/test/util/test_utils.cc:160: Failure
Expected equality of these values:
  expected_data[i]
    Which is: true
  actual_data[i]
    Which is: true
Element 1 mismatch for Y

/local/mnt/workspace/repo/onnxruntime-qnn/onnxruntime/test/util/test_utils.cc:160: Failure
Expected equality of these values:
  expected_data[i]
    Which is: true
  actual_data[i]
    Which is: true
Element 2 mismatch for Y

... (same pattern repeated for elements 3 through 8) ...
```

qti-shubham: I am thinking of using `QNN Greater Op` instead of `QNN Cast`. It's more reliable and consistent with tests.

qti-shubham: From latest logs, after the bool normalized test checks :
TestCastFloatToBoolHTP — [OK] on V73. 
TestCastFloat16ToBoolHTP — FAILED with num_ep_nodes=0, num_nodes=1 — QNN EP rejects the node on v73.

qti-chuteng:
Hi @qti-shubham, due to CCI, we couldn't put any our internal information on the PR comment and code. Could you remove the JIRA link and any related information from this PR? thanks!

qti-shubham:
For CI Win Arm64 device to pass, I have to do following hacks:
1. CAST(FP32->BOOL): I have to update the `test_utils.cc` bool accuracy check to uint8 dtype and compare against 0
2. CAST(FP32->BOOL): I have to disable test on v73

However, both of above operations works fine on HUB `Snapdragon X Elite CRD` with tungsten backend.

qti-shubham:
The alternate PR using Greater(Abs(x), 0) for which CI is passing without hacks, also faced v79 HTP kernel bug with 0.0 comparison : https://github.com/onnxruntime/onnxruntime-qnn/pull/627

The only viable solution remaining this this PR only :
image

qti-shubham:
The cast op passes on v69 and v73 windows arm64 device. It looks like a false alarm from CI
image

I feel it is safe to use and the right translation as well. 
This is what Qairt-converter do as well.

Tested Inference jobs = 
v69 = https://dev.aihub.qualcomm.com/jobs/j57xqvwvg
win arm v73 = https://dev.aihub.qualcomm.com/jobs/jgdoj79rp
v75 = https://dev.aihub.qualcomm.com/jobs/jg9w691wp
v79 = https://dev.aihub.qualcomm.com/jobs/jp1erql8g

qti-shubham:
Other models also passed which were previously having 0% accuracy on v79 device:
Bert base = https://dev.aihub.qualcomm.com/jobs/jgdoj076p 
Albert = https://dev.aihub.qualcomm.com/jobs/j5wqjxm45 

image

onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
yath1: Could you pls confirm if this is still an issue?

onnxruntime/test/providers/qnn/cast_test.cc
yath1: Okay, v69 does have fp but looks like you have addressed it below.

yath1 (approved):
Looks good to me. I have one concern with fp16 precision which I have raised below but please feel free to address it separately if it needs more changes. Thanks for the fix :)

PR #624, opened 2026-07-09
[CI] Fix corrupt QAIRT SDK cache poisoning CI builds

huaychou:
### Description

- Refactor FileCache.fetch into a retry loop (3 attempts, exponential backoff) that re-fetches on network errors, checksum mismatches, or non-archive bodies.
- Validate both on cache-hit and after each download; self-heal by purging and re-downloading anything that fails checksum or isn't a valid tar/zip.
- Add an install() backstop that purges the cache file and raises a clear, actionable error if an unrecognized archive still slips through.
- Pin sha256 for the qairt package in packages.yml (computed from a verified fresh download) so a bad body never gets committed to cache.

### Motivation and Context

- CI builds fail with "v2.47.0.260601.zip has unknown archive format" because a corrupt/partial QAIRT SDK download got written into the persistent runner cache and is reused on every run. Two gaps in package_manager.py allowed this: the qairt package had no checksum (so FileCache.fetch wrote whatever bytes came back), and cache hits were returned without any validation, so a once-poisoned cache stayed poisoned forever.

### DIFF
--- a/qcom/packages.yml
+++ b/qcom/packages.yml
@@ -207,6 +207,7 @@ python_314_windows_arm64:
 qairt:
   version: 2.47.0.260601
   url: https://softwarecenter.qualcomm.com/api/download/software/sdks/Qualcomm_AI_Runtime_Community/All/{version}/v{version}.zip
+  sha256: d3497e110eae82c35a9152a93c0a18bbede402aaf9faa7a97c8079eb0f522b01
   content_root: qairt/{version}
   bindir: bin
 qualcomm_device_cloud_sdk:
--- a/qcom/packages.yml
+++ b/qcom/packages.yml
@@ -155,18 +155,22 @@ onnx_models:
 ort_prebuilt_windows_x64: &ort_prebuilt_common
   version: 1.26.0
   url: https://github.com/microsoft/onnxruntime/releases/download/v{version}/onnxruntime-win-x64-{version}.zip
+  sha256: 6ebe99b5564bf4d029b6e93eac9ff423682b6212eade769e9ca3f685eaf500b4
   content_root: onnxruntime-win-x64-{version}
 ort_prebuilt_windows_arm64:
   <<: *ort_prebuilt_common
   url: https://github.com/microsoft/onnxruntime/releases/download/v{version}/onnxruntime-win-arm64-{version}.zip
+  sha256: 852e89621fb752b261821dda131042ed7c7fa18d8bb06768bd4eb7fa7086d87f
   content_root: onnxruntime-win-arm64-{version}
 ort_prebuilt_linux_x64:
   <<: *ort_prebuilt_common
   url: https://github.com/microsoft/onnxruntime/releases/download/v{version}/onnxruntime-linux-x64-{version}.tgz
+  sha256: 1254da24fb389cf39dc0ff3451ab48301740ffbfcbaf646849df92f80ee92c57
   content_root: onnxruntime-linux-x64-{version}
 ort_prebuilt_linux_aarch64:
   <<: *ort_prebuilt_common
   url: https://github.com/microsoft/onnxruntime/releases/download/v{version}/onnxruntime-linux-aarch64-{version}.tgz
+  sha256: 34ff1c2d0f12e2cf3d33a0c5f82e39792e1d581fbd6968fd7c30d173654be01a
   content_root: onnxruntime-linux-aarch64-{version}
 python_311_windows_x86_64: &py_x86_64_common
   version: 3.11.9
@@ -213,3 +217,4 @@ qairt:
 qualcomm_device_cloud_sdk:
   version: 0.3.0
   url: https://softwarecenter.qualcomm.com/api/download/software/tools/Qualcomm_Device_Cloud_SDK/Windows/{version}/qualcomm_device_cloud_sdk-{version}.zip
+  sha256: 80e690591a576b8bcbe0357e3a2d364a03768cc801acd15358b4b740e43a9a1b
--- a/qcom/scripts/all/package_manager.py
+++ b/qcom/scripts/all/package_manager.py
@@ -13,6 +13,7 @@ import subprocess
 import tarfile
 import tempfile
 import time
+import urllib.error
 import urllib.parse
 import urllib.request
 import zipfile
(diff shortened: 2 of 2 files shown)


tirupath-qti (approved):

qti-chuteng:
Hi @huaychou, I forgot to paste the comment yesterday, could you help review these comment and see if it is necessary to update in the following PR?

---

### 🟪 Architecture Fit

**[D-2]** `🟠 Concern` `[Arch]` — `package_manager.py:213` (branch), `:108-110`, `:55-57`; made dead by `:447`
`__parse_config` (:447) now mandates a checksum on every package, so `__fetch` always passes a non-None checksum, so `has_checksum` (:108) is always `True`, so the guard `if not has_checksum and expects_archive and not _looks_like_archive(path)` (:213) can never fire at runtime. That makes `_looks_like_archive`, `expects_archive`, and the `has_checksum` plumbing an entire sub-feature no package can reach. The motivating bug (qairt previously had no checksum) is fully fixed by the sha256 pin alone: cache-hit validation now sees a mismatch on the poisoned file, purges, and re-downloads. Evidence: the comment at :212 still says "With no checksum to rely on (e.g. the qairt SDK)" — but the same PR adds a sha256 to qairt in `packages.yml`. The comment is contradicted by its own diff.
*Better design:* Pick one contract. Since checksums are now mandatory, drop the no-checksum archive-sniff branch (and `has_checksum`, `expects_archive`, `_looks_like_archive`), letting checksum equality be the sole cache-hit validator. If defense-in-depth against a mis-pinned hash is wanted, keep the sniff but make it unconditional and drop the `not has_checksum` gate — don't leave a branch that is dead by construction. Fix the stale qairt comment at :212.

**[D-4]** `🟠 Concern` `[Arch]` — `package_manager.py:447-455` (guard), called from `:263` (`__init__`) and `:275` (`clean`)
Checksum presence is a download-integrity concern, but it's now enforced at the single chokepoint every entrypoint passes through just to read the YAML. If a future package is added without a checksum, `--print-content-dir qairt` fails even though qairt itself is fully checksummed, and `clean` (which only removes stale dirs, never downloads) fails too. The very command you'd use to recover from a bad state (`clean`) is itself blocked by the parse-time gate. Today every package has a checksum, so no current operation regresses (hence Concern), but it will surprise the next contributor who adds a legitimately-unhashable source.
*Better design:* Enforce the checksum requirement where the value is consumed — in `__fetch`/`fetch` (the download path) — or as a dedicated CI lint / unit test over `packages.yml`, not as a precondition of parsing. If parse-time fail-fast is deliberate, add a code comment making the whole-CLI coupling explicit.


### 🟩 Maintainability

**[D-6]** `🔵 Improvement` `[Maint]` — `package_manager.py:116, :123, :139, :154-161, :175, :199-203, :212`
The same `(checksums, has_checksum, expects_archive)` trio repeats in every signature in the download chain (`fetch → __download_with_retries → __download_once → __validate_file`). Worse, `has_checksum` is not independent state — it's `any(c[0] is not None for c in checksums)` (computed once at :108), carried through four functions purely to avoid recomputation. `__download_with_retries` does nothing with the tuple but hand it to `__download_once`.
*Better design:* Introduce a small `ValidationSpec` dataclass (fields `checksums`, `expects_archive`; `has_checksum` as a `@property`) built once in `fetch`, passed as one argument, with `validate(path) -> str | None` as a method. Collapses three params to one across four signatures and gives "what is valid" a single home — which is also where D-2's dead branch becomes obvious.

huaychou:
> Hi @huaychou, I forgot to paste the comment yesterday, could you help review these comment and see if it is necessary to update in the following PR?
> 
> ### 🟪 Architecture Fit
> **[D-2]** `🟠 Concern` `[Arch]` — `package_manager.py:213` (branch), `:108-110`, `:55-57`; made dead by `:447` `__parse_config` (:447) now mandates a checksum on every package, so `__fetch` always passes a non-None checksum, so `has_checksum` (:108) is always `True`, so the guard `if not has_checksum and expects_archive and not _looks_like_archive(path)` (:213) can never fire at runtime. That makes `_looks_like_archive`, `expects_archive`, and the `has_checksum` plumbing an entire sub-feature no package can reach. The motivating bug (qairt previously had no checksum) is fully fixed by the sha256 pin alone: cache-hit validation now sees a mismatch on the poisoned file, purges, and re-downloads. Evidence: the comment at :212 still says "With no checksum to rely on (e.g. the qairt SDK)" — but the same PR adds a sha256 to qairt in `packages.yml`. The comment is contradicted by its own diff. _Better design:_ Pick one contract. Since checksums are now mandatory, drop the no-checksum archive-sniff branch (and `has_checksum`, `expects_archive`, `_looks_like_archive`), letting checksum equality be the sole cache-hit validator. If defense-in-depth against a mis-pinned hash is wanted, keep the sniff but make it unconditional and drop the `not has_checksum` gate — don't leave a branch that is dead by construction. Fix the stale qairt comment at :212.
> 
> **[D-4]** `🟠 Concern` `[Arch]` — `package_manager.py:447-455` (guard), called from `:263` (`__init__`) and `:275` (`clean`) Checksum presence is a download-integrity concern, but it's now enforced at the single chokepoint every entrypoint passes through just to read the YAML. If a future package is added without a checksum, `--print-content-dir qairt` fails even though qairt itself is fully checksummed, and `clean` (which only removes stale dirs, never downloads) fails too. The very command you'd use to recover from a bad state (`clean`) is itself blocked by the parse-time gate. Today every package has a checksum, so no current operation regresses (hence Concern), but it will surprise the next contributor who adds a legitimately-unhashable source. _Better design:_ Enforce the checksum requirement where the value is consumed — in `__fetch`/`fetch` (the download path) — or as a dedicated CI lint / unit test over `packages.yml`, not as a precondition of parsing. If parse-time fail-fast is deliberate, add a code comment making the whole-CLI coupling explicit.
> 
> ### 🟩 Maintainability
> **[D-6]** `🔵 Improvement` `[Maint]` — `package_manager.py:116, :123, :139, :154-161, :175, :199-203, :212` The same `(checksums, has_checksum, expects_archive)` trio repeats in every signature in the download chain (`fetch → __download_with_retries → __download_once → __validate_file`). Worse, `has_checksum` is not independent state — it's `any(c[0] is not None for c in checksums)` (computed once at :108), carried through four functions purely to avoid recomputation. `__download_with_retries` does nothing with the tuple but hand it to `__download_once`. _Better design:_ Introduce a small `ValidationSpec` dataclass (fields `checksums`, `expects_archive`; `has_checksum` as a `@property`) built once in `fetch`, passed as one argument, with `validate(path) -> str | None` as a method. Collapses three params to one across four signatures and gives "what is valid" a single home — which is also where D-2's dead branch becomes obvious.

Addressed in https://github.com/onnxruntime/onnxruntime-qnn/pull/632

PR #625, opened 2026-07-09
[QNN EP] Support LPBQ encodings for MatMulNBits op

qti-ashimaj:
### Description
This PR adds support for the LPBQ (BLOCKWISE_EXPANSION) encoding path in the QNN EP MatMulNBits operator builder.

Route block-quantized weights to the HTP LPBQ encodings (`QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION`) when supported enabling the HTP backend to use its native low-precision block quantization kernel.

* matmulnbits_op_builder: Selects the encoding based on the weight and activation datatype and other necessary conditions for LPBQ. If selected, creates the QNN_QUANTIZATION_ENCODING_BLOCKWISE_EXPANSION encoding.
* matmulnbits_test: Adds lpbq tests for MatMulNBits op.
* qnn_ep_utils: Adds QDQ selector support for MatMulNBits when present with DQ and Q nodes in case activations are quantized so that they can be fused to become a node group.

### Motivation and Context
Previously, the QNN EP MatMulNBits builder only supported the BW_FLOAT_BLOCK path (FP16 activation + block-quantized INT4 weight). This PR adds the LPBQ path that is triggered when:

1. The activation is INT16-quantized (uint16 or int16)
2. The weight is 4-bit signed and symmetric
3. The provider option enable_block_quant_weight_optimization=1 is set


qti-chuteng:
### 🟠 Significant Concern

**[D-1]** `[Arch]` `qnn_ep_utils.cc` (selector) + `matmulnbits_op_builder.cc` + `qnn_quant_params_wrapper.cc:627`
The fuse gate and the encoding gate disagree. `OrtMatMulNBitsNodeGroupSelector::Check` runs at partition time with no access to provider options and never checks `enable_block_quant_weight_optimization`, but the builder's LPBQ encoding gate does (`Init` returns early at :627 when the option is off). The option defaults to **false**, so a 16-bit-quantized-activation + INT4-symmetric-weight MatMulNBits still fuses on the default config, then falls to the BwFloatBlock branch — feeding a 16-bit activation into a Conv2D the code documents as "only supports FP16" (`matmulnbits_op_builder.cc:567`). This is a behavior change on the default path (before, such a model would not fuse and would run the activation as float). Drive both gates from the same condition set, or decline structurally in `IsOpSupported`; add an option-OFF + 16-bit-activation negative test.

**[D-2]** `[Arch]` `matmulnbits_op_builder.cc` (`AreZeroPointsSymmetric`) + `qnn_ep_utils.cc:355` (`IsMatMulNBitsZeroPointSymmetric`)
The "are the packed zero-points symmetric for this bitwidth" decision is implemented twice (even the `1u << (bits-1)` arithmetic is duplicated), and the two copies **return opposite verdicts on the failure/absent corner**: the builder returns `true` when zp is unresolvable (→ emits LPBQ), the selector returns `false` when zp is present-but-unresolvable (→ rejects fusion) — the exact "selector fuses but builder emits the wrong encoding" hazard. They coincide today only because zero_points are always constant initializers, and nothing enforces that. Extract the bitwidth→expected-packed-byte computation into one shared helper and align the failure return; keep the layer-specific data access local.

**[D-3]** `[Arch]` `qnn_ep_utils.h:481` + `qnn_ep_utils.cc:434,441`
The new selector reuses the family's `allow_16bit_`/`allow_4bit_weight_` names but **inverts their meaning**: every sibling writes `if (!allow_16bit_ && Is16BitIntType(dt)) return false;` (flag = *permit*), while this one writes `if (allow_16bit_ && !Is16BitIntType(dt)) return false;` (flag = *require*). A maintainer writing `(false,false)` expecting the family meaning gets the check skipped. The flags are also vestigial — the only registration uses the default `(true,true)` and no call site varies them. Drop the flags and inline the hard requirements (`if (!Is16BitIntType(...)) return false;`, `if (bits != 4) return false;`); if configurability is ever needed, name them `require_16bit_act_` / `require_4bit_weight_`.

**[D-4]** `[Completeness]` `test/providers/qnn/matmulnbits_test.cc`
All four new tests are HTP-only positive LPBQ cells (symmetric zp, option ON). Untested branches: asymmetric zp → fallback, bits!=4 with 16-bit activation → no fuse, and option OFF (D-1's default high-risk path). The tests also assert E2E accuracy only, not that BLOCKWISE_EXPANSION was actually emitted, so a silent fallback that stays numerically close passes green. And the selector / `IsMatMulNBitsZeroPointSymmetric` are backend-agnostic C++ run at partition time for every backend, yet exercised only inside the arm64/linux `#if`. Add negative cells + a backend-agnostic (x86/CPU) selector test, and if feasible assert the emitted encoding.

---

### 🔵 Improvement

**[D-5]** `[Arch]` `matmulnbits_op_builder.cc:225-247` (`ProcessInputs`)
To reuse the conv LPBQ conversion, the builder fabricates a synthetic `OrtNodeUnitIODef` (scale=real initializer, zero_point=nullptr, INT4, block_size) purely to drive `Init()`'s block-quant branch, then consumes only `.quant_param`. Sound reuse, but the seam is leaky — correctness depends on knowing Init reads exactly those fields and that `zero_point=nullptr` suppresses Init's own symmetric re-check. Call `QnnQuantParamsWrapper::Init(...)` directly, or add a first-class API that builds LPBQ params from (scale, block_size, axis); at minimum comment why the fake def is needed and why zero_point is nulled.

**[D-6]** `[Maint]` `matmulnbits_op_builder.cc` + `matmul_op_builder.cc:22` + `qlinear_matmul_op_builder.cc:37` (et al.)
The new `IsQuant16bit` is byte-identical to at least two existing copies, with further inline versions in batchnorm/conv/qnn_bq_utils/qnn_model_wrapper. Pre-existing debt the PR adds a 4th+ copy to — a good moment to converge it into one shared helper in `qnn_utils.h` (sibling of the ONNX-side `Is16BitIntType`).

**[D-7]** `[Completeness]` `matmulnbits_op_builder.cc:255` (baseline) → `AreZeroPointsSymmetric`
The old GPU check compared each zp byte against a hardcoded `0b10001000` (with no `bits==4` guard); the new helper derives the expected byte from `bits`. Identical for bits=4, but a bits=8 symmetric GPU model was **rejected** before and is **accepted** now — a silent broadening of a pre-existing path. Keep an explicit `bits==4` guard on the GPU branch if it's meant to stay 4-bit-only, or document the intentional change.

**[D-8]** `[Completeness]` `docs/execution_providers/QNN-ExecutionProvider.md`
The LPBQ fusions table documents the sibling Gemm/MatMul patterns but not the new DQ(act)→MatMulNBits→Q LPBQ fusion, and the `com.microsoft.MatMulNBits` row has empty notes despite the new option-gated 16-bit-activation/INT4-symmetric HTP path. Add a MatMulNBits row mirroring the Gemm/MatMul rows and note the gating option.

qti-ashimaj:
> ### 🟠 Significant Concern
> **[D-1]** `[Arch]` `qnn_ep_utils.cc` (selector) + `matmulnbits_op_builder.cc` + `qnn_quant_params_wrapper.cc:627` The fuse gate and the encoding gate disagree. `OrtMatMulNBitsNodeGroupSelector::Check` runs at partition time with no access to provider options and never checks `enable_block_quant_weight_optimization`, but the builder's LPBQ encoding gate does (`Init` returns early at :627 when the option is off). The option defaults to **false**, so a 16-bit-quantized-activation + INT4-symmetric-weight MatMulNBits still fuses on the default config, then falls to the BwFloatBlock branch — feeding a 16-bit activation into a Conv2D the code documents as "only supports FP16" (`matmulnbits_op_builder.cc:567`). This is a behavior change on the default path (before, such a model would not fuse and would run the activation as float). Drive both gates from the same condition set, or decline structurally in `IsOpSupported`; add an option-OFF + 16-bit-activation negative test.
> 
> **[D-2]** `[Arch]` `matmulnbits_op_builder.cc` (`AreZeroPointsSymmetric`) + `qnn_ep_utils.cc:355` (`IsMatMulNBitsZeroPointSymmetric`) The "are the packed zero-points symmetric for this bitwidth" decision is implemented twice (even the `1u << (bits-1)` arithmetic is duplicated), and the two copies **return opposite verdicts on the failure/absent corner**: the builder returns `true` when zp is unresolvable (→ emits LPBQ), the selector returns `false` when zp is present-but-unresolvable (→ rejects fusion) — the exact "selector fuses but builder emits the wrong encoding" hazard. They coincide today only because zero_points are always constant initializers, and nothing enforces that. Extract the bitwidth→expected-packed-byte computation into one shared helper and align the failure return; keep the layer-specific data access local.
> 
> **[D-3]** `[Arch]` `qnn_ep_utils.h:481` + `qnn_ep_utils.cc:434,441` The new selector reuses the family's `allow_16bit_`/`allow_4bit_weight_` names but **inverts their meaning**: every sibling writes `if (!allow_16bit_ && Is16BitIntType(dt)) return false;` (flag = _permit_), while this one writes `if (allow_16bit_ && !Is16BitIntType(dt)) return false;` (flag = _require_). A maintainer writing `(false,false)` expecting the family meaning gets the check skipped. The flags are also vestigial — the only registration uses the default `(true,true)` and no call site varies them. Drop the flags and inline the hard requirements (`if (!Is16BitIntType(...)) return false;`, `if (bits != 4) return false;`); if configurability is ever needed, name them `require_16bit_act_` / `require_4bit_weight_`.
> 
> **[D-4]** `[Completeness]` `test/providers/qnn/matmulnbits_test.cc` All four new tests are HTP-only positive LPBQ cells (symmetric zp, option ON). Untested branches: asymmetric zp → fallback, bits!=4 with 16-bit activation → no fuse, and option OFF (D-1's default high-risk path). The tests also assert E2E accuracy only, not that BLOCKWISE_EXPANSION was actually emitted, so a silent fallback that stays numerically close passes green. And the selector / `IsMatMulNBitsZeroPointSymmetric` are backend-agnostic C++ run at partition time for every backend, yet exercised only inside the arm64/linux `#if`. Add negative cells + a backend-agnostic (x86/CPU) selector test, and if feasible assert the emitted encoding.
> 
> ### 🔵 Improvement
> **[D-5]** `[Arch]` `matmulnbits_op_builder.cc:225-247` (`ProcessInputs`) To reuse the conv LPBQ conversion, the builder fabricates a synthetic `OrtNodeUnitIODef` (scale=real initializer, zero_point=nullptr, INT4, block_size) purely to drive `Init()`'s block-quant branch, then consumes only `.quant_param`. Sound reuse, but the seam is leaky — correctness depends on knowing Init reads exactly those fields and that `zero_point=nullptr` suppresses Init's own symmetric re-check. Call `QnnQuantParamsWrapper::Init(...)` directly, or add a first-class API that builds LPBQ params from (scale, block_size, axis); at minimum comment why the fake def is needed and why zero_point is nulled.
> 
> **[D-6]** `[Maint]` `matmulnbits_op_builder.cc` + `matmul_op_builder.cc:22` + `qlinear_matmul_op_builder.cc:37` (et al.) The new `IsQuant16bit` is byte-identical to at least two existing copies, with further inline versions in batchnorm/conv/qnn_bq_utils/qnn_model_wrapper. Pre-existing debt the PR adds a 4th+ copy to — a good moment to converge it into one shared helper in `qnn_utils.h` (sibling of the ONNX-side `Is16BitIntType`).
> 
> **[D-7]** `[Completeness]` `matmulnbits_op_builder.cc:255` (baseline) → `AreZeroPointsSymmetric` The old GPU check compared each zp byte against a hardcoded `0b10001000` (with no `bits==4` guard); the new helper derives the expected byte from `bits`. Identical for bits=4, but a bits=8 symmetric GPU model was **rejected** before and is **accepted** now — a silent broadening of a pre-existing path. Keep an explicit `bits==4` guard on the GPU branch if it's meant to stay 4-bit-only, or document the intentional change.
> 
> **[D-8]** `[Completeness]` `docs/execution_providers/QNN-ExecutionProvider.md` The LPBQ fusions table documents the sibling Gemm/MatMul patterns but not the new DQ(act)→MatMulNBits→Q LPBQ fusion, and the `com.microsoft.MatMulNBits` row has empty notes despite the new option-gated 16-bit-activation/INT4-symmetric HTP path. Add a MatMulNBits row mirroring the Gemm/MatMul rows and note the gating option.

**[D-1]** The checks in `OrtMatMulNBitsNodeGroupSelector::Check` are temporary WAs till the support for BQ W4A16 is not merged https://github.com/onnxruntime/onnxruntime-qnn/pull/604. The QDQ selector checks will all be relaxed.

**[D-2]** Same as **[D-1]**. The symmetric zp check will be removed. Updated the unresolvable zp case to return failure for both.

**[D-3]** Removed the flags.

**[D-4]** Its difficult to assert the encodings emitted in the current test suite.

**[D-5]** Added a detailed comment on making a synthetic `OrtNodeUnitIODef`. In calling the `Init` function directly or a factory api, it will lead to duplication of code and add to difficulty in maintaining the code in future.

**[D-6]** Added a util function and removed the inline function from the existing op builders.

**[D-7]** `IsOpSupported` already checks for `bits=4` for `is_gpu_backend` at the beginning.

**[D-8]** The LPBQ fusion table documents about the patterns having lpbq encodings rather than the onnx bq encodings. Those correspond to lpbqmatmul_fusion.cc and lpbqgemm_fusion.cc files.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -707,7 +716,8 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
     const int64_t qnn_axis = 1 - axis;
 
     *this = QnnQuantParamsWrapper::LowPowerBlockwise(per_channel_scales, per_block_int_scales, lpbq_offsets,
-                                                     qnn_axis, /*block_scale_bitwidth=*/is_int4_type ? 4 : 8);
+                                                     qnn_axis, /*block_scale_bitwidth=*/4);  // LPBQ conversion only supports INT4;
+                                                                                             // guarded by the check at the start of this block
```
qti-kromero, line 721: which check is this referring to?

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -338,6 +338,62 @@ bool IsQOrDQScalePositiveConstantScalar(const OrtGraph* graph, const OrtApi& ort
...
+
+  std::vector<const OrtValueInfo*> inputs(num_inputs);
+  if (ort_api.Node_GetInputs(node, inputs.data(), inputs.size()) != nullptr) {
+    return false;
+  }
+
+  // inputs[3] is the optional zero_points initializer.
+  const OrtValue* zp_init = GetInitializerFromValueInfo(graph, ort_api, inputs[3]);
+  if (zp_init == nullptr) {
+    return false;
```
qti-kromero: how do we get to this condition? zp is not from initializer?

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -338,6 +338,62 @@ bool IsQOrDQScalePositiveConstantScalar(const OrtGraph* graph, const OrtApi& ort
...
+  }
+
+  uint8_t* zp_data = nullptr;
+  if (ort_api.GetTensorMutableData(const_cast<OrtValue*>(zp_init), (void**)&zp_data) != nullptr ||
+      zp_data == nullptr || num_elems == 0) {
+    ort_api.ReleaseTensorTypeAndShapeInfo(shape_info);
+    return false;
+  }
+
+  // For INT4 (bits=4): symmetric ZP is 8, packed two per byte as 8|(8<<4)=0b10001000=136.
```
qti-kromero: should this be calling AreZeroPointsSymmetric? it seems like there might be some opportunity for reuse there but haven't looked to closely

onnxruntime/test/providers/qnn/matmulnbits_test.cc
```diff
@@ -364,6 +385,55 @@ TEST_F(QnnHTPBackendTests, MatMulNBits_M1_N2_K64_B8_BS64_ZP) {
   params.has_zero_point = true;
   RunMatMulNBitsTest<8>(params, "htp");
 }
+
```
qti-kromero, line 522: might be good to have some tests for invalid operators which trigger failure branches above

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
quic-calvnguy: Looks like zp_init can be nullptr if inputs[3] is not a valid input tensor name:

https://github.com/onnxruntime/onnxruntime-qnn/blob/15ca4e80282068495151c13c0969827172d5ce9a/onnxruntime/core/providers/qnn/qnn_ep_utils.cc#L106-L113
https://github.com/onnxruntime/onnxruntime-qnn/blob/15ca4e80282068495151c13c0969827172d5ce9a/onnxruntime/core/providers/qnn/qnn_ep_utils.cc#L243-L279

Line 360 looks like it is retrieving the initializer for the zp which will later be used to init the zp (line 377)

quic-calvnguy (approved):
LGTM beyond the negative testing Kyle suggested

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: https://github.com/onnxruntime/onnxruntime-qnn/blob/0039c77be33c3f9f1a55654f3835d520f3ebbe33/onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc#L632-L637

LPBQ is only supported for INT4 weights

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
qti-ashimaj: Yes, if inputs[3] or zp is present, it should come from a constant initializer and we should be able to read it. If not, we cannot check for symmetric value and hence return false

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
qti-ashimaj: This check has been introduced temporarily as a WA till the PR https://github.com/onnxruntime/onnxruntime-qnn/pull/675 does not get merged. Since BQ only supports FP16 activations in mainline, we are adding additional checks to fuse QDQ nodes only when we have 4-bit symmetric weights and 16-bit activations.

onnxruntime/test/providers/qnn/matmulnbits_test.cc
qti-ashimaj: When we add negative tests (eg., bits !=4, asymmetric zp etc.), the QDQ nodes will not get fused and automatically fallback to the `BW_FLOAT_BLOCK` encodings and hence will run on HTP. `ExpectedEPNodeAssignment::All` will still be valid and the TCs will pass. There isn't a straightforward way in the test suite to check whether it was the LPBQ or the BQ float encodings that got populated.

huaychou (commented):
Please update the PR on the latest mainline since https://github.com/onnxruntime/onnxruntime-qnn/pull/675 is merged.

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
```diff
@@ -649,29 +649,38 @@ Ort::Status QnnQuantParamsWrapper::Init(const QnnModelWrapper& qnn_model_wrapper
...
+                  "Only axis 0 or 1 is supported for block quantization LPBQ conversion");
+
     // Get scale tensor shape to determine block/channel dimensions.
     // Scale tensor may be rank 2 (e.g., MatMul/Gemm) or higher rank (e.g., Conv with rank-4 weights).
-    // Only the first two dimensions (indexed by onnx_axis and 1 - onnx_axis) are used for LPBQ conversion.
+    // Only the first two dimensions (indexed by axis and 1 - axis) are used for LPBQ conversion.
     const std::vector<int64_t> scale_shape =
         utils::GetInitializerShape(ort_quant_params->scale, qnn_model_wrapper.GetOrtApi());
-    RETURN_IF_NOT(scale_shape.size() >= 2 && scale_shape.size() <= 4,
+    // MatMulNBits can sometimes produce rank-1 scales; other ops (MatMul, Conv, Gemm) always use rank-2+.
```
huaychou, line 664: According to https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.MatMulNBits, I think MatMulNBits shouldn't produce rank-1 scales.

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -134,6 +134,39 @@ void UnpackDataToDatatype(const std::vector<uint8_t>& packed_data,
     }
   }
 }
+
+// Returns true if all packed zero_points are symmetric (i.e., each element equals 2^(bits-1)).
+// MatMulNBits stores zero_points as packed sub-byte integers in uint8 bytes
+// (e.g., two 4-bit values per byte, four 2-bit values per byte).
+// The expected symmetric value per byte is computed by packing (kByteBits/bits) copies of 2^(bits-1).
+bool AreZeroPointsSymmetric(QnnModelWrapper& qnn_model_wrapper, const std::string& zp_tensor_name,
```
huaychou: Can we move this to qnn_utils?

onnxruntime/core/providers/qnn/builder/qnn_quant_params_wrapper.cc
qti-ashimaj: Yes, as per the official docs, scales should be rank-2. But we do have models having rank-1 scales.

Transformer block
image

LM head from MSFT model
image

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
qti-ashimaj: Done

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
```diff
@@ -464,26 +421,112 @@ Ort::Status MatMulNBitsOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapp
...
+            RETURN_IF_ERROR(bq::AddInt16ToFp16DequantForActivation(qnn_model_wrapper,
+                                                                   input_names[0],
+                                                                   fp16_act_name,
+                                                                   do_op_validation,
+                                                                   "MatMulNBits"));
+            input_names[0] = fp16_act_name;
+          }
+        }
       }
+      // 2.6 Create and register the weight tensor wrapper.
```
huaychou: nit: 2.6 is used in L510

onnxruntime/core/providers/qnn/builder/qnn_utils.h
```diff
@@ -890,6 +890,14 @@ Ort::Status DequantizeInt32BiasToFp16(gsl::span<const uint8_t> raw_int32_bytes,
                                       gsl::span<const float> scales,
                                       std::vector<uint8_t>& fp16_bytes);
 
+// Returns true if all packed zero_points in the given initializer tensor are symmetric
+// (i.e., each sub-byte element equals 2^(bits-1)).
+// MatMulNBits stores zero_points as packed sub-byte integers in uint8 bytes
+// (e.g., two 4-bit values per byte, four 2-bit values per byte).
+bool AreZeroPointsSymmetric(QnnModelWrapper& qnn_model_wrapper,
```
huaychou: nit: can be AreZeroPointsSymmetricConstant

huaychou (approved):

onnxruntime/test/providers/qnn/matmulnbits_test.cc
qti-ashimaj: Negative TCs have been added which will trigger `BW_FLOAT_BLOCK` encodings but still run on HTP

onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc
qti-ashimaj: Done

onnxruntime/core/providers/qnn/builder/qnn_utils.h
qti-ashimaj: Updated

qti-ashwshan:
LGTM. Updated the branch for CI to run.

qti-ashwshan (approved):

PR #626, opened 2026-07-09
[QNN EP] Add SkipSimplifiedLayerNormalization op builder

qti-ashwshan:
### Description
- Decomposes com.microsoft.SkipSimplifiedLayerNormalization into QNN_OP_ELEMENT_WISE_BINARY(ADD) + QNN_OP_RMS_NORM. Supports 2D/3D inputs with optional bias. Includes float HTP unit tests.


onnxruntime/test/providers/qnn/skip_simplified_layernormalization_test.cc
qti-mattsinc, line 1: Can you add GPU tests here? I think the GPU backend should support this op decomposition, so it would be nice to include it in the tests

onnxruntime/core/providers/qnn/builder/opbuilder/skip_simplified_layernormalization_op_builder.cc
```diff
@@ -0,0 +1,231 @@
...
+
+#include "core/providers/qnn/builder/op_builder_factory.h"
+#include "core/providers/qnn/builder/opbuilder/base_op_builder.h"
+#include "core/providers/qnn/builder/qnn_model_wrapper.h"
+#include "core/providers/qnn/builder/qnn_utils.h"
+
+namespace onnxruntime {
+namespace qnn {
+
+// Handles com.microsoft.SkipSimplifiedLayerNormalization.
```
yath1, line 12: I am curious, why do we want to add support for this op? Its not an official onnx op

qti-ashwshan: These ops are part of a lot of LLM models in the oga path.

onnxruntime/test/providers/qnn/skip_simplified_layernormalization_test.cc
qti-ashwshan: Yes. I am planning on adding GPU coverage as well

onnxruntime/core/providers/qnn/builder/opbuilder/skip_simplified_layernormalization_op_builder.cc
yath1: ONNX does not guarantee any backward compatibility, and the op support can be changed for these ops. This would lead us into a slippery slope. Are you not able to represent the op using onnx ops in the default domain?

tirupath-qti: Thanks @yath1 for highlighting the risk with Contrib operators. But there are some important operators widely present in LLM models and failing to support them leads to CPU fallbacks which further impact the overall perf. So, we are carefully adding support for important Contrib operators.

yath1: There's actually a huge list of contrib ops. I am afraid if we start introducing them, I think we would end up getting requests for multiple ops from non-default domains. Moreover, this PR is actually decomposing the op into ops in default domain. Could we just do that as a preprocessing step before taking the model through ORT? Also I am not sure how there can be any perf issue in this case because we are essentially decomposing the op into standard ops. But in the future, if we end up getting a custom op for SkipSimplifiedLayerNormalization in HTP, we can add a fusion pass in EP which would again give us same perf on device.

qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🟠 Concern` `[Arch]` — `qnn_ep_utils.cc:L1745`
The new builder carries a complete quantized path: `EmitAddNode` copies `output_info.quant_param` onto the intermediate sum tensor (`skip_simplified_layernormalization_op_builder.cc:143-145`), and the beta-dummy block has an `IsQuantized()` branch producing `PerTensor(1,0)` (`:230-233`). That path only executes when a DQ→SkipSimplifiedLayerNormalization→Q subgraph is selected as a QDQ node unit — but the `rmsnorm_ops` selector (`qnn_ep_utils.cc:1745`) registers only `RMSNormalization` and `SimplifiedLayerNormalization`; the Skip variant is absent (grep confirms count = 0). So a quantized model's SkipSimplifiedLayerNorm is never grouped, never reaches this builder as a quantized unit, and silently falls back to CPU. This is the exact PR #338 ⟦C4.2⟧ shape. All six tests are float, so the contradiction is undetectable in CI.
*Better design:* Decide the scope. In scope → add `{"SkipSimplifiedLayerNormalization", {}}` to `rmsnorm_ops` + QDQ (u8/u16) test + fix D-3. Float-only → strip the `quant_param.Copy()`/`IsQuantized()`/`PerTensor` branches and note it. Either way, code and registration must agree.

**[D-3]** `🔵 Improvement` `[Arch]` — `skip_simplified_layernormalization_op_builder.cc:L143`
The intermediate sum tensor is built to "inherit shape/dtype/quant from input[0]". For a quantized graph, `sum = input + skip` has roughly double the dynamic range of `input`; encoding it with input[0]'s scale/zero-point clips/saturates the sum, silently corrupting output — the ⟦C4.1⟧ / PR #417 shape. The second Add (sum + bias) compounds it. The bug is dormant today (D-1 makes the quant path unreachable, so on float `quant_param` is empty and the copy is a no-op), but fixing D-1 without fixing this ships quantized corruption.
*Better design:* Let QNN derive the intermediate's encoding, or use the QDQ node-group's boundary qparams. When `expose_sum` names the sum as graph output[3], its qparam must come from the ONNX output[3]'s Q, not input[0].

### 🟩 Maintainability

**[D-2]** `🟠 Concern` `[Maint]` — `skip_simplified_layernormalization_op_builder.cc:L219`
The RMSNorm half of this decomposition — the SDK<2.49 beta-dummy block (`:219-245`), the epsilon scalar (`:259`), and the axes param (`:263-274`) — is copied almost verbatim from `rmsnormalization_op_builder.cc`. There is now no single owner for "how QNN-EP emits an RMSNorm on this SDK"; when the beta-dummy contract moves, a maintainer must edit two files that don't reference each other (change amplification). The copy is *already* drifting — it dropped the sibling's load-bearing rationale comment (why `QNN_SDK_VERSION_MINOR` is used because SDK 2.47/2.48 share API 2.36) and the `#else` VERBOSE-log branch.
*Better design:* Extract the beta-dummy emission (taking a scale `TensorInfo` + `input_names`) into a shared static helper both builders call; likewise factor the epsilon+axes emission. Reusing the whole class is legitimately impossible (SkipSLN feeds the intermediate Add output as the RMSNorm input), but these sub-blocks are self-contained free functions that extract trivially.

### 🔵 Completeness

**[D-4]** `🔵 Improvement` `[Completeness]` — `skip_simplified_layernormalization_test.cc`
Tests only have `QnnHTPBackendTests` (guarded `__aarch64__ || _M_ARM64 || __linux__`) and `QnnGPUBackendTests` (`_M_ARM64`). There is no unguarded `QnnCPUBackendTests` cell, but the direct sibling `rmsnormalization_test.cc` has one. SkipSimplifiedLayerNorm's own logic (Add-decomposition, output[3] passthrough, training-output rejection) has zero coverage on an x86 dev host.
*Better design:* Add an unguarded `QnnCPUBackendTests` float cell mirroring the sibling's CPU cases.

**[D-5]** `🔵 Improvement` `[Completeness]` — `docs/execution_providers/QNN-ExecutionProvider.md:L542`
The QNN-EP supported-operators table lists `com.microsoft:SimplifiedLayerNormalization` (line 542) but not `com.microsoft:SkipSimplifiedLayerNormalization`. A newly enabled contrib op should appear in the user-facing list with its constraints noted (rank ≤ 3, no training outputs).
*Better design:* Add a `com.microsoft:SkipSimplifiedLayerNormalization` row near the SimplifiedLayerNormalization entry.

tirupath-qti:
> ### 🟪 Architecture Fit
> **[D-1]** `🟠 Concern` `[Arch]` — `qnn_ep_utils.cc:L1745` The new builder carries a complete quantized path: `EmitAddNode` copies `output_info.quant_param` onto the intermediate sum tensor (`skip_simplified_layernormalization_op_builder.cc:143-145`), and the beta-dummy block has an `IsQuantized()` branch producing `PerTensor(1,0)` (`:230-233`). That path only executes when a DQ→SkipSimplifiedLayerNormalization→Q subgraph is selected as a QDQ node unit — but the `rmsnorm_ops` selector (`qnn_ep_utils.cc:1745`) registers only `RMSNormalization` and `SimplifiedLayerNormalization`; the Skip variant is absent (grep confirms count = 0). So a quantized model's SkipSimplifiedLayerNorm is never grouped, never reaches this builder as a quantized unit, and silently falls back to CPU. This is the exact PR #338 ⟦C4.2⟧ shape. All six tests are float, so the contradiction is undetectable in CI. _Better design:_ Decide the scope. In scope → add `{"SkipSimplifiedLayerNormalization", {}}` to `rmsnorm_ops` + QDQ (u8/u16) test + fix D-3. Float-only → strip the `quant_param.Copy()`/`IsQuantized()`/`PerTensor` branches and note it. Either way, code and registration must agree.
> 
> **[D-3]** `🔵 Improvement` `[Arch]` — `skip_simplified_layernormalization_op_builder.cc:L143` The intermediate sum tensor is built to "inherit shape/dtype/quant from input[0]". For a quantized graph, `sum = input + skip` has roughly double the dynamic range of `input`; encoding it with input[0]'s scale/zero-point clips/saturates the sum, silently corrupting output — the ⟦C4.1⟧ / PR #417 shape. The second Add (sum + bias) compounds it. The bug is dormant today (D-1 makes the quant path unreachable, so on float `quant_param` is empty and the copy is a no-op), but fixing D-1 without fixing this ships quantized corruption. _Better design:_ Let QNN derive the intermediate's encoding, or use the QDQ node-group's boundary qparams. When `expose_sum` names the sum as graph output[3], its qparam must come from the ONNX output[3]'s Q, not input[0].
> 
> ### 🟩 Maintainability
> **[D-2]** `🟠 Concern` `[Maint]` — `skip_simplified_layernormalization_op_builder.cc:L219` The RMSNorm half of this decomposition — the SDK<2.49 beta-dummy block (`:219-245`), the epsilon scalar (`:259`), and the axes param (`:263-274`) — is copied almost verbatim from `rmsnormalization_op_builder.cc`. There is now no single owner for "how QNN-EP emits an RMSNorm on this SDK"; when the beta-dummy contract moves, a maintainer must edit two files that don't reference each other (change amplification). The copy is _already_ drifting — it dropped the sibling's load-bearing rationale comment (why `QNN_SDK_VERSION_MINOR` is used because SDK 2.47/2.48 share API 2.36) and the `#else` VERBOSE-log branch. _Better design:_ Extract the beta-dummy emission (taking a scale `TensorInfo` + `input_names`) into a shared static helper both builders call; likewise factor the epsilon+axes emission. Reusing the whole class is legitimately impossible (SkipSLN feeds the intermediate Add output as the RMSNorm input), but these sub-blocks are self-contained free functions that extract trivially.
> 
> ### 🔵 Completeness
> **[D-4]** `🔵 Improvement` `[Completeness]` — `skip_simplified_layernormalization_test.cc` Tests only have `QnnHTPBackendTests` (guarded `__aarch64__ || _M_ARM64 || __linux__`) and `QnnGPUBackendTests` (`_M_ARM64`). There is no unguarded `QnnCPUBackendTests` cell, but the direct sibling `rmsnormalization_test.cc` has one. SkipSimplifiedLayerNorm's own logic (Add-decomposition, output[3] passthrough, training-output rejection) has zero coverage on an x86 dev host. _Better design:_ Add an unguarded `QnnCPUBackendTests` float cell mirroring the sibling's CPU cases.
> 
> **[D-5]** `🔵 Improvement` `[Completeness]` — `docs/execution_providers/QNN-ExecutionProvider.md:L542` The QNN-EP supported-operators table lists `com.microsoft:SimplifiedLayerNormalization` (line 542) but not `com.microsoft:SkipSimplifiedLayerNormalization`. A newly enabled contrib op should appear in the user-facing list with its constraints noted (rank ≤ 3, no training outputs). _Better design:_ Add a `com.microsoft:SkipSimplifiedLayerNormalization` row near the SimplifiedLayerNormalization entry.

@qti-chuteng ,
Regarding D3, the QNN EP should not calculate encodings for decomposed non data mover operators. If PR #417  is doing any such Quant param calculation, it is incorrect.

PR #629, opened 2026-07-10
[QNN EP] Fuse RTR-only SpaceToDepth after Layout Transform

yuhuchua-qti:
## Summary
- A bare Reshape→Transpose→Reshape (CRD-mode `SpaceToDepth` with no adjacent Conv/Transpose) was never fused into a QNN `SpaceToDepth`. The `node_count==3` gate in `spacetodepth_fusion.cc` unconditionally skipped RTR-only in every `GetCapability` pass to protect the 5-node Conv→RTR→Conv pattern, leaving standalone-RTR models (e.g. a customer denoise model) unfused.
- Restrict the skip to the **1st (pre-Layout-Transform)** `GetCapability` pass only; allow RTR-only fusion in the **2nd (post-LT)** pass, where the absence of wrapping NCHW↔NHWC Transposes makes fusion safe (`CreateOrValidateOnQnn` adds its own pre/post Transpose for the NHWC S2D).
- Plumb `QnnModelWrapper::IsPostLayoutTransform()` from a per-session `get_capability_call_count_` counter (`>1 == post-LT`); `ComposeGraph` sets it `true` (Compile always runs post-LT).
- Add `SpaceToDepthFusion_BareRTR_Float_CRD` HTP test; `DISABLED_SpaceToDepthFusion_BareRTR_Float_DCR` skipped for AISW-175353 (same DCR-kernel accuracy issue that already blocks `DISABLED_SpaceToDepthFusion_Float_DCR`).
- Add VERBOSE `GetCapability pass #N (post-LT=…)` log line so silent regressions (e.g. ORT eliding the 2nd pass) are visible.

## Test plan
- [x] `SpaceToDepthFusion_BareRTR_Float_CRD` on QnnHTP (x86 host — placement + DLC-dump; on-device numeric TBD)
- [x] `DISABLED_SpaceToDepthFusion_BareRTR_Float_DCR` on QnnHTP (blocked by AISW-175353 — same kernel accuracy issue as pre-existing `DISABLED_SpaceToDepthFusion_Float_DCR` and `DISABLED_SpaceToDepthFusion_UnequalBlockSize_DCR`)
- [x] Existing `SpaceToDepthFusion_Wrapped4Node_*` / Conv→RTR→Conv cases — no regression (byte-identical DLC vs main)
- [x] x86_64 build + DLC dump: customer denoise model now emits `SpaceToDepth` (via `_spacetodepth_nhwc_{pre,s2d,post}` path); full QNN placement, zero CPU fallback
- [x] ARM64 on-device numeric verification (pending)

### Notes on the BareRTR test harness
The `BuildBareRTRSpaceToDepthTestCase` helper emits a side-branch 1x1 Conv (unrelated to the RTR chain) alongside R→T→R. This is a **test-harness workaround** for a pre-existing ORT limitation: a graph with no layout-sensitive ops causes pass 1 to return empty capabilities (individual R/T/R fail on HTP because rank-6 tensors are unsupported), so ORT never issues the 2nd `GetCapability` pass, and the post-LT-gated bare-RTR fusion is unreachable. Real customer models (e.g. `denoise_HD`) with `Conv`/`MaxPool` elsewhere in the graph are unaffected. Same limitation existed on `main`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model.cc
@@ -297,13 +297,7 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
                                                       *context.model_settings,
                                                       context.tensor_name_overrides,
                                                       trace_collector.get(),
-                                                      // ComposeGraph always runs during Compile, which is
-                                                      // strictly after ORT's Layout Transformer. RTR-only
-                                                      // SpaceToDepth fusion claimed in the 2nd GetCapability
-                                                      // pass must also be built here, so this wrapper reports
-                                                      // post-layout-transform = true (see
-                                                      // QnnModelWrapper::IsPostLayoutTransform() and
-                                                      // spacetodepth_fusion.cc's node_count==3 gate).
+                                                      // Compile runs strictly after Layout Transformer.
                                                       /*is_post_layout_transform=*/true);
 
   qnn::profile::ProfilingInfo profiling_info;
--- a/onnxruntime/core/providers/qnn/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model.cc
@@ -297,7 +297,6 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
                                                       *context.model_settings,
                                                       context.tensor_name_overrides,
                                                       trace_collector.get(),
-                                                      // Compile runs strictly after Layout Transformer.
                                                       /*is_post_layout_transform=*/true);
 
   qnn::profile::ProfilingInfo profiling_info;
--- a/onnxruntime/core/providers/qnn/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model.cc
@@ -296,7 +296,15 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
                                                       qnn_backend_manager_->GetQnnBackendType(),
                                                       *context.model_settings,
                                                       context.tensor_name_overrides,
-                                                      trace_collector.get());
+                                                      trace_collector.get(),
+                                                      // ComposeGraph always runs during Compile, which is
+                                                      // strictly after ORT's Layout Transformer. RTR-only
+                                                      // SpaceToDepth fusion claimed in the 2nd GetCapability
+                                                      // pass must also be built here, so this wrapper reports
+                                                      // post-layout-transform = true (see
+                                                      // QnnModelWrapper::IsPostLayoutTransform() and
+                                                      // spacetodepth_fusion.cc's node_count==3 gate).
+                                                      /*is_post_layout_transform=*/true);
 
   qnn::profile::ProfilingInfo profiling_info;
 #ifdef QNN_SYSTEM_PROFILE_API_ENABLED
--- a/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
@@ -620,6 +620,7 @@ class QnnModelWrapper {
   // Null when tracing is disabled.
   OpTraceCollector* op_trace_collector_ = nullptr;
 
+  // A flag for model wrapper users (e.g., op builders, node group fusions) to know whether pre- or post-layout transform.
   bool is_post_layout_transform_ = false;
 };  // QnnModelWrapper
 
--- a/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
@@ -69,11 +69,11 @@ class QnnModelWrapper {
         graph_inputs_(graph_inputs),
         graph_outputs_(graph_outputs),
         qnn_backend_type_(qnn_backend_type),
-        is_post_layout_transform_(is_post_layout_transform),
         model_settings_(model_settings),
         api_ptrs_(ApiPtrs{api_ptrs.ort_api, api_ptrs.ep_api, api_ptrs.model_editor_api}),
         tensor_name_overrides_(tensor_name_overrides),
-        op_trace_collector_(op_trace_collector) {
+        op_trace_collector_(op_trace_collector),
+        is_post_layout_transform_(is_post_layout_transform) {
     // Invariant: validator interface and handle must both be set or both be null.
     // They are populated together by QnnBackendManager::LoadQnnSerializerBackend() (QnnIr flow).
     assert((validator_backend_handle == nullptr) ==
@@ -389,9 +389,6 @@ class QnnModelWrapper {
 
   QnnBackendType GetQnnBackendType() const { return qnn_backend_type_; }
 
-  /// True if this GetCapability call is post-Layout-Transform. Derived from a pass
-  /// counter (count > 1), so it's a conservative proxy: never falsely post-LT, but
-  /// under-reports when ORT skips the 2nd pass (layout-neutral graph, empty 1st pass).
   bool IsPostLayoutTransform() const { return is_post_layout_transform_; }
 
   const OrtGraph& GetOrtGraph() const { return ort_graph_; }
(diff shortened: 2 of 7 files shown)


qti-shubham (approved):
I liked the implementation. 
Once this PR merges in : https://github.com/onnx/onnx/pull/7848 , we should remove all 3 cases and do Node Group Fusion at Pre-optimization phase.

yuhuchua-qti:
> I liked the implementation. Once this PR merges in : [onnx/onnx#7848](https://github.com/onnx/onnx/pull/7848) , we should remove all 3 cases and do Node Group Fusion at Pre-optimization phase.

Agreed. It's better to remove all 3 cases once Onnx S2D has mode attribute.
Thanks.

onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc
```diff
@@ -100,6 +100,8 @@ static std::unordered_map<std::string, std::vector<FusionFunc>> fusions = {
     {"Erf", {GeluFusion::TryFusion}},
     {"ReduceMean", {LayerNormFusion::TryFusion}},
     {"Einsum", {ReshapeEinsumReshapeNodeGroup::TryFusion}},
+    // Both match rank-6 Reshape->Transpose->Reshape but are disjoint by construction:
+    // Rank6ToRank5 requires the last dim unchanged, whereas SpaceToDepth changes it.
```
minfhong-qti: Not sure the motivation and benefit of adding this comment here. Frankly speaking, this comment does not provide any useful info of the difference between two fusions.
Since I suppose all developers understand the order matters here, please re-consider whether the comment is beneficial.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -296,7 +296,9 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
                                                       qnn_backend_manager_->GetQnnBackendType(),
                                                       *context.model_settings,
                                                       context.tensor_name_overrides,
-                                                      trace_collector.get());
+                                                      trace_collector.get(),
+                                                      // Compile runs strictly after Layout Transformer.
```
minfhong-qti: Nit. The comment is redundant.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -68,6 +69,7 @@ class QnnModelWrapper {
         graph_inputs_(graph_inputs),
         graph_outputs_(graph_outputs),
         qnn_backend_type_(qnn_backend_type),
+        is_post_layout_transform_(is_post_layout_transform),
```
minfhong-qti: Prefer to move to last.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -387,6 +389,11 @@ class QnnModelWrapper {
 
   QnnBackendType GetQnnBackendType() const { return qnn_backend_type_; }
 
+  /// True if this GetCapability call is post-Layout-Transform. Derived from a pass
+  /// counter (count > 1), so it's a conservative proxy: never falsely post-LT, but
+  /// under-reports when ORT skips the 2nd pass (layout-neutral graph, empty 1st pass).
```
minfhong-qti: I think what comment describes here is irrelevant to this class. Prefer to remove it.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1898,6 +1901,16 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
     return nullptr;
   }
 
+  // See get_capability_call_count_ field doc. Empty passes still count.
+  ++ep->get_capability_call_count_;
```
minfhong-qti: Should it be after the below number of nodes check?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1520,7 +1520,10 @@ OrtStatus* QnnEp::GetSupportedNodes(const OrtGraph* graph,
                                                 model_outputs,
                                                 qnn_backend_manager_->GetQnnBackendType(),
                                                 model_settings_,
-                                                &tensor_name_overrides_);
+                                                &tensor_name_overrides_,
+                                                /*op_trace_collector=*/nullptr,
+                                                /*is_post_layout_transform=*/
+                                                get_capability_call_count_ > 1);
```
```suggestion
                                                /*is_post_layout_transform=*/get_capability_call_count_ > 1);
```
Nit.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -293,6 +293,12 @@ class QnnEp : public OrtEp, public ApiPtrs {
   // times (e.g. initial pass + EPContext re-pass).
   bool hnrd_warning_emitted_ = false;
 
+  // Counts top-level GetCapability passes. NHWC-preferring EPs get two passes
+  // (pre- then post-LayoutTransformer). Post-LT-only fusions (e.g. RTR-only
+  // SpaceToDepth) read this via QnnModelWrapper::IsPostLayoutTransform() (count > 1).
+  // Incremented in GetCapabilityImpl for top-level graphs only (parent_node == nullptr).
+  size_t get_capability_call_count_ = 0;
```
minfhong-qti: How about make it a bool type and use `gsl::finally` to automatically flipped while leaving?

minfhong-qti (commented):
General comment. Please carefully review the added comments as I found out Claude easily adds excessive but irrelevant/redundant ones.

onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc
yuhuchua-qti: Agreed — removed.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
yuhuchua-qti: Removed — the argument name /*is_post_layout_transform=*/true is already self-documenting.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
yuhuchua-qti: Done — moved to last.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
yuhuchua-qti: Moved — the state mutation (now a gsl::finally marker per your #7 suggestion) is placed after the num_nodes == 0 gate. Empty-graph early-return no longer touches the flag.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
yuhuchua-qti: Applied. Thanks.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
yuhuchua-qti: Adopted. 
size_t get_capability_call_count_ → bool is_post_layout_transform_, flipped via gsl::finally on GetCapabilityImpl exit (placed after the num_nodes == 0 gate — see #5). 
GetSupportedNodes reads the bool directly instead of computing count > 1. 
The pass #N verbose log became "GetCapability pre-LT/post-LT" — the pass-number granularity wasn't being consumed anywhere.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -615,6 +619,8 @@ class QnnModelWrapper {
   // QnnModel::ComposeGraph (stack-allocated unique_ptr).
   // Null when tracing is disabled.
   OpTraceCollector* op_trace_collector_ = nullptr;
+
+  bool is_post_layout_transform_ = false;
```
minfhong-qti, line 624:
```suggestion
  // A flag for model wrapper users (e.g., op builders, node group fusions) to know whether pre- or post-layout transform.
  bool is_post_layout_transform_ = false;
```
Nit.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -293,6 +293,9 @@ class QnnEp : public OrtEp, public ApiPtrs {
   // times (e.g. initial pass + EPContext re-pass).
   bool hnrd_warning_emitted_ = false;
 
+  // Flipped to true on first GetCapability exit via gsl::finally in GetCapabilityImpl.
```
minfhong-qti:
```suggestion
  // A flag to indicate whether current GetCapability call is after layout transform, flipped to true once exiting the first call.
```
Nit.

qti-chuteng (approved):

minfhong-qti (approved):

PR #107, opened 2026-03-04
Add share_resource_optimization option

qti-chuteng:
- Add new 'share_resource_optimization' option to enable SEQUENTIAL_WITHOUT_VA_OPTIMIZATION
- Deprecate 'enable_vtcm_backup_buffer_sharing' with warning message
- Unify both options at earliest entry point for consistent behavior
- Maintain backward compatibility: either option set to 1 enables both
- Update all usage points: API checks, context management, SetupBackend calls

PR #630, opened 2026-07-12
[QNN-EP] Set the default to unroll LSTM at ORT

chunghow-qti:
### Description

Set the default to unroll LSTM  at ORT

### Changes
- Rename session option `disable_htp_monolithic_lstm` → `enable_htp_monolithic_lstm` and invert its default so LSTM defaults to unrolling per-timestep `QNN_OP_LSTM` cells (+ `Pack`) at the QNN EP layer, with the HTP-native monolithic LSTM kernel now opt-in.
- Update `lstm_op_builder.cc` to branch the lowering (unrolled vs. monolithic) on the new option, including backend gating (HTP/IR use the flag; CPU/GPU always unroll) and the `QNN_HTP_GRAPH_CONFIG_OPTION_MONOLITHIC_LSTM` graph-config toggle.
- Update `cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch` with new option name
- Re-enable 6 previously `DISABLED_` optional-output LSTM tests. They were disabled pending a fix for [microsoft/onnxruntime#27301](https://github.com/microsoft/onnxruntime/pull/27301) (illegal memory access in `GetOutputIndex` with optional outputs). That PR itself wasn't merged, but the same issue was fixed by [microsoft/onnxruntime#27644](https://github.com/microsoft/onnxruntime/pull/27644) (merged 2026-03-16), which is included in the `ort_core` version this repo currently pins (`v1.26.0`). The fix is present, so these tests can be safely re-enabled.
- Restore unrolled-path coverage in `lstm_test.cc`: the non-suffixed FP16 sanity tests (reverse/bidirectional/optional-input cases) had been left exercising the monolithic path and now default back to the unrolled path (matching their name and the new default), removing a duplicate `_monolithic_lstm`-equivalent test in the process. Also add a `layout==1` CPU tripwire test (`LSTM_FP32_layout1_forward`) asserting the LSTM correctly falls back off QNN EP.
- Update `QNN-ExecutionProvider.md` to document the renamed option, including a note for the trade off between monolithic  lstm on/off.

### Motivation and Context

There are 3 kinds of LSTM use cases:
1. Unroll LSTM at ORT (enable_htp_monolithic_lstm == false)
2. Unroll LSTM at HTP (enable_htp_monolithic_lstm == false)
3. Run Monolithic LSTM on HTP (enable_htp_monolithic_lstm == true)

When `enable_htp_monolithic_lstm == false`. since there is an inference time regression with "Unroll LSTM at HTP" compared to "Unroll LSTM at ORT", set the default to "Unroll LSTM at ORT".

### DIFF
--- a/cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
+++ b/cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
@@ -16,7 +16,7 @@ index f08b974d60..a50d771234 100644
              ORT_THROW("Wrong value for htp_arch. select from: " + str);
            }
 -        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization" || key == "extended_udma") {
-+        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization" || key == "extended_udma" || key == "disable_htp_monolithic_lstm") {
++        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization" || key == "extended_udma" || key == "enable_htp_monolithic_lstm") {
            std::unordered_set<std::string> supported_options = {"0", "1"};
            if (supported_options.find(value) == supported_options.end()) {
              std::ostringstream str_stream;
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -205,10 +205,12 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
 |'0'|Default. Disabled. Inference with fp32 precision if it's fp32 model.|
 |'1'|Enable the float32 model to be inferenced with fp16 precision.|
 
-|`"disable_htp_monolithic_lstm"`|Description|
+|`"enable_htp_monolithic_lstm"`|Description|
 |---|---|
-|'0'|Default. HTP uses the monolithic LSTM graph configuration.|
-|'1'|Disable the monolithic LSTM graph configuration on HTP.|
+|'0'|Default. LSTM is expanded into per-timestep cells at the ORT layer.|
+|'1'|Run the monolithic LSTM kernel on HTP (single graph node).|
+
+Warning: Enabling HTP Monolithic LSTM may improve session creation time, but this improvement may come with a regression in inference performance.
 
 |`"enable_htp_spill_fill_buffer"`|Description|
 |---|---|
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
@@ -250,6 +250,12 @@ Ort::Status LSTMOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
                                          const OrtNodeUnit& node_unit,
                                          const Ort::Logger& logger) const {
   ORT_UNUSED_PARAMETER(logger);
+
+  // Both unrolled/monolithic paths need X's shape to read seq_length/batch_size/input_size at compile time
+  // Reject a dynamic/symbolic X shape here rather than hard-failing later in ComposeGraph.
+  TensorInfo x_tensor_info = {};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[0], x_tensor_info));
+
   if (node_unit.Inputs().size() > 4 && node_unit.Inputs()[4].Exists()) {
     TensorInfo tensor_info = {};
     RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[4], tensor_info));
@@ -337,6 +343,20 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
   const uint32_t seq_length = layout == 0 ? input_tensor_infos[0].shape[0] : input_tensor_infos[0].shape[1];
   const int32_t direction_idx = input_tensor_infos[1].shape[0] < 2 || direction == "forward" ? 0 : 1;
 
+  // When false (default), lower LSTM as ORT-QNN-EP side per-timestep unrolled QNN_OP_LSTM cells + Pack.
+  // When true, lower as a single monolithic QNN_OP_LSTM node and rely on HTP's native
+  // monolithic LSTM kernel (enabled via enable_htp_monolithic_lstm provider option, which also
+  // toggles the QNN_HTP_GRAPH_CONFIG_OPTION_MONOLITHIC_LSTM graph config).
+  //
+  // The monolithic path is only meaningful for the HTP/NPU backend.
+  // The IR backend is also allowed to honor the flag
+  // so that a serialized DLC can faithfully mirror the HTP monolithic graph — otherwise an
+  // IR-produced DLC (always unrolled) could not be compared against the HTP monolithic one.
+  // All other backends (CPU/GPU) have no monolithic kernel and always use the unrolled lowering.
+  const bool enable_htp_monolithic_lstm = (IsNpuBackend(qnn_model_wrapper.GetQnnBackendType()) ||
+                                           IsIrBackend(qnn_model_wrapper.GetQnnBackendType())) &&
+                                          qnn_model_wrapper.GetModelSettings().enable_htp_monolithic_lstm;
+
   // params
   std::vector<std::string> param_names;
 
@@ -354,8 +374,9 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
                                       QNN_OP_LSTM_PARAM_OUTPUT_CLIP_THRESHOLD, param_names));
 
   // time_major
-  // set to true since current builder only support time major lstm(layout=0)
-  RETURN_IF_ERROR(AddQnnScalar<bool>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), true,
+  // monolithic QNN_OP_LSTM consumes the whole 3D sequence at once (time_major=true);
+  // the unrolled path feeds one 2D timestep per cell node (time_major=false).
+  RETURN_IF_ERROR(AddQnnScalar<bool>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), enable_htp_monolithic_lstm,
                                      QNN_OP_LSTM_PARAM_TIME_MAJOR, param_names));
 
   // input_gate_qscale
@@ -388,8 +409,10 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
 
   qnn_model_wrapper.AddTensorWrapper(std::move(null_tensor_wrapper));
 
-  std::vector<std::string> qnn_lstm_input_names(25, null_tensor_name);
-  qnn_lstm_input_names[0] = input_names[0];
+  std::vector<std::string> qnn_lstm_input_names(enable_htp_monolithic_lstm ? 25 : 24, null_tensor_name);
+  if (enable_htp_monolithic_lstm) {
+    qnn_lstm_input_names[0] = input_names[0];
+  }
   // input W
   {
     // QNN in[1] = ONNX in[1][direction, 2*hidden_size:3*hidden_size, :]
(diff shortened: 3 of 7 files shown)


cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
```diff
@@ -16,7 +16,7 @@ index f08b974d60..a50d771234 100644
              ORT_THROW("Wrong value for htp_arch. select from: " + str);
            }
 -        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization" || key == "extended_udma") {
-+        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization" || key == "extended_udma" || key == "disable_htp_monolithic_lstm") {
++        } else if (key == "enable_htp_fp16_precision" || key == "offload_graph_io_quantization" || key == "extended_udma" || key == "enable_htp_monolithic_lstm") {
```
kuanyul-qti, line 19: I think we might not need this file anymore.

kuanyul-qti:
### 🟠 Significant Concern

**[D-1]** `[Arch]` `lstm_op_builder.cc (AddUnidirectionLSTM) + qnn_execution_provider.cc`
Before this PR the op-builder emitted a single QNN_OP_LSTM node for every backend and `disable_htp_monolithic_lstm` only toggled the HTP-gated graph config. This PR moves the monolithic-vs-unrolled branch into the op-builder reading one global `ModelSettings` flag with no backend gate, so the new default (unrolled) also rewrites LSTM lowering on CPU/GPU/IR (seq_length cells + Pack) even though the motivating regression is HTP-only — and a non-HTP user's only opt-out (`enable_htp_monolithic_lstm=1`) also pushes the HTP-only `QNN_HTP_GRAPH_CONFIG_OPTION_MONOLITHIC_LSTM` config. Better direction: gate the op-builder branch on the HTP/NPU backend (keep single-node lowering for non-HTP), or explicitly re-scope + re-document the flag and back the non-HTP change with data.

**[D-3]** `[Completeness]` `test/providers/qnn/lstm_test.cc:617, 645–869`
Verified against the file: `RunHtpFp16LSTMOpTest` now defaults to unrolled, but every non-`_monolithic_lstm` fp16 test except `LSTM_Fp16_sanity_forward` was flipped to monolithic (reverse at L617, all bidirectional_* at L645–869). So the new-default unrolled path has only one fp16 test (forward, at 0.51 tol); the riskiest new code (reverse index math, bidirectional Concat over Pack) has zero fp16 coverage. QDQ unrolled coverage is solid, making fp16 the load-bearing gap. Better direction: add fp16 unrolled reverse + at least one bidirectional test, keeping the `_monolithic_lstm` tests as opt-in coverage.

---

### 🔵 Improvement

**[D-4]** `[Completeness]` `test/providers/qnn/lstm_test.cc`
Every HTP test with a real peephole tensor (`std::ref(P_def)`) is either `_monolithic_lstm` (now monolithic) or `DISABLED_`; all un-suffixed HTP default-path tests pass `std::nullopt` for P. So the unrolled path's per-cell peephole slicing (QNN in[18..20]) is exercised only on CPU-FP32, never HTP QDQ/fp16. (Logic is correct — see verified strengths; this is a coverage gap.) Add one HTP QDQ(u8) and one HTP fp16 default-path test with a real `P_def`.

**[D-5]** `[Completeness]` `lstm_op_builder.cc (AddUnidirectionLSTM loop / IsOpSupported)`
The unrolled path materializes `seq_length` QNN_OP_LSTM nodes at compile time from `input_tensor_infos[0].shape[0]`, but `IsOpSupported` does not reject a symbolic/dynamic X sequence dim. The new default depends on a concrete compile-time seq_length more strongly than the monolithic path did. Add an explicit guard that declines non-static (or zero) seq dim so unsupported models cleanly fall back to CPU.

**[D-6]** `[Arch]` `qnn_def.h + qnn_model_wrapper.h + qnn_execution_provider.cc`
The single monolithic-vs-unrolled decision now feeds two consumers via two members (`HtpGraphConfigs_t` for the finalize-time graph config, `ModelSettings` for the op-builder branch), kept consistent by one mirror line — a latent sync obligation where a future edit touching only one would silently desync graph config from builder path. Acceptable minimal wiring for this PR; longer term, read from one source.

chunghow-qti:
> ### 🟠 Significant Concern
> **[D-1]** `[Arch]` `lstm_op_builder.cc (AddUnidirectionLSTM) + qnn_execution_provider.cc` Before this PR the op-builder emitted a single QNN_OP_LSTM node for every backend and `disable_htp_monolithic_lstm` only toggled the HTP-gated graph config. This PR moves the monolithic-vs-unrolled branch into the op-builder reading one global `ModelSettings` flag with no backend gate, so the new default (unrolled) also rewrites LSTM lowering on CPU/GPU/IR (seq_length cells + Pack) even though the motivating regression is HTP-only — and a non-HTP user's only opt-out (`enable_htp_monolithic_lstm=1`) also pushes the HTP-only `QNN_HTP_GRAPH_CONFIG_OPTION_MONOLITHIC_LSTM` config. Better direction: gate the op-builder branch on the HTP/NPU backend (keep single-node lowering for non-HTP), or explicitly re-scope + re-document the flag and back the non-HTP change with data.
> 
> **[D-3]** `[Completeness]` `test/providers/qnn/lstm_test.cc:617, 645–869` Verified against the file: `RunHtpFp16LSTMOpTest` now defaults to unrolled, but every non-`_monolithic_lstm` fp16 test except `LSTM_Fp16_sanity_forward` was flipped to monolithic (reverse at L617, all bidirectional_* at L645–869). So the new-default unrolled path has only one fp16 test (forward, at 0.51 tol); the riskiest new code (reverse index math, bidirectional Concat over Pack) has zero fp16 coverage. QDQ unrolled coverage is solid, making fp16 the load-bearing gap. Better direction: add fp16 unrolled reverse + at least one bidirectional test, keeping the `_monolithic_lstm` tests as opt-in coverage.
> 
> ### 🔵 Improvement
> **[D-4]** `[Completeness]` `test/providers/qnn/lstm_test.cc` Every HTP test with a real peephole tensor (`std::ref(P_def)`) is either `_monolithic_lstm` (now monolithic) or `DISABLED_`; all un-suffixed HTP default-path tests pass `std::nullopt` for P. So the unrolled path's per-cell peephole slicing (QNN in[18..20]) is exercised only on CPU-FP32, never HTP QDQ/fp16. (Logic is correct — see verified strengths; this is a coverage gap.) Add one HTP QDQ(u8) and one HTP fp16 default-path test with a real `P_def`.
> 
> **[D-5]** `[Completeness]` `lstm_op_builder.cc (AddUnidirectionLSTM loop / IsOpSupported)` The unrolled path materializes `seq_length` QNN_OP_LSTM nodes at compile time from `input_tensor_infos[0].shape[0]`, but `IsOpSupported` does not reject a symbolic/dynamic X sequence dim. The new default depends on a concrete compile-time seq_length more strongly than the monolithic path did. Add an explicit guard that declines non-static (or zero) seq dim so unsupported models cleanly fall back to CPU.
> 
> **[D-6]** `[Arch]` `qnn_def.h + qnn_model_wrapper.h + qnn_execution_provider.cc` The single monolithic-vs-unrolled decision now feeds two consumers via two members (`HtpGraphConfigs_t` for the finalize-time graph config, `ModelSettings` for the op-builder branch), kept consistent by one mirror line — a latent sync obligation where a future edit touching only one would silently desync graph config from builder path. Acceptable minimal wiring for this PR; longer term, read from one source.

[D-1] Bind `enable_htp_monolithic_lstm` with NPU backend & IR backend,
[D-3] done
[D-4] done
[D-5] done
[D-6] add a comment for this.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -895,12 +895,17 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
 
-  // HTP monolithic lstm
-  htp_graph_configs_.disable_htp_monolithic_lstm = ParseBoolOption(ort_api,
-                                                                   session_options_,
-                                                                   FormatEPConfigKey("disable_htp_monolithic_lstm"),
-                                                                   false,
-                                                                   logger_);
+  // HTP monolithic lstm. Default false: lower LSTM as ORT-side per-timestep unrolled cells
+  // (expand at ORT). Set to true to run the monolithic LSTM kernel on HTP.
+  htp_graph_configs_.enable_htp_monolithic_lstm = ParseBoolOption(ort_api,
```
huaychou, line 901: **[M-2] (Major)** Old key disable_htp_monolithic_lstm removed with no backward compatibility fallback

**Issue Description**
The PR removes `disable_htp_monolithic_lstm` entirely with no fallback logic. This is a breaking change: existing users who set `disable_htp_monolithic_lstm=0` (explicit monolithic-on) or `disable_htp_monolithic_lstm=1` (monolithic-off) will have their behavior silently changed after upgrading.

The default behavior is also inverted:
- Old: `disable=false` (default) → monolithic **enabled**
- New: `enable=false` (default) → monolithic **disabled**

This means all users who never explicitly set this key will see a behavior change without any indication.

**Impact**
Silent behavior change may cause perf or accuracy regressions in production deployments after upgrading. Per the team's session config key rename convention (PR #107 pattern), the old key should be kept as a fallback in the same PR and cleaned up in a later release.

**Suggested fix:**

In `qnn_execution_provider.cc`, keep parsing `disable_htp_monolithic_lstm` as a fallback, with the new key taking priority when both are set. Map old `"1"` (disable monolithic) → `enable=false`; old `"0"` → `enable=true`. Log a deprecation warning when the old key is used.

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
```diff
@@ -663,28 +686,126 @@ Ort::Status LSTMOpBuilder::AddUnidirectionLSTM(QnnModelWrapper& qnn_model_wrappe
...
+      }
+
+      std::vector<uint32_t> qnn_lstm_output_shape = {batch_size, hidden_size};
+      std::vector<std::string> qnn_lstm_output_names = {
+          utils::UniqueNameGenerator().New(node_unit, "_QNN_LSTM_output_all_hidden_state_" + std::to_string(sequence_idx) + "_" + direction),
+          utils::UniqueNameGenerator().New(node_unit, "_QNN_LSTM_output_cell_state_" + std::to_string(sequence_idx) + "_" + direction),
+          utils::UniqueNameGenerator().New(node_unit, "_QNN_LSTM_output_hidden_state_" + std::to_string(sequence_idx) + "_" + direction)};
+      qnn_lstm_input_names[10] = qnn_lstm_output_names[2];  // update initial_h for next cell
+      qnn_lstm_input_names[11] = qnn_lstm_output_names[1];  // update initial_c for next cell
+      qnn_all_hidden_state_names[sequence_idx] = qnn_lstm_output_names[2];
```
huaychou, line 758: **[N-1] (Minor)** Unrolled path Pack for all_hidden uses out[2] (o_t) instead of out[0] (h_t) — semantic needs clarification

**Issue Description**
In the unrolled path, ONNX output Y (all hidden states) is formed by Packing per-timestep `qnn_lstm_output_names[2]` (variable named `_output_hidden_state_`, corresponding to QNN LSTM out[2]). However, the file header comment describes QNN LSTM outputs as:
- out[0]: h_t (hidden state output — should be the per-timestep value for ONNX Y)
- out[2]: o_t (possibly output gate or an alias)

The monolithic path uses `kOutputPositionMap[0] = 0` (i.e., out[0] = h_t) as the source for ONNX Y, but the unrolled path uses out[2] to form the same Y output. Two paths use different output indices to represent the same ONNX semantic (Y = all hidden states). This may be semantically consistent (QNN 2D LSTM out[0] and out[2] may be identical), or it may be an h_t / o_t confusion bug.

The tests currently pass, possibly because: a) QNN 2D LSTM out[0] == out[2], or b) tolerance masks the difference.

**Impact**
If QNN LSTM 2D out[0] and out[2] differ semantically, the unrolled path produces Y as output-gate activation (o_t) rather than true hidden state (h_t), causing a silent accuracy error.

**Suggested fix:**

Clarify in comments whether QNN LSTM 2D out[0] (h_t) and out[2] (o_t) are identical, and why the unrolled path uses out[2] rather than out[0] to construct ONNX Y. If they are the same, add an explanatory comment. If they differ, switch to using out[0] for the Pack.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -895,12 +895,17 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+                                                                  session_options_,
+                                                                  FormatEPConfigKey("enable_htp_monolithic_lstm"),
```
huaychou: **[N-2] (Minor)** enable_htp_monolithic_lstm uses inline string literals; prefer named constexpr constants

**Issue Description**
`ParseBoolOption` uses an inline string literal `FormatEPConfigKey("enable_htp_monolithic_lstm")`, while other similar keys (e.g., `enable_framework_op_trace_`) use `constexpr const char*` constants. The difference is minor when the key appears only once, but named constants prevent typos and make cross-site grep easier when keys appear in multiple places (e.g., logs, tests, docs).

**Impact**
A misspelled key string silently fails at runtime with no compiler error.

**Suggested fix:**

Declare `static constexpr const char* kEnableHtpMonolithicLstm = "enable_htp_monolithic_lstm";` and replace all inline string literal uses.

chenweng-quic (approved):

cmake/patches/ort_core/0004-Update-argument-for-monolithic-lstm.patch
chunghow-qti: Will keep it once CI migration complete

chenweng-quic (approved):

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -205,10 +205,12 @@ Refer to the [QAIRT SDK documentation](https://docs.qualcomm.com/doc/80-63442-10
...
 
-|`"disable_htp_monolithic_lstm"`|Description|
+|`"enable_htp_monolithic_lstm"`|Description|
 |---|---|
-|'0'|Default. HTP uses the monolithic LSTM graph configuration.|
-|'1'|Disable the monolithic LSTM graph configuration on HTP.|
+|'0'|Default. LSTM is expanded into per-timestep cells at the ORT layer.|
+|'1'|Run the monolithic LSTM kernel on HTP (single graph node).|
+
+The two settings trade off session creation time against inference time. Per-timestep expansion (`'0'`, default) emits `O(seq_length)` nodes per direction, so session creation time grows with sequence length; the monolithic kernel (`'1'`) keeps a constant node count, so session creation is faster, at the cost of slightly slower inference. For models with long sequences (e.g. audio/ASR), consider setting `enable_htp_monolithic_lstm=1`.
```
qti-chuteng: Remove the detailed explainantion before we confirm there is no room for improvement.
We only can add the brief warning, like enablement may improve the session creation time but may regress the inference time.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -895,12 +895,17 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+                                                                  false,
+                                                                  logger_);
+  // Mirror into model settings so the LSTM op builder can pick the unrolled vs monolithic path.
```
qti-chuteng: This comment is not needed

huaychou (commented):
Did we only support monolithic LSTM before? I think this PR add the support for unrolled LSTM. If so, do we need to add unit tests for unrolled LSTM?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: As discussed with @qti-chuteng, Monolithic LSTM was never shipped in any release. In the latest 2.4.0 release, there is no disable_htp_monolithic_lstm option. The option only exists temporarily on the main branch.

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: sure

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: sure

onnxruntime/core/providers/qnn/builder/opbuilder/lstm_op_builder.cc
chunghow-qti: add a comment
// Single-timestep cell, so out[0] (all_hidden) == out[2] (final hidden state); reuse out[2].

onnxruntime/core/providers/qnn/qnn_execution_provider.cc

chunghow-qti:
> Did we only support monolithic LSTM before? I think this PR add the support for unrolled LSTM. If so, do we need to add unit tests for unrolled LSTM?

yes, the default change to unroll at ORT. so all default UTs are for unrolliung at ORT.

qti-chuteng (approved):

PR #631, opened 2026-07-12
[QNN EP] Fix OOB read in OrtPadNodeGroupSelector::Check on Pad with 0 dq inputs

qti-yuduo:
## Summary
- `OrtPadNodeGroupSelector::Check` only rejected `dq_nodes.size() > 2`, never `== 0`. A Pad node whose `data` input is not DQ-quantized but whose output feeds a `QuantizeLinear` passes `CheckQDQNodes` (0 == 0) and then dereferences `dq_nodes[0]` on an empty vector, crashing with SIGSEGV.
- Fix: also reject `num_dq_inputs < 1`, matching the existing comment ("Pad can have 1 or 2 dq input").

## Root cause
Hit via a real model (Moonshine decoder, w8a8 QDQ) containing float32 `Pad` nodes fed by `Transpose`/`Concat` (not `DequantizeLinear`) whose output feeds `QuantizeLinear` directly — `dq_nodes` empty, `q_nodes` non-empty.


qti-chuteng (approved):

qti-yuduo:
Thanks for review @qti-chuteng!

PR #635, opened 2026-07-14
[QNN EP] Read concrete 6D shape from ValueInfo for SpaceToDepth fusion

yuhuchua-qti:
## Summary
- `HasSpaceToDepthCoreSignature` previously read the Reshape shape from initializer
  bytes and hard-coded `-1` handling to batch dim only (PR #567). That rejects the
  common exporter pattern `reshape(N, -1, H/b, b, W/b, b)` where `-1` sits at the
  channel dim.
- Switch to reading `Reshape1`'s output ValueInfo (concrete 6D from ORT shape
  inference) so `-1` at any position is accepted as long as shape inference
  resolved it. The 4D check likewise uses the concrete `output_shape` we already
  read at the top of the function, removing the `shape_4d` initializer path.
- This aligns `spacetodepth_fusion` with the shape-reading convention already used
  by `reshape_transpose_rank5`, `gather_transpose_reshape_fusion`, `reshape_gemm_fusion`,
  and 5 other node-group fusions in this directory — `spacetodepth` was the only
  one still going through `GetInitializerDataAsInt64` for shape reasoning.

## Test plan
- [x] Added `SpaceToDepthFusion_{Float,QDQ}_CRD_ChannelPlaceholder` under
      `QnnHTPBackendTests`, exercising `-1` at channel in wrapped Conv->RTR->Conv
- [x] 18/18 SpaceToDepth fusion cases pass on linux x86_64 HTP QEMU
- [x] Lint clean (`python qcom/build_and_test.py lint_and_fix`)

## Relation to other PRs
- Independent of #629 (bare RTR-only fusion). Both target different exporter
  patterns and do not need to land together.
- Follows #567 (which added `-1`-at-batch support). This PR generalises that fix.


qti-chuteng:
### 🟪 Architecture Fit

**[D-3]** `🔵 Improvement` `[Arch]` — `spacetodepth_fusion.cc` (removed loops, old lines 185-200)
After removing the positivity loop, `dim==0` is no longer rejected locally — `GetOnnxShape` (`qnn_model_wrapper.cc:723`) returns false only on `dim < 0` and lets `0` through. `b0/b1` stay guarded by the retained `b0 < 1 || b1 < 1`, and `h_div`/`w_div`/`r_c` are pinned by the equality checks, so a zero slips through only if the concrete input tensor itself carries a 0 dim (e.g. C=0/H=0) — degenerate, and rejected downstream by QNN validation. Genuine but effectively-unreachable narrowing of the old guard's coverage (⟦C8.3⟧).
*Better design:* If zero-size tensors are out of scope (they are), add a `> 0` assertion alongside the top-of-function rank-4 `input_shape` check so the decline stays local and self-documenting.


### 🟩 Simplicity

**[D-1]** `🔵 Improvement` `[Simplicity]` — `utils.cc:214` (decl `utils.h:90`)
This diff deletes all three call sites of `GetInitializerDataAsInt64` in `spacetodepth_fusion.cc` (lines 175, 176, 372). A repo-wide grep confirms it now has **zero callers** — only its definition, declaration, and its own name in a log string remain. The author cleanly dropped the paired `#include <limits>`, but left a dead public helper behind. This is the highest-leverage cut in the diff: the PR's own framing ("spacetodepth was the last fusion using it") is the signal the helper's job is done. Leaving it makes future readers think it's live and lets it drift into wrong reference code (⟦C8.1⟧). Not a blocker — it's a general-purpose util a future fusion could reuse.
*Better design:* Delete the definition (`utils.cc:214`) and declaration (`utils.h:90`) in this same PR — git restores it if needed; or add a `// Retained for <PR/why>` note plus a tracking reference if intentionally kept.

### 🔵 Completeness

**[D-2]** `🔵 Improvement` `[Completeness]` — `spacetodepth_fusion_test.cc:596-627`
Only a CRD channel-placeholder test was added; the symmetric DCR variant is missing. The fix lives in the perm-agnostic core gate `HasSpaceToDepthCoreSignature` (accepts both DCR `{0,3,5,1,2,4}` and CRD `{0,1,3,5,2,4}`), and this file pairs nearly every scenario DCR+CRD (Float, QDQ, U16, UnequalBlock). The channel-placeholder `-1` sits at `shape_6d[1]` regardless of perm, so CRD already exercises the fixed line — but it's a small gap against the file's own symmetry convention (⟦C6.1⟧).
*Better design:* Add a DCR channel-placeholder test (one `RunSpaceToDepthFusionTest` call, perm `{0,3,5,1,2,4}`, `use_channel_placeholder=true`). Cheap. Not a blocker.

onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
```diff
@@ -230,9 +211,7 @@ bool HasSpaceToDepthCoreSignature(
   const int64_t expected_c = c * b0 * b1;
   const std::array<int64_t, 4> expected_shape_4d = {n, expected_c, h / b0, w / b1};
```
minfhong-qti:
```suggestion
  const std::array<int64_t, 4> expected_output_shape = {n, expected_c, h / b0, w / b1};
```
Nit.

onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
```diff
@@ -171,50 +170,32 @@ bool HasSpaceToDepthCoreSignature(
...
-  if (!shape_6d.has_value() || !shape_4d.has_value()) {
+  // Read the *resolved* 6D shape from Reshape1's output ValueInfo (== Transpose's input).
+  // Shape inference has already replaced any -1 in the Reshape shape initializer with the
+  // concrete value the exporter meant, which lets us accept exporter-convenience patterns
+  // like reshape(1, -1, H, b, W, b) without special-casing -1 at each position. Same
+  // reasoning for the 4D shape via output_shape above.
+  std::vector<uint32_t> shape_6d;
+  if (!qnn_model_wrapper.GetOnnxShape(reshape1.Outputs()[0].shape, shape_6d) || shape_6d.size() != kRank6) {
     return false;
   }
```
minfhong-qti: Combine with above `input_shape` and `output_shape`.

onnxruntime/test/providers/qnn/qnn_node_group/spacetodepth_fusion_test.cc
minfhong-qti, line 28: Add another shape parameter here instead of `use_channel_placeholder`, and add testcases other than -1 in batch or channel.
In addition, I believe the logic in LN40 is no longer needed with this new parameter.

onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
yuhuchua-qti: Done — renamed expected_shape_4d → expected_output_shape and applied the suggested one-liner.

onnxruntime/core/providers/qnn/builder/qnn_node_group/spacetodepth_fusion.cc
yuhuchua-qti: Done — the three GetOnnxShape reads (input_shape / output_shape / shape_6d) are now combined into a single if-block with the rank checks inline.

onnxruntime/test/providers/qnn/qnn_node_group/spacetodepth_fusion_test.cc
yuhuchua-qti: Done. 
Replaced the use_channel_placeholder bool with explicit reshape1_shape / reshape2_shape params so a -1 can sit at any position. Removed the LN40 std::replace (no longer needed). 
Added cases beyond batch/channel: _Reshape1HeightPlaceholder (-1 at H/b) and _Reshape2ChannelPlaceholder (-1 in the rank-4 second Reshape), plus a DCR channel variant. 
All placeholder tests were also renamed to a consistent Reshape{1,2}<Dim>Placeholder scheme.

yuhuchua-qti:
> ### 🟪 Architecture Fit
> **[D-3]** `🔵 Improvement` `[Arch]` — `spacetodepth_fusion.cc` (removed loops, old lines 185-200) After removing the positivity loop, `dim==0` is no longer rejected locally — `GetOnnxShape` (`qnn_model_wrapper.cc:723`) returns false only on `dim < 0` and lets `0` through. `b0/b1` stay guarded by the retained `b0 < 1 || b1 < 1`, and `h_div`/`w_div`/`r_c` are pinned by the equality checks, so a zero slips through only if the concrete input tensor itself carries a 0 dim (e.g. C=0/H=0) — degenerate, and rejected downstream by QNN validation. Genuine but effectively-unreachable narrowing of the old guard's coverage (⟦C8.3⟧). _Better design:_ If zero-size tensors are out of scope (they are), add a `> 0` assertion alongside the top-of-function rank-4 `input_shape` check so the decline stays local and self-documenting.
> 
> ### 🟩 Simplicity
> **[D-1]** `🔵 Improvement` `[Simplicity]` — `utils.cc:214` (decl `utils.h:90`) This diff deletes all three call sites of `GetInitializerDataAsInt64` in `spacetodepth_fusion.cc` (lines 175, 176, 372). A repo-wide grep confirms it now has **zero callers** — only its definition, declaration, and its own name in a log string remain. The author cleanly dropped the paired `#include <limits>`, but left a dead public helper behind. This is the highest-leverage cut in the diff: the PR's own framing ("spacetodepth was the last fusion using it") is the signal the helper's job is done. Leaving it makes future readers think it's live and lets it drift into wrong reference code (⟦C8.1⟧). Not a blocker — it's a general-purpose util a future fusion could reuse. _Better design:_ Delete the definition (`utils.cc:214`) and declaration (`utils.h:90`) in this same PR — git restores it if needed; or add a `// Retained for <PR/why>` note plus a tracking reference if intentionally kept.
> 
> ### 🔵 Completeness
> **[D-2]** `🔵 Improvement` `[Completeness]` — `spacetodepth_fusion_test.cc:596-627` Only a CRD channel-placeholder test was added; the symmetric DCR variant is missing. The fix lives in the perm-agnostic core gate `HasSpaceToDepthCoreSignature` (accepts both DCR `{0,3,5,1,2,4}` and CRD `{0,1,3,5,2,4}`), and this file pairs nearly every scenario DCR+CRD (Float, QDQ, U16, UnequalBlock). The channel-placeholder `-1` sits at `shape_6d[1]` regardless of perm, so CRD already exercises the fixed line — but it's a small gap against the file's own symmetry convention (⟦C6.1⟧). _Better design:_ Add a DCR channel-placeholder test (one `RunSpaceToDepthFusionTest` call, perm `{0,3,5,1,2,4}`, `use_channel_placeholder=true`). Cheap. Not a blocker.

[D-1] Deleted the now-dead GetInitializerDataAsInt64 (both the utils.cc definition and the utils.h declaration). Confirmed zero remaining callers across the QNN EP.

[D-2] Added the symmetric DCR channel-placeholder test (SpaceToDepthFusion_QDQ_DCR_Reshape1ChannelPlaceholder, perm {0,3,5,1,2,4}). QDQ-only, since HTP has no accurate float DCR SpaceToDepth kernel — matching the existing suite's convention of carrying no Float DCR case.

[D-3] Added a local zero-size guard right after the rank-4 checks: std::any_of(... d == 0) on input_shape/output_shape, so the decline stays local and self-documenting instead of deferring a degenerate 0 to downstream QNN validation.

qti-chuteng (approved):

minfhong-qti (approved):

PR #644, opened 2026-07-16
[QNN EP] Add Softplus to QDQ-fusion unary_ops selector allowlist

qti-yuduo:
## Summary
- Softplus's op-builder (`simple_op_builder.cc`) already fully supports quantized Softplus via `QNN_OP_ELEMENT_WISE_NEURON` — the same dispatch path used by `Sigmoid`/`Tanh` — but `Softplus` was missing from `OrtSelectorManager::CreateSelectors()`'s `unary_ops` QDQ-fusion allowlist in `qnn_ep_utils.cc`.
- Without a matching selector, ORT never recognizes the `DequantizeLinear -> Softplus -> QuantizeLinear` node group as fusable, so Softplus runs as unfused float32 bracketed by its own Quantize/Dequantize pair instead of collapsing into a single fused HTP op.
- Adds `"Softplus"` to the allowlist (one line), matching how `Sigmoid`/`Tanh`/`Softmax`/etc. are already registered.

### DIFF
--- a/onnxruntime/test/providers/qnn/simple_op_test.cc
+++ b/onnxruntime/test/providers/qnn/simple_op_test.cc
@@ -568,9 +568,8 @@ TEST_F(QnnHTPBackendTests, UnaryOp_Softplus_U16) {
 // Dequantize pair. With offload_graph_io_quantization=0, the graph's Q/DQ input and output
 // boundary nodes are always present (1 Quantize + 1 Dequantize) even when fused; an extra
 // pair beyond that indicates Softplus itself failed to fuse.
-// Regression test for https://github.com/qcom-ai-hub/tetracode/issues/20283, where Softplus
-// was missing from the QDQ-fusion selector's unary_ops allowlist even though the op-builder
-// fully supports quantized Softplus.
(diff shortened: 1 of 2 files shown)


onnxruntime/test/providers/qnn/simple_op_test.cc
```diff
@@ -563,6 +563,44 @@ TEST_F(QnnHTPBackendTests, UnaryOp_Softplus_U16) {
                          true);
 }
 
+// Check that QNN fuses DQ -> Softplus -> Q into a single quantized ElementWiseNeuron op,
+// instead of leaving Softplus running as unfused float32 bracketed by its own Quantize/
+// Dequantize pair. With offload_graph_io_quantization=0, the graph's Q/DQ input and output
+// boundary nodes are always present (1 Quantize + 1 Dequantize) even when fused; an extra
+// pair beyond that indicates Softplus itself failed to fuse.
+// Regression test for https://github.com/qcom-ai-hub/tetracode/issues/20283, where Softplus
```
yath1: Minor: Could we remove this internal link and only explain what the issue this change is solving instead?

yath1 (approved):
lgtm other than one minor comment.

PR #517, opened 2026-06-11
[QNN EP] Add prepare_and_load session option for single-session AOT inference

qti-mbadnara:
# Description

Add a new session config flag `ep.qnnexecutionprovider.enable_htp_prepare_and_load=1` that performs AOT compilation and context loading within a single ORT session. This allows large models (e.g., LLMs) to bypass the single process-domain (PD) memory limit imposed by the QNN JIT flow without requiring two separate sessions.

## Motivation / Context

QNN EP's JIT flow places all graph splits in a single QNN process domain (PD), which exhausts memory for large models. The existing AOT workaround requires two sessions:
1. Session 1: `prepare_only=1` — compile and emit `_ctx.onnx`
2. Session 2: load `_ctx.onnx` for inference (QNN spreads splits across multiple PDs)

This two-session flow is cumbersome for embedded customers who want a single API call. The new `prepare_and_load` option performs both steps internally:
1. Compile the model (JIT path)
2. Extract the QNN context binary
3. Release the compile-time single-PD context
4. Reload the binary via the AOT path (multi-PD)
5. Session is immediately ready for inference

### Behavior matrix

| `ep.context_enable` | `enable_htp_prepare_and_load` | Behavior |
|---|---|---|
| 0 | 0 | Existing JIT flow (unchanged) |
| 1 | 0 | Existing AOT flow (unchanged) |
| 0 | 1 | **New Path A:** compile → reload from memory → infer. No persisted artifact. |
| 1 | 1 | **New Path B:** compile → write `_ctx.onnx` + `.bin` → reload → infer. Artifact persists. |

**Constraints:**
- Mutually exclusive with `enable_htp_prepare_only`. Setting both to `'1'` raises an error.
- Setting `enable_htp_prepare_and_load=1` with `ep.context_enable=0` AND an explicit `ep.context_file_path` raises an error (contradictory: "don't persist" + "here's where to persist").
- If the input model is already a pre-compiled context model (`_ctx.onnx`), `enable_htp_prepare_and_load` is silently ignored with a warning — the model loads directly via the existing AOT path.
- `ep.context_embed_mode` is fully respected: with `embed_mode=1`, the binary is embedded inside the `.onnx`; with `embed_mode=0`, a separate `_qnn.bin` is written.

## Files Changed

| File | Change |
|---|---|
[...]

PR #645, opened 2026-07-16
[QNN EP] Add HTP Graph Splitting support

qti-mbadnara:
## Description

- Adds three new HTP provider options: `enable_htp_graph_splitting`, `htp_graphsplitter_num_prepare_threads`, and `htp_graph_splitting_kway_partitions`
- When `enable_htp_graph_splitting=1`, the QNN HTP backend splits the model graph into independently-prepareable sub-graphs, reducing context preparation time by ~30% (single-thread) and ~78% (8 threads)
- Feature availability is gated by `QNN_HTP_GRAPH_SPLITTING_AVAILABLE` macro (defined in `qnn_def.h` when compiling against QAIRT SDK 2.49+) — builds against older SDKs compile and run cleanly with graph splitting silently disabled
- Parse-time warning emitted if `enable_htp_graph_splitting=1` is set but the build was compiled against SDK < 2.49
- Parse-time warning emitted if `htp_graphsplitter_num_prepare_threads` or `htp_graph_splitting_kway_partitions` are set while `enable_htp_graph_splitting=0`
- Config array passed to `QnnContext_create` is built dynamically (vector) to avoid nullptr holes in a null-terminated list

## Motivation and Context

HTP Graph Splitting is a QNN HTP feature targeting large single-core LLMs (Llama 2/3). Context binary preparation is the dominant latency cost for these models at load time. Graph splitting parallelizes this by partitioning the graph into transformer-block-sized sub-graphs that can be prepared independently. Inference behavior is unchanged — only prepare-time is affected.

The feature is effective in both JIT (compile-and-run) and AOT (context binary generation) workflows, including on-device AOT. It has no effect when loading from an already-compiled context binary.

The feature requires QAIRT SDK 2.49+ at both compile time and runtime. This PR introduces the ORT session config surface so applications can opt in without code changes beyond setting provider options.

## Provider Options Reference

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `enable_htp_graph_splitting` | `"0"` / `"1"` | `"0"` | Enable HTP graph splitting. Requires QAIRT SDK 2.49+ at runtime. |
| `htp_graphsplitter_num_prepare_threads` | integer string | `"8"` | Parallel threads for sub-graph preparation. Only effective when `enable_htp_graph_splitting=1`. `"0"` = auto. |
| `htp_graph_splitting_kway_partitions` | integer string | `"4"` | Number of sub-graph partitions. `"0"` = SDK default. Sets `GPE_KWAY_PARTITIONS` env var before context creation. Only effective when `enable_htp_graph_splitting=1`. |

## Implementation Notes

- `QNN_HTP_GRAPH_SPLITTING_AVAILABLE` macro added to `qnn_def.h` — replaces all raw `#if defined(QNN_SDK_VERSION_MAJOR) && ...>= 49` guards
- `ParseUint32ConfigEntry` helper added alongside existing `ParseBoolOption` for consistent uint32 option parsing with warning log on invalid input
- Config array is built via `std::vector::push_back` (no conditional-nullptr entries) to safely handle null-terminated list semantics in `QnnContext_create`
- `htp_graph_splitting_kway_partitions` has no QNN API counterpart — it maps to the `GPE_KWAY_PARTITIONS` process environment variable, which is the only SDK-provided interface for this setting

### onnxruntime_perf_test

Basic graph splitting with defaults:
```bash
.\onnxruntime_perf_test.exe -e qnn -i "backend_type|htp enable_htp_graph_splitting|1" model.onnx
```

Full configuration:
```bash
.\onnxruntime_perf_test.exe -e qnn -i "backend_type|htp enable_htp_graph_splitting|1 htp_graphsplitter_num_prepare_threads|8 htp_graph_splitting_kway_partitions|4" model.onnx
```

Baseline comparison (graph splitting disabled):
```bash
.\onnxruntime_perf_test.exe -e qnn -i "backend_type|htp enable_htp_graph_splitting|0" model.onnx
```

### C++ API

```cpp
Ort::SessionOptions so;
so.AddConfigEntry("ep.qnnexecutionprovider.enable_htp_graph_splitting", "1");
so.AddConfigEntry("ep.qnnexecutionprovider.htp_graphsplitter_num_prepare_threads", "8");
so.AddConfigEntry("ep.qnnexecutionprovider.htp_graph_splitting_kway_partitions", "4");
so.AppendExecutionProvider("QnnExecutionProvider", {{"backend_type", "htp"}});
```

Or as provider options:
```cpp
ProviderOptions opts;
opts["backend_type"]                       = "htp";
opts["enable_htp_graph_splitting"]         = "1";
opts["htp_graphsplitter_num_prepare_threads"] = "8";
opts["htp_graph_splitting_kway_partitions"]   = "4";
```

### Python

```python
import onnxruntime as ort

so = ort.SessionOptions()
so.add_session_config_entry("ep.qnnexecutionprovider.enable_htp_graph_splitting", "1")
so.add_session_config_entry("ep.qnnexecutionprovider.htp_graphsplitter_num_prepare_threads", "8")
so.add_session_config_entry("ep.qnnexecutionprovider.htp_graph_splitting_kway_partitions", "4")

session = ort.InferenceSession(
    "model.onnx", so,
    providers=["QNNExecutionProvider"],
    provider_options=[{"backend_type": "htp"}]
)
```


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1299,14 +1302,49 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
...
+#endif
+
   const QnnContext_Config_t* configs[] = {&context_priority_config,
 #if QNN_API_VERSION_MAJOR == 2 && (QNN_API_VERSION_MINOR >= 26)
                                           &context_config_resource_sharing,
                                           &resource_sharing_opt_type_config,
                                           &context_config_weight_sharing,
+#endif
+#if defined(QNN_SDK_VERSION_MAJOR) && QNN_SDK_VERSION_MAJOR == 2 && defined(QNN_SDK_VERSION_MINOR) && QNN_SDK_VERSION_MINOR >= 49
+                                          enable_gpe ? &context_config_gpe_vtcm : nullptr,
```
quic-calvnguy: Please verify this will not cause any errors. I _think_ it should be fine since QNN assumes this is a nullptr-terminated list.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1521,12 +1562,36 @@ Ort::Status QnnBackendManager::CreateContext(bool enable_htp_weight_sharing,
...
+#endif
+
   const QnnContext_Config_t* npu_context_configs[] = {
       &context_priority_config,
       &context_config_weight_sharing,
       &context_config_extended_udma,
       &context_config_prepare_only,
       enable_htp_ref_weight_sharing ? &context_config_ref_weight_sharing : nullptr,
+#if defined(QNN_SDK_VERSION_MAJOR) && QNN_SDK_VERSION_MAJOR == 2 && defined(QNN_SDK_VERSION_MINOR) && QNN_SDK_VERSION_MINOR >= 49
+      enable_gpe ? &context_config_gpe : nullptr,
```
quic-calvnguy: Please verify that this is not broken when the previous option is a nullptr. Similar to other comment: QNN reads these configs in as a nullptr-terminated list

quic-calvnguy: Looks like enable_htp_ref_weight_sharing is currently always true for now as groundwork for FCB so this should be fine.

@minfhong-qti FYI for future FCB work

quic-calvnguy (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Agreed, based on offline discussion can be taken up for future FCB work

minfhong-qti: In `QnnEp`, the flag `enable_htp_ref_weight_sharing` is always true indeed. However, it is used in FCB flow only. In non-FCB flow, the parameter `enable_htp_ref_weight_sharing` here could be false. https://github.com/onnxruntime/onnxruntime-qnn/blob/main/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc#L2082

quic-calvnguy (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Done!

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -332,6 +332,19 @@ For more information, see the [Parallel Graph Preparation](#parallel-graph-prepa
 |'0'|Default. Disabled.|
 |'1'|Compile the model and save the QNN context binary, but skip inference. `OnRunStart`, `OnRunEnd`, `CreateState`, and `SetDynamicOptions` are all no-ops. Useful for a compile-once/run-later workflow. Requires `ep.context_enable=1`; silently disabled with a warning if context cache is not enabled.|
 
+|`"htp_enable_gpe"`|Description|
```
tirupath-qti: since htp is not explicitly using GraphSplitting instead of GPE in its API, ORT EP should also use the same terminology instead of using a different name for this feature. Change GPE with GraphSplitting or GraphSplittingConfig in all the places.

for this provider option, let us use "enable_htp_graph_splitting".

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -332,6 +332,19 @@ For more information, see the [Parallel Graph Preparation](#parallel-graph-prepa
...
+|---|---|
+|'0'|Default. Disabled.|
+|'1'|Enable Graph Program Executor (GPE): the HTP backend splits the LLM graph into independently-prepareable sub-graphs and packages them as a program in the context binary. Reduces context-binary prepare time by ~30% (single-thread) and ~78% (8 threads). Requires QAIRT SDK 2.49+ at runtime; using this option with an older runtime will cause context creation to fail. Only effective during context-binary generation (`ep.context_enable=1`); no impact when loading from a cached binary.|
```
tirupath-qti: how about the JIT usecase? If this works only in AOT workflow, the information should be explicit. btw, does it work in On-Device AOT workflow?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1299,26 +1302,66 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
...
-  const QnnContext_Config_t* configs[] = {&context_priority_config,
+#if defined(QNN_SDK_VERSION_MAJOR) && QNN_SDK_VERSION_MAJOR == 2 && defined(QNN_SDK_VERSION_MINOR) && QNN_SDK_VERSION_MINOR >= 49
+  QnnContext_Config_t context_config_gpe_vtcm = QNN_CONTEXT_CONFIG_INIT;
+  QnnHtpContext_CustomConfig_t gpe_custom_config_vtcm;
+  if (enable_gpe) {
+    gpe_custom_config_vtcm.option = QNN_HTP_CONTEXT_CONFIG_OPTION_GRAPH_SPLITTING_CONFIGS;
+    gpe_custom_config_vtcm.graphSplittingConfigs.graphSplittingEnabled = true;
+    gpe_custom_config_vtcm.graphSplittingConfigs.numPrepareThreads = gpe_num_prepare_threads;
+    context_config_gpe_vtcm.option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
+    context_config_gpe_vtcm.customConfig = &gpe_custom_config_vtcm;
```
tirupath-qti: where is the gpe_kway_partitions parameter programmed in HTP config or custom config ?

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -332,6 +332,19 @@ For more information, see the [Parallel Graph Preparation](#parallel-graph-prepa
...
+
+|`"htp_gpe_num_prepare_threads"`|Description|
```
tirupath-qti: htp_graphsplitter_num_prepare_threads

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: graphSplittingEnabled and numPrepareThreads go through the QNN custom config API, while kway_partitions goes through the process environment because that's the only interface the SDK provides for it.

docs/execution_providers/QNN-ExecutionProvider.md
qti-mbadnara: Fixed!

docs/execution_providers/QNN-ExecutionProvider.md
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1240,6 +1240,53 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                    false,
                                                    logger_);
 
+  // GPE (Graph Program Executor / Graph Splitting). Requires SDK 2.49+ at runtime.
+  {
+    std::string gpe_str;
+    GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                   FormatEPConfigKey("enable_htp_graph_splitting"), "0", gpe_str);
+    enable_htp_graph_splitting_ = gpe_str == "1";
```
minfhong-qti: Leverage `ParseBoolOption`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1240,6 +1240,53 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                   FormatEPConfigKey("htp_graphsplitter_num_prepare_threads"), "", gpe_threads_str);
+    user_set_graph_splitting_threads = !gpe_threads_str.empty();
+    if (user_set_graph_splitting_threads) {
+      try {
+        unsigned long val = std::stoul(gpe_threads_str);
+        htp_graphsplitter_num_prepare_threads_ = val > 0 ? static_cast<uint32_t>(val) : 8u;
+      } catch (...) {
+        htp_graphsplitter_num_prepare_threads_ = 8;
+      }
```
minfhong-qti: Maybe worth defining a function for parsing string-to-unsignedlong and log warning message if any thing goes wrong and use default value.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1240,6 +1240,53 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    std::string kway_str;
+    GetSessionConfigEntryOrDefault(ort_api, session_options_,
+                                   FormatEPConfigKey("htp_graph_splitting_kway_partitions"), "", kway_str);
+    user_set_kway = !kway_str.empty();
+    if (user_set_kway) {
+      try {
+        htp_graph_splitting_kway_partitions_ = static_cast<uint32_t>(std::stoul(kway_str));
+      } catch (...) {
+        htp_graph_splitting_kway_partitions_ = 4;
+      }
```
minfhong-qti: Ditto.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1299,26 +1302,66 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
...
+    graph_splitting_custom_config_vtcm.graphSplittingConfigs.numPrepareThreads = graphsplitter_num_prepare_threads;
+    context_config_graph_splitting_vtcm.option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
+    context_config_graph_splitting_vtcm.customConfig = &graph_splitting_custom_config_vtcm;
+  }
+#else
+  ORT_UNUSED_PARAMETER(enable_htp_graph_splitting);
+  ORT_UNUSED_PARAMETER(graphsplitter_num_prepare_threads);
+  ORT_UNUSED_PARAMETER(graph_splitting_kway_partitions);
+  if (enable_htp_graph_splitting) {
+    ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_WARNING,
```
minfhong-qti: Prefer to log this during parsing instead of waiting until backend setup.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1299,26 +1302,66 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
   QnnContext_Config_t context_priority_config = QNN_CONTEXT_CONFIG_INIT;
   RETURN_IF_ERROR(SetQnnContextConfig(context_priority_, context_priority_config));
 
-  const QnnContext_Config_t* configs[] = {&context_priority_config,
+#if defined(QNN_SDK_VERSION_MAJOR) && QNN_SDK_VERSION_MAJOR == 2 && defined(QNN_SDK_VERSION_MINOR) && QNN_SDK_VERSION_MINOR >= 49
```
minfhong-qti: Define a macro in `qnn_def.h` indicating that GPE is available in current QAIRT build.

minfhong-qti:
Just curious. Will GPE be conflict with the parallel prepare implemented in `QnnEp` (i.e., option `num_graph_prepare_threads`)?

qti-mbadnara:
> Just curious. Will GPE be conflict with the parallel prepare implemented in `QnnEp` (i.e., option `num_graph_prepare_threads`)?

It wont conflict. Both can work independently

docs/execution_providers/QNN-ExecutionProvider.md
qti-mbadnara: I believe JIT is also supported. Will make the necessary changes

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -669,6 +693,30 @@ QnnEp::QnnEp(QnnEpFactory& factory,
           "enable_htp_prepare_only=1 requires ep.context_enable=1. "
           "prepare_only mode only generates the context model for ahead-of-time compilation.");
     }
+
+    std::string prepare_and_load_str;
```
minfhong-qti: It seems this is from another PR?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1240,6 +1288,44 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+  enable_htp_graph_splitting_ = ParseBoolOption(ort_api,
+                                                session_options_,
+                                                FormatEPConfigKey("enable_htp_graph_splitting"),
+                                                false,
+                                                logger_);
+#ifndef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
+  if (enable_htp_graph_splitting_) {
+    ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
+                "enable_htp_graph_splitting=1 was set but this build was compiled against QAIRT SDK < 2.49. "
+                "Graph splitting is not available and the option will be ignored.");
```
minfhong-qti, line 1278: Prefer to set it false here even though it should not be used later.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mbadnara: Thanks for catching, it leaked from #517

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mbadnara: Done!

minfhong-qti (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1299,26 +1302,60 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
...
+  configs_vec.push_back(&context_config_weight_sharing);
+#endif
+#ifdef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
+  if (enable_htp_graph_splitting) {
+    configs_vec.push_back(&context_config_graph_splitting_vtcm);
+  }
+#endif
+  configs_vec.push_back(nullptr);
+
+#ifdef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
```
qti-ashwshan, line 1335: GPE_KWAY_PARTITIONS is a global environment variable. This block sets it when > 0 but never unsets it when == 0. If a prior session in the same process set it to e.g. 4, a subsequent session with kway_partitions=0 will silently inherit the stale 4 because the env var persists ? Please check that logic and add a test for that as well if possible.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -332,6 +332,19 @@ For more information, see the [Parallel Graph Preparation](#parallel-graph-prepa
...
+
+|`"htp_graphsplitter_num_prepare_threads"`|Description|
+|---|---|
+|Integer string ≥ 1, default `"8"`|Number of threads used to prepare sub-graphs in parallel. Only effective when `enable_htp_graph_splitting=1`. `"0"` means auto (uses `min(hardware_concurrency, number_of_splits)`). Higher values reduce prepare time at the cost of peak CPU/memory usage during preparation.|
+
+|`"htp_graph_splitting_kway_partitions"`|Description|
+|---|---|
+|Integer string, default `"4"`|Number of sub-graphs (k-way partitions) the HTP backend splits the model into during context creation. `"0"` leaves the decision to the SDK. Only effective when `enable_htp_graph_splitting=1`. Equivalent to setting the `GPE_KWAY_PARTITIONS` environment variable; setting this provider option is preferred since it applies per-session without process-wide side effects (note: implementation sets the env var immediately before context creation).|
```
yath1, line 346: Couple of high level questions: 
1. Where can I find the docs for `GPE_KWAY_PARTITIONS`? 
2. Is `enable_htp_graph_splitting` same as setting env variable `GPE_KWAY_PARTITIONS`? If so, why are we exposing this to the user through QNN-EP as a provider option? We can just document that setting `GPE_KWAY_PARTITIONS` would split the graph into k subgraphs and let the user code handle it, isnt it? EP would be agnostic to this flag instead of post-processing it unnecessarily.

qti-mbadnara: Hey @yath1 

 | Option | What it does |
  |---|---|
  | `enable_htp_graph_splitting` | **Enables** graph splitting |
  | `GPE_KWAY_PARTITIONS` | Env var that controls how many partitions (K) to split into when splitting is already enabled |

Without `enable_htp_graph_splitting=1`, `GPE_KWAY_PARTITIONS` has no effect at all

qti-ashwshan (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1299,26 +1302,70 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
...
   RETURN_IF_ERROR(SetQnnContextConfig(context_priority_, context_priority_config));
 
-  const QnnContext_Config_t* configs[] = {&context_priority_config,
+#ifdef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
+  QnnContext_Config_t context_config_graph_splitting_vtcm = QNN_CONTEXT_CONFIG_INIT;
+  QnnHtpContext_CustomConfig_t graph_splitting_custom_config_vtcm;
+  if (enable_htp_graph_splitting) {
+    graph_splitting_custom_config_vtcm.option = QNN_HTP_CONTEXT_CONFIG_OPTION_GRAPH_SPLITTING_CONFIGS;
+    graph_splitting_custom_config_vtcm.graphSplittingConfigs.graphSplittingEnabled = true;
+    graph_splitting_custom_config_vtcm.graphSplittingConfigs.numPrepareThreads = graphsplitter_num_prepare_threads;
```
huaychou, line 1311: Just wondering, will this config be included in QAIRT 2.49 or 2.50?

qti-mbadnara: I believe from 2.50. #698 adds a gate when the option is not present in the SDK

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1299,26 +1302,70 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
   QnnContext_Config_t context_priority_config = QNN_CONTEXT_CONFIG_INIT;
   RETURN_IF_ERROR(SetQnnContextConfig(context_priority_, context_priority_config));
 
-  const QnnContext_Config_t* configs[] = {&context_priority_config,
+#ifdef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
```
minfhong-qti, line 1305: Happen to checking this function. In my understanding, `CreateContextVtcmBackupBufferSharingEnabled` is expected to only be called when loading context binaries.
I would like to confirm that is graph splitting feature a runtime feature?

qti-mbadnara: Thanks for catching this! Have opened #778 to address this!

PR #649, opened 2026-07-17
Layernorm fusion for trailing axis by adding pre and post transpose

nagesh-qti:
### Description
Adding Layernorm fusion for trailing axis by adding pre and post transpose.

Non-trailing axis support — the main feature change. When TryFusion matches a valid LayerNorm pattern, but the axes are non-trailing (e.g., axes={1} on a rank-4 tensor), CreateOrValidateOnQnn now wraps the
  QNN_OP_LAYER_NORM node with:
  - a pre-Transpose that permutes the input to move the normalized dims to the trailing positions, and
  - a post-Transpose that restores the original layout.

Gamma and beta are validated and squeezed against the original ONNX axes — they are never transposed because they describe only the normalized dimensions

### Motivation and Context
The QNN QNN_OP_LAYER_NORM operator only accepts axes that form a trailing suffix of the input tensor's dimensions — for example, axes=[-1] on any rank, or axes=[-2,-1] on a 4D tensor. This is fine for the
  common NLP layout [batch, seq, hidden] where normalization is always over the last axis, but fails for models using NCHW layout [N, C, H, W] where normalization may be over the channel
  axis (axes={1}). Before this patch, such models would silently fall back to executing all 9 decomposed ops individually on HTP instead of being fused into a single LayerNorm, losing the performance and
  precision benefits of the hardware op.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
@@ -595,46 +595,12 @@ static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qmw,
                              std::move(axes_shape), std::move(trailing_axes_copy));
 
   if (validate) {
-    if (needs_transpose) {
-      // Validate pre-transpose: root_input -> ln_input (transposed shape).
-      QnnParamWrapper pre_tp_param(node_units[0]->Index(), node_units[0]->Name(),
-                                   QNN_OP_TRANSPOSE_PARAM_PERM,
-                                   std::vector<uint32_t>{static_cast<uint32_t>(pre_perm.size())},
-                                   std::vector<uint32_t>(pre_perm));
-      RETURN_IF_ERROR(qmw.ValidateQnnNode(node_name + "_pre_tp",
-                                          QNN_OP_PACKAGE_NAME_QTI_AISW,
-                                          QNN_OP_TRANSPOSE,
-                                          {input_tensor.GetQnnTensor()},
-                                          {ln_input_tensor.GetQnnTensor()},
-                                          {pre_tp_param.GetQnnParam()}));
-
-      // Validate LayerNorm with trailing axes on the transposed tensor.
-      RETURN_IF_ERROR(qmw.ValidateQnnNode(node_name,
-                                          QNN_OP_PACKAGE_NAME_QTI_AISW,
-                                          QNN_OP_LAYER_NORM,
-                                          {ln_input_tensor.GetQnnTensor(),
-                                           gamma_tensor.GetQnnTensor(),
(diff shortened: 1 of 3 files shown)


qti-mattsinc:
@nagesh-qti, FYI I also have a change for LN fusion up at #618. Since you are also working on this fusion builder, could you review it?

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -475,6 +440,49 @@ static Ort::Status ValidateAndSqueezeScaleShape(
   return Ort::Status();
 }
 
+// Returns true if the axes form a contiguous trailing suffix of [0, rank).
+// e.g. axes={2,3} on rank-4 is trailing; axes={1} on rank-4 is non-trailing.
+static bool AreAxesTrailing(gsl::span<const uint32_t> axes, size_t rank) {
+  if (axes.empty() || rank == 0) return false;
```
qti-mattsinc: There are a lot of uses of `for`s and `if`s without `{ }` in this PR. Can we change to using braces around the inner expression, even if you want to keep it on the same line? For a consistent style and to avoid goto fail type bugs

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -499,7 +505,40 @@ static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qmw,
   // The validated shape is then squeezed to 1D for QNN.
   std::vector<uint32_t> input_shape;
   RETURN_IF_NOT(qmw.GetOnnxShape(root_input.shape, input_shape), "Cannot get input shape.");
+  const size_t rank = input_shape.size();
+
+  // QNN LayerNorm only supports trailing axes. When the ONNX axes are non-trailing,
```
qti-mattsinc: nit.

```suggestion
  // QNN LayerNorm on HTP only supports trailing axes. When the ONNX axes are non-trailing,
```

onnxruntime/core/providers/qnn/builder/qnn_node_group/layer_norm_fusion.cc
```diff
@@ -499,7 +505,40 @@ static Ort::Status CreateOrValidateOnQnn(QnnModelWrapper& qmw,
...
+  // wrap the op with a pre-Transpose (moves axes to the end) and a post-Transpose (restores layout).
+  const bool needs_transpose = !AreAxesTrailing(axes, rank);
```
qti-mattsinc, line 530: We will want to include a check for `IsNpuBackend`. But that can happen after #618 lands because there will be a couple merge conflicts. Just noting this for now.

PR #651, opened 2026-07-17
[QNN EP] Add support for Attention Op

### LINKED ISSUE #646: QNN EP does not support the standard ONNX Attention op (opset 23/24)
### Describe the issue

The QNN execution provider does not claim the standard ONNX `Attention` operator (`ai.onnx` domain, introduced in opset 23, here opset 24). A model containing it has every `Attention` node assigned to the CPU EP; with CPU fallback disabled, session creation fails. Other opset-23/24 ops on the same model (e.g. `RMSNormalization`, `Gelu`, `MatMul`, `Softmax`, `Transpose`, blocked-`DequantizeLinear` QDQ) compile onto HTP fine, so the fused `Attention` op is the sole blocker for running standard-ONNX transformer decoders on the Hexagon NPU.

### Urgency

Moderate. It forces attention (a large fraction of decoder compute) onto CPU, or requires manually decomposing `Attention` into primitives before compiling for QNN.

### Platform and versions

- OS: Windows 11 ARM64, Snapdragon X (HTP V73)
- onnxruntime: 1.27.0
- onnxruntime-qnn: 2.4.0 (QNN SDK 2.48.40)
- Backend: QnnHtp.dll

### To reproduce

Create a minimal model with a single static-shaped opset-24 `Attention` node (3D Q/K/V `[1,8,64]`, `q_num_heads=4`, `kv_num_heads=4`, `is_causal=1`), register the QNN EP plugin, and create a session with CPU fallback disabled:

```python
import onnxruntime as ort, onnxruntime_qnn as oqnn
ort.register_execution_provider_library("QNNExecutionProvider", oqnn.get_library_path())
npu = [d for d in ort.get_ep_devices()
       if d.ep_name == "QNNExecutionProvider" and d.device.type == ort.OrtHardwareDeviceType.NPU]
so = ort.SessionOptions()
so.add_provider_for_devices(npu, {})
so.add_session_config_entry("session.disable_cpu_ep_fallback", "1")
ort.InferenceSession("attention24.onnx", so)
```

### Actual result

```
FAIL : This session contains graph nodes that are assigned to the default CPU EP,
but fallback to CPU EP has been explicitly disabled by the user.
```

With fallback enabled, the `Attention` node silently runs on CPU while the surrounding ops run on HTP.

### Expected result

The QNN EP claims the standard ONNX `Attention` op and lowers it to the HTP backend (as it does for `RMSNormalization` / `Gelu` / `MatMul` / `Softmax`).

### Notes

`RMSNormalization` (opset 24), `Gelu` (opset 20), `MatMul`, `Softmax`, `Transpose`, and blocked-`DequantizeLinear` QDQ all compile onto HTP on this same setup, so support for the fused `Attention` op is the remaining gap. As a workaround the op can be decomposed into `Reshape` / `Transpose` / `MatMul` / `Softmax` / `Add` (+ `Tile` for GQA), which does run on HTP.

qti-mbadnara:
## Description

Adds `AttentionOpBuilder` to the QNN EP, enabling the standard ONNX `ai.onnx::Attention` operator (opset 23 and 24) to run on the Hexagon HTP and Adreno GPU backends. The builder decomposes the fused Attention node into elementary QNN ops for HTP, and for the GPU backend routes to the native `QNN_OP_GROUP_QUERY_ATTENTION` kernel when possible.

---

## Motivation & Context

The QNN EP had no registration for `ai.onnx::Attention`. Every Attention node fell back to CPU EP; with `session.disable_cpu_ep_fallback=1` session creation failed outright:

```
FAIL : This session contains graph nodes that are assigned to the default CPU EP,
but fallback to CPU EP has been explicitly disabled by the user.
```

---

## Routing

The builder uses a two-track strategy based on backend and node attributes:

| Condition | Path |
|---|---|
| GPU backend + `n_q % n_kv == 0` + `is_causal=1` + no `softcap`/`attn_mask`/`qk_output` + `present_key` declared | **Native** `QNN_OP_GROUP_QUERY_ATTENTION` |
| All other cases (HTP, non-causal, softcap, attn_mask, no present_key) | **Decomposition** into elementary QNN ops |

The native path covers both MHA (`n_q == n_kv`) and GQA/MQA (`n_q > n_kv`). Both 3D BSH and 4D BNSH inputs are supported: for 4D, `Transpose+Reshape` nodes are inserted before Q/K/V and a `Reshape+Transpose` is inserted after Y to bridge the layout.

**Why decomposition for HTP:** QNN SDK ≤ 2.49 has no MHA primitive for HTP. `QNN_OP_GROUP_QUERY_ATTENTION` is GPU-only (Adreno). The ONNX `Attention` spec defines a reference body of `Scale + MatMul + Softmax + MatMul`; this PR implements that body as QNN node sequences.

**Why decomposition on GPU for some cases:** The native `QNN_OP_GROUP_QUERY_ATTENTION` does not support `softcap` or an additive `attn_mask` input, and is always causal. Cases that need these features fall to the decomposition path, which runs on GPU using individual QNN ops.

---

## GPU Native Path (`QNN_OP_GROUP_QUERY_ATTENTION`)

A single node is emitted with:
- **Params**: `NUM_HEADS`, `KV_NUM_HEADS`, `SCALE`, `DO_ROTARY=0`
- **Inputs**: query (BSH), synthetic `seqlens_k` (INT32 `[B]` = `S_past + S_k − 1`), synthetic `total_seq_len` (INT32 0D scalar = `S_past + S_k`), key, value, past_key/past_value (4D BNSH, passed directly)
- **Outputs**: Y (BSH), present_key / present_value (4D BNSH, when KV cache is active)

`seqlens_k` and `total_seq_len` are synthesized from static input shapes; they are not present in the ONNX `ai.onnx::Attention` signature but required by QNN GQA.

For 4D BNSH inputs, the builder transparently inserts `Transpose(0,2,1,3)+Reshape` before Q/K/V (converting `[B,n,S,hs]` → `[B,S,n*hs]`) and `Reshape+Transpose(0,2,1,3)` after Y (converting `[B,S,n*hs]` → `[B,n,S,hs]`). past_key/past_value are already BNSH and are passed straight through.

---

### 3D inputs `[B, S, n·hs]` (BSH layout)

| Step | QNN Op | Output shape |
|------|--------|-------------|
| 1 | Static `sqrt(scale)` scalar | — |
| 2 | `ElementWiseMultiply` Q·√s | `[B, S_q, n_q·hs]` |
| 3 | `ElementWiseMultiply` K·√s *(no KV cache)* | `[B, S_k, n_kv·hs]` |
| 4–5 | `Reshape` + `Transpose(0,2,1,3)` Q → BNSH | `[B, n_q, S_q, hs]` |
| 6–7 | `Reshape` + `Transpose(0,2,1,3)` K → BNSH | `[B, n_kv, S_k, hs]` |
| 8 | KV cache: `Concat(past_key, K, axis=2)` *(cache only)* | `[B, n_kv, S_total, hs]` |
| 9 | GQA expand K: `Reshape→Tile→Transpose→Reshape` *(GQA only)* | `[B, n_q, S_k, hs]` |
| 10 | `ElementWiseMultiply` K·√s *(KV cache path only, after concat+expand)* | `[B, n_q, S_total, hs]` |
| 11 | `MatMul(Q, K^T, transpose_in1=true)` | `[B, n_q, S_q, S_k]` |
| 12 | Softcap: `ElementWiseDivide→Tanh→ElementWiseMultiply` *(if softcap!=0)* | `[B, n_q, S_q, S_k]` |
| 13 | Static causal mask `ElementWiseAdd` *(is_causal=1)* | `[B, n_q, S_q, S_k]` |
| 14 | User `attn_mask ElementWiseAdd` *(if provided)* | `[B, n_q, S_q, S_k]` |
| 15 | `Softmax(axis=3)` | `[B, n_q, S_q, S_k]` |
| 16–17 | `Reshape` + `Transpose(0,2,1,3)` V → BNSH | `[B, n_kv, S_k, v_hs]` |
| 18 | KV cache: `Concat(past_value, V, axis=2)` *(cache only)* | `[B, n_kv, S_total, v_hs]` |
| 19 | GQA expand V *(GQA only)* | `[B, n_q, S_k, v_hs]` |
| 20 | `MatMul(attn_weights, V)` | `[B, n_q, S_q, v_hs]` |
| 21–22 | `Transpose(0,2,1,3)` + `Reshape` Y | `[B, S_q, n_q·v_hs]` |

The `qk_matmul_output_mode` capture points (per published ONNX 1.23 spec — note that `cmake/external/onnx` is pinned at v1.20.1 pre-errata which has the opposite ordering):
- mode 0 — raw QK scores (after step 11, before softcap)
- mode 1 — post-softcap, pre-mask (after step 12)
- mode 2 — post-softcap+mask, pre-softmax (after step 14)
- mode 3 — post-softmax / attn_weights (after step 15)

For **4D inputs** `[B, n, S, hs]` (BNSH layout): steps 4–7 and 21–22 are skipped — inputs are already in BNSH format.

**Operation order — softcap before masks (per ONNX 1.23 spec):** Step 12 applies softcap to the raw QK scores _before_ masks. Applying masks before softcap would clamp `-1e9` causal mask values to `-softcap`, making masked positions visible to softmax.

**GQA head expansion** (floor-division semantics matching the ONNX function body):
- `K[b, kv*ratio+r, s, h] = K_orig[b, kv, s, h]` → `[K0,K0,...,K1,K1,...]`
- 4D approach (avoids 5D HTP limitations): `Reshape[B,1,n_kv,S*hs]→Tile[1,ratio,1,1]→Transpose(0,2,1,3)→Reshape[B,n_q,S,hs]`
- GQA expansion runs **after** the KV concat so that `present_key`/`present_value` retain `kv_num_heads` per the ONNX spec.

**Causal mask** (`is_causal=1`): Static `[1, 1, S_q, S_k]` lower-triangular float tensor (broadcast over `[B, n_q, S_q, S_k]` in ADD), values `0.0` for `j ≤ i+offset`, `-1e9` otherwise. `-inf` is avoided because HTP V73 HVX does not reliably propagate IEEE-754 `-inf` through ADD. Offset = `S_past` when KV cache is active, `0` otherwise.

**Softcap** (`softcap != 0`): `ElementWiseDivide(scores, sc) → QNN_OP_TANH → ElementWiseMultiply(result, sc)`.

**KV cache ordering:** `present_key = Concat(past_key, K_raw)` at `kv_num_heads` heads, then GQA-expanded to `n_q` heads, then scaled by √s. This ensures past and current keys are treated uniformly, and `present_key`/`present_value` outputs carry `kv_num_heads` as the spec requires.

---

## Supported Configuration

| Feature | HTP | GPU |
|---------|-----|-----|
| Input rank | 3D BSH or 4D BNSH | 3D BSH or 4D BNSH |
| Opsets | 23, 24 | 23, 24 |
| Dtypes | float32, float16 | float32 |
| MHA (`n_q == n_kv`, `is_causal=1`) | Decomposition ✓ | Native `QNN_OP_GROUP_QUERY_ATTENTION` ✓ |
| GQA / MQA (`n_q > n_kv`, `is_causal=1`) | Decomposition ✓ | Native `QNN_OP_GROUP_QUERY_ATTENTION` ✓ |
| Any attention with `is_causal=0` | Decomposition ✓ | Decomposition ✓ |
| `scale` | Any positive float ✓ | Any positive float ✓ |
| `attn_mask` (float additive bias) | ✓ | Decomposition ✓ |
| `softcap` | via `QNN_OP_TANH` ✓ | via `QNN_OP_TANH` (decomposition) ✓ |
| KV cache (`past_key`/`present_key`) | ✓ | ✓ (native path includes past/present) |
| `qk_matmul_output` modes 0–3 | ✓ | Decomposition ✓ |
| Shape | Static only | Static only |

---

## Unsupported / Rejected

- Dynamic shape dimensions on Q, K, or V
- `scale <= 0`
- `nonpad_kv_seqlen` — padding-mask computation not implemented; rejected outright
- `softmax_precision` — cross-dtype softmax accumulation not implemented; rejected outright
- `past_key` without `present_key` (or vice versa)
- Boolean `attn_mask` — spec requires `Where(mask, 0, -inf)` conversion; builder emits raw ADD, so bool masks are rejected. Pre-convert to a float additive bias before passing to this EP.
- `double` and `bfloat16` dtypes — only float32 and float16 are supported; other types would silently corrupt scalar buffer helpers (causal mask, sqrt_scale, softcap)


qti-mattsinc:
@qti-mbadnara, can we map onnx Attention onto QNN_OP_GROUP_QUERY_ATTENTION for the GPU backend?

> Why decomposition, not a native QNN op:
QNN SDK ≤ 2.49 has no MHA primitive for HTP. QNN_OP_GROUP_QUERY_ATTENTION is GPU-only (Adreno). The ONNX Attention spec defines a reference body of Scale + MatMul + Softmax + MatMul; this PR implements that body as QNN node sequences.

HTP may not support this, but GPU can. And perf on GPU using the QNN GQA op will be better than the decomposed sequence.

johnpaultaken:
I agree that the onnx Attention need to map to QNN_OP_GROUP_QUERY_ATTENTION in the normal case. It can resort to this splitting only for the special HTP case temporarily because it would be beneficial for HTP also to implement GQA eventually. I guess the onnx document specifies splitting Attention op to Scale + MatMul + Softmax + MatMul only as a temporary measure until implementations can catch up with the full Attention op kernels. Afaik the GQA is as generic as the new ONNX Attention op. Still we can also add the new ONNX Attention op to QNN if desired. Please let us know your thoughts.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -207,6 +236,213 @@ Ort::Status AttentionOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper
...
+static bool ShouldUseNativeGQA(QnnBackendType backend,
+                               size_t q_rank,
+                               uint32_t n_q, uint32_t n_kv,
+                               int64_t is_causal,
+                               float softcap,
+                               bool has_attn_mask,
+                               bool has_qk_output) {
+  return IsGpuBackend(backend) &&
+         q_rank == 3 &&                    // QNN GQA uses BSH; 4D BNSH → decomposition
+         n_q > n_kv && n_q % n_kv == 0 &&  // true GQA per https://arxiv.org/pdf/2305.13245
```
qti-mattsinc: I don't think we need the `n_q > n_kv` check.

`QNN_OP_GROUP_QUERY_ATTENTION` doesn't impose that restriction and explicitly calls out that MHA is expressible through the op.

> kv_num_heads
> 
> Number of key and value attention heads. Must be a divisor of num_heads. Equal to num_heads for standard multi-head attention; less for grouped-query attention.

onnxruntime/test/providers/qnn/attention_test.cc
```diff
@@ -554,6 +554,181 @@ TEST_F(QnnHTPBackendTests, Attention_DebugOutput_Mode3) {
 
 #endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
 
+// ===========================================================================
+// GPU Attention tests
+//
+// Two categories:
+//   Native   — GPU + true GQA (n_q > n_kv, divisible) + is_causal=1
```
qti-mattsinc, line 558: Same as other comment regarding the exclusion of `n_q == n_kv`

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -207,6 +236,213 @@ Ort::Status AttentionOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper
...
 
+// ---------------------------------------------------------------------------
+// GPU native GQA path
+//
+// When backend=GPU, input layout is 3D BSH, the node is true GQA
+// (n_q > n_kv, divisible), is_causal=1, and no features QNN GQA cannot
+// express (softcap/attn_mask/qk_output), a single
+// QNN_OP_GROUP_QUERY_ATTENTION node is emitted instead of the decomposed
+// graph.  All other cases (MHA, non-causal, softcap, HTP, 4D inputs) fall
+// through to decomposition unchanged.
```
qti-mattsinc: Same as other comment regarding the exclusion of `n_q == n_kv`

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -207,6 +236,213 @@ Ort::Status AttentionOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper
...
+
+static bool ShouldUseNativeGQA(QnnBackendType backend,
+                               size_t q_rank,
+                               uint32_t n_q, uint32_t n_kv,
+                               int64_t is_causal,
+                               float softcap,
+                               bool has_attn_mask,
+                               bool has_qk_output) {
+  return IsGpuBackend(backend) &&
+         q_rank == 3 &&                    // QNN GQA uses BSH; 4D BNSH → decomposition
```
qti-mattsinc: I wonder if we can support the 4D case by inserting reshape/transpose ops. Can be left for future work, though

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1548 @@
...
+// Decomposition builder for ai.onnx::Attention (opset 23/24).
+//
+// 4D inputs [B, n, S, hs] (BNSH layout, no reshape needed)
+// 3D inputs [B, S, n*hs] (BSH layout, reshape + transpose to BNSH)
+// GQA/MQA, KV cache (past/present), softcap, qk_matmul_output
+//
+// The SDK version guard matches GroupQueryAttentionOpBuilder so that this class
+// is compiled away when building against an SDK without the required op-set
+// version metadata.
+#if !(QNN_OPSET_VERSION_MAJOR < 2 || (QNN_OPSET_VERSION_MAJOR == 2 && QNN_OPSET_VERSION_MINOR <= 11))
```
qti-mattsinc: It would be better for this guard only to surround the native QNN GQA path and no the whole op builder. The decomposition path doesn't require opset 2.12 because it won't produce `QNN_OP_GROUP_QUERY_ATTENTION`

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1548 @@
...
+  TensorInfo q_info{};
+  RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(inputs[0], q_info));
+  const size_t q_rank = q_info.shape.size();
+  RETURN_IF(q_rank != 3 && q_rank != 4,
+            "Attention: Q input must be rank 3 ([B,S_q,n_q*hs]) or rank 4 ([B,n_q,S_q,hs])");
+
+  // ---- Reject dynamic (0-dim) shapes on Q, K, V ----
+  for (uint32_t d : q_info.shape) {
+    RETURN_IF(d == 0,
+              "Attention: Q input contains a dynamic (0) dimension; only static shapes are supported");
```
qti-mattsinc: Is it necessary to check for dynamic shapes in the op builder? Won't this be caught elsewhere in the EP?

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1548 @@
...
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(t)),
+                  ("Failed to add seqlens_k: " + seqlens_k_name).c_str());
+  }
+
+  total_seq_len_name = utils::UniqueNameGenerator().New(node_unit, "_gqa_total_seq_len");
+  {
+    std::vector<uint8_t> bytes(sizeof(int32_t));
+    *reinterpret_cast<int32_t*>(bytes.data()) = total_val;
+    // 0D shape (empty dims vector) — QNN requires a scalar, same override used
+    // by GroupQueryAttentionOpBuilder for com.microsoft::GroupQueryAttention.
```
qti-mattsinc: I wonder if we should consider fixing the root of the issue in `GetOnnxShape`, which is the reason why scalar inputs like `total_sequence_length` are being set to rank 1 tensors and then requiring workarounds like this in the op builders:

https://github.com/onnxruntime/onnxruntime-qnn/blob/3e0c61b090a8710b43c6694353c43ccf4d9d32d0/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc#L716-L720

Removing it could break different op builders, though. Might be worth investigating in a follow up

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1548 @@
...
+    TensorInfo past_k_info{};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(onnx_inputs[4], past_k_info));
+    S_past = past_k_info.shape[2];  // [B, n_kv, S_past, hs]
+  }
+
+  auto AddNull = [&](const char* suffix) -> Ort::Status {
+    const std::string name = utils::UniqueNameGenerator().New(node_unit, suffix);
+    input_names.push_back(name);
+    // Use QNN_DATATYPE_UNDEFINED — matches GroupQueryAttentionOpBuilder's null tensor
+    // pattern. QNN_DATATYPE_FLOAT_32 (from MakeNull) causes validator error 3110.
```
qti-mattsinc: > matches GroupQueryAttentionOpBuilder's null tensor pattern

But the GQA op builder *is* using `MakeNull`:

https://github.com/onnxruntime/onnxruntime-qnn/blob/3e0c61b090a8710b43c6694353c43ccf4d9d32d0/onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc#L158

> QNN_DATATYPE_FLOAT_32 (from MakeNull) causes validator error 3110

Really? Why don't we see this when using `MakeNull` in the GQA op builder?

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1548 @@
...
+}
+
+// Emit a single QNN_OP_GROUP_QUERY_ATTENTION node (GPU native path).
+// For 4D BNSH inputs: inserts Transpose(0,2,1,3)+Reshape before Q/K/V to
+// produce the 3D BSH layout QNN GQA expects, and Reshape+Transpose after
+// the Y output to restore the 4D BNSH shape expected by the ONNX graph.
+// past_key/past_value are always 4D BNSH and are passed straight through.
+// Output shapes are derived from input shapes — not from GetTensorInfo on
+// outputs — because the test models do not run ONNX shape inference on
+// ai.onnx::Attention and the output value_info has no shape.
```
qti-mattsinc: I would have thought that by the time the EP gets the graph from ORT core, all tensor shapes are resolved (even if the onnx file itself didn't have shape inference run on it prior). I don't see other op builders with this caveat, and I don't see why this would be unique to Attention?

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1548 @@
...
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(QnnTensorWrapper::MakeNull(gqa_y_name)),
+                  "Failed to add null Y output.");
+  }
+  output_names.push_back(gqa_y_name);
+
+  // present_key [B, n_kv, S_total, hs] and present_value [B, n_kv, S_total, v_hs]
+  // QNN GPU GQA requires real (non-null) tensors for all 3 output slots.
+  // When the ONNX model does not declare present_key/present_value, allocate
+  // private NATIVE scratch tensors so QNN can write to them — the caller
+  // never reads these, but the op validator requires them to exist.
```
qti-mattsinc: nit. I think it would be a bit better to phrase it like this:

```suggestion
  // When the ONNX model does not declare present_key/present_value, allocate
  // private NATIVE scratch tensors so QNN can write to them — the caller
  // never reads these, but the QNN op def for GQA requires them.
```

Also, the fact that the QNN op def marks past_key and past_value as optional, while requiring present_key and present_value seems like an oversight, esp. given that the ORT contrib op def allows the present KV$ to be optional. I'll ask the QNN API team about it.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1548 @@
...
+                          std::vector<uint32_t>(*cache_shapes[i - 1]));
+      RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(pw)),
+                    "Failed to add private cache scratch tensor.");
+    }
+  }
+
+  // ---- Emit QNN_OP_GROUP_QUERY_ATTENTION ----
+  // Validation is intentionally left on (do_op_validation passed through).
+  // If past_key/past_value are STATIC initializers the GPU validator returns
+  // error 3110 and IsOpSupported fails — the node falls to CPU EP rather than
```
qti-mattsinc: That's a bit surprising to me, especially if there is a runtime crash. Can you file a GPU backend ticket with the full details so that we can fix it?

onnxruntime/test/providers/qnn/attention_test.cc
```diff
@@ -0,0 +1,700 @@
...
+
+// ---------------------------------------------------------------------------
+// Native GQA path — requires KV cache with APP_WRITE (dynamic) past_key/past_value.
+// GPU QNN GQA kernel needs live cache buffers; STATIC initializers are rejected.
+// ---------------------------------------------------------------------------
+
+// GQA 3D with KV cache, causal — native QNN_OP_GROUP_QUERY_ATTENTION.
+// past_key/past_value are dynamic (is_initializer=false) → APP_WRITE as required by GPU.
+// Q [1,8,32] (4 heads × 8), K/V [1,8,16] (2 heads × 8), S_past=4.
+TEST_F(QnnGPUBackendTests, Attention_GPU_GQA_3D_Native_KVCache) {
```
qti-mattsinc: Looks like this is failing in CI because the GPU backend doesn't support unpacked QKV in 2.48. It should be supported in 2.50, so you can disable this UT for now.

onnxruntime/test/providers/qnn/attention_test.cc
```diff
@@ -0,0 +1,700 @@
...
+           TestInputDef<float>({1, 8, 16}, false, -1.0f, 1.0f)},
+          {test::MakeAttribute("q_num_heads", static_cast<int64_t>(4)),
+           test::MakeAttribute("kv_num_heads", static_cast<int64_t>(2)),
+           test::MakeAttribute("is_causal", static_cast<int64_t>(0))}),
+      opts, /*opset_version=*/24,
+      EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(1e-3f)});
+}
+
+// GQA 3D with softcap — no softcap param in QNN GQA → decomposition.
+TEST_F(QnnGPUBackendTests, Attention_GPU_GQA_3D_Decompose_Softcap) {
```
qti-mattsinc, line 670: This test is failing in CI with an accuracy issue. If it looks like a GPU backend issue, then you can disable this UT for now and create a ticket.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Yes fixed it!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Yes, agreed will remove this!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Yup fixed it!

onnxruntime/test/providers/qnn/attention_test.cc
qti-mbadnara: Sounds good!

onnxruntime/test/providers/qnn/attention_test.cc
qti-mbadnara: Yes will disable it for not if the failure still exists after the fix!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Done!

onnxruntime/test/providers/qnn/attention_test.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Yup will add support!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Yeahh makes sense, will move this!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Will fix the implementation and raise a corresponding GPU ticket if needed!

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mattsinc: Actually, now that I check the QNN op def, it looks like present_key and present_value are both marked correctly as optional. So, why do we need this workaround to provide scratch tensors?

> but the op validator requires them to exist.

I guess the original comment here was referring to the GPU backend validator and not the Core validator for GQA? It's true that the GPU backend currently doesn't support GQA w/o KV$ (it's a new op so we will probably add this support soon), but I don't think we should try to work around this in the op builder. But maybe I am missing something.

qti-mattsinc: (Re-reading the original comment now, I see that yes it was referring to the GPU validator.) The reason I'm hesitant to have this workaround in the op builder is that KV$ tensors can be quite large. And if we are running a non-iterative model that isn't using a KV$ for it's attention blocks, then by doing this workaround we could be 'tricking' the GPU backend into allocating and writing to a lot of unused memory. My feeling is that it would be better to leave the workaround out and just get this absent-KV$ support added to QnnGpu.

What do you think?

qti-mattsinc (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
qti-mbadnara: Sounds good, agree with you. Make sense fixing the GPU Backend support rather than adding workarounds in ORT QNN EP

qti-mattsinc (changes requested):
Please fix decomposition ordering with softcap

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1541 @@
...
+                                    /*is_graph_output=*/false, do_op_validation));
+    scores_cur = attn_masked_out;
+  }
+
+  // ---- qk_matmul_output mode 1: post-mask, pre-softcap ----
+  if (has_qk_output && qk_mode == 1) {
+    qk_captured = scores_cur;
+  }
+
+  // ---- Softcap (applied AFTER masks per ONNX spec) ----
```
qti-mattsinc: @qti-mbadnara, Softcap should go before the mask addition according to the onnx spec:

```
The following pattern is applied by this operator:
      Q          K          V
      |          |          |
Q*sqrt(scale) K*sqrt(scale) |
      |          |          |
      |       Transpose     |
      |          |          |
      ---MatMul---          |
            |               |
  softcap (if provided)     |
            |               |
 at_mask---Add              |
            |               |
         Softmax            |
            |               |
            -----MatMul------
                   |
                   Y
```

This is why the GPU softcap test was failing; it wasn't a backend issue.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1539 @@
...
+  const std::string qk_out = utils::UniqueNameGenerator().New(node_unit, "_qk_out");
+  const std::vector<uint32_t> qk_shape = {B, n_q, S_q, S_k};
+  RETURN_IF_ERROR(AddMatMulNode(qnn_model_wrapper, node_unit,
+                                q_cur, k_cur, qk_out,
+                                qk_shape, dtype, q_quant,
+                                /*transpose_in1=*/true, do_op_validation));
+  std::string scores_cur = qk_out;
+
+  // ---- qk_matmul_output mode 0: raw post-QK scores (pre-mask, pre-softcap) ----
+  // Per ONNX spec the pipeline is: QK → masks → softcap → softmax
```
qti-mattsinc: nit.

```suggestion
  // Per ONNX spec the pipeline is: QK → softcap → masks → softmax
```

qti-mattsinc (approved):
LGTM

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+              for (uint32_t j = 0; j < S_k; ++j) {
+                const size_t idx = ((static_cast<size_t>(b) * n_q + h) * S_q + i) * S_k + j;
+                mask_ptr[idx] = (j <= i + offset) ? static_cast<uint16_t>(0u) : neg_raw;
+              }
+            }
+          }
+        }
+      } else {
+        constexpr float large_neg = -1e9f;
+        mask_bytes.resize(total * sizeof(float));
```
qti-yuduo, line 1384: **Blocking — buffer undersize for `double`.**

This `else` writes 4 bytes/elem, but the tensor below is declared with the real `dtype`. `T1` is `all_float_types_ir4()` = {bfloat16, float16, float, double}, and the EP maps both through (`qnn_utils.cc:896`: DOUBLE→`FLOAT_64`, 8 bytes/elem). So a fp64 model declares a tensor twice the size of its backing buffer. bf16 gets a wrong bit pattern.

Nothing rejects these — `IsOpSupported` has no dtype check. Note adding `ProcessDataTypes` would not help: `CheckCpu/Htp/GpuDataTypes` are all unconditional no-ops (`base_op_builder.cc:93-104`).

Suggest gating to fp16/fp32 in `IsOpSupported`. Same pattern at L723 and L1074.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+    RETURN_IF_ERROR(AddBinaryOpNode(qnn_model_wrapper, node_unit,
+                                    QNN_OP_ELEMENT_WISE_ADD,
+                                    scores_cur, causal_mask_name, masked_out,
+                                    qk_shape, dtype, q_quant,
+                                    /*is_graph_output=*/false, do_op_validation));
+    scores_cur = masked_out;
+  }
+
+  // ---- User attention mask: ADD ----
+  if (has_attn_mask) {
```
qti-yuduo, line 1413: **Blocking — boolean `attn_mask` is added as 0/1.**

Type `U` admits `tensor(bool)`, and the spec converts it first: `Where(attn_mask, ScalarZero, FloatNegInf)` (`onnx/defs/nn/utils.cc:174`). This emits a raw `ELEMENT_WISE_ADD`, so a bool mask shifts masked logits by -1 instead of excluding them.

BOOL→`BOOL_8` maps through (`qnn_utils.cc:897`) and no dtype gate stops it. This one is not caught by op validation — Add on a correctly-shaped `BOOL_8` tensor is a *valid* QNN graph that computes the wrong result, so it fails silently.

Either convert via Select, or reject bool masks.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+                                                       k_transposed_shape,
+                                                       dtype, k_quant,
+                                                       do_op_validation,
+                                                       /*is_for_input=*/false));
+    k_cur = k_transposed;
+  }
+  // k_cur is now [B, n_kv, S_k, hs] in BNSH layout.
+
+  // ---- GQA expansion of K (if GQA) ----
+  if (is_gqa) {
```
qti-yuduo: **GQA expansion runs before the KV concat**, so `k_cur` has `n_q` heads while `past_key` has `kv_num_heads`. Axis-1 sizes disagree whenever `is_gqa` (L1016), making the L1178 concat invalid.

Spec declares `past_key`/`present_key` with `kv_num_heads` (`defs.cc:3470-3481`) and expands *after* the Concat for exactly this reason.

Not silent — `AddKVConcatNode` honors `do_op_validation`, so GQA+KV-cache is rejected at GetCapability and falls back to CPU. Still a perf cliff with a misleading reason. Suggest moving the expansion after the concat, or rejecting `is_gqa && has_past_key` explicitly.

Uncovered by tests: the only KV test is `Attention_KVCache_4D` with `n_q == n_kv`, and the GQA+KV test is `DISABLED_`.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+                                     head_ratio,
+                                     do_op_validation));
+    k_cur = k_expanded;
+  }
+
+  // ---- KV cache concat for K ----
+  std::string k_present_name;
+  if (has_past_key) {
+    const uint32_t S_k_total = S_past + S_k;
+    const std::vector<uint32_t> k_present_shape = {B, n_q, S_k_total, hs};
```
qti-yuduo: `present_key` declared with `n_q`; spec says `kv_num_heads` (`defs.cc:3498-3511`). Coincides only when `n_q == n_kv`. Same at L1255, L1484, L1496.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+  const std::vector<uint32_t> qk_shape = {B, n_q, S_q, S_k};
+  RETURN_IF_ERROR(AddMatMulNode(qnn_model_wrapper, node_unit,
+                                q_cur, k_cur, qk_out,
+                                qk_shape, dtype, q_quant,
+                                /*transpose_in1=*/true, do_op_validation));
+  std::string scores_cur = qk_out;
+
+  // ---- qk_matmul_output mode 0: raw post-QK scores (pre-mask, pre-softcap) ----
+  // Per ONNX spec the pipeline is: QK → softcap → masks → softmax
+  // mode 0 = raw QK, mode 1 = post-mask pre-softcap,
```
qti-yuduo: This comment says mode 1 is "post-mask pre-softcap", but the code correctly does **post-softcap, pre-mask** — matching the current spec ("after the softcap operation (before mask addition)").

The prose is the pre-errata semantics. Worth fixing plus citing the **1.23 errata** and noting `cmake/external/onnx` is pinned pre-errata — otherwise the next reader greps the submodule and "fixes" correct code back to broken. That is exactly what I did on first pass.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+  // mode 2 = post-softcap,  mode 3 = post-softmax.
+  std::string qk_captured;  // The intermediate captured for qk_matmul_output.
+  if (has_qk_output && qk_mode == 0) {
+    qk_captured = scores_cur;
+  }
+
+  // ---- Softcap BEFORE masks (per ONNX spec) ----
+  // From the ONNX spec function body: MatMul → softcap → attn_mask+Add → Softmax.
```
qti-yuduo, line 1331: Ordering here matches the published spec but not the vendored submodule. Worth saying so explicitly.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+                mask_ptr[idx] = (j <= i + offset) ? static_cast<uint16_t>(0u) : neg_raw;
+              }
+            }
+          }
+        }
+      } else {
+        constexpr float large_neg = -1e9f;
+        mask_bytes.resize(total * sizeof(float));
+        float* mask_ptr = reinterpret_cast<float*>(mask_bytes.data());
+        for (uint32_t b = 0; b < B; ++b) {
```
qti-yuduo: Mask content is identical across `B` and `n_q`, but a full `[B, n_q, S_q, S_k]` STATIC tensor is materialized. At B=1, n_q=32, S=2048 that is ~537 MB baked into the context binary; `[1, 1, S_q, S_k]` broadcasts identically at 1/32 the size.

Also `-1e9f` rather than `-inf` means a fully-masked row yields a uniform distribution instead of zeros. Looks deliberate (avoids NaN) — worth a comment.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+static Ort::Status AddGQAExpandNode(QnnModelWrapper& qnn_model_wrapper,
+                                    const OrtNodeUnit& node_unit,
+                                    const std::string& in_name,
+                                    const std::string& out_name,
+                                    const std::vector<uint32_t>& in_shape,   // [B, n_kv, S, hs]
+                                    const std::vector<uint32_t>& out_shape,  // [B, n_q,  S, hs]
+                                    Qnn_DataType_t dtype,
+                                    const QnnQuantParamsWrapper& quant_param,
+                                    uint32_t head_ratio,
+                                    bool /*do_op_validation*/) {
```
qti-yuduo: `do_op_validation` is accepted and discarded — all four sub-nodes hardcode `false`. So the GQA expansion is never backend-validated, which is part of why the L1161 mismatch surfaces at the concat instead of here. Either honor it or drop the parameter.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+  //   [0]=Q [1]=K [2]=V [3]=attn_mask (if present) [4]=past_key (if present) ...
+  const bool has_attn_mask = (onnx_inputs.size() > 3 && onnx_inputs[3].Exists());
+  const std::string& q_in = input_names[0];
+  const std::string& k_in = input_names[1];
+  const std::string& v_in = input_names[2];
+  // Offset into input_names for optional inputs (computed by walking present flags).
+  size_t opt_idx = 3;
+  const std::string attn_mask_in = has_attn_mask ? input_names[opt_idx++] : std::string{};
+  const std::string past_key_in = has_past_key ? input_names[opt_idx++] : std::string{};
+  const std::string past_value_in = has_past_key ? input_names[opt_idx++] : std::string{};
```
qti-yuduo, line 1069: `past_value_in` gated on `has_past_key`, while `ProcessInputs` (L223) gates registration on `inputs[5].Exists()`. `has_past_value` is computed at L79 and unused. Agrees today only because `IsOpSupported` enforces both-or-neither; relaxing that validator would silently corrupt this `opt_idx` walk.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+  // input[5] = past_value (optional, KV cache)
+  if (onnx_inputs.size() > 5 && onnx_inputs[5].Exists()) {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, onnx_inputs[5], logger, input_names));
+  }
+  // input[6] = nonpad_kv_seqlen (optional, opset 24)
+  if (onnx_inputs.size() > 6 && onnx_inputs[6].Exists()) {
+    RETURN_IF_ERROR(ProcessInput(qnn_model_wrapper, onnx_inputs[6], logger, input_names));
+  }
+
+  return Ort::Status();
```
qti-yuduo, line 256: `nonpad_kv_seqlen` is registered as a QNN input but never consumed, so the spec's padding mask is silently skipped. The PR description says it is rejected — it isn't. Same for `softmax_precision` (zero occurrences in the file; spec uses it at `defs.cc:3722-3724`). Suggest rejecting both in `IsOpSupported`.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: MIT
+
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#include <numeric>
```
qti-yuduo: `<numeric>` unused (no `accumulate`/`iota`).

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(QnnTensorWrapper::MakeNull(gqa_y_name)),
+                  "Failed to add null Y output.");
+  }
+  output_names.push_back(gqa_y_name);
+
+  // present_key and present_value (output slots 1-2).
+  // QNN GPU GQA requires real (non-null) tensors for all 3 output slots.
+  // When the ONNX model does not declare present_key/present_value, allocate
+  // private NATIVE scratch tensors so QNN can write to them — the caller
+  // never reads these, but the QNN op def for GQA requires them.
```
qti-yuduo: Duplicated comment block — the second statement contradicts the first.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+
+#else  // SDK < 2.12: no QNN_OP_GROUP_QUERY_ATTENTION — always use decomposition
+
+static bool ShouldUseNativeGQA(QnnBackendType, uint32_t, uint32_t,
+                               int64_t, float, bool, bool, bool) {
+  return false;
+}
+
+Ort::Status AttentionOpBuilder::ProcessInputsNativeGQA(const AttentionOpBuilder&,
+                                                       QnnModelWrapper&,
```
qti-yuduo, line 598: This pre-2.48 stub returns success without emitting a node. Unreachable today since `ShouldUseNativeGQA` is false here, but a silent-success codegen stub is one refactor from an empty graph. Suggest returning an error.

onnxruntime/core/providers/qnn/builder/opbuilder/attention_op_builder.cc
```diff
@@ -0,0 +1,1523 @@
...
+                                                {},
+                                                do_op_validation),
+                "Failed to create output identity Reshape node.");
+  return Ort::Status();
+}
+
+// ---------------------------------------------------------------------------
+// ProcessAttributesAndOutputs — emit the full decomposed attention graph
+// ---------------------------------------------------------------------------
+Ort::Status AttentionOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_model_wrapper,
```
qti-yuduo, line 982: This function runs to L1516 — **557 lines** (file is 1523 vs the 800-line guideline). It is a linear pipeline that splits cleanly along the seams already commented: prepare-QK / apply-masks / emit-outputs.

Related: much of `EmitNativeGQANode` duplicates the pre-existing `group_query_attention_op_builder.cc`, which does the slot mapping declaratively (`constexpr auto qnn_idx_to_onnx = std::array{...}` + one `MakeNull` loop) and already has the `total_sequence_length` 0D override and a `seqlens_k` rank-normalization precedent for the TODO at L272. Reusing that table would cut most of the hand-unrolled null plumbing.

qti-yuduo (commented):
Reviewed against the ONNX Attention spec and the QNN SDK headers. Architecture and the two-path split look right, and the SDK 2.12 guard is load-bearing (local `QnnOpDef.h` is 2.4.0 with no `GROUP_QUERY` symbols).

The softcap-before-mask order is **correct** — worth noting the vendored `cmake/external/onnx` is pinned at v1.20.1, which is pre-1.23-errata and has the opposite order. I initially flagged this as a bug against the submodule before checking the published spec. See the comment on L1282.

Two blocking issues (L1336 dtype buffer, L1370 bool mask). Everything else is minor.

Also: `IsOpSupported` re-runs the decomposition with `do_op_validation=true`, so most shape mismatches become a clean CPU fallback rather than a bad compile. The bool-mask issue is the exception — it builds a valid graph that computes the wrong thing.

onnxruntime/test/providers/qnn/attention_test.cc
```diff
@@ -0,0 +1,757 @@
...
+           TestInputDef<float>({1, 8, 64}, false, -1.0f, 1.0f)},
+          {test::MakeAttribute("q_num_heads", static_cast<int64_t>(4)),
+           test::MakeAttribute("kv_num_heads", static_cast<int64_t>(4)),
+           test::MakeAttribute("is_causal", static_cast<int64_t>(1))}),
+      opts, /*opset_version=*/24,
+      EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(1e-3f)});
+}
+
+// GQA 4D BNSH, causal, no KV cache → native (EmitNativeGQANode inserts Transpose+Reshape).
+// DISABLED: after 4D→3D transforms the QNN node has unpacked K/V inputs, which is not
```
qti-yuduo: Comment says `DISABLED:` (blocked until QAIRT 2.50) but the test is not actually disabled — so it either passes for an undocumented reason or is failing CI.

Coverage gaps that map to the findings above: no GQA+KV-cache (L1161), no bool mask (L1370), no fp16/double (L1336), opset 23 never exercised though docs claim 23-24, no softcap+attn_mask combination, and qk modes 1-2 untested (the two the errata redefined).

qti-mbadnara:
Hi @qti-yuduo thanks for the detailed review! Fixed the blockers.

qti-yuduo (approved):

PR #654, opened 2026-07-19
Match QNN GQA OpDef for total_sequence_length input shape

vjatoth-qti:
- The QNN GroupQueryAttention op def requires total_sequence_length (input[2]) to be a 0D scalar, but the ONNX com.microsoft.GroupQueryAttention op provides it as a 1D tensor with shape [1]. This mismatch causes op validation failure at runtime

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
@@ -131,10 +131,27 @@ Ort::Status GroupQueryAttentionOpBuilder::ProcessInputs(QnnModelWrapper& qnn_mod
       8u,  // sin_cache
       9u   // position_ids
   };
+  constexpr size_t kQnnTotalSeqLenIdx = 2;  // index of total_sequence_length in qnn_idx_to_onnx
 
-  for (const auto onnx_idx : qnn_idx_to_onnx) {
+  for (size_t qnn_idx = 0; qnn_idx < qnn_idx_to_onnx.size(); ++qnn_idx) {
+    const auto onnx_idx = qnn_idx_to_onnx[qnn_idx];
     if (onnx_inputs.size() > onnx_idx && onnx_inputs[onnx_idx].Exists()) {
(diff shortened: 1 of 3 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -270,6 +270,32 @@ Ort::Status GroupQueryAttentionOpBuilder::ProcessAttributesAndOutputs(QnnModelWr
...
+                                                          1u,
+                                                          std::multiplies<uint32_t>());
+      RETURN_IF_NOT(total_seq_len_total_size == 1, "Unexpected shape for total_sequence_length");
+      const auto& total_seq_len_input_name = input_names[2];
+      bool is_total_seq_len_graph_input = qnn_model_wrapper.IsGraphInput(total_seq_len_input_name);
+      std::string total_seq_len_reshaped = onnxruntime::qnn::utils::UniqueNameGenerator().New(total_seq_len_input_name, "_reshaped");
+      RETURN_IF_ERROR(qnn_model_wrapper.AddReshapeNode(total_seq_len_input_name,
+                                                       total_seq_len_reshaped,
+                                                       total_seq_len_info.shape,
+                                                       {},  // reshape to 0D scalar
```
chunghow-qti: I don't think htp backend's validation will accept Reshape node with rank 0 output tensor.

**Suggested fix:** 
build the tensor directly with an empty ({}) shape instead of reshaping into one. Also need to fix GetQnnTensorDataSizeInBytes in qnn_utils.cc, which
  currently throws on empty shape.

vjatoth-qti: Updated

qti-mattsinc (approved):

johnpaultaken (approved):
@tirupath-qti @qti-mbadnara  we need this change for the 9D EP certification. Otherwise our LLMs will break due to mismatch between GQA contrib op definition and ModelBuilder export.

qti-mbadnara:
@vjatoth-qti @johnpaultaken Seems like coverage tests are failing in CI.

tirupath-qti (approved):

PR #656, opened 2026-07-20
[QNN EP] Add HTP shared-memory zero-copy support

tirupath-qti:
### Description
- Wire factory-level shared allocator (OrtEpFactory::CreateAllocator, EpDevice_AddAllocatorInfo at library-load time)
- Decouple shared-allocator init from context-cache generation
- Add per-tensor one-shot WARNING on zero-copy fallback
- Add ORT Core patch for correct HOST_ACCESSIBLE memory routing in PluginExecutionProvider (GetOrtDeviceForPluginEp fallback + GetOrtDeviceByMemType override)
- Add 12 C++ unit tests


qti-chuteng:
Hi @kuanyul-qti @minfhong-qti, could you help review this PR?

cmake/patches/ort_core/0005-plugin-ep-and-allocation-planner-zero-copy.patch
```diff
@@ -0,0 +1,84 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
minfhong-qti: Just curious where the implementation in this patch is referring to.
I only find this commit https://github.com/microsoft/onnxruntime/commit/a827f9c34cfb854d37d41243cadd70d599abd468 which seems relevant.
In addition, if any patch is necessary currently, will it be removed someday after ORT Core uplevel?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -2258,6 +2260,974 @@ TEST_F(QnnHTPBackendTests, io_binding_qnn_htp_shared_offset) {
     binding.ClearBoundOutputs();
   }
 }
+
+namespace {
```
minfhong-qti, line 2205: Prefer to separate the new testcases in a new file.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -142,6 +151,18 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetSupportedDevicesImpl(OrtEpFactory* this
...
+    // Advertise HOST_ACCESSIBLE memory for NPU devices so ORT Core auto-registers
+    // a shared allocator at library-load time.
+    const auto device_type = factory->ort_api.HardwareDevice_Type(device);
+    if (device_type == OrtHardwareDeviceType_NPU &&
+        factory->rpcmem_library_ != nullptr &&
+        factory->host_accessible_memory_info_ != nullptr) {
+      if (OrtStatus* alloc_status = factory->ep_api.EpDevice_AddAllocatorInfo(
+              ep_device, factory->host_accessible_memory_info_.get())) {
+        return alloc_status;
+      }
```
minfhong-qti:
```suggestion
      RETURN_IF_NOT_NULL(factory->ep_api.EpDevice_AddAllocatorInfo(
          ep_device, factory->host_accessible_memory_info_.get()));
```

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -100,6 +100,15 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
     ort_api.ReleaseMemoryInfo(mem_info);
   }
   host_accessible_memory_info_ = MemoryInfoUniquePtr(mem_info, ort_api.ReleaseMemoryInfo);
+
+  // Probe RPCMEM once at load time; enables factory-level shared allocator advertisement.
+  try {
+    rpcmem_library_ = std::make_shared<qnn::RpcMemLibrary>();
```
minfhong-qti, line 139: There is another `RpcMemLibrary` in `QnnEp`. I'm wondering what is the difference and are both of them used during execution?

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -100,6 +100,15 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
...
+    qnn_allocator_type_ = qnn::QnnAllocatorType::HTP_SHARED;
```
minfhong-qti: There is a provider option controlling whether to use HTP share memory. Wouldn't this force to use regardless whether the option is set?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -533,14 +550,20 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
         logger,
         *qnn_backend_manager_,
         static_cast<const OrtMemoryInfo*>(input_tensor_mem_info),
         const_cast<void*>(raw_data), qnn_input_info.tensor_byte_size,
         graph_info_->GraphContext(),
-        qnn_inputs.back()));
+        qnn_inputs.back(),
+        fell_back_to_raw));
+
+    if (fell_back_to_raw && qnn_backend_manager_->GetQnnAllocatorType() != QnnAllocatorType::NONE) {
```
minfhong-qti: Why not move this check inside `BindQnnTensorMemoryToOrtValueMemory` and log?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -445,7 +445,9 @@ static Ort::Status BindQnnTensorMemoryToOrtValueMemory(const OrtApi& ort_api,
                                                        const OrtMemoryInfo* ort_value_memory_info,
                                                        void* ort_value_data, uint32_t ort_value_data_size,
                                                        Qnn_ContextHandle_t qnn_context,
-                                                       Qnn_Tensor_t& qnn_tensor) {
+                                                       Qnn_Tensor_t& qnn_tensor,
+                                                       bool& fell_back_to_raw) {
+  fell_back_to_raw = false;
```
kuanyul-qti: Do we need this line given that the passed `fell_back_to_raw` is false?

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -341,6 +355,32 @@ void ORT_API_CALL QnnEpFactory::ReleaseEpImpl(OrtEpFactory* /*this_ptr*/, OrtEp*
   delete dummy_ep;
 }
 
+OrtStatus* ORT_API_CALL QnnEpFactory::CreateAllocatorImpl(_In_ OrtEpFactory* this_ptr,
```
kuanyul-qti, line 389: In https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc#L855, `CreateAllocator` prefers the `OrtEp` function when it is available and falls back to the `OrtEpFactory` implementation otherwise. Since `QnnEp::CreateAllocatorImpl` has already been defined, under what circumstances will this function be executed?

qti-chuteng:
### 🟪 Architecture Fit

**[D-1]** `🔴 Blocker` `[Arch]` — `qnn_provider_factory.cc` (ctor ~L107; ReleaseAllocatorImpl ~L384; deleted main:L322 copy-back)
`ReleaseAllocatorImpl` picks the delete cast (`HtpSharedMemoryAllocator*` vs `Dx12SharedMemoryAllocator*`) by reading `factory->qnn_allocator_type_`. On main that field was synced per-session by the `CreateEpImpl` copy-back, so it reflected DX12 when the user chose DX12. This PR deletes the copy-back **and** hard-sets the field to `HTP_SHARED` in the ctor whenever libcdsprpc loads — but `QnnEp::CreateAllocatorImpl` (qnn_execution_provider.cc:2958) still builds a real `Dx12SharedMemoryAllocator`. On WoS/ARM64 where libcdsprpc loads, a DX12 allocator is released down the HTP branch (`delete static_cast<HtpSharedMemoryAllocator*>(dx12_ptr)`) → wrong-type delete / UB; if libcdsprpc is absent the field stays NONE → DX12 hits the unknown-type branch → leak. The deleted line's comment documented it as load-bearing. ⟦C8.3⟧ (+⟦C8.4⟧ one-way door → auto-raised).
*Better design:* dispatch release on the type actually created (tag the allocator, or key off its own OrtMemoryInfo mem-type), or restore a factory-side record of the DX12 selection. The decouple must re-wire DX12 in the same PR.

**[D-2]** `🟠 Concern` `[Arch]` — `cmake/patches/ort_core/0005-*.patch` (ep_plugin_provider_interfaces.cc)
The new `GetOrtDeviceByMemType` override lives on the generic base `PluginExecutionProvider`, so any EP with `default_device_.MemType()==HOST_ACCESSIBLE` now re-routes `CPUInput/CPUOutput/Default` to its default device. A non-QNN plugin EP that advertises a HOST_ACCESSIBLE default but genuinely needs CPU staging has its explicitly-CPU-requested tensors redirected — defeating the `CPUInput/CPUOutput` contract for that class. It's a shared ORT-core one-way door; the blast radius is empty only by accident of today's EP roster, not by any invariant the patch enforces. ⟦C8.4⟧.
*Better design:* scope the routing to the QNN memory-info signature (or a capability flag), or at minimum restrict the override to `OrtMemTypeDefault`; add an ORT-core test across the plugin-EP matrix (incl. a device-memory EP) proving neutrality.

**[D-3]** `🟠 Concern` `[Arch]` — `qnn_execution_provider.cc:1305-1320`
main's `if (!context_cache_enabled_)` guard skipped the RPCMEM load in ctx-gen mode (x86 offline-prepare, no libcdsprpc). `RpcMemLibrary`'s ctor throws when it can't load. This PR removes the guard and does `make_shared<qnn::RpcMemLibrary>()` unconditionally with no try/catch — so x86 offline ctx-gen + `enable_htp_shared_memory_allocator=1` throws out of the QnnEp ctor → CreateEpImpl → ORT_FAIL → session creation fails where it used to degrade. The PR's own intent ('a ctx-gen session may also run inference') encourages carrying the flag during ctx-gen, and the file-mapped-weights path 200 lines up (1108-1118) already uses the graceful `try/catch → disable + WARN` idiom this code ignores. ⟦C8.4⟧ (#138) + Sibling-Consistency.
*Better design:* wrap in try/catch mirroring lines 1108-1118 (leave `qnn_allocator_type_=NONE`, WARN-once, continue), or re-gate on execution mode.

**[D-4]** `🟠 Concern` `[Arch]` — `qnn_provider_factory.cc` GetSupportedDevicesImpl (~L155-163); deleted CreateEpImpl loop
Old code advertised in `CreateEpImpl` gated on `factory->qnn_allocator_type_ != NONE` (i.e. only if the user opted in). New code advertises in `GetSupportedDevices` gated only on NPU + rpcmem-loaded, independent of the session flag. Combined with the patch's `GetOrtDeviceForPluginEp` fallback, every QNN NPU session's default device becomes HOST_ACCESSIBLE — even sessions that never set the flag, moving their CPU tensors from the CPU arena to RPCMEM. Yet without the opt-in, `model_settings_.htp_shared_memory` stays false, so the binding still copies — a split-brain: RPCMEM allocated but still memcpy'd. ⟦C8.3⟧/⟦C8.4⟧.
*Better design:* gate advertisement on the same opt-in the memhandle binding uses, or make always-on an explicit documented default and verify the RPCMEM-allocated-but-RAW-branch case is harmless. Reconcile the factory-level and EP-level switches into one story.

**[D-9]** `🔵 Improvement` `[Arch]` — `qnn_provider_factory.cc:104-110`
The factory-ctor load-time probe `try { ... } catch(...) { rpcmem_library_=nullptr; }`: the partial state on failure is consistent (a highlight), but it catches everything incl. `std::bad_alloc` and logs nothing — so on a genuine NPU box where rpcmem should load but fails (permissions, missing service), zero-copy is silently disabled with no trace, while the EP path fails loudly. It also runs a doomed-and-swallowed dlopen at library-load on every non-NPU / x86 offline process (#138 shape). ⟦C4.1⟧/⟦C6.2⟧/⟦C8.4⟧.
*Better design:* emit a one-shot WARNING on catch (and consider narrowing to `const std::exception&`); cheap-gate the probe when no NPU-class device is present.

### 🟩 Simplicity

**[D-6]** `🔵 Improvement` `[Simplicity]` — `qnn_model.cc:113-115,164-188`; `qnn_model.h:210-217`
`BindQnnTensorMemoryToOrtValueMemory` already knows at its RAW branch that it fell back; the new `bool& fell_back_to_raw` out-param exports that private fact so the caller re-derives it and duplicates the same 3-line guard in both the input and output loops. And `WarnZeroCopyFallbackOnce` uses a `std::mutex` + `std::unordered_set<std::string>`, whereas the codebase's sibling warn-once (`hnrd_warning_emitted_` bool, `std::call_once`) uses a single bool; the mutex is redundant because the sole caller `ExecuteGraph` is already serialized by `graph_exec_mutex_`. ⟦C2.1⟧/⟦C6.2⟧/⟦C8.1⟧.
*Better design:* delete the out-param, the set, and the mutex; have the bind fn emit the warning at its own RAW branch guarded by a single `bool zero_copy_fallback_warned_` in the `hnrd_warning_emitted_` style — one cut removes the out-param, both duplicated guards, the unbounded set, and the redundant mutex.

**[D-7]** `🔵 Improvement` `[Simplicity]` — `qnn_provider_factory.cc:311-335` vs `qnn_execution_provider.cc:2951-2977`
The `HtpSharedMemoryAllocator` construction is typed twice (factory + EP; rule-of-two → extract `MakeHtpSharedAllocator(memory_info, rpcmem_library)`), and `RpcMemLibrary` is dlopen'd once in each ctor for the same process-global libcdsprpc. ⟦C1.1⟧/⟦C8.1⟧.
*Better design:* extract a shared `MakeHtpSharedAllocator()` for both CreateAllocator paths; have the EP borrow the factory's `rpcmem_library_` instead of re-probing.

### 🔵 Completeness

**[D-5]** `🟠 Concern` `[Completeness]` — `onnxruntime/test/providers/qnn/qnn_basic_test.cc` (all 12 new tests)
The headline claim (MEMHANDLE across the CPU-EP↔QNN-EP edge) is exercised only by the two cross-partition tests — both `#if !defined(__aarch64__)&&!defined(_M_ARM64) → GTEST_SKIP` AND wrapped in `#if defined(_WIN32)`. So: x86 Win → RPCMEM unloadable → skip; x86 Linux → skip; **Linux-arm64 → `#if _WIN32` false → empty body → trivial pass**. The other 10 set `enable_htp_shared_memory_allocator=1`, which throws in `RpcMemLibrary` on any NPU-less host → GTEST_SKIP. On every CI job without a physical NPU, all 12 are non-coverage, and the host-testable pure logic (D-1/D-2 routing, tiebreak, warn-once) has no PR-gate-runnable test. ⟦C6.1⟧.
*Better design:* extract the pure decision logic and unit-test it host-agnostically (mem-type→allocator eligibility; the `MemType()==HOST_ACCESSIBLE` tiebreak truth-table over {CPU, GPU-DEFAULT, CPU-HOST_ACCESSIBLE}; warn-once); add a host-CI assertion that the factory degrades to NONE when libcdsprpc is absent; fix the Linux-arm64 empty-body false-green.

**[D-8]** `🔵 Improvement` `[Completeness]` — `docs/execution-providers/QNN-ExecutionProvider.md` (unchanged)
Three user-visible changes ship undocumented: (1) `enable_htp_shared_memory_allocator` now coexists with `ep.context_enable=1`; (2) the shared allocator is now advertised at library-load time and usable before any session via `CreateSharedAllocator`; (3) a new one-shot WARNING on CPU-backed fallback. The PR does not touch this file. ⟦C8.3⟧ sibling coverage.
*Better design:* update the provider-options / shared-memory section for ctx-gen coexistence, load-time availability, and the fallback WARNING (and how to avoid it — allocate from QnnHtpShared).

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: After https://github.com/microsoft/onnxruntime/pull/32244 for perftest on ort-core merged, it will select the htp allocator even when enable_htp_shared_memory_allocator was not requested. 
I believe this should be correctly handled here.

cmake/patches/ort_core/0005-plugin-ep-and-allocation-planner-zero-copy.patch
chunghow-qti: Removed this ORT Core patch. The current ORT Core already provides the required behavior, so this PR no longer carries a duplicate patch or needs a future uplevel to remove it.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: Applied the suggestion. This now uses RETURN_IF_NOT_NULL(factory->ep_api.EpDevice_AddAllocatorInfo(...)).

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: This now uses a single factory-owned RpcMemLibrary that is created on demand. An allocator created through OrtEnv before session creation and a later opt-in QnnEp session share the same handle, rather than each loading RPCMEM independently.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: Clarified and fixed the semantics: the factory advertisement only makes the pre-session QnnHtpShared allocator available through OrtEnv. Only a QnnEp session with enable_htp_shared_memory_allocator=1 uses HOST_ACCESSIBLE as its default memory device and enables QNN MEMHANDLE binding.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
chunghow-qti: Moved this into BindQnnTensorMemoryToOrtValueMemory. The out-parameter and the duplicated input/output guards are gone. Since bind+execute is serialized by graph_exec_mutex_, one bool now emits one warning per model/session.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
chunghow-qti: Removed the fell_back_to_raw out-parameter and its initialization. The fallback decision and one-shot warning are now fully encapsulated by the bind helper.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: Confirmed: for an existing QnnEp session, ORT Core prefers QnnEp::CreateAllocator. The factory path is used by OrtEnv::CreateSharedAllocator to obtain QnnHtpShared before session creation. Core still releases allocators from both paths through the factory ReleaseAllocator, so dispatch now uses the allocator callback to distinguish HTP from DX12 instead of session state.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: Addressed: the factory no longer sets qnn_allocator_type_ to HTP. It only makes the pre-session QnnHtpShared allocator available on demand; enable_htp_shared_memory_allocator still controls whether the session uses HOST_ACCESSIBLE as its default memory device and enables MEMHANDLE binding.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
chunghow-qti: Addressed in 42838b13f4. The HTP shared-memory scenarios now live in the dedicated `qnn_htp_shared_memory_test.cc` file, and `qnn_basic_test.cc` matches the current main branch.

chunghow-qti:
Addressed the remaining review feedback in `42838b13f4`:

- D-1: allocator release dispatches by the concrete allocator callback, avoiding HTP/DX12 wrong-type deletion.
- D-2: removed the duplicated ORT Core patch now that the required changes are upstream.
- D-3/D-9: RPCMEM is factory-owned and loaded lazily on first allocation; missing RPCMEM no longer prevents EP registration or non-shared-memory workflows.
- D-4: factory advertisement only exposes the pre-session allocator. Session-level HOST_ACCESSIBLE routing and MEMHANDLE binding still require `enable_htp_shared_memory_allocator=1`.
- D-5: added host-runnable deferred-loading unit tests, fixed Linux ARM64 NPU-device selection, and moved device coverage into `qnn_htp_shared_memory_test.cc` for Windows/Linux ARM64.
- D-6: fallback detection and the one-shot warning remain encapsulated in the bind helper.
- D-7: factory and session allocator callbacks now share `CreateHtpSharedMemoryAllocator`, and lifetime tests cover allocators across session destruction and environment-allocator release.
- D-8: documented pre-session allocation, context-generation coexistence, and fallback-warning behavior.

No build was run for this update.

onnxruntime/test/providers/qnn/qnn_htp_shared_memory_test.cc
```diff
@@ -0,0 +1,1118 @@
...
+  for (auto& th : threads) {
+    th.join();
+  }
+
+  // Primary assertion: no crashes or exceptions under contention. Each Run
+  // must complete without throwing; the QNN HTP backend serializes
+  // graphExecute via graph_exec_mutex_, and the mem-handle manager is
+  // protected by mem_handles_mutex_, so neither should deadlock or throw.
+  //
+  // Note: output correctness is not asserted here because the QNN HTP
```
kuanyul-qti: **[M-1] (Major)** Concurrent Run test deliberately ignores incorrect output

This test creates four distinct shared-memory input/output buffer pairs to exercise one session while memhandles change across concurrent callers, but it only checks that `Run()` does not throw. The comment explicitly says the backend may not recompute for new memhandles and therefore omits output validation. That is not an ignorable backend detail: a successful ORT run returning a previous or another invocation's result is an execution-correctness failure, and the populated `expected_y` values are currently never used.

**Impact:**The test passes even if every concurrent Run produces stale or incorrect output, leaving regressions in shared-memory memhandle caching or bind/execute ordering undetected.

**Suggested fix:**

Compare every thread's `y_data` with its `expected_y` after each `Run()`, or at least after all threads join. If switching registered memhandles genuinely yields stale output, fix the EP registration/binding lifecycle (or explicitly reject that usage) instead of treating incorrect results as acceptable behavior.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1602,7 +1618,31 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    const auto shared_allocator_type = qnn_backend_manager_->GetQnnAllocatorType();
+    if (shared_allocator_type != qnn_allocator_type_) {
+      const std::string message =
+          "Cannot share QNN backend manager with a different shared-memory allocator mode. "
+          "The existing shared manager uses '" +
+          std::string{qnn::QnnAllocatorTypeToString(shared_allocator_type)} +
+          "', but this session requested '" +
+          std::string{qnn::QnnAllocatorTypeToString(qnn_allocator_type_)} + "'.";
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_ERROR, message.c_str());
+      throw std::runtime_error(message);
```
huaychou: **[C-1] (Critical)** Reusing a shared QnnBackendManager calls ResetLogger() before validating allocator-mode compatibility, leaving a dangling logger pointer once the new check throws

When `reusing_shared_backend_manager` is true, the constructor immediately does:

```
qnn_backend_manager_ = shared_qnn_backend_manager;
qnn_backend_manager_->ResetLogger(logger_);
```

`QnnBackendManager::ResetLogger()` (in `qnn_backend_manager.h`) stores `&logger_` as a raw pointer (`logger_ptr_ = &logger;`), where `logger_` is this `QnnEp` instance's own member (`const Ort::Logger logger_;`) — not something owned by the shared manager or by ORT Core.

Later in the same constructor (line 1631, new code added by this PR), if this session's `qnn_allocator_type_` does not match the shared manager's existing allocator type, the constructor throws `std::runtime_error`. At that point the `QnnEp` object under construction (and its `logger_` member) is unwound and destroyed. The *shared*, still-alive `QnnBackendManager` — which may be actively used by another, unrelated session in the same `htp_share_resource_optimization=1` group — is left with `logger_ptr_` pointing at freed memory. Any subsequent `ORT_CXX_LOG_PTR(logger_ptr_, ...)` call from that other session (there are many throughout `qnn_backend_manager.cc`, including SSR-recovery and graph-execution paths) dereferences a dangling pointer.

**Impact**: Use-after-free / dangling-pointer read that can crash the process or corrupt behavior for a completely unrelated, still-running session, triggered simply by creating a second session in the same resource-sharing group with a different `enable_htp_shared_memory_allocator`/`enable_dx12_shared_memory_allocator` setting than the first. The new regression test `Ctor_SharedBackendManagerWithDifferentAllocatorMode_Throws` only asserts that the exception is thrown; it does not exercise or detect the dangling-pointer aftermath, so this defect ships without test coverage.

**Suggested fix:**

Resolve `qnn_allocator_type_` for this session before touching the shared manager, and only call `ResetLogger()` once the allocator-mode compatibility check has passed (e.g. read `shared_qnn_backend_manager->GetQnnAllocatorType()` directly — without going through `qnn_backend_manager_`/`ResetLogger` — and throw before ever assigning `qnn_backend_manager_ = shared_qnn_backend_manager` or calling `ResetLogger`). Add a test that, after the expected throw, verifies the first (owner) session can still safely log through the shared manager.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -164,6 +210,16 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetSupportedDevicesImpl(OrtEpFactory* this
     ep_devices[num_ep_devices++] = ep_device;
     factory->ep_devices_.push_back(ep_device);
 
+    // Advertise HOST_ACCESSIBLE memory for NPU devices so OrtEnv can create a
+    // QnnHtpShared allocator before a QNN session exists. Loading RPCMEM is deferred
+    // until that allocator (or an opted-in session) actually needs it.
+    const auto device_type = factory->ort_api.HardwareDevice_Type(device);
+    if (device_type == OrtHardwareDeviceType_NPU &&
+        factory->host_accessible_memory_info_ != nullptr) {
```
huaychou: **[M-1] (Major)** GPU/DX12 shared-memory allocator's EpDevice_AddAllocatorInfo registration was removed with no replacement, unlike the new NPU/HTP path

Before this PR, `QnnEpFactory::CreateEpImpl` called `EpDevice_AddAllocatorInfo(ep_device, host_accessible_memory_info_.get())` for every `ep_device` in the factory whenever a session set `factory->qnn_allocator_type_` to either `HTP_SHARED` or `DX12_SHARED`. This PR deletes that block entirely (no longer needed for HTP now that `GetSupportedDevicesImpl` registers HOST_ACCESSIBLE memory for NPU devices up front), but does not add an equivalent for GPU devices — `GetSupportedDevicesImpl`'s new registration block is gated on `device_type == OrtHardwareDeviceType_NPU` only.

Net effect: `enable_dx12_shared_memory_allocator=1` sessions no longer get `EpDevice_AddAllocatorInfo` called for their GPU `OrtEpDevice` at all — neither eagerly (the old per-session path, now deleted) nor lazily (the new path only covers NPU). None of the existing DX12 tests (`get_allocator_qnn_gpu_shared`, `io_binding_qnn_gpu_shared` in `qnn_basic_test.cc`) were touched or re-verified by this PR's commits, and no new DX12/GPU test was added alongside the new HTP/NPU coverage.

**Impact**: (unverified — confirm before acting) depending on whether ORT Core's session-level `Ort::Allocator(session, memory_info)` / env-level shared-allocator resolution for a plugin EP relies on the `OrtEpDevice`'s registered allocator infos, this may silently break the DX12 shared-memory allocator path for GPU sessions. Even if it still works through some other resolution path, the asymmetry (NPU handled, GPU not) deserves a deliberate confirmation rather than a silent omission, and DX12 zero-copy has zero test coverage after this change.

**Suggested fix:**

Either add the equivalent `EpDevice_AddAllocatorInfo` registration for GPU devices in `GetSupportedDevicesImpl` (mirroring the NPU block), or explicitly confirm — and note in the PR description — that DX12/GPU allocator advertisement is intentionally out of scope for this PR and file a follow-up. Either way, run (or add) a GPU/DX12 test on Windows-ARM64 hardware to confirm `get_allocator_qnn_gpu_shared`/`io_binding_qnn_gpu_shared` still pass after this change.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -216,6 +223,12 @@ class QnnEp : public OrtEp, public ApiPtrs {
   mutable std::optional<uint32_t> htp_power_config_id_;
   mutable std::mutex config_id_mutex_;
 
+  QnnEpFactory& factory_;
+
+  // Non-null only when enable_htp_shared_memory_allocator is enabled for this
+  // session. The factory still exposes QnnHtpShared before session creation.
+  const OrtMemoryDevice* default_memory_device_ = nullptr;
```
huaychou: **[T-1] (Nit)** Stale duplicate // const QnnEpFactory& factory_; comment left directly below the now-real factory_ member

This PR uncomments `factory_{factory}` in the constructor init list and adds a real `QnnEpFactory& factory_;` member (line 226). The old, now-redundant `// const QnnEpFactory& factory_;` comment directly below it (line 232) was left in place — it now reads as if the member is still commented-out/unused, which is confusing for anyone skimming the header.

**Suggested fix:**

Delete the stale `// const QnnEpFactory& factory_;` comment on line 232 — it duplicates and contradicts the real member declared just above it.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -122,6 +122,52 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
     ort_api.ReleaseStatus(status);
   }
   host_accessible_memory_info_ = MemoryInfoUniquePtr(mem_info, ort_api.ReleaseMemoryInfo);
+
+  // Keep this available before session creation so OrtEnv can create a shared
+  // allocator for Python I/O binding. ORT creates advertised allocators while
+  // registering the EP library, so the allocator defers loading RPCMEM until
+  // its first allocation.
+  CreateAllocator = CreateAllocatorImpl;
```
minfhong-qti: Move above with others.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -122,6 +122,52 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
...
+  return rpcmem_library_;
+}
+
+OrtStatus* QnnEpFactory::CreateHtpSharedMemoryAllocator(
+    const OrtMemoryInfo* memory_info,
+    std::shared_ptr<qnn::RpcMemLibrary> rpcmem_library,
+    OrtAllocator** allocator) noexcept {
+  *allocator = nullptr;
+
+  try {
```
minfhong-qti: Unlike above where `RpcMemLibrary` may throw error, is there any function may throw error below?

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -60,6 +60,11 @@ class QnnEp : public OrtEp, public ApiPtrs {
   friend struct GenieNodeComputeInfo;
   friend class QnnEpFactory;
 
+#if defined(QNN_EP_INTERNAL_SYMBOL_ACCESS)
```
minfhong-qti: I do not prefer adding logic only for testing.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1563,20 +1561,38 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
-      ORT_CXX_LOGF(logger_,
-                   ORT_LOGGING_LEVEL_INFO,
-                   "Context cache is enabled in this session (via %s); the HTP shared memory allocator will be disabled"
-                   " as no allocations are expected to be made.",
-                   kOrtSessionOptionEpContextEnable);
+      // Enable shared allocator regardless of context-generation mode; a session that
+      // generates a context binary may still run inference in the same session.
+      qnn_allocator_type_ = qnn::QnnAllocatorType::HTP_SHARED;
+      model_settings_.htp_shared_memory = true;
+      default_memory_device_ = ep_api.MemoryInfo_GetMemoryDevice(factory.GetHostAccessibleMemoryInfo());
```
minfhong-qti: It seems unnecessary to record this as we can always acquire in `GetDefaultMemoryDeviceImpl`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1602,7 +1618,31 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_ERROR, message.c_str());
+      throw std::runtime_error(message);
+    }
+  } else {
+    // An allocator mode belongs to the QNN context. A manager that is later
+    // shared must retain this first session's mode instead of allowing a
+    // subsequent session to overwrite it.
+    qnn_backend_manager_->SetQnnAllocatorType(qnn_allocator_type_);
+
+    // Publish only after fixing the allocator mode. ContextCreateAsyncCallback
```
minfhong-qti: It seems not necessary to move it here?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1563,20 +1561,38 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
-    // This library is only necessary for the inference (for the shared memory allocator), if we are in context
-    // generation stage, there is no need to load it as no allocations will be made.
-    if (!context_cache_enabled_) {
-      rpcmem_library_ = std::make_shared<qnn::RpcMemLibrary>();
-      qnn_allocator_type_ = qnn::QnnAllocatorType::HTP_SHARED;
+    // Context generation must preserve the MEMHANDLE graph I/O contract even on
+    // hosts that cannot load RPCMEM (for example, offline generation on x86).
+    // RPCMEM availability only controls whether this session can allocate and
+    // bind HTP shared memory locally.
+    model_settings_.htp_shared_memory = context_cache_enabled_;
```
minfhong-qti: Shouldn't it always be true here?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1563,20 +1561,38 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    if (rpcmem_library_ == nullptr) {
+      if (context_cache_enabled_) {
+        ORT_CXX_LOGF(logger_, ORT_LOGGING_LEVEL_WARNING,
+                     "Unable to load RPCMEM; the local HTP shared memory allocator is disabled, "
+                     "but the generated context will retain the shared-memory graph I/O contract: %s",
+                     rpcmem_error.c_str());
+      } else {
+        ORT_CXX_LOGF(logger_, ORT_LOGGING_LEVEL_WARNING,
+                     "Unable to load RPCMEM; disabling HTP shared memory allocator: %s",
+                     rpcmem_error.c_str());
```
minfhong-qti: Why `htp_shared_memory` isn't set to false here?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1563,20 +1561,38 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+    // Context generation must preserve the MEMHANDLE graph I/O contract even on
+    // hosts that cannot load RPCMEM (for example, offline generation on x86).
+    // RPCMEM availability only controls whether this session can allocate and
+    // bind HTP shared memory locally.
+    model_settings_.htp_shared_memory = context_cache_enabled_;
+
+    std::string rpcmem_error;
+    rpcmem_library_ = factory.GetOrCreateRpcMemLibrary(rpcmem_error);
+    if (rpcmem_library_ == nullptr) {
+      if (context_cache_enabled_) {
```
minfhong-qti: Should we use platform (i.e., x86 or not) to determine whether it is expected not loadable?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1542,13 +1547,6 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                      skip_backend_op_validation,
                                      reused_io_limit_mb},
         ApiPtrs{ort_api, ep_api, model_editor_api}, logger_);
-    // Publish for later sessions. Always publish when htp_share_resource_optimization_==1,
-    // even for a terminator session, because ContextCreateAsyncCallback retrieves the backend
-    // manager from the singleton during SetupBackend (GetCapability). The terminator reset for
-    // all sharing paths is deferred to after SetupBackend completes (in GetCapabilityImpl).
-    if (htp_share_resource_optimization_ == 1) {
```
huaychou: Why do we remove this?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1520,8 +1521,12 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
   const bool use_shared_backend_mgr =
       ((context_cache_enabled_ && share_ep_contexts_) || htp_share_resource_optimization_ == 1);
-  if (use_shared_backend_mgr && SharedContext::GetInstance().GetSharedQnnBackendManager()) {
-    qnn_backend_manager_ = SharedContext::GetInstance().GetSharedQnnBackendManager();
+  const auto shared_qnn_backend_manager = use_shared_backend_mgr
+                                              ? SharedContext::GetInstance().GetSharedQnnBackendManager()
+                                              : nullptr;
+  const bool reusing_shared_backend_manager = shared_qnn_backend_manager != nullptr;
+  if (reusing_shared_backend_manager) {
+    qnn_backend_manager_ = shared_qnn_backend_manager;
```
huaychou: **[C-1] (Critical)** Reusing a shared QnnBackendManager calls ResetLogger() before validating allocator-mode compatibility, leaving a dangling logger pointer once the new check throws

When `reusing_shared_backend_manager` is true, the constructor immediately does:

```
qnn_backend_manager_ = shared_qnn_backend_manager;
qnn_backend_manager_->ResetLogger(logger_);
```

`QnnBackendManager::ResetLogger()` (in `qnn_backend_manager.h`) stores `&logger_` as a raw pointer (`logger_ptr_ = &logger;`), where `logger_` is this `QnnEp` instance's own member (`const Ort::Logger logger_;`) — not something owned by the shared manager or by ORT Core.

Later in the same constructor (line 1631, new code added by this PR), if this session's `qnn_allocator_type_` does not match the shared manager's existing allocator type, the constructor throws `std::runtime_error`. At that point the `QnnEp` object under construction (and its `logger_` member) is unwound and destroyed. The *shared*, still-alive `QnnBackendManager` — which may be actively used by another, unrelated session in the same `htp_share_resource_optimization=1` group — is left with `logger_ptr_` pointing at freed memory. Any subsequent `ORT_CXX_LOG_PTR(logger_ptr_, ...)` call from that other session (there are many throughout `qnn_backend_manager.cc`, including SSR-recovery and graph-execution paths) dereferences a dangling pointer.

**Impact**: Use-after-free / dangling-pointer read that can crash the process or corrupt behavior for a completely unrelated, still-running session, triggered simply by creating a second session in the same resource-sharing group with a different `enable_htp_shared_memory_allocator`/`enable_dx12_shared_memory_allocator` setting than the first. The new regression test `Ctor_SharedBackendManagerWithDifferentAllocatorMode_Throws` only asserts that the exception is thrown; it does not exercise or detect the dangling-pointer aftermath, so this defect ships without test coverage.

**Suggested fix:**

Resolve `qnn_allocator_type_` for this session before touching the shared manager, and only call `ResetLogger()` once the allocator-mode compatibility check has passed (e.g. read `shared_qnn_backend_manager->GetQnnAllocatorType()` directly — without going through `qnn_backend_manager_`/`ResetLogger` — and throw before ever assigning `qnn_backend_manager_ = shared_qnn_backend_manager` or calling `ResetLogger`). Add a test that, after the expected throw, verifies the first (owner) session can still safely log through the shared manager.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou: I think we shouldn't throw in EP

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -216,6 +223,12 @@ class QnnEp : public OrtEp, public ApiPtrs {
   mutable std::optional<uint32_t> htp_power_config_id_;
   mutable std::mutex config_id_mutex_;
 
+  QnnEpFactory& factory_;
+
+  // Non-null only when enable_htp_shared_memory_allocator is enabled for this
+  // session. The factory still exposes QnnHtpShared before session creation.
+  const OrtMemoryDevice* default_memory_device_ = nullptr;
+
   // const QnnEpFactory& factory_;
```
huaychou:
```suggestion
```

onnxruntime/core/providers/qnn/qnn_provider_factory.h
```diff
@@ -61,28 +87,29 @@ class QnnEpFactory : public OrtEpFactory, public ApiPtrs {
...
 
-  // CPU allocator so we can control the arena behavior. optional as ORT always provides a CPU allocator if needed.
   using MemoryInfoUniquePtr = std::unique_ptr<OrtMemoryInfo, std::function<void(OrtMemoryInfo*)>>;
   MemoryInfoUniquePtr host_accessible_memory_info_;
 
+  // Created on the first allocation or opted-in session request. Merely
+  // registering the factory and its environment allocator must remain usable on
+  // hosts without RPCMEM (for example, offline context generation on x86).
+  std::shared_ptr<qnn::RpcMemLibrary> rpcmem_library_;
+  std::mutex rpcmem_library_mutex_;
```
huaychou: I think this is duplicated with qnn_allocator.h. Could you check if this is necessary?

onnxruntime/core/providers/qnn/qnn_execution_provider.h
chunghow-qti: removed

onnxruntime/core/providers/qnn/qnn_execution_provider.h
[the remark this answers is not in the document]
chunghow-qti: done, and update to const reference, then update below in QnnEpFactory to mutable
```
mutable std::shared_ptr<qnn::RpcMemLibrary> rpcmem_library_;
mutable std::mutex rpcmem_library_mutex_;
```

chunghow-qti: done, and update to const reference, then update below in QnnEpFactory to mutable
```
mutable std::shared_ptr<qnn::RpcMemLibrary> rpcmem_library_;
mutable std::mutex rpcmem_library_mutex_;

File: onnxruntime/core/providers/qnn/qnn_execution_provider.cc (hunk 16 above)
Comment (chunghow-qti) [reply onnxruntime/core/providers/qnn/qnn_execution_provider.cc]: sure, updated.

File: onnxruntime/core/providers/qnn/qnn_execution_provider.cc (hunk 10 above)
Comment (chunghow-qti) [reply onnxruntime/core/providers/qnn/qnn_execution_provider.cc]: Thanks. This is for the case when multiple session shared a backend manager (i.e. "ep.context_enable=1 & ep.share_ep_contexts=1" or  
"htp_shared_resource_optimization = 1"). 

My original thought was that it would be strange for some sessions to use memHandle while others use clientBuf, so I chose to fail fast here.

but im free to accept suggestion here. what is the concern if we throw in EP?

File: onnxruntime/core/providers/qnn/qnn_provider_factory.cc [hunk 25]
```diff
@@ -164,6 +205,16 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetSupportedDevicesImpl(OrtEpFactory* this
     ep_devices[num_ep_devices++] = ep_device;
     factory->ep_devices_.push_back(ep_device);
 
+    // Advertise HOST_ACCESSIBLE memory for NPU devices so OrtEnv can create a
+    // QnnHtpShared allocator before a QNN session exists. Loading RPCMEM is deferred
+    // until that allocator (or an opted-in session) actually needs it.
+    const auto device_type = factory->ort_api.HardwareDevice_Type(device);
+    if (device_type == OrtHardwareDeviceType_NPU &&
+        factory->host_accessible_memory_info_ != nullptr) {
```
quic-calvnguy: factory->host_accessible_memory_info will never be nullptr on creation of the EP factory unless ORT API fails to create memory info:
https://github.com/onnxruntime/onnxruntime-qnn/pull/656/changes#diff-076fc19d6a7a68ab829f45e20af67c04b1a86d560afa8de2a0cc78c2cc666e2cR113-R126

This will result in ORT API requesting EP Session to create an allocator regardless of EP session configuration:
https://github.com/microsoft/onnxruntime/blob/v1.30.0/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc#L220-L234
https://github.com/microsoft/onnxruntime/blob/v1.30.0/onnxruntime/core/session/plugin_ep/ep_plugin_provider_interfaces.cc#L890-L905

quic-calvnguy: Relevant PR:
https://github.com/onnxruntime/onnxruntime-qnn/pull/859

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
quic-calvnguy: Do we know the call flow to EP Factory's CreateAllocator() and this function from ORT Core if Ortvalues are allocated before any session?

quic-calvnguy: @chunghow-qti 
PR #895 has been merged and you will need to rebase onto main to get the changes.

Once you rebase, I believe you can remove these lines of code:
https://github.com/onnxruntime/onnxruntime-qnn/blob/main/onnxruntime/core/providers/qnn/qnn_execution_provider.cc#L3570-L3580

Your changes on for lines 3598-3605 accomplish the same thing.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: I believe the GPU team is planning to use a different memory_info (QnnHtpShared) than HTP. The session-level DX12 allocator should continue to work as usual. However, supporting the DX12 allocator in this scenario would require a different memory_info, which I believe is out of scope for this PR.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: fix to preserve the session-level dx-12 allocator. and will discuss about the sperate memory info name with GPU team.

onnxruntime/test/providers/qnn/qnn_htp_shared_memory_test.cc
chunghow-qti: done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: fix to preserve the session-level dx-12 allocator. and will discuss about if we should include the sperate memory info name with GPU team for factory level shared allocator

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
chunghow-qti: Yes. Although the lazy path does not load RPCMEM here, std::make_unique and the allocator constructors may still throw. This helper is used from noexcept C ABI callbacks, so it catches C++ exceptions and converts them to OrtStatus instead of allowing std::terminate().

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Thanks for review. Rebased.

Yes. Before a session exists, `QnnEp::CreateAllocatorImpl()` cannot be called because there is no `QnnEp` instance yet.

The pre-session flow is:

```
RegisterExecutionProviderLibrary()
  -> QnnEpFactory::GetSupportedDevicesImpl()
     (advertises HOST_ACCESSIBLE memory on the NPU OrtEpDevice)
  -> Environment::RegisterExecutionProviderLibrary()
  -> Environment::CreateSharedAllocatorImpl()
  -> OrtEpFactory::CreateAllocator()
  -> QnnEpFactory::CreateAllocatorImpl()
```

ORT stores that allocator in the environment. The application then obtains it through `GetSharedAllocator()` (or explicitly replaces it through `CreateSharedAllocator()`) and calls `allocator->Alloc()` before constructing the `OrtValue`. Creating the `OrtValue` around that buffer does not call either allocator-creation callback.

`QnnEp::CreateAllocatorImpl()` is only used after session creation. During `PluginExecutionProvider::CreatePreferredAllocators()`, ORT prefers the session EP `CreateAllocator` callback and falls back to the factory callback only if the EP callback is absent.

`htp_shared_memory_factory_allocator_pre_session` covers the explicit pre-session factory path, and `htp_shared_memory_env_auto_register` covers the automatically registered environment allocator path.

PR #658, opened 2026-07-22
[QNN EP] Fix onnxruntime_perf_test HTP shared-memory zero-copy on plugin-EP path

chunghow-qti:
## Description
`onnxruntime_perf_test`'s `QnnHtpShared` allocator path routes model inputs through
the QNN EP's HTP shared-memory (RPCMEM) allocator for zero-copy inference. It works
when the EP is registered the traditional way (`-e qnn`) but is silently broken when
registered via the plugin-EP mechanism (`--plugin_eps`): `provider_name_` is never
`kQnnExecutionProvider`, so 
* the `-i` branch that sets `device_memory_name_ =
"QnnHtpShared"` is skipped, and every input stays CPU-backed (`clientBuf`, per-frame
copy). 
* Separately, real `.pb` test data is loaded onto plain CPU memory by the shared
`TestCase` infrastructure, so even generated (`-I`) inputs were the only ones that
could reach zero-copy.

## Fix

Delivered as ort_core patch `0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch`:

1. In the plugin-EP branch, detect `enable_htp_shared_memory_allocator|1` and set
   `device_memory_name_ = "QnnHtpShared"`.
2. Build the `QnnHtpShared` `OrtMemoryInfo` with `OrtDeviceAllocator`, matching how the
   QNN EP registers the shared allocator, so inputs take the `memHandle` path.
3. Intercept in `PreLoadTestData` (the common sink for both generated `-I` and real
   `.pb` inputs) and re-wrap CPU-backed input tensors onto the shared allocator via a
   new `CopyToSharedAllocator` helper, so real-data runs reach zero-copy too. Tensors
   already HOST_ACCESSIBLE, non-tensors, and string tensors are left unchanged.

## Test plan

- [x] All GQA inputs switch from `clientBuf` to `memHandle` — generated (`-I`) path.
- [x] All GQA inputs switch from `clientBuf` to `memHandle` — real `.pb` data path.
- [x] Patch applies cleanly against pristine `ort_test_session.{cc,h}` (`git apply`
      and GNU `patch --dry-run`).

> Note: this patch targets ort_core 1.27.0; it may need updating on ort_core uplevel.

### DIFF
--- a/cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
+++ b/cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
@@ -9,8 +9,10 @@ skipped. device_memory_name_ stays empty, the shared-memory allocator block is
 never entered, and every model input is CPU-backed -> per-frame copy (no zero-copy).
 
 Fix:
-1. In the plugin-EP branch, detect enable_htp_shared_memory_allocator|1 in the EP
-   runtime config string and set device_memory_name_ = "QnnHtpShared".
+1. In the plugin-EP branch, parse the EP runtime config string with the existing
+   test::utils::ParseSessionConfigs helper and check for
+   enable_htp_shared_memory_allocator == "1", then set
+   device_memory_name_ = "QnnHtpShared".
 2. In the allocator block, build the QnnHtpShared OrtMemoryInfo with the
    OrtDeviceAllocator type (matching how the QNN EP registers the shared
    allocator) so inputs take the zero-copy memHandle path.
@@ -37,23 +39,27 @@ index f67d922..0acdaaf 100644
  #include <limits>
  #include <fstream>
  #include <set>
-@@ -110,6 +111,22 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
+@@ -110,6 +111,26 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
          device_memory_name_.empty()) {
        device_memory_name_ = CUDA;
      }
 +
 +    // In plugin-EP mode provider_name_ is not kQnnExecutionProvider, so the QNN
-+    // "-i" option-parsing branch below is skipped. Detect the shared-allocator option
-+    // here so inputs go through the QnnHtpShared (RPCMEM) allocator for zero-copy.
-+    // Note: enable_htp_shared_memory_allocator is a QNN-EP-specific option; no other
-+    // plugin EP uses this key, so matching on it is sufficient to identify the QNN EP.
++    // "-i" option-parsing branch below is skipped. Parse the EP runtime config
++    // string here so inputs go through the QnnHtpShared (RPCMEM) allocator for
++    // zero-copy. Note: enable_htp_shared_memory_allocator is a QNN-EP-specific
++    // option; no other plugin EP uses this key, so matching on it is sufficient
++    // to identify the QNN EP.
 +    if (device_memory_name_.empty()) {
 +#ifdef _MSC_VER
 +      const std::string qnn_opt_string = ToUTF8String(performance_test_config.run_config.ep_runtime_config_string);
 +#else
 +      const std::string qnn_opt_string = performance_test_config.run_config.ep_runtime_config_string;
 +#endif
-+      if (qnn_opt_string.find("enable_htp_shared_memory_allocator|1") != std::string::npos) {
++      std::unordered_map<std::string, std::string> qnn_plugin_options;
++      test::utils::ParseSessionConfigs(qnn_opt_string, qnn_plugin_options);
++      auto it = qnn_plugin_options.find("enable_htp_shared_memory_allocator");
++      if (it != qnn_plugin_options.end() && it->second == "1") {
 +        device_memory_name_ = "QnnHtpShared";
 +      }
 +    }
--- a/cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
+++ b/cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
@@ -24,7 +24,7 @@ Fix:
    take the memHandle path. Tensors already on a HOST_ACCESSIBLE allocator, non-tensor
    values, and string tensors are left unchanged.
 
-TODO: Verified against ort_core 1.27.0; re-check this patch on future ort core upleveling since the target file may change.
+TODO: re-check this patch on future ort core upleveling since the target file may change.
 
 ---
 diff --git a/onnxruntime/test/perftest/ort_test_session.cc b/onnxruntime/test/perftest/ort_test_session.cc
--- a/cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
+++ b/cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
@@ -1,0 +1,149 @@
+From: QNN EP team <noreply@qualcomm.com>
+Date: Mon, 21 Jul 2026 00:00:00 +0000
+Subject: [PATCH] perftest: enable QnnHtpShared zero-copy input on plugin-EP path
+
+When onnxruntime_perf_test registers the QNN EP via --plugin_eps, provider_name_
+is not kQnnExecutionProvider, so the "-i" QNN option-parsing branch (which sets
+device_memory_name_ = "QnnHtpShared" on enable_htp_shared_memory_allocator|1) is
+skipped. device_memory_name_ stays empty, the shared-memory allocator block is
+never entered, and every model input is CPU-backed -> per-frame copy (no zero-copy).
+
+Fix:
+1. In the plugin-EP branch, detect enable_htp_shared_memory_allocator|1 in the EP
+   runtime config string and set device_memory_name_ = "QnnHtpShared".
+2. In the allocator block, build the QnnHtpShared OrtMemoryInfo with the
+   OrtDeviceAllocator type (matching how the QNN EP registers the shared
+   allocator) so inputs take the zero-copy memHandle path.
+3. Real .pb test data is loaded onto plain CPU memory by the shared TestCase
+   infrastructure, bypassing this session's allocator, so generated (-I) inputs
+   were the only ones reaching zero-copy. Intercept in PreLoadTestData (the common
+   sink for all input paths) and, when the QnnHtpShared allocator is active, re-wrap
+   CPU-backed input tensors onto it via CopyToSharedAllocator so real-data runs also
+   take the memHandle path. Tensors already on a HOST_ACCESSIBLE allocator, non-tensor
+   values, and string tensors are left unchanged.
+
+TODO: Verified against ort_core 1.27.0; re-check this patch on future ort core upleveling since the target file may change.
+
+---
+diff --git a/onnxruntime/test/perftest/ort_test_session.cc b/onnxruntime/test/perftest/ort_test_session.cc
+index f67d922..0acdaaf 100644
+--- a/onnxruntime/test/perftest/ort_test_session.cc
++++ b/onnxruntime/test/perftest/ort_test_session.cc
+@@ -5,6 +5,7 @@
+
+ #include "ort_test_session.h"
+ #include <algorithm>
++#include <cstring>
+ #include <limits>
+ #include <fstream>
+ #include <set>
+@@ -110,6 +111,22 @@ OnnxRuntimeTestSession::OnnxRuntimeTestSession(Ort::Env& env, std::random_device
+         device_memory_name_.empty()) {
+       device_memory_name_ = CUDA;
+     }
++
++    // In plugin-EP mode provider_name_ is not kQnnExecutionProvider, so the QNN
++    // "-i" option-parsing branch below is skipped. Detect the shared-allocator option
++    // here so inputs go through the QnnHtpShared (RPCMEM) allocator for zero-copy.
++    // Note: enable_htp_shared_memory_allocator is a QNN-EP-specific option; no other
++    // plugin EP uses this key, so matching on it is sufficient to identify the QNN EP.
++    if (device_memory_name_.empty()) {
++#ifdef _MSC_VER
++      const std::string qnn_opt_string = ToUTF8String(performance_test_config.run_config.ep_runtime_config_string);
++#else
++      const std::string qnn_opt_string = performance_test_config.run_config.ep_runtime_config_string;
++#endif
++      if (qnn_opt_string.find("enable_htp_shared_memory_allocator|1") != std::string::npos) {
++        device_memory_name_ = "QnnHtpShared";
++      }
++    }
+   }
+
+   provider_name_ = performance_test_config.machine_config.provider_type_name;
+@@ -979,6 +994,10 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
+     Ort::MemoryInfo memory_info(nullptr);  // Default initialize, will be overwritten
+     if (device_memory_name_ == CUDA) {
+       memory_info = Ort::MemoryInfo(device_memory_name_.data(), OrtArenaAllocator, 0, OrtMemTypeDefault);
++    } else if (device_memory_name_ == "QnnHtpShared") {
++      // HOST_ACCESSIBLE (RPCMEM) allocator: OrtDeviceAllocator matches how the QNN EP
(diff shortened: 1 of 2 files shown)


cmake/patches/ort_core/0005-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,124 @@
...
++
++    // When the QNN plugin EP is registered with enable_htp_shared_memory_allocator|1
++    // (passed via the -i EP runtime config string), the traditional
++    // "provider_name_ == kQnnExecutionProvider" parsing branch below is skipped
++    // (provider_name_ is empty in plugin-EP mode). Without setting device_memory_name_
++    // here, the shared-memory allocator block later is never entered and every model
++    // input is CPU-backed -> per-frame copy instead of zero-copy. Detect the option
++    // from the runtime config string and route inputs through the QnnHtpShared allocator.
++    if (device_memory_name_.empty()) {
++#ifdef _WIN32
```
kuanyul-qti:
```suggestion
+#ifdef _MSC_VER
```
Use `_MSC_VER` to keep it consistent with the other occurrence in the file.

cmake/patches/ort_core/0005-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,65 @@
...
+@@ -979,6 +997,12 @@ select from 'TF8', 'TF16', 'UINT8', 'FLOAT', 'ITENSOR'. \n)");
+     Ort::MemoryInfo memory_info(nullptr);  // Default initialize, will be overwritten
+     if (device_memory_name_ == CUDA) {
+       memory_info = Ort::MemoryInfo(device_memory_name_.data(), OrtArenaAllocator, 0, OrtMemTypeDefault);
++    } else if (device_memory_name_ == "QnnHtpShared") {
++      // Match the OrtMemoryInfo the QNN EP registers for its HTP shared allocator
++      // (OrtDeviceAllocator + OrtMemTypeCPU). This is the same construction the QNN
++      // shared-memory unit tests use to obtain a HOST_ACCESSIBLE (RPCMEM) allocator,
++      // so inputs allocated from it take the zero-copy memHandle path.
++      memory_info = Ort::MemoryInfo(device_memory_name_.data(), OrtDeviceAllocator, 0, OrtMemTypeCPU);
```
kuanyul-qti: Nit: `OrtMemTypeCPU` and `OrtMemTypeCPUOutput` have the same value, so either keeping this change or dropping it would be fine.

chunghow-qti: done

cmake/patches/ort_core/0005-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
chunghow-qti: done

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,148 @@
...
++    // "-i" option-parsing branch below is skipped. Detect the shared-allocator option
++    // here so inputs go through the QnnHtpShared (RPCMEM) allocator for zero-copy.
++    if (device_memory_name_.empty()) {
++#ifdef _MSC_VER
++      const std::string qnn_opt_string = ToUTF8String(performance_test_config.run_config.ep_runtime_config_string);
++#else
++      const std::string qnn_opt_string = performance_test_config.run_config.ep_runtime_config_string;
++#endif
++      if (qnn_opt_string.find("enable_htp_shared_memory_allocator|1") != std::string::npos) {
++        device_memory_name_ = "QnnHtpShared";
```
huaychou, line 63: **[N-1] (Minor)** Plugin-EP shared-memory detection does not confirm the plugin is actually the QNN EP

The new detection logic (`qnn_opt_string.find("enable_htp_shared_memory_allocator|1")`) searches the runtime config string of *any* plugin EP. If a future non-QNN plugin EP happens to include the same substring in its config string, perf_test will attempt to use the `QnnHtpShared` allocator and likely fail.

**Impact:** No effect on current behavior (only the QNN EP carries this key), but if a future plugin EP accidentally includes this substring in its config string, perf_test could attempt to use the wrong allocator.

**Suggested fix:**

Consider adding a defensive plugin-name check, or add a clarifying comment explaining why this key is QNN-EP-specific (e.g., `// QNN-EP specific option; no other plugin EP uses this key`) to document the implicit assumption and alert future maintainers.

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,148 @@
...
+3. Real .pb test data is loaded onto plain CPU memory by the shared TestCase
+   infrastructure, bypassing this session's allocator, so generated (-I) inputs
+   were the only ones reaching zero-copy. Intercept in PreLoadTestData (the common
+   sink for all input paths) and, when the QnnHtpShared allocator is active, re-wrap
+   CPU-backed input tensors onto it via CopyToSharedAllocator so real-data runs also
+   take the memHandle path. Tensors already on a HOST_ACCESSIBLE allocator, non-tensor
+   values, and string tensors are left unchanged.
+
+TODO: We may need update this patch when ort core uplevel, since there is some changes different to current release (1.26.0)
+
```
huaychou, line 28: **[T-1] (Nit)** Uplevel TODO is only in the patch header comment and not tracked

The patch header contains `TODO: We may need update this patch when ort core uplevel, since there is some changes different to current release (1.26.0)`, warning that this patch may need updating during future ORT core uplevels. However, this TODO is buried inside the `.patch` file's commit message body, making it easy to overlook during the upleveling workflow.

**Suggested fix:**

Add an inline comment next to the patch application line in `onnxruntime_external_deps.cmake` (e.g., `# TODO: review on ORT core uplevel — see patch header`), or file a tracking ticket, so the patch maintenance reminder is visible during the ORT core upleveling workflow.

cmake/patches/ort_core/0005-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,148 @@
...
+   allocator) so inputs take the zero-copy memHandle path.
+3. Real .pb test data is loaded onto plain CPU memory by the shared TestCase
+   infrastructure, bypassing this session's allocator, so generated (-I) inputs
+   were the only ones reaching zero-copy. Intercept in PreLoadTestData (the common
+   sink for all input paths) and, when the QnnHtpShared allocator is active, re-wrap
+   CPU-backed input tensors onto it via CopyToSharedAllocator so real-data runs also
+   take the memHandle path. Tensors already on a HOST_ACCESSIBLE allocator, non-tensor
+   values, and string tensors are left unchanged.
+
+TODO: We may need update this patch when ort core uplevel, since there is some changes different to current release (1.26.0)
```
huaychou:
```suggestion
TODO: We may need update this patch when ort core upleveling, since there are some changes different to current release (1.26.0)
```

huaychou (commented):
Please also create an PR on ORT core

cmake/patches/ort_core/0005-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
chunghow-qti: Done

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
chunghow-qti: Done

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
```diff
@@ -43,8 +43,9 @@ Ort::Status GroupQueryAttentionOpBuilder::IsOpSupported(QnnModelWrapper& qnn_mod
                                                         const Ort::Logger& logger) const {
   ORT_UNUSED_PARAMETER(logger);
 
-  RETURN_IF_NOT(IsGpuBackend(qnn_model_wrapper.GetQnnBackendType()),
-                "GroupQueryAttention is only supported with the GPU backend");
+  auto backend_type = qnn_model_wrapper.GetQnnBackendType();
+  RETURN_IF_NOT(IsGpuBackend(backend_type) || IsNpuBackend(backend_type),
+                "GroupQueryAttention is only supported with the GPU backend and HTP backend");
```
huaychou: Why do we have GQA change in this PR?

cmake/patches/ort_core/0005-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,148 @@
...
+   allocator) so inputs take the zero-copy memHandle path.
+3. Real .pb test data is loaded onto plain CPU memory by the shared TestCase
+   infrastructure, bypassing this session's allocator, so generated (-I) inputs
+   were the only ones reaching zero-copy. Intercept in PreLoadTestData (the common
+   sink for all input paths) and, when the QnnHtpShared allocator is active, re-wrap
+   CPU-backed input tensors onto it via CopyToSharedAllocator so real-data runs also
+   take the memHandle path. Tensors already on a HOST_ACCESSIBLE allocator, non-tensor
+   values, and string tensors are left unchanged.
+
+TODO: We need updat this patch when ort core upleveling, since there are some changes different to current release (1.26.0)
```
huaychou:
```suggestion
TODO: We need update this patch when ort core upleveling, since there are some changes different to current release (1.26.0)
```

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -148,7 +148,7 @@ size_t GetQnnTensorDataSizeInBytes(size_t num_elements, Qnn_DataType_t element_t
 size_t GetQnnTensorDataSizeInBytes(gsl::span<const uint32_t> shape, Qnn_DataType_t element_type) {
   // TODO can we just treat empty shape as a scalar?
   if (shape.empty()) {
-    ORT_CXX_API_THROW("Empty shape not allowed.", ORT_EP_FAIL);
+    return GetQnnTensorDataSizeInBytes(1, element_type);
```
huaychou: **[T-3] (Nit)** qnn_utils.cc TODO comment is now resolved but was not removed

`qnn_utils.cc` line 150 still has `// TODO can we just treat empty shape as a scalar?`, but this PR already answers the question by replacing the throw with a scalar-size return. The residual TODO is now a misleading stale comment and should be removed.

**Suggested fix:**

Remove the TODO comment at line 150.

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,148 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
huaychou: Just wondering, where does the `0000000000000000000000000000000000000000` come from?

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
chunghow-qti: Removed.

onnxruntime/core/providers/qnn/builder/opbuilder/group_query_attention_op_builder.cc
chunghow-qti: Removed

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
chunghow-qti: This is the standard git format-patch placeholder for the parent commit SHA. Since the patch was hand-authored (not generated from a real commit) as an ort_core patch, the From <hash> line just carries the all-zero placeholder. It's ignored by git apply / patch, so it has no effect on how the patch is applied.

cmake/patches/ort_core/0005-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
chunghow-qti: done

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
chunghow-qti: comment added

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
huaychou: I think we can remove this line if there is no effect.

chunghow-qti: Done

chunghow-qti:
Rebase and commit.

qti-chuteng (approved):

qti-chuteng:
### 🟩 Simplicity
**[D-1]** `🟠 Concern` `[Simplicity]` — `0006-...patch` (patched `ort_test_session.cc` constructor block)
The new plugin-EP detection does `qnn_opt_string.find("enable_htp_shared_memory_allocator|1") != npos` — a raw substring scan of the whole un-parsed config blob. But the *same file*'s traditional `-e qnn` branch (`ort_test_session.cc:274/358`) already parses this key the robust way: `ParseSessionConfigs` builds a key→value map, validates the value against `{"0","1"}` (`ORT_THROW` otherwise), then checks `key==... && value=="1"`. The same design decision is now encoded a second time in a shallower, more fragile form. Concrete divergences: (1) `find("...|1")` false-matches `...allocator|10`/`|11`; (2) the plugin path has no value validation at all; (3) whitespace variants behave differently from the sibling.
*Better design:* In the plugin branch, call the same `ParseSessionConfigs` into a local map and check `provider_options["enable_htp_shared_memory_allocator"] == "1"` — identical semantics/validation, and it converges both detection sites onto one parser. The helper is already included and used one branch over.

**[D-5]** `🔵 Improvement` `[Completeness]` — PR #658 description
The PR body says it delivers `0005-perftest-...` and "targets ort_core 1.26.0". The actual file is `0006-...` and its header says "Verified against ort_core 1.27.0" — which matches the real pin (`cmake/deps.txt:38` = v1.27.0), and 0005 is already taken by `0005-Suppress-C4875`. The committed artifacts are all correct; only the prose is stale, but a reviewer trusting the body would suspect either a numbering collision or a version-behind validation.
*Better design:* Fix the PR body to say 0006 / 1.27.0 so description, filename, and patch header all agree. No code change needed.

chunghow-qti:
Done

cmake/patches/ort_core/0006-perftest-QnnHtpShared-zerocopy-plugin-ep.patch
```diff
@@ -0,0 +1,155 @@
...
+   allocator) so inputs take the zero-copy memHandle path.
+3. Real .pb test data is loaded onto plain CPU memory by the shared TestCase
+   infrastructure, bypassing this session's allocator, so generated (-I) inputs
+   were the only ones reaching zero-copy. Intercept in PreLoadTestData (the common
+   sink for all input paths) and, when the QnnHtpShared allocator is active, re-wrap
+   CPU-backed input tensors onto it via CopyToSharedAllocator so real-data runs also
+   take the memHandle path. Tensors already on a HOST_ACCESSIBLE allocator, non-tensor
+   values, and string tensors are left unchanged.
+
+TODO: Verified against ort_core 1.27.0; re-check this patch on future ort core upleveling since the target file may change.
```
huaychou: If this patch targets at ort_core 1.27.0, do we need to uplevel ort_core first before applying this patch?

chunghow-qti: Updated, thanks!

kuanyul-qti (approved):

chunghow-qti:
This change fixes the htp_shared_memory issue in onnxruntime_perf_test.exe. We should be aware of this change, as there is currently an HTP shared memory issue in QAIRT 2.49 when an EP_Context takes both clientBuf and memHandle.

PR #659, opened 2026-07-22
[QNN EP] Use native QNN Rope Operator instead of decomposition

qti-ashwshan:
### Description
- Replace StridedSlice/Split/Gather/ElementWiseBinary/Concat decomposition with a single call to the native HTP QNN_OP_ROTARY_EMBEDDING op
- Reject scale != 1.0 and is_packed_batching != 0 in IsOpSupported, since the native op has no equivalent param and there's no decomposition fallback anymore
- Add RotaryEmbedding_Rank3Input test to exercise the new layout adapter path

### Motivation and Context
- Leveraging the QNN Native Rope Operator instead of decomposing to multiple primitive ops


onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -109,21 +98,17 @@ Ort::Status RotaryEmbeddingOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_w
...
 
-  // Validate rotary_embedding_dim
   if (rotary_embedding_dim == 0) {
     rotary_embedding_dim = head_size;
   }
-  RETURN_IF_NOT(rotary_embedding_dim > 0 && rotary_embedding_dim <= static_cast<int64_t>(head_size),
-                "rotary_embedding_dim must be > 0 and <= head_size");
-  RETURN_IF_NOT(rotary_embedding_dim % 2 == 0,
-                "rotary_embedding_dim must be even");
+  RETURN_IF_NOT(rotary_embedding_dim >= 2 && rotary_embedding_dim <= static_cast<int64_t>(head_size),
```
qti-mbadnara, line 114: So do we support odd dimensions now? If so can you add a unit test for this

onnxruntime/test/providers/qnn/rope_op_test.cc
```diff
@@ -65,33 +100,210 @@ TEST_F(QnnHTPBackendTests, RotaryEmbedding_Basic) {
...
+      TestInputDef<int64_t>({batch_size, seq_len}, false, std::vector<int64_t>{0, 1, 2, 3}),
+      cos_def_fp16,
+      sin_def_fp16,
+      {test::MakeAttribute("interleaved", int64_t{0}),
+       test::MakeAttribute("num_heads", num_heads),
+       test::MakeAttribute("rotary_embedding_dim", rotary_dim)},
+      /*opset_version*/ 23,
+      ExpectedEPNodeAssignment::All,
+      kDefaultRopeOpToleranceFp16);
+}
```
qti-mbadnara, line 125: Are we not testing any FP32 inputs?

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -29,59 +29,50 @@ class RotaryEmbeddingOpBuilder final : public BaseOpBuilder {
...
-                                       bool do_op_validation) const ORT_MUST_USE_RESULT;
 };
 
 Ort::Status RotaryEmbeddingOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
                                                     const OrtNodeUnit& node_unit,
                                                     const Ort::Logger& logger) const {
   ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, "Validating RotaryEmbedding op for QNN EP");
 
-  // Only support HTP backend for now
+  RETURN_IF_NOT(node_unit.Domain() != "com.microsoft",
```
qti-mbadnara: NIT: Change the description to reflect this as well. On first glance of the description, I could see only the inital rejection `Reject scale != 1.0 and is_packed_batching != 0 in IsOpSupported`

qti-yuduo: Dropping the `com.microsoft` domain is a silent support regression. The op registry keys on op type only, so contrib nodes do reach this builder, and the pre-PR code read the contrib input order `[input, position_ids, cos_cache, sin_cache]` — its only enabled test was `kMSDomain`/FP16 with `ExpectedEPNodeAssignment::All`, and that test is deleted here. `docs/execution_providers/QNN-ExecutionProvider.md:641` still advertises `com.microsoft:RotaryEmbedding|HTP backend only`, so contrib-op exports will now quietly fragment back to CPU.

Fix: either drop that docs row and call the change out in the PR description, or keep the domain and remap the input indices. If remapping, note the contrib op also allows 1D `position_ids` of shape `(1)` and has `scale` / `is_packed_batching` attributes that need rejecting when non-default (the old builder ignored both). Also prefer the `kMSDomain` constant over the raw string literal.

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -29,59 +29,50 @@ class RotaryEmbeddingOpBuilder final : public BaseOpBuilder {
...
                                   const Ort::Logger& logger) const ORT_MUST_USE_RESULT;
-
-  Ort::Status DecomposeRotaryEmbedding(QnnModelWrapper& qnn_model_wrapper,
-                                       const OrtNodeUnit& node_unit,
-                                       std::vector<std::string>&& input_names,
-                                       const Ort::Logger& logger,
-                                       bool do_op_validation) const ORT_MUST_USE_RESULT;
 };
 
 Ort::Status RotaryEmbeddingOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
```
qti-yuduo, line 41: `IsOpSupported` never calls `BaseOpBuilder::IsOpSupported` / `AddToModelBuilder(..., do_op_validation=true)`, so `do_op_validation` is `false` on every `CreateQnnNode` and nothing is checked against the backend. Anything the hand-rolled checks below miss now surfaces as a `graphFinalize` failure inside `Compile` — session creation throws instead of the node falling back to CPU. That mattered less with the decomposition in place; now the full constraint set of the native op applies.

Fix: end `IsOpSupported` with `BaseOpBuilder::IsOpSupported(...)` (as `GroupQueryAttentionOpBuilder` does) so unmodeled constraints degrade to CPU fallback rather than a hard session-creation error.

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
qti-yuduo, line 114: The evenness guard was lost — `main` had `rotary_embedding_dim % 2 == 0` in addition to the range check. With `rotary_embedding_dim=5, head_size=8`, `expected_cache_dim = 5/2 = 2` truncates, so a cache with last dim 2 passes validation and only 4 of the 5 elements get rotated. Same hole for 4D input with odd `head_size` and no attribute (ONNX requires `head_size` even for 4D, and nothing checks it now).

Fix: restore `RETURN_IF_NOT(rotary_embedding_dim % 2 == 0, ...)` alongside the range check.

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -109,21 +98,17 @@ Ort::Status RotaryEmbeddingOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_w
...
   std::vector<uint32_t> cos_shape;
-  RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[2].shape, cos_shape),
+  RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[1].shape, cos_shape),
                 "Cannot get cos_cache shape");
   std::vector<uint32_t> sin_shape;
-  RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[3].shape, sin_shape),
+  RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[2].shape, sin_shape),
                 "Cannot get sin_cache shape");
 
   RETURN_IF_NOT(cos_shape.size() == 2 || cos_shape.size() == 3,
```
qti-yuduo: Cache rank isn't tied to `position_ids` presence — rank 2 or 3 is accepted in both modes. The backend requires rank 3 `[batch, seq, rd/2]` when `position_ids` is absent and rank 2 `[max_pos, rd/2]` when present, and ONNX fixes this the same way. Combined with the missing backend validation, a mismatch is reported as supported and then aborts at finalize instead of falling back.

Fix: gate the accepted rank on `has_position_ids` (rank 2 with, rank 3 without).

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -139,10 +124,9 @@ Ort::Status RotaryEmbeddingOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_w
...
                 "sin_cache last dimension must equal rotary_embedding_dim/2");
 
-  // Validate position_ids (always at index 1)
-  if (inputs[1].Exists()) {
+  if (inputs.size() > 3 && inputs[3].Exists()) {
     std::vector<uint32_t> pos_ids_shape;
-    RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[1].shape, pos_ids_shape),
+    RETURN_IF_NOT(qnn_model_wrapper.GetOnnxShape(inputs[3].shape, pos_ids_shape),
                   "Cannot get position_ids shape");
     RETURN_IF_NOT(pos_ids_shape.size() == 2,
```
qti-yuduo, line 148: `position_ids` is only rank-checked; its dims aren't compared against `[batch_size, seq_len]`, which the backend does enforce. A graph whose `position_ids` is `[B, S']` with `S' != seq_len` is claimed supported and then fails hard at finalize.

Fix: compare `pos_ids_shape` to `{batch_size, seq_len}` here.

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -169,759 +153,133 @@ Ort::Status RotaryEmbeddingOpBuilder::ValidateInputShapes(QnnModelWrapper& qnn_m
...
-    RETURN_IF_NOT(qnn_model_wrapper.CreateQnnNode(
-                      mul_sin_x1_name,
-                      QNN_OP_PACKAGE_NAME_QTI_AISW,
-                      QNN_OP_ELEMENT_WISE_BINARY,
-                      std::move(mul_input_names),
-                      {sin_x1},
-                      std::move(mul_sin_x1_param_names),
-                      do_op_validation),
-                  "Failed to create Multiply node for sin*x1");
+  if (rotary_embedding_dim != 0) {
```
qti-yuduo: This branch means the QNN param is omitted whenever the ONNX attribute is absent or 0 — which is the common case, since ONNX defaults it to 0 meaning "full rotation over `head_size`" and most exporters omit it. We then rely on the QNN default equalling `head_size`, which isn't documented in `QnnOpDef.h`; HTP has a `rotary_embedding_dim must be greater than 1` constraint, so a backend default of 0 would fail at finalize. All six new tests set the attribute explicitly, so this path is never exercised.

Fix: drop the `!= 0` guard and always emit the resolved value (`head_size` when the attribute is 0 — already computed above), and/or add a test that omits the attribute.

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -169,759 +153,133 @@ Ort::Status RotaryEmbeddingOpBuilder::ValidateInputShapes(QnnModelWrapper& qnn_m
...
-                      mul_sin_x2_name,
-                      QNN_OP_PACKAGE_NAME_QTI_AISW,
-                      QNN_OP_ELEMENT_WISE_BINARY,
-                      std::move(mul_input_names),
-                      {sin_x2},
-                      std::move(mul_sin_x2_param_names),
-                      do_op_validation),
-                  "Failed to create Multiply node for sin*x2");
+  std::vector<std::string> rope_input_names = {native_input, input_names[1], input_names[2]};
+  const bool has_position_ids = (inputs.size() > 3 && inputs[3].Exists());
```
qti-yuduo, line 258: `has_position_ids` correctly guards `input_names[3]` here, but `ProcessInputs` isn't overridden and `RotaryEmbedding` isn't in the required-input-count map, so `GetInputCountQnnRequired` returns `Inputs().size()` and `ProcessInput` runs on the absent def — `MakeTensorWrapper` → `GetTensorInfo` then fails during `Compile`, after `IsOpSupported` already claimed the node. Only affects models passing an explicit `""` 4th input; the omitted-trailing form is fine.

Fix: override `ProcessInputs` and skip (or `QnnTensorWrapper::MakeNull`) the absent optional, as `attention_op_builder.cc` does.

onnxruntime/test/providers/qnn/rope_op_test.cc
```diff
@@ -52,10 +88,9 @@ static void RunRopeOpTest(const TestInputDef<DataType>& input_def,
                   EPVerificationParams{expected_ep_assignment, ElementwiseAbsoluteVerifier(abs_err)});
 }
 
-// Basic test with FP16 data type (QNN EP only supports FP16 for RotaryEmbedding)
+// Basic 4D input with position_ids, interleaved=0, full rotation.
 TEST_F(QnnHTPBackendTests, RotaryEmbedding_Basic) {
```
qti-yuduo, line 92: FP16 coverage is gone. The deleted test was FP16 — with the comment "QNN EP only supports FP16 for RotaryEmbedding" and a `SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V68)` guard — while all six new tests are FP32, and `IsOpSupported` still advertises FP16.

Fix: add an FP16 variant of this test, keeping the arch skip guard.

onnxruntime/core/providers/qnn/builder/opbuilder/rotary_embedding_op_builder.cc
```diff
@@ -169,759 +170,141 @@ Ort::Status RotaryEmbeddingOpBuilder::ValidateInputShapes(QnnModelWrapper& qnn_m
...
-                      QNN_OP_ELEMENT_WISE_BINARY,
-                      std::move(mul_input_names),
-                      {sin_x2},
-                      std::move(mul_sin_x2_param_names),
-                      do_op_validation),
-                  "Failed to create Multiply node for sin*x2");
+  // Step 2: build the native QNN_OP_ROTARY_EMBEDDING node.
+  // Native input order: [X, cos_cache, sin_cache, position_ids(optional)].
+  std::vector<std::string> rope_input_names = {native_input, input_names[2], input_names[3]};
+  const bool has_position_ids = inputs[1].Exists();
```
qti-arnadesh: The new builder handles position IDs as optional.
I think we should add coverage for the absent-input case.

qti-mbadnara (approved):
LGTM

qti-yuduo (approved):

PR #671, opened 2026-07-27
[QNN EP] BatchNorm: promote to F32 when any ONNX param is per-channel

qti-ashwshan:
### Description
- QNN BN folds gamma/beta/mean/var into fused_scale/fused_bias with a single per-tensor QNN quant scale
- If any ONNX BN param is per-channel, squeezing into per-tensor loses precision
- Extend GetBatchNormFloatExecution to trigger float-promotion when bias/mean/var are per-channel (not just scale)
- Add regression test: BatchNorm2D_PerTensorGamma_PerChannelVar_U16
- Follow-up: emit per-channel fused params natively for U16 activation


onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -500,24 +500,40 @@ void OverrideParamTypeForRequantize(Qnn_DataType_t x_dtype,
 // Single source of truth for float execution, shared by ProcessInputs (stores params) and
 // ProcessAttributesAndOutputs (emits the op).
 //   - has_float_output: quantized input, no output Q -> float island; BN emits float directly.
-//   - use_float_params: also covers u8/u16 input with per-channel scale, whose fused weight
-//     gamma/sqrt(var+eps) can overflow a single per-tensor requant scale.
+//   - use_float_params: also covers u8/u16 input where any ONNX BN param (scale/bias/mean/var)
```
tirupath-qti: now it also covers not only per-channel params but also any float param. this description needs to be updated.

onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -500,24 +500,35 @@ void OverrideParamTypeForRequantize(Qnn_DataType_t x_dtype,
...
+  const bool any_param_per_channel = scale_info.quant_param.IsPerChannel() ||
+                                     bias_info.quant_param.IsPerChannel() ||
+                                     mean_info.quant_param.IsPerChannel() ||
+                                     var_info.quant_param.IsPerChannel();
+  const bool has_unquantized_params = is_quantized_op &&
+                                      (!mean_info.quant_param.IsQuantized() ||
+                                       !var_info.quant_param.IsQuantized());
   const bool use_float_params =
       has_float_output ||
       (is_quantized_op &&
```
yath1: Nit: (not introduced in your PR) is_quantized_op is essentially "is_input_quantized", can we rename it as such?  
Also, If is_input_quantized==false (ie input is float irrespective of what the output precision is), we need to set use_float_params to true.

onnxruntime/test/providers/qnn/batchnormalization_test.cc
```diff
@@ -883,6 +883,95 @@ TEST_F(QnnHTPBackendTests, BatchNorm2D_NearZeroVariance_U8) {
                        qdq_model_fn, provider_options, 21, ExpectedEPNodeAssignment::All);
 }
 
+// Regression: gamma is per-tensor but var is per-channel. The fused weight
```
yath1, line 886: Generally, all the params are quantized as either per-channel or per-tensor (except bias which can be float). If scale is per-channel so would mean and var. But okay to explicitly check to avoid incorrectly changing the mode in EP (per-tensor to per-channel and vice-versa). Maybe add other negative case to solidify testing - ie variance in per-tensor and scale in per-channel.

onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -500,24 +500,35 @@ void OverrideParamTypeForRequantize(Qnn_DataType_t x_dtype,
...
+                                                   const TensorInfo& mean_info,
+                                                   const TensorInfo& var_info,
                                                    const TensorInfo& output_info) {
   const bool is_quantized_op = input_info.quant_param.IsQuantized();
   const bool has_float_output = is_quantized_op && !output_info.quant_param.IsQuantized();
+  const bool any_param_per_channel = scale_info.quant_param.IsPerChannel() ||
+                                     bias_info.quant_param.IsPerChannel() ||
+                                     mean_info.quant_param.IsPerChannel() ||
+                                     var_info.quant_param.IsPerChannel();
+  const bool has_unquantized_params = is_quantized_op &&
```
yath1: Nit: Could you add a comment in code on why bias_info is not checked for has_unquantized_params here?

yath1 (commented):
Left a few minor comments. Could you also explain if you saw this issue in any quantized model? General assumption is all params are similarly quantized (per-tensor or per-channel)

tirupath-qti:
> Left a few minor comments. Could you also explain if you saw this issue in any quantized model? General assumption is all params are similarly quantized (per-tensor or per-channel)

In a quantized model, we have seen scale and bias are per-tensor quantized but input_var and input_mean are left in float.

qti-ashwshan:
Hi @yath1 and @tirupath-qti  
Addressed all the nits. Please re-review. thanks

onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
```diff
@@ -500,24 +500,36 @@ void OverrideParamTypeForRequantize(Qnn_DataType_t x_dtype,
...
+                                     mean_info.quant_param.IsPerChannel() ||
+                                     var_info.quant_param.IsPerChannel();
+  // bias excluded: OverrideParamTypeForRequantize already converts float bias to int32.
+  const bool has_unquantized_params = is_input_quantized &&
+                                      (!mean_info.quant_param.IsQuantized() ||
+                                       !var_info.quant_param.IsQuantized());
   const bool use_float_params =
       has_float_output ||
-      (is_quantized_op &&
+      (is_input_quantized &&
```
yath1, line 529: This logic still needs updates. Like I raised earlier, if input is in float and output is quantized, this logic would not set use_float_params to True. Please update it

qti-ashwshan: Yes, but here float input + quantized output doesn't happen in QDQ models since NodeUnit wrapping ensures input is quantized when output is. So we can't set use_float_params=true here because it emits a Dequantize on the input, which would fail on a float tensor right ?

yath1: This situation would be present in mixed precision cases. We would need to relax the selectors, add converts as needed. Feel free to resolve this conversation (probably unrelated to the scope of this PR), I can make this change as part of a different PR I am working on now.

qti-ashwshan: Yeah, sounds good!

qti-ashwshan: Yeah I think this would not be in the scope of this PR so I will mak this as resolved. Thanks for the review!

qti-kromero (approved):

PR #674, opened 2026-07-28
Reconcile RMSNorm scale rank with QNN RmsNorm OpDef

qti-yuduo:
## Summary

QNN's `RmsNorm` OpDef requires **`rank(gamma) == size(axes)`**, while ONNX `RMSNormalization` (opset 23) allows `Scale` to be *any* shape unidirectionally broadcastable to `X`. A graph with `X [1, 50, 1024]` and `scale [1, 1, 1024]` (axis `-1`) is therefore perfectly valid ONNX but is rejected by QNN op validation with `QNN_OP_PACKAGE_ERROR_VALIDATION_FAILURE` (error code `3110`).

`RMSNormalizationOpBuilder` previously passed the scale through verbatim, and the SDK-<2.49 dummy-beta path copied `scale_info.shape`, propagating the bad rank to beta as well. Every such node was therefore rejected during partitioning and fell back to CPU EP, fragmenting the graph into one QNN subgraph per surviving span.

This PR makes the EP reconcile the rank, which is the correct layer for it: the EP is what knows the QNN OpDef.

## Changes

`onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc`

1. **`ProcessInputs` — squeeze leading 1-dims off the scale** to reach `size(axes)`:
   - **static scale**: re-declared at the reduced rank via a new `QnnTensorWrapper`. Dropping leading 1-dims doesn't change the element layout, so no `Reshape` node is needed — only the declared dims change. Emitted under a `_squeeze`-suffixed unique name so a scale shared with another consumer keeps its original rank there.
   - **dynamic scale**: `qnn_model_wrapper.AddReshapeNode(...)` — the established in-graph shape-reconciliation idiom already used by the conv, softmax, instancenorm, einsum and rotary_embedding builders.
2. **Dummy beta inherits the squeezed rank** — `scale_info.shape` is updated post-squeeze, so the existing `beta_shape = scale_info.shape` becomes correct with no extra logic.
3. A scale that genuinely cannot be reconciled (non-1 leading dim) is rejected in `ProcessInputs` with an explicit message. Because `IsOpSupported` builds the node with `do_op_validation=true`, that rejection propagates out during partitioning and the node is simply never claimed — no separate `IsOpSupported` check is needed.
4. Only per-tensor quantization is accepted on a scale that needs squeezing. Axis-bearing and per-block encodings describe the pre-squeeze rank and there is no helper to remap one for a rank reduction (`QnnQuantParamsWrapper` has `HandleUnsqueeze`/`HandleTranspose` but no rank-reducing counterpart), so they are rejected rather than silently emitted misaligned.

The reconciliation is driven by `size(axes)`, not by a hardcoded rank, so rank-2/3/4 scales are all handled by the same path.

## Motivation — real model

Found while triaging **Pi05ActionExpert** (the π0.5 action expert) in [qcom-ai-hub/tetracode#20549](https://github.com/qcom-ai-hub/tetracode/issues/20549). Its 18 transformer blocks each normalize `X [1, 50, 1024]` with a scale of shape `[1, 1, 1024]` — 37 `RMSNormalization` sites in total, every scale a *computed* `Add` output (the `weight + 1` idiom), so constant folding cannot flatten it. Measured on the real model through the full compile pipeline:

| Configuration | `RMSNormalization` ops | QNN subgraphs (DLCs) | Result |
|---|---|---|---|
| Rank-3 scale (as shipped) | 37 | **38** | ❌ 107 × `backendValidateOpConfig() failed ... error code 3110`, 37 nodes on CPU EP |
| RMSNorm fusion disabled | 0 (37 `ReduceMean` decompositions) | **1** | ✅ compiles, but loses the fused op |
| Rank-1 scale | 37 | **1** | ✅ compiles, `RmsNorm` retained |

The constraint is **rank**, not dims — verified by isolation:

```
gamma=[1024]     x=[1,50,1024]  -> PASS   (rank == size(axes))
gamma=[1]        x=[1,50,1024]  -> PASS   (rank 1 accepted even though dim != 1024)
gamma=[1,1024]   x=[1,50,1024]  -> FAIL 3110
gamma=[1,1,1024] x=[1,50,1024]  -> FAIL 3110
gamma=[1,1,1]    x=[1,50,1024]  -> FAIL 3110
```

After the equivalent rank fix, the DLC contains 37 `RmsNorm` ops with `axes: [2]` and rank-1 `[1024]` betas, and numerics match the original ONNX on CPU EP to `max_abs_diff = 7.15e-07` / `max_rel_diff = 4.61e-07`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
@@ -1,6 +1,7 @@
 // Copyright (c) Qualcomm. All rights reserved.
 // Licensed under the MIT License.
 
+#include <algorithm>
 #include <cassert>
 #include <optional>
 #include <string>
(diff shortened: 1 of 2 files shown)


yath1 (approved):
lgtm. Nice doc! :)

PR #677, opened 2026-07-28
[QNN EP] Restore file mapping in SSR context recovery

qti-hungjuiw:
### Description

- File mapping continuity after SSR
- Test coverage for file mapping recovery
    - Add `QnnMockSSR_WasFileMappingUsedForRecovery()` export to `QnnMockSSR.dll` — intercepts `contextCreateFromBinaryWithCallback` after SSR injection (`g_ssr_fired == true`) to distinguish recovery calls from initial load calls.
    - `SSRGraphExecuteEpContextNonEmbedMode` now asserts that file mapping was used during recovery when the platform supports it.

### Motivation and Context

- After SSR, RecoverFromSSR recreated the QNN context using only the direct read path (`ReadContextBinIfValid` + `contextCreateFromBinary`), even when the session was originally loaded with **file mapping** enabled. This meant SSR recovery silently lost the performance benefit of file mapping on production Windows ARM64 devices.                                                                                                                                                                                                      
- The **WindowsFileMapper** already keeps the MapViewOfFile alive across SSR events (the file mapping itself is session-scoped, not context-scoped), so recovery can safely reuse the existing mapped memory pointer via `GetContextBinMappedMemoryPtr` and re-register it with the new context via the DMA callbacks.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -1281,6 +1281,27 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::BuildContextBinaryConfigs(
+    int64_t max_spill_fill_size,
+    Qnn_ContextHandle_t first_group_handle,
+    QnnConfigsBuilder<QnnContext_Config_t, QnnHtpContext_CustomConfig_t>& configs_builder) {
+  gsl::not_null<QnnContext_Config_t*> priority_cfg = configs_builder.PushConfig();
+  RETURN_IF_ERROR(SetQnnContextConfig(context_priority_, *priority_cfg));
+
+  if (max_spill_fill_size > 0) {
+    gsl::not_null<QnnHtpContext_CustomConfig_t*> spill_custom = configs_builder.PushCustomConfig();
+    spill_custom->option = QNN_HTP_CONTEXT_CONFIG_OPTION_REGISTER_MULTI_CONTEXTS;
+    spill_custom->groupRegistration.firstGroupHandle = first_group_handle;
+    spill_custom->groupRegistration.maxSpillFillBuffer = max_spill_fill_size;
+
+    gsl::not_null<QnnContext_Config_t*> spill_cfg = configs_builder.PushConfig();
+    spill_cfg->option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
+    spill_cfg->customConfig = spill_custom;
+  }
+
+  return Ort::Status();
+}
+
 Ort::Status QnnBackendManager::CreateContextHandleFromBinary(
     void* bin_buffer,
     uint64_t buffer_length,
@@ -1357,22 +1378,8 @@ Ort::Status QnnBackendManager::ReloadContextForSSR(const std::string& context_bi
                                                    const qnn::EpContextIoDispatch& io_dispatch) {
   QnnConfigsBuilder<QnnContext_Config_t, QnnHtpContext_CustomConfig_t> configs_builder(
       QNN_CONTEXT_CONFIG_INIT, QnnHtpContext_CustomConfig_t{});
-
-  auto* priority_cfg = configs_builder.PushConfig();
-  RETURN_IF_ERROR(SetQnnContextConfig(context_priority_, *priority_cfg));
-
-  // Spill-fill: re-register buffer allocation for the recovered context.
-  // Always a new group (firstGroupHandle = 0x0) since SSR invalidated all prior handles.
-  if (max_spill_fill_size > 0) {
-    auto* spill_custom = configs_builder.PushCustomConfig();
-    spill_custom->option = QNN_HTP_CONTEXT_CONFIG_OPTION_REGISTER_MULTI_CONTEXTS;
-    spill_custom->groupRegistration.firstGroupHandle = 0x0;
-    spill_custom->groupRegistration.maxSpillFillBuffer = max_spill_fill_size;
-
-    auto* spill_cfg = configs_builder.PushConfig();
-    spill_cfg->option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
-    spill_cfg->customConfig = spill_custom;
-  }
+  // SSR always starts a new spill-fill group (firstGroupHandle = 0x0): all prior handles are invalid.
+  RETURN_IF_ERROR(BuildContextBinaryConfigs(max_spill_fill_size, 0x0, configs_builder));
 
   void* bin_buffer = nullptr;
   uint64_t buffer_length = 0;
@@ -1963,34 +1970,16 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
 
   } else {
 #endif
-    QnnContext_Config_t qnn_context_config = QNN_CONTEXT_CONFIG_INIT;
-    RETURN_IF_ERROR(SetQnnContextConfig(context_priority_, qnn_context_config));
-
-    // Register spill fill buffer for multi context
-    QnnContext_Config_t spill_fill_config = QNN_CONTEXT_CONFIG_INIT;
-
-    // The spill fill buffer is available since 2.28, API version starts from 2.21
-#if QNN_API_VERSION_MAJOR == 2 && (QNN_API_VERSION_MINOR >= 21)
-    QnnHtpContext_CustomConfig_t custom_config;
-    custom_config.option = QNN_HTP_CONTEXT_CONFIG_OPTION_REGISTER_MULTI_CONTEXTS;
-    QnnHtpContext_GroupRegistration_t group_info;
+    // Join the existing spill-fill group if one exists; otherwise start a new one.
     size_t current_contexts_size = GetQnnContextSize();
-    // set to 0x0 (new group) if this is the first context, otherwise point to the first context handle
-    // note that we already move the context with max spill fill size to the beginning of the list
-    group_info.firstGroupHandle = (max_spill_fill_size > 0 && current_contexts_size > 0) ? GetQnnContext(0) : 0x0;
-    group_info.maxSpillFillBuffer = max_spill_fill_size;  // Max spill-fill buffer across contexts. Must be >0
-    custom_config.groupRegistration = group_info;
-    spill_fill_config.option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
-    spill_fill_config.customConfig = &custom_config;
-
-#endif
-
-    QnnContext_Config_t* spill_fill_config_pointer = max_spill_fill_size > 0 ? &spill_fill_config : nullptr;
-    ORT_CXX_LOG_PTR(logger_ptr_,
-                    ORT_LOGGING_LEVEL_VERBOSE,
+    Qnn_ContextHandle_t first_group_handle =
+        (max_spill_fill_size > 0 && current_contexts_size > 0) ? GetQnnContext(0) : 0x0;
+    ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_VERBOSE,
                     ("Max spill fill buffer size: " + std::to_string(max_spill_fill_size)).c_str());
 
-    const QnnContext_Config_t* context_configs[] = {&qnn_context_config, spill_fill_config_pointer, nullptr};
+    QnnConfigsBuilder<QnnContext_Config_t, QnnHtpContext_CustomConfig_t> configs_builder(
+        QNN_CONTEXT_CONFIG_INIT, QnnHtpContext_CustomConfig_t{});
+    RETURN_IF_ERROR(BuildContextBinaryConfigs(max_spill_fill_size, first_group_handle, configs_builder));
 
     qnn::profile::ProfilingInfo profiling_info;
 #ifdef QNN_SYSTEM_PROFILE_API_ENABLED
@@ -2000,8 +1989,8 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
 #endif
 
     RETURN_IF_ERROR(CreateContextHandleFromBinary(bin_buffer, buffer_length, use_file_mapping,
-                                                  context_configs, context_bin_filepath, io_dispatch,
-                                                  context));
+                                                  configs_builder.GetQnnConfigs(), context_bin_filepath,
+                                                  io_dispatch, context));
 
 #ifdef QNN_SYSTEM_PROFILE_API_ENABLED
     if (ProfilingEnabled()) {
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -1281,6 +1281,76 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::CreateContextHandleFromBinary(
+    void* bin_buffer,
+    uint64_t buffer_length,
+    bool use_file_mapping,
+    const QnnContext_Config_t** context_configs,
+    const std::string& context_bin_filepath,
+    const qnn::EpContextIoDispatch& io_dispatch,
+    Qnn_ContextHandle_t& context) {
+  RETURN_IF(nullptr == qnn_interface_.contextCreateFromBinary,
+            "Invalid function pointer for contextCreateFromBinary.");
+
+  Qnn_ErrorHandle_t rt = QNN_SUCCESS;
+
+#ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
+  if (use_file_mapping && file_mapper_) {
+    RETURN_IF(nullptr == qnn_interface_.contextCreateFromBinaryWithCallback,
+              "Invalid function pointer for contextCreateFromBinaryWithCallback.");
+
+    auto notify_param_ptr = std::make_unique<FileMappingCallbackInfo_t>(bin_buffer, buffer_length, this);
+
+    Qnn_ContextBinaryCallback_t callbacks;
+    callbacks.type = QNN_CONTEXT_CALLBACK_DMA_BUFFER;
+    callbacks.dmaBufferCallback.version = QNN_CONTEXT_CALLBACK_DMA_BUFFER_VERSION_1;
+    callbacks.dmaBufferCallback.v1.dataProvide = MapDmaDataCallback;
+    callbacks.dmaBufferCallback.v1.dataRelease = ReleaseDmaDataCallback;
+    callbacks.dmaBufferCallback.v1.notifyParam = reinterpret_cast<void*>(notify_param_ptr.get());
+
+    file_mapping_notify_params_.push_back(std::move(notify_param_ptr));
+
+    rt = qnn_interface_.contextCreateFromBinaryWithCallback(backend_handle_,
+                                                            device_handle_,
+                                                            context_configs,
+                                                            &callbacks,
+                                                            bin_buffer,
+                                                            static_cast<Qnn_ContextBinarySize_t>(buffer_length),
+                                                            &context,
+                                                            profile_backend_handle_,
+                                                            NULL);
+    if (rt != QNN_SUCCESS) {
+      ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_WARNING,
+                      ("contextCreateFromBinaryWithCallback failed (" + QnnErrorHandleToString(rt) +
+                       "). Retrying with direct read.")
+                          .c_str());
+    }
+  }
+#endif
+
+  if (!use_file_mapping || rt != QNN_SUCCESS) {
+    // Read from file if no pre-read buffer was supplied or if file mapping failed.
+    std::vector<char> file_buffer;
+    if (bin_buffer == nullptr || rt != QNN_SUCCESS) {
+      RETURN_IF_ERROR(ReadContextBinIfValid(context_bin_filepath, file_buffer, io_dispatch));
+      bin_buffer = static_cast<void*>(file_buffer.data());
+      buffer_length = static_cast<uint64_t>(file_buffer.size());
+    }
+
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1266,6 +1266,96 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
...
+    Qnn_ContextBinaryCallback_t callbacks;
+    callbacks.type = QNN_CONTEXT_CALLBACK_DMA_BUFFER;
+    callbacks.dmaBufferCallback.version = QNN_CONTEXT_CALLBACK_DMA_BUFFER_VERSION_1;
+    callbacks.dmaBufferCallback.v1.dataProvide = MapDmaDataCallback;
+    callbacks.dmaBufferCallback.v1.dataRelease = ReleaseDmaDataCallback;
+    callbacks.dmaBufferCallback.v1.notifyParam = reinterpret_cast<void*>(notify_param_ptr.get());
+
+    file_mapping_notify_params_.push_back(std::move(notify_param_ptr));
+
+    rt = qnn_interface_.contextCreateFromBinaryWithCallback(backend_handle_,
```
kuanyul-qti, line 1334: **[D-1] (D)** [Arch] SSR recovery path missing blob version gate — behavioral divergence from initial-load sibling

`LoadCachedQnnContextFromBuffer` calls `GetGraphInfoAndBinVersion` to extract `blob_version`, then checks `MinVersionMet(blob_version, {3,3,3})` before ever invoking `contextCreateFromBinaryWithCallback`: if the binary is < 3.3.3, it logs a clear warning and falls back to direct read. The new SSR recovery path `CreateContextFromFilePath` has no such gate — it calls `contextCreateFromBinaryWithCallback` unconditionally, relying only on `rt != QNN_SUCCESS` as a fallback.

The existing code treats the version check as a *precondition*, not error recovery, because the QNN runtime's behavior when passed a pre-3.3.3 binary to a newer API is not guaranteed to be a clean error return. This gives the SSR recovery path a weaker behavioral contract than the initial-load sibling: a session using a pre-3.3.3 binary correctly falls back to direct read on initial load but may encounter undefined behavior after SSR.

**Suggested fix:**

Add the same version gate before calling `contextCreateFromBinaryWithCallback` — mirror the initial-load sibling: call `GetGraphInfoAndBinVersion` to get `blob_version`, check `MinVersionMet(blob_version, {3,3,3})`, log a warning and fall through to direct read if not met. Both `GetSystemContextHandle()` and `GetGraphInfoAndBinVersion()` are private methods on `QnnBackendManager`, so no interface change is needed.

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
```diff
@@ -137,6 +143,15 @@ TEST_F(QnnMockSSRBackendTests, SSRGraphExecuteEpContextNonEmbedMode) {
...
                        run_session_opts);
 
+  // Verify that SSR recovery used file mapping (contextCreateFromBinaryWithCallback)
+  // when the feature is available.  On platforms without file mapping support this
+  // assertion is vacuously true (the function always returns false there).
+#ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
+  EXPECT_TRUE(QnnMockSSR_WasFileMappingUsedForRecovery())
+      << "SSR recovery should have used contextCreateFromBinaryWithCallback (file mapping) "
+         "but it did not. Check CreateContextFromFilePath in qnn_backend_manager.cc.";
+#endif
```
kuanyul-qti, line 182: **[D-2] (D)** [Completeness] No test for pre-3.3.3 blob + file-mapping-enabled direct-read fallback path

The existing test `SSRGraphExecuteEpContextNonEmbedMode` asserts that file mapping is used (under `QNN_FILE_MAPPED_WEIGHTS_AVAILABLE`). No test covers the path where blob version < 3.3.3 causes `CreateContextFromFilePath` to proactively fall back to direct read even when `file_mapped_weights_enabled_` is true. This path only becomes meaningful once D-1 is fixed, but a test cell is needed to confirm the fallback works correctly and does not leave the session in a broken state.

**Suggested fix:**

Add a mock SSR test mode where the `contextCreateFromBinaryWithCallback` interceptor returns a version-unsupported error, and assert: (a) `WasFileMappingUsedForRecovery() == false`, and (b) session outputs are still correct (direct-read fallback produced a working context). This is a mutation test for the version-gate path added by D-1.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: After the move, `#include HTP/QnnHtpContext.h` in `qnn_model.cc` becomes a dead include.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: Done. Thanks for pointing this out!

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
qti-hungjuiw: Done. Added `QnnMockSSRBackendTests.SSRGraphExecuteEpContextNonEmbedModeFileMappingRejected`. Thanks!

onnxruntime/core/providers/qnn/builder/qnn_model.cc
qti-hungjuiw: Fixed. Thanks!

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
```diff
@@ -161,6 +172,91 @@ TEST_F(QnnMockSSRBackendTests, SSRGraphExecuteEpContextNonEmbedMode) {
...
+  ASSERT_TRUE(std::filesystem::exists(context_model_file))
+      << "Context model file was not generated: " << context_model_file;
+
+  // -----------------------------------------------------------------------
+  // Step 2: Load the context model via QnnMockSSR.dll with the callback API
+  //         rejected, simulating a pre-3.3.3 blob.  Recovery must succeed via
+  //         the direct-read fallback.
+  // -----------------------------------------------------------------------
+#ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
+  QnnMockSSR_SetSimulateOldBlobVersion(true);
```
huaychou, line 241: **[M-1] (Major)** Test flag g_simulate_old_blob_version not reset when test exits early

**Problem**

In `SSRGraphExecuteEpContextNonEmbedModeFileMappingRejected`, line 237 calls `QnnMockSSR_SetSimulateOldBlobVersion(true)` before executing `TestQDQModelAccuracy`. If `TestQDQModelAccuracy` triggers an internal `ASSERT_*` failure (which returns from the test function under gtest convention), the reset `QnnMockSSR_SetSimulateOldBlobVersion(false)` at line 257 is never reached.

Per the comment at `qnn_mock_ssr.cc` line 48: “NOT reset by QnnInterface_getProviders; tests must set/clear this explicitly.” — there is no safety net. If this test fails, subsequent SSR tests sharing the same mock DLL (e.g. `SSRGraphExecuteEpContextNonEmbedMode`) will have file mapping unexpectedly rejected during recovery.

**Impact**

State contamination across the test suite: a failure in this test leaves `g_simulate_old_blob_version` stuck at `true`, causing subsequent SSR tests to receive `QNN_COMMON_ERROR_NOT_SUPPORTED` during recovery and fail or produce incorrect results, potentially masking real bugs. Severity: Major.

**Suggested fix:**

Move the SetSimulateOldBlobVersion(true)/false pair into a RAII guard so the flag is always reset even on early test exit:

```cpp
#ifdef QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
  struct OldBlobGuard {
    ~OldBlobGuard() { QnnMockSSR_SetSimulateOldBlobVersion(false); }
  } old_blob_guard;
  QnnMockSSR_SetSimulateOldBlobVersion(true);
#endif
```

Alternatively, unconditionally call `QnnMockSSR_SetSimulateOldBlobVersion(false)` in the test fixture's `TearDown()` so the flag is always cleared regardless of test outcome.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -206,6 +206,13 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   Ort::Status ReadContextBinIfValid(const std::string& context_bin_filepath,
                                     std::vector<char>& buffer);
```
huaychou: **[T-1] (Nit)** Double space in header comment

**Problem**

The header comment above the new `CreateContextFromFilePath` declaration at line 208 contains a double space: `// paths.  Used by SSR recovery` (two spaces after the period instead of one).

**Suggested fix:**

```
// paths. Used by SSR recovery so it matches the file mapping behavior of the initial load.
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -206,6 +206,13 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
 
+  // Creates a new QNN context from a binary file, handling both file-mapped and direct-read
+  // paths.  Used by SSR recovery so it matches the file mapping behavior of the initial load.
```
huaychou:
```suggestion
  // paths. Used by SSR recovery so it matches the file mapping behavior of the initial load.
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1268,6 +1268,120 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::CreateContextFromFilePath(const std::string& context_bin_filepath,
+                                                         int64_t max_spill_fill_size,
+                                                         Qnn_ContextHandle_t& new_context) {
+  QnnContext_Config_t qnn_context_config = QNN_CONTEXT_CONFIG_INIT;
+  RETURN_IF_ERROR(SetQnnContextConfig(context_priority_, qnn_context_config));
+
+#if QNN_API_VERSION_MAJOR == 2 && (QNN_API_VERSION_MINOR >= 21)
```
huaychou: Just wondering, is this still required?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1268,6 +1268,120 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
...
+                                                device_handle_,
+                                                context_configs,
+                                                static_cast<void*>(buffer.data()),
+                                                static_cast<Qnn_ContextBinarySize_t>(buffer_length),
+                                                &new_context,
+                                                profile_backend_handle_);
+  }
+
+  RETURN_IF(QNN_SUCCESS != rt,
+            ("SSR recovery: contextCreateFromBinary failed. Error: " + QnnErrorHandleToString(rt)).c_str());
```
huaychou: I think the error message is not related to `CreateContextFromFilePath` failure.

onnxruntime/test/providers/qnn/ssr/qnn_mock_ssr.cc
```diff
@@ -36,6 +36,18 @@ extern "C" void SetMockSSRAlwaysCrash(bool always_crash) {
   g_always_crash_mode = always_crash;
 }
 
+// Tracks whether contextCreateFromBinaryWithCallback was invoked during SSR recovery
+// (i.e., after g_ssr_fired became true).  Reset alongside g_ssr_fired each session.
+// Only meaningful when QNN_FILE_MAPPED_WEIGHTS_AVAILABLE is defined (Windows ARM64,
+// QNN >= 2.32); always false on other platforms since the API doesn't exist there.
```
huaychou: Can be removed.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1268,6 +1268,120 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
...
+                                              static_cast<Qnn_ContextBinarySize_t>(buffer_length),
+                                              blob_version,
+                                              graph_count,
+                                              &graphs_info));
+    if (!MinVersionMet(blob_version, {3, 3, 3})) {
+      std::string version_str = std::to_string(blob_version.major) + "." +
+                                std::to_string(blob_version.minor) + "." +
+                                std::to_string(blob_version.patch);
+      ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_WARNING,
+                      ("SSR recovery: context binary version " + version_str +
```
huaychou, line 1411: Why do we have SSR recovery error message in `CreateContextFromFilePath`?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1268,6 +1268,120 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::CreateContextFromFilePath(const std::string& context_bin_filepath,
```
huaychou: Just wondering, what is the difference between `CreateContextFromFilePath` and `LoadCachedQnnContextFromBuffer`?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
qti-hungjuiw: Fixed. Thanks!

onnxruntime/test/providers/qnn/ssr/qnn_mock_ssr.cc
qti-hungjuiw: Done. Thanks!

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
qti-hungjuiw: Done. Thanks!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: Thanks for pointing this out. I rename the `CreateContextFromFilePath` to `ReloadContextForSSR` to ensure this function is SSR-specific.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: I renamed the `CreateContextFromFilePath` to `ReloadContextForSSR` to ensure this function is SSR-specific. This function re-acquires a fresh Context handle from context binary to help recover from SSR.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: Hi, I renamed the `CreateContextFromFilePath` to `ReloadContextForSSR`.
- The `ReloadContextForSSR` only focuses on instances required by SSR recovery (i.e. QnnContext).
- `LoadCachedQnnContextFromBuffer` also recreates the `QnnModel` and update `qnn_models` map, which is not required.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -208,7 +208,16 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   // callback is dispatched instead of reading from disk.
   Ort::Status ReadContextBinIfValid(const std::string& context_bin_filepath,
                                     std::vector<char>& buffer,
-                                    const qnn::EpContextIoDispatch& io_dispatch);
+                                    const qnn::EpContextIoDispatch& io_dispatch = qnn::EpContextIoDispatch(nullptr));
```
tirupath-qti: is this default selection related to SSR ?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
tirupath-qti: since ABI EP works for QAIRT releases after 2.42, do we still need this kind of check?

qti-hungjuiw: Removed. Thanks for pointing this out!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1281,6 +1281,117 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
...
+
+    void* bin_buffer = nullptr;
+    RETURN_IF_ERROR(file_mapper_->GetContextBinMappedMemoryPtr(context_bin_filepath, &bin_buffer));
+
+    // Cannot use contextCreateFromBinaryWithCallback() unless context bin version is >= 3.3.3
+    auto sys_ctx_handle = GetSystemContextHandle();
+    RETURN_IF(sys_ctx_handle == nullptr, "System context handle is null.");
+    Qnn_Version_t blob_version = {0, 0, 0};
+    uint32_t graph_count = 0;
+    QnnSystemContext_GraphInfo_t* graphs_info = nullptr;
```
huaychou: **[N-1] (Minor)** <code class="c">graphs_info</code> out-param declared but unused in <code class="c">ReloadContextForSSR</code>

```
// qnn_backend_manager.cc (ReloadContextForSSR new block)
uint32_t graph_count = 0;
QnnSystemContext_GraphInfo_t* graphs_info = nullptr;  // declared but never read

RETURN_IF_ERROR(GetGraphInfoAndBinVersion(sys_ctx_handle.get(),
                                          bin_buffer,
                                          static_cast<Qnn_ContextBinarySize_t>(buffer_length),
                                          blob_version,
                                          graph_count,
                                          &graphs_info));
```

`ReloadContextForSSR` only needs `blob_version` to check the 3.3.3 minimum; `graphs_info` and `graph_count` are declared but never read. A reader must verify this is not a latent missing-`DeserializeGraphInfoFromBinaryInfo` bug.

**Impact**: Minor readability issue only. No functional impact: `graphs_info` points into data owned by `sys_ctx_handle` and is released when that handle goes out of scope.

**Suggested fix:**

Consider adding a lightweight `GetBlobVersion()` helper returning only the version, or annotate the declaration: `/* not used — only blob_version is needed here */` to prevent future readers from suspecting a missing `DeserializeGraphInfoFromBinaryInfo` call.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
qti-hungjuiw: Thanks for pointing this out. Removed this since we now place the default `io_dispatch` argument in `ReloadContextForSSR`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1281,6 +1281,117 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
...
+  QnnContext_Config_t spill_fill_config = QNN_CONTEXT_CONFIG_INIT;
+  QnnHtpContext_CustomConfig_t custom_config;
+  custom_config.option = QNN_HTP_CONTEXT_CONFIG_OPTION_REGISTER_MULTI_CONTEXTS;
+  QnnHtpContext_GroupRegistration_t group_info;
+  group_info.firstGroupHandle = 0x0;  // New group after SSR — this is the only context
+  group_info.maxSpillFillBuffer = max_spill_fill_size;
+  custom_config.groupRegistration = group_info;
+  spill_fill_config.option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
+  spill_fill_config.customConfig = &custom_config;
+  QnnContext_Config_t* spill_fill_ptr = max_spill_fill_size > 0 ? &spill_fill_config : nullptr;
```
huaychou: nit: we can check max_spill_fill_size first to avoid redundant configs

huaychou (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
minfhong-qti: Probabaly `QnnConfigsBuilder`?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1281,6 +1281,117 @@ Ort::Status QnnBackendManager::ReadContextBinIfValid(const std::string& context_
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::ReloadContextForSSR(const std::string& context_bin_filepath,
```
minfhong-qti: I think this function is highly duplicate with `LoadCachedQnnContextFromBuffer`. Can we extract common parts as another function?

huaychou (approved):

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: Sure, we extracted the common part `BuildContextBinaryConfigs` and `CreateContextHandleFromBinary` from `LoadCachedQnnContextFromBuffer` and `ReloadContextForSSR`.

The `BuildContextBinaryConfigs` can handle new configs (i.e. `reused_io_limit_mb`) from the `QnnBackendManager` for `LoadCachedQnnContextFromBuffer` and `ReloadContextForSSR` in the future so that we do not need to have duplicate update.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-hungjuiw: Sure, we now use `QnnConfigsBuilder` for `ReloadContextForSSR` and `LoadCachedQnnContextFromBuffer`.

minfhong-qti (approved):

qti-chuteng (approved):

PR #678, opened 2026-07-29
[QNN EP] Release folded Q/DQ chain inputs to keep peak memory flat

qti-yuduo:
## Summary

Constant folding of standalone Q/DQ node units (#339) registers each hop's result as a STATIC tensor but never releases the hop's input, so every intermediate of an N-hop constant chain is retained until the whole graph is composed. Peak working set therefore grows linearly in chain length.

This releases a hop's input once the fold has provably made it unreachable: when the node unit is the input's sole consumer and the input is not a graph output. Any other reader still resolves the name through `model_tensors_map_`, whether it folds too or emits a real QNN op, so the sole-consumer test is what makes the release safe. A failed or unknown consumer count leaves the tensor in place, matching the behavior before the optimization existed.

## Test plan

Peak working set, 24-weight model with consumed fp32 volume held constant:

| chain hops | before | after |
|---|---|---|
| 1 | 668 MB | 638 MB |
| 3 | 908 MB | 639 MB |
| 5 | 1151 MB | 645 MB |
| 8 | 1512 MB | 647 MB |

- Output checksums bit-identical to the pre-fix fold at every chain length (memory-only change).
- #339's regression tests pass; mixed fan-out (two consumers, one folding) correctly retains the payload and matches CPU.
- Two `QnnUnit_ModelWrapperTest` cases added for `ReleaseTensorWrapper`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
@@ -159,6 +159,14 @@ bool QnnModelWrapper::AddParamWrapper(QnnParamWrapper&& param_wrapper) {
   return true;
 }
 
+void QnnModelWrapper::ReleaseTensorWrapper(const std::string& tensor_name) {
+  if (model_tensors_map_.erase(tensor_name) == 0) {
+    return;
+  }
+  ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE,
+              ("Released folded constant tensor: " + tensor_name).c_str());
+}
+
 const QnnTensorWrapper& QnnModelWrapper::GetQnnTensorWrapper(const std::string& tensor_name) {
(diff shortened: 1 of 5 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -159,6 +159,14 @@ bool QnnModelWrapper::AddParamWrapper(QnnParamWrapper&& param_wrapper) {
   return true;
 }
 
+void QnnModelWrapper::ReleaseTensorWrapper(const std::string& tensor_name) {
+  if (model_tensors_map_.erase(tensor_name) == 0) {
```
yath1, line 163: Nit: Better to error out if we cant find `tensor_name` in the map.

yath1 (commented):
Curious how the memory savings help with LLMs? Any data points on what was before and after the change?

PR #686, opened 2026-07-31
[QNN EP] Add HTP option fp16_clamp_overflow to avoid NaN/Inf on Glymur

huaychou:
### Description

- The option is intentionally undocumented (changes HTP numerical behavior) and gated on QNN API version >= 2.38 (QAIRT 2.49). On older SDKs the option is parsed but logs a warning and is ignored.

### Motivation and Context

- On Glymur, fp16 Conv operations on HTP can overflow fp16 range and produce NaN/Inf outputs. This adds a new HTP-backend-specific session config option `fp16_clamp_overflow` (default OFF) that sets the QNN_HTP_GRAPH_CONFIG_OPTION_FP16_CLAMP_OVERFLOW graph config to clamp such overflow.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_def.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_def.h
@@ -137,7 +137,7 @@ typedef struct HtpGraphConfigs {
   HtpGraphFinalizationOptimizationMode htp_graph_finalization_opt_mode = HtpGraphFinalizationOptimizationMode::kDefault;
   bool enable_htp_fp16_precision = false;
   bool enable_htp_monolithic_lstm = false;
-  bool fp16_clamp_overflow = false;
+  bool enable_htp_fp16_clamp_overflow = false;  // Intentionally undocumented; for internal/diagnostic use only.
 } HtpGraphConfigs_t;
 
 enum class QnnBackendType : uint8_t {
--- a/onnxruntime/core/providers/qnn/builder/qnn_def.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_def.h
@@ -30,6 +30,13 @@ namespace qnn {
 #define QNN_SYSTEM_DLC_API_ENABLED
 #endif  // QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 37
 
+// QNN_HTP_GRAPH_CONFIG_OPTION_FP16_CLAMP_OVERFLOW is available from QNN API 2.38
+// (QAIRT 2.49).
+#if QNN_API_VERSION_MAJOR > 2 || \
+    (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 38)
+#define QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
+#endif
+
 #if defined(_WIN32) && (defined(__aarch64__) || defined(_M_ARM64))
 #if QNN_API_VERSION_MAJOR > 2 || ((QNN_API_VERSION_MAJOR) == 2 && (QNN_API_VERSION_MINOR >= 32))
 #define QNN_FILE_MAPPED_WEIGHTS_AVAILABLE
@@ -123,6 +130,7 @@ typedef struct HtpGraphConfigs {
   HtpGraphFinalizationOptimizationMode htp_graph_finalization_opt_mode = HtpGraphFinalizationOptimizationMode::kDefault;
   bool enable_htp_fp16_precision = false;
   bool enable_htp_monolithic_lstm = false;
+  bool fp16_clamp_overflow = false;
 } HtpGraphConfigs_t;
 
 enum class QnnBackendType : uint8_t {
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -941,6 +941,14 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                                       FormatEPConfigKey(kEnableHtpFp16ClampOverflow),
                                                                       false,
                                                                       logger_);
+#ifndef QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
+  // Warn once at parse time (not per graph) if the option was requested but the SDK lacks support.
+  if (htp_graph_configs_.enable_htp_fp16_clamp_overflow) {
+    ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
+                "enable_htp_fp16_clamp_overflow was requested but the QNN HTP SDK in use does "
+                "not support it (requires QAIRT >= 2.49). Ignoring.");
+  }
+#endif
 
   // Try to parse multi-SoC HTP options first. If not multi-SoC htp_arch/soc_model is given, fallback to normal parsing.
   ParsePerSocHtpConfigs();
@@ -1760,10 +1768,6 @@ void QnnEp::InitQnnHtpGraphConfigs(
       gsl::not_null<QnnGraph_Config_t*> graph_config = configs_builder.PushConfig();
       graph_config->option = QNN_GRAPH_CONFIG_OPTION_CUSTOM;
       graph_config->customConfig = htp_fp16_clamp_config;
-#else
-      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
-                  "enable_htp_fp16_clamp_overflow was requested but the QNN HTP SDK in use does "
-                  "not support it (requires QAIRT >= 2.49). Ignoring.");
 #endif
     }
   }
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -442,7 +442,8 @@ void QnnEp::ParsePerSocHtpConfigs() {
     qnn::HtpGraphConfigs_t config{vtcm_size_in_mb_per_soc[idx],
                                   htp_graph_configs_.htp_graph_finalization_opt_mode,
                                   htp_graph_configs_.enable_htp_fp16_precision,
-                                  htp_graph_configs_.enable_htp_monolithic_lstm};
+                                  htp_graph_configs_.enable_htp_monolithic_lstm,
+                                  htp_graph_configs_.fp16_clamp_overflow};
     htp_graph_configs_per_soc_.push_back(std::move(config));
   }
 
@@ -905,6 +906,17 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                                   false,
                                                                   logger_);
   model_settings_.enable_htp_monolithic_lstm = htp_graph_configs_.enable_htp_monolithic_lstm;
+
+  // HTP fp16 clamp overflow. Default false. On HTP Arch v79+ (e.g. Glymur/v81),
+  // fp16 Conv accumulator overflow becomes NaN and propagates to the output
+  // (pre-v79 kept it finite). When true, saturate such overflow to the fp16 max
+  // value instead of NaN. Requires QAIRT SDK > 2.49.
+  static constexpr const char* kFp16ClampOverflow = "fp16_clamp_overflow";
+  htp_graph_configs_.fp16_clamp_overflow = ParseBoolOption(ort_api,
+                                                           session_options_,
+                                                           FormatEPConfigKey(kFp16ClampOverflow),
+                                                           false,
+                                                           logger_);
 
   // Try to parse multi-SoC HTP options first. If not multi-SoC htp_arch/soc_model is given, fallback to normal parsing.
   ParsePerSocHtpConfigs();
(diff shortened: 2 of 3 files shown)


onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -1192,6 +1192,36 @@ TEST_F(QnnHTPBackendTests, Convf32_PerChannelQDQChainConstWeight_NonIdentity_Reg
                   EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(1e-3f)});
 }
 
+// Smoke test for the enable_htp_fp16_clamp_overflow HTP option
+#ifdef QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
+TEST_F(QnnHTPBackendTests, Conv_Fp16ClampOverflow_Smoke) {
+#if defined(_WIN32)
+  SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V75);
```
kuanyul-qti, line 1200: 1. Should it be `QNN_HTP_DEVICE_ARCH_V79`?
2. Why does this line need to be wrapped with `_WIN32`?

huaychou: 1. It works on V79+ (inclusive), so I think skip less than or equal to V75 is correct.
2. Removed.

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -1192,6 +1192,34 @@ TEST_F(QnnHTPBackendTests, Convf32_PerChannelQDQChainConstWeight_NonIdentity_Reg
                   EPVerificationParams{ExpectedEPNodeAssignment::All, ElementwiseAbsoluteVerifier(1e-3f)});
 }
 
+// Smoke test for the enable_htp_fp16_clamp_overflow HTP option
+#ifdef QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
```
kuanyul-qti: **[C-1] (Critical)** <code class="c">QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE</code> is never defined in the <code class="c">conv_test.cc</code> TU — the new test is never compiled

Problem Code

```
// conv_test.cc — the file's entire include list is just these two
#include "test/providers/qnn/qnn_test_utils.h"
#include "gtest/gtest.h"

// …but the macro is defined on the core side, in qnn_def.h:
// qnn_def.h:42-45
#if QNN_API_VERSION_MAJOR > 2 || \
    (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 38)
#define QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
#endif

// conv_test.cc:1196 — this condition is always false
#ifdef QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
TEST_F(QnnHTPBackendTests, Conv_Fp16ClampOverflow_Smoke) { … }
#endif
```

Root Cause

The macro is `#define`d only inside `onnxruntime/core/providers/qnn/builder/qnn_def.h`, and it is **not** a CMake-injected compile definition (confirmed: the string appears nowhere under `cmake/`, `tools/`, or `qcom/`).

`conv_test.cc` includes only `qnn_test_utils.h` and `gtest.h`. `qnn_test_utils.h` pulls in `onnxruntime_cxx_api.h`, `HTP/QnnHtpDevice.h`, `QnnDevice.h`, `QnnInterface.h`, `QnnLog.h`, `QnnTypes.h` and friends — **no path reaches `qnn_def.h`**. A repo-wide scan shows only `core/providers/qnn/**` and `test/providers/qnn/unit/**` include `qnn_def.h`, and `conv_test.cc` lives in `test/providers/qnn/` (not `unit/`).

So there are exactly two ways this macro could have become visible in this TU, and neither holds: (1) the sole `#define`, in `qnn_def.h`, is not in this TU's include closure; (2) no CMake compile definition for it exists.

**Impact**: the `#ifdef` is permanently false, so `Conv_Fp16ClampOverflow_Smoke` **is never compiled into the test binary on any platform or SDK version**. Running `--gtest_filter=*Fp16ClampOverflow*` reports "0 tests ran", meaning this PR ships **zero test coverage** for the new feature. Worse, the diff *looks* tested, which will mislead later reviewers and CI into believing it was validated — exactly the [D02] failure mode: "0 tests registered = 0 validation, not a pass".

**Suggested fix:**

The test itself **does not reference any symbol that requires the newer SDK to compile** — it only sets `provider_options` strings. The cleanest fix is therefore to **delete the `#ifdef`/`#endif` entirely**: on older SDKs the EP takes the `#else` branch, logs a warning and ignores the option, the Conv still computes correctly, and the test still passes. This also satisfies [T17] ("prefer `SKIP_IF` / a runtime gate over a raw `#ifdef` so coverage stays visible").

If compile-time gating really is wanted, the macro must be made visible to this TU — e.g. add `#include "core/providers/qnn/builder/qnn_def.h"` to `conv_test.cc`, or promote the macro to a CMake compile definition (mirroring how `QNN_SDK_VERSION_MAJOR/MINOR` are injected). Either way, actually run `--gtest_filter=*Fp16ClampOverflow*`, confirm the test count is > 0, and paste the result into the PR description.

onnxruntime/test/providers/qnn/conv_test.cc
```diff
@@ -1192,6 +1192,34 @@ TEST_F(QnnHTPBackendTests, Convf32_PerChannelQDQChainConstWeight_NonIdentity_Reg
...
+  ProviderOptions provider_options;
+  provider_options["backend_type"] = "htp";
+  provider_options["offload_graph_io_quantization"] = "0";
+  provider_options["enable_htp_fp16_precision"] = "1";
+  provider_options["enable_htp_fp16_clamp_overflow"] = "1";
+#if defined(__linux__) && !defined(__aarch64__)
+  provider_options["soc_model"] = std::to_string(QNN_SOC_MODEL_SM8850);
+#endif
+
+  auto input_def = TestInputDef<float>({1, 2, 4, 4}, false, GetFloatDataInRange(-1.0f, 1.0f, 32));
```
kuanyul-qti, line 1210: **[M-1] (Major)** The smoke test's input range cannot overflow fp16 — even if compiled it does not validate the feature (no kill-test fence)

Problem Code

```
auto input_def   = TestInputDef<float>({1, 2, 4, 4}, false, GetFloatDataInRange(-1.0f, 1.0f, 32));
auto weights_def = TestInputDef<float>({2, 2, 2, 2}, true,  GetFloatDataInRange(-1.0f, 1.0f, 16));
auto bias_def    = TestInputDef<float>({2}, true, {1.0f, -1.0f});
```

Both input and weights are confined to `[-1, 1]` and the kernel is `2×2×2` (8 MACs), so the accumulator's absolute value is bounded by roughly `8 × 1 × 1 + 1 = 9`. The fp16 max is about `65504` — four orders of magnitude away. **This test numerically cannot trigger an fp16 accumulator overflow**, so it never reaches the clamp path this PR adds.

Put differently: flipping `provider_options["enable_htp_fp16_clamp_overflow"]` to `"0"`, or even ripping the whole feature out of the EP, would leave this test passing identically. It verifies "enabling the option does not break anything", not "the option fixes the NaN".

**Impact**: this violates [G02] ("a bug fix needs a reproducing test plus a kill-test fence — temporarily disable the fix and confirm the test fails"), and the PR description records no kill-test fence result. Once merged, neither a QAIRT-side behavioral regression nor an EP-side failure to actually deliver the config to the backend would be caught by CI.

**Suggested fix:**

Use magnitudes that actually overflow, and assert the output is finite rather than comparing against the CPU EP (CPU runs fp32 and will not overflow, so the two are not expected to match). For example scale the weights or input to the `~300` range so the accumulator exceeds `65504`, then verify the output contains **no NaN/Inf**. Also record the kill-test fence in the PR description: with `enable_htp_fp16_clamp_overflow` set to `"0"` the test should fail (NaN) on a v79+ device, and pass with `"1"`. If no Glymur device is currently available to complete this, say so explicitly in the PR description and attach the manual verification log.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1738,6 +1750,22 @@ void QnnEp::InitQnnHtpGraphConfigs(
...
+#ifdef QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
+      gsl::not_null<QnnHtpGraph_CustomConfig_t*> htp_fp16_clamp_config = configs_builder.PushCustomConfig();
+      htp_fp16_clamp_config->option = QNN_HTP_GRAPH_CONFIG_OPTION_FP16_CLAMP_OVERFLOW;
+      htp_fp16_clamp_config->fp16ClampOverflow = true;
+
+      gsl::not_null<QnnGraph_Config_t*> graph_config = configs_builder.PushConfig();
+      graph_config->option = QNN_GRAPH_CONFIG_OPTION_CUSTOM;
+      graph_config->customConfig = htp_fp16_clamp_config;
+#else
+      ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
```
kuanyul-qti: **[N-2] (Minor)** The old-SDK warning sits inside the per-graph loop — it repeats once per partition

`InitQnnHtpGraphConfigs()` is called inside the `for (graph_idx …)` loop in `CompileOnnxModel()` (`qnn_execution_provider.cc`), and on the multi-SoC path `CompileOnnxModel` is itself invoked once per SoC. So on an older SDK, a model split into N partitions emits the identical warning N times (N×SoC in the multi-SoC case).

**Impact**: pure log noise — functional behaviour is unaffected. But for large models (partition counts can reach the dozens) it drowns out other useful warnings and hurts log readability.

**Suggested fix:**

Move this "requested but unsupported by the SDK" warning to the **option parsing site** (near `qnn_execution_provider.cc:939`, which runs once per session and is already within reach of the same `#if` condition); or keep it in place but emit it only once via `std::call_once` or a `mutable bool` latch. The former is cleaner and also tells the user at session-creation time that the option is being ignored.

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -37,6 +37,13 @@ namespace qnn {
 #define QNN_HTP_GRAPH_SPLITTING_AVAILABLE
 #endif
 
+// QNN_HTP_GRAPH_CONFIG_OPTION_FP16_CLAMP_OVERFLOW is available from QNN API 2.38
+// (QAIRT 2.49).
+#if QNN_API_VERSION_MAJOR > 2 || \
+    (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 38)
+#define QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
```
kuanyul-qti, line 44: Would it be possible to move this line to L38? That location is also wrapped by QAIRT SDK >= 2.49.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -930,6 +931,17 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                                   logger_);
   model_settings_.enable_htp_monolithic_lstm = htp_graph_configs_.enable_htp_monolithic_lstm;
 
+  // HTP fp16 clamp overflow. Default false. On HTP Arch v79+ (e.g. Glymur/v81),
+  // fp16 Conv accumulator overflow becomes NaN and propagates to the output
+  // (pre-v79 kept it finite). When true, saturate such overflow to the fp16 max
+  // value instead of NaN. Requires QAIRT SDK >= 2.49.
+  static constexpr const char* kEnableHtpFp16ClampOverflow = "enable_htp_fp16_clamp_overflow";
+  htp_graph_configs_.enable_htp_fp16_clamp_overflow = ParseBoolOption(ort_api,
```
kuanyul-qti, line 939: Just wondering whether this should be an option, or whether it can be enabled by default when defining `QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE`.

onnxruntime/core/providers/qnn/builder/qnn_def.h
huaychou: We’d like to keep it separate so we avoid conflicts and maintain a clean git history.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou: No, we have discussed with HTP team, and it cannot be enabled by default. The reason is that we need to follow IEEE compliance by default.

onnxruntime/test/providers/qnn/conv_test.cc
huaychou: It is hard to overflow fp16 in real cases since it is not overflow in the output but overflow during HTP accumulator. In addition, there is no real Glymur in CI machines. Therefore, I think smoke test is enough.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou: Fixed

onnxruntime/test/providers/qnn/conv_test.cc
huaychou: Fixed

kuanyul-qti (approved):

qti-chuteng (approved):

PR #688, opened 2026-07-31
[QNN EP] Revert standalone constant Q/DQ folding (#339)

qti-yuduo:
## Summary

- Behaviorally revert the generic standalone constant `QuantizeLinear`/`DequantizeLinear` folding introduced by #339.
- Keep representable per-tensor Q/DQ tensors compact and emit them as QNN operations instead of retaining materialized STATIC copies.
- Retain one narrow compatibility exception: an initializer-backed, single-hop per-channel DQ may fold because QNN cannot represent it standalone and existing Conv paths require it.
- Keep DQ-wrapped Clip bounds and mixed-quant-type BatchNormalization working by grouping and consuming their compact parameters at the owning operator.
- Remove the chained per-channel Conv tests that asserted #339's behavior.

Rebased onto `a0620c84c3` and squashed to a single commit. Net **+123 / -290** across 9 files.

## Why revert

The reported model has 255 direct initializer-to-DQ weights, **all per-tensor**. Their compact storage is 1.046 GiB, while folding them to FP32 materializes 4.184 GiB. The expanded buffers remain live during graph construction and raise peak working set substantially.

Note on scoping: only #339's `ExplicitOpCheck` relaxation was per-channel-gated. The fold itself (`CanFoldConstantQdq`) accepted any standalone constant Q/DQ, so per-tensor weights folded too — which is why this model, containing zero per-channel constant DQ, regressed as hard as it did.

This version restores the pre-#339 path for every representable standalone Q/DQ node and removes constant `QuantizeLinear` folding as well.

## Memory A/B

Windows ARM64 / HTP V73 (Snapdragon X Elite), `onnxruntime_perf_test` on the reported 1.1 GB model, two runs per arm (spread < 30 MB):

| Build | Peak working set | Session creation |
|---|---:|---:|
| `main` @ `a0620c84c3` (#339 folding) | 7,712,866,304 B / 7.183 GiB | 156.3 s |
| This revert @ `482fbcb016` | 6,068,539,392 B / 5.652 GiB | 165.4 s |

**Recovers 1.53 GiB (-20.9%).** For reference, the pre-#339 control measured earlier on base `b5826e06` was 5.872 GiB, so this is at or below the pre-regression level.

Session creation grows ~6% because the 255 DQ ops are now compiled by the backend instead of folded away. That is the pre-#339 cost profile being restored.

## Why this is not a literal `git revert`

The fold call site in `simple_op_builder.cc` was the **only** producer of folded constants, so a bare revert makes `IsFoldedConstant()` permanently false and silently regresses two downstream consumers. Both now obtain their compact parameters via QDQ grouping instead:

- **Clip with DQ-wrapped constant min/max** (`bf5ef01036`): the Clip selector accepts an empty output Q so the group forms, and the builder reads the bounds directly.
- **BatchNormalization with mixed input/output quant types** (`513f6c127a`, tetracode #20348): the selector no longer rejects the type mismatch, and the builder runs those in float internally (`Dequantize -> BN in F32 -> Quantize`).

Gating the retained per-channel fold on a *real initializer* rather than an effectively-constant input is what stops it from chaining hop by hop.

## Intentional tradeoff

This PR does not preserve #339's original Qwen input-signature fix. The original `initializer -> DQ -> Q -> DQ -> Conv` topology assigns only 4 of 6 nodes on both CPU and HTP, and the final constant-derived weight can become a QNN `APP_WRITE` input. That is the expected old behavior for this revert.

A future fix should collapse that chain while retaining compact quantized storage; it should not restore hop-by-hop FP32 materialization.

## Validation

- Windows ARM64 QNN provider test build clean (`-WX`).
- Affected CPU/HTP matrix: **154 passed, 1 architecture-gated skip** (`Conv_Fp16ClampOverflow_Smoke`), including all four regression guards the revert has to keep alive (BatchNorm mixed-dtype x2, Clip QDQ const min/max x2) and the new `ConstantPerTensorDequantizeStaysOp`.
- 8 `ConvLPBQ_U16Int4_*` HTP failures are **pre-existing**: they reproduce identically on untouched `main` with the same `CreateQnnInputOutputTensors` / `weight_quant error code: 1000`.
- `unit/qnn_ep_utils_test.cc` changes are behind `QNN_EP_INTERNAL_SYMBOL_ACCESS` and do not compile locally — CI-only, unverified on the dev box.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/batchnormalization_op_builder.cc
@@ -447,14 +447,7 @@ class BatchNormalizationOpBuilder : public BaseOpBuilder {
 
 namespace {
 
-// Helper to check if a BatchNorm param is constant - either direct initializer or through a DQ node.
-//
-// node_unit.GetDQNodes() is only populated for a QDQGroup-type NodeUnit. When the surrounding QDQ
-// selector rejects the group (e.g. BatchNormalizationNodeGroupSelector requires the quantized input
-// and output element types to match; mixed-bitwidth BN like u8-in/u16-out does not), BatchNorm becomes
-// a SingleNode-type NodeUnit with an empty GetDQNodes(). In that case the param's standalone DQ node
-// is visited (and constant-folded via TryFoldConstantQDQ) as its own NodeUnit before BN, in topological
-// order, so IsEffectivelyConstantInput (which also checks IsFoldedConstant) still recognizes it.
+// Helper to check if a BatchNorm param is constant, either directly or through a DQ in its group.
 bool IsParamConstant(const QnnModelWrapper& qnn_model_wrapper,
                      const OrtNodeUnit& node_unit,
                      const std::string& name) {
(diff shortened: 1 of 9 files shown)


qti-yuduo:
will address differently.

PR #690, opened 2026-08-03
[QNN_EP] Support for Castlike op.

pratha-qti:
Description

  Adds QNN execution provider support for the ONNX CastLike operator by reusing the
  existing CastOpBuilder.

  - onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc: relaxed
  ProcessInputs's input-count check to accept 1 input (Cast) or 2 inputs (CastLike). Only
  inputs[0] is ever processed — CastLike's second input (the type-reference tensor) is
  intentionally never added to the QNN graph, since its value is unused; ONNX type
  inference has already propagated its dtype onto the node's output type by the time the
  builder runs.
  - onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.h: added a
  {"CastLike", QNN_OP_CAST} entry to the GetQnnOpType map so CastLike resolves to the same
  native QNN op as Cast.
  - onnxruntime/core/providers/qnn/builder/op_builder_factory.cc: registered
  CreateCastOpBuilder("CastLike", *this) alongside the existing Cast registration.
  - onnxruntime/test/providers/qnn/cast_test.cc: added
  BuildCastLikeTestCase/RunCastLikeOpTest helpers and CPU/HTP test coverage (int32↔float,
  int64→int32, and float→bool to exercise the existing FP-to-Bool NotEqual fallback path
  with the extra CastLike input present).

  No changes were needed to CastOpBuilder's FP→Bool NotEqual workaround, INT64/FP64
  output-narrowing logic, or the CastLoneQFusion node-group fusion — all already operate
  only on inputs[0]/outputs[0] and are unaffected by the extra input.

  Motivation and Context

  QNN EP had no op-builder registration for CastLike, so every CastLike node in a graph was
  rejected in GetCapabilityImpl and fell back to CPU EP — regardless of how trivial the
  cast. This showed up heavily in diffusion-style models (e.g. Stable Diffusion UNet),
  where ONNX export commonly inserts CastLike around GroupNormalization/MultiHeadAttention
  scale and bias tensors to keep them in sync with the surrounding activation dtype. In one
  repro model this produced 100+ unsupported CastLike nodes, fragmenting the graph into
  many small CPU/NPU partitions and hurting both accuracy-parity confidence and
  performance.

  Functionally, CastLike only differs from Cast in how the target dtype is expressed (a
  second input's dtype vs. a to attribute) — by builder time, ORT's type inference has
  already resolved both to a concrete Outputs()[0].type. So no new QNN op or casting
  behavior was needed; this just teaches QNN EP's op-builder factory to route CastLike to
  the same handler already used and validated for Cast.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
+++ b/onnxruntime/core/providers/qnn/builder/op_builder_factory.cc
@@ -15,6 +15,7 @@ OpBuilderRegistrations::OpBuilderRegistrations() {
   CreateArgMaxMinOpBuilder("ArgMin", *this);
   CreateBatchNormalizationOpBuilder("BatchNormalization", *this);
   CreateCastOpBuilder("Cast", *this);
+  CreateCastOpBuilder("CastLike", *this);
   CreateClipOpBuilder("Clip", *this);
   CreateConcatOpBuilder("Concat", *this);
   CreateConvOpBuilder("Conv", *this);
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.h
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.h
@@ -140,6 +140,7 @@ class BaseOpBuilder : public IOpBuilder {
         {"AveragePool", QNN_OP_POOL_AVG_2D},
         {"BatchNormalization", QNN_OP_BATCHNORM},
         {"Cast", QNN_OP_CAST},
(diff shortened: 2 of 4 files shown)


qti-chuteng (dismissed):

qti-chuteng (approved):

qti-mbadnara:
Hi @pratha-qti Can you please update the docs for the supported ops?

pratha-qti:
sure

PR #692, opened 2026-08-03
Ort-D3d interop import memory feature.

johnpaultaken:
### Description
Import memory from D3d for onnxruntime inferencing input or output.

### Motivation and Context
Implementing FR https://github.com/microsoft/onnxruntime/issues/26821


onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -501,6 +511,41 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetHardwareDeviceIncompatibilityDetailsImp
...
+  // Create the external resource importer
+  auto* importer = new (std::nothrow) QnnExternalResourceImporterImpl(device_id, factory->ort_api);
+  if (importer == nullptr) {
+    return factory->ort_api.CreateStatus(ORT_FAIL, "Failed to create external resource importer");
+  }
+
+  *out_importer = importer;
+  return nullptr;
+#else
+  return factory->ort_api.CreateStatus(
```
qti-mattsinc, line 547: From the ORT docs for `CreateExternalResourceImporterForDevice`:

> This is an optional EP capability. If the EP does not support external resource
> import, out_importer is set to nullptr and the function returns success (nullptr
> status). This allows callers to use the simple "if (status != nullptr) handle_error()"
> pattern and check out_importer separately for capability detection.

In the `#else` branch, we are already returning `out_import = nullptr` correctly, but we should return success instead of an error to signal that it's unsupported but nothing went wrong.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -86,6 +90,12 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
   ValidateCompiledModelCompatibilityInfo = ValidateCompiledModelCompatibilityInfoImpl;
   GetHardwareDeviceIncompatibilityDetails = GetHardwareDeviceIncompatibilityDetailsImpl;
 
+#ifdef _WIN32
+  CreateExternalResourceImporterForDevice = CreateExternalResourceImporterForDeviceImpl;
+#else
+  CreateExternalResourceImporterForDevice = nullptr;
+#endif
+
```
qti-mattsinc, line 99: nit. We already have `#ifdef _WIN32` in `CreateExternalResourceImporterForDeviceImpl` and we can just do:

```suggestion
  CreateExternalResourceImporterForDevice = CreateExternalResourceImporterForDeviceImpl
```

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -501,6 +511,41 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetHardwareDeviceIncompatibilityDetailsImp
...
+#ifdef _WIN32
+  // Since CreateExternalResourceImporterForDeviceImpl doesn't take memory_device
+  // as a parameter (unlike CreateSyncStreamForDeviceImpl), and we don't have direct API
+  // to extract device ID from OrtEpDevice, we currently use 0.
+  // In the future, we could extract device ID from ep_device using OrtApi::EpDevice_MemoryInfo
+  // and then query the resulting OrtMemoryInfo for device ID.
+  int device_id = 0;
+
+  // Create the external resource importer
+  auto* importer = new (std::nothrow) QnnExternalResourceImporterImpl(device_id, factory->ort_api);
```
qti-mattsinc: nit. For consistency with the rest of the code, we should use `make_unique` inside a try / catch followed by `release` instead of `new (std::nothrow)`

onnxruntime/core/providers/qnn/qnn_external_resource_importer.h
```diff
@@ -0,0 +1,160 @@
...
+
+  static void* ORT_API_CALL GetHandleImpl(
+      _In_ OrtSyncStreamImpl* this_ptr) noexcept;
+
+  static void ORT_API_CALL ReleaseImpl(
+      _In_ OrtSyncStreamImpl* this_ptr) noexcept;
+
+  int device_id_;
+  const OrtApi& ort_api_;
+};
```
qti-mattsinc, line 156: AFAICT, `QnnSyncStreamImpl` is not being used anywhere. Does it need to be included in this PR?

johnpaultaken:
https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/30842614354

onnxruntime/core/providers/qnn/qnn_external_resource_importer.h
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * Derives from OrtExternalSemaphoreHandle (base struct) and adds QNN-specific fields.
+ */
+struct QnnExternalSemaphoreHandle : OrtExternalSemaphoreHandle {
+  ComPtr<ID3D12Fence> d3d12_fence_;
+
+  QnnExternalSemaphoreHandle();
+
+  static void ORT_API_CALL ReleaseCallback(_In_ OrtExternalSemaphoreHandle* handle) noexcept;
+};
```
qti-mattsinc, line 48: AFAICT, `QnnExternalSemaphoreHandle` is not being used anywhere. Does it need to be included in this PR?

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+
+bool ORT_API_CALL QnnExternalResourceImporterImpl::CanImportSemaphoreImpl(
+    _In_ const OrtExternalResourceImporterImpl* /*this_ptr*/,
+    _In_ OrtExternalSemaphoreType /*type*/) noexcept {
+  // Currently not implemented.
+  return false;
+  /*
+  // Supports D3D12 fence only.
+  return type == ORT_EXTERNAL_SEMAPHORE_D3D12_FENCE;
+  */
```
qti-mattsinc, line 259: remove commented code

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+    default:
+      return impl.ort_api_.CreateStatus(ORT_INVALID_ARGUMENT,
+                                       "Invalid external semaphore type");
+  }
+  handle->ep_device = nullptr;
+  handle->descriptor = *desc;
+
+  *out_handle = handle.release();
+  return nullptr;
+  */
```
qti-mattsinc, line 311: remove commented code

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+
+  if (handle == nullptr || sync_stream == nullptr) {
+    return impl.ort_api_.CreateStatus(ORT_INVALID_ARGUMENT,
+                                     "handle and sync_stream cannot be nullptr");
+  }
+
+  auto* sem_handle = static_cast<QnnExternalSemaphoreHandle*>(handle);
+
+  return nullptr;
+  */
```
qti-mattsinc, line 343: remove commented code

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
```
qti-mattsinc, line 366: remove commented code

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+  *out_handle = nullptr;
+
+  if (desc->offset_bytes != 0) {
+    return impl.ort_api_.CreateStatus(ORT_INVALID_ARGUMENT,
+                                      "Not supported importing resource with non-zero offset");
+  }
+
+  // Create and return the derived handle
+  std::unique_ptr<QnnExternalMemoryHandle> handle{
+      new (std::nothrow) QnnExternalMemoryHandle()};
```
qti-mattsinc, line 132: nit. For consistency with the rest of the code, we should use `make_unique` inside a try / catch followed by `release` instead of `new (std::nothrow)`

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+          IID_PPV_ARGS(&(handle->d3d12_resource_)));
+
+      if (!SUCCEEDED(hr)) {
+        return impl.ort_api_.CreateStatus(ORT_INVALID_ARGUMENT,
+                                          "Invalid external memory handle");
+      }
+
+      break;
+    }
+    case ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_HEAP:
```
qti-mattsinc: There are also Vulkan memory types in the latest ORT header:

https://github.com/microsoft/onnxruntime/blob/6ac1a853925fae3924d98691683fa33883eb2f5e/include/onnxruntime/core/session/onnxruntime_c_api.h#L1061-L1066

So, I think it would make more sense to remove the explicit `case` for the unsupported type w/ fallthrough and instead just let it hit `default` directly. Otherwise it would look strange to someone reading the code why it enumerates D3D12_HEAP but not the VK types (or any future additions).

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+    default:
+      return impl.ort_api_.CreateStatus(ORT_INVALID_ARGUMENT,
+                                        "Invalid external memory handle type");
```
qti-mattsinc: nit.

```suggestion
                                        "Unsupported external memory handle type");
```

Invalid makes it sound like an invalid instantiation of `OrtExternalMemoryHandleType`

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+  }
+
+  // Calculate the data pointer with offset
+  void* data_ptr = static_cast<char*>((void*)qnn_handle) + tensor_desc->offset_bytes;
+
+  // Create memory info for the GPU device. Using the existing name DML works for us.
+  OrtMemoryInfo* memory_info = nullptr;
+  OrtStatus* status = impl.ort_api_.CreateMemoryInfo("DML", OrtDeviceAllocator,
+                                                     impl.device_id_, OrtMemTypeDefault,
+                                                     &memory_info);
```
qti-mattsinc: I don't see why we would want to use the DML memory info. Why not add a memory info of type `OrtDeviceMemoryType_DEFAULT` to the GPU device in during EP creation (`QnnEpFactory::CreateEpImpl`) using `EpDevice_AddAllocatorInfo` and later use that memory info here?

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+  const auto* qnn_handle = static_cast<const QnnExternalMemoryHandle*>(mem_handle);
+
+  // Calculate tensor size
+  size_t element_count = 1;
+  for (size_t i = 0; i < tensor_desc->rank; ++i) {
+    element_count *= static_cast<size_t>(tensor_desc->shape[i]);
+  }
+
+  size_t element_size = 0;
+  switch (tensor_desc->element_type) {
```
qti-mattsinc: Data types that have multiple elements packed into one byte are not being handled here. So, trying to create a tensor with elem type `ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT4` on imported memory would fail.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2931,10 +2932,10 @@ Ort::Status QnnBackendManager::GetOrRegisterContextMemHandle(Qnn_ContextHandle_t
   if (did_register) {
     // The cleanup lambda is the same for both HTP and DX12: unregister the QNN mem handle when the allocation is freed.
     auto unregister_mem_handle =
-        [shared_memory_address,
+        [memory_address,
          weak_backend_manager = weak_from_this(),
          weak_context_handle_record = std::weak_ptr{context_handle_record}](
-            void* /* allocation_base_address */) {
+            void* /* memory_address */) {
```
qti-mattsinc: Why change the name of `allocation_base_address`? I know it's commented out currently, but it would otherwise be shadowing the `memory_address` in the lambda capture list.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -243,7 +245,7 @@ TEST(QnnEP, ParseOpPackages_AbsolutePath) {
   ASSERT_EQ(op_packages.size(), 1u);
   EXPECT_EQ(op_packages[0].op_type, "MyOp");
   EXPECT_EQ(op_packages[0].path, tmp_dll.string());
-  EXPECT_EQ(op_packages[0].interface, "MyAddOpPackageInterfaceProvider");
+  // EXPECT_EQ(op_packages[0].interface, "MyAddOpPackageInterfaceProvider");
```
qti-mattsinc: Why is this commented out?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -290,7 +292,7 @@ TEST(QnnEP, ParseOpPackages_AbsolutePath_NotYetOnDisk) {
   ASSERT_EQ(op_packages.size(), 1u);
   EXPECT_EQ(op_packages[0].op_type, "MyOp");
   EXPECT_EQ(op_packages[0].path, non_existent_path);
-  EXPECT_EQ(op_packages[0].interface, "MyAddOpPackageInterfaceProvider");
+  // EXPECT_EQ(op_packages[0].interface, "MyAddOpPackageInterfaceProvider");
```
qti-mattsinc: Why is this commented out?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -323,7 +325,7 @@ TEST(QnnEP, ParseOpPackages_RelativePath) {
 #else
   EXPECT_EQ(op_packages[0].path, "foo.so");
 #endif
-  EXPECT_EQ(op_packages[0].interface, "MyAddOpPackageInterfaceProvider");
+  // EXPECT_EQ(op_packages[0].interface, "MyAddOpPackageInterfaceProvider");
```
qti-mattsinc: Why is this commented out?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -3236,6 +3238,345 @@ TEST_F(QnnGPUBackendTests, DISABLED_io_binding_qnn_gpu_shared_offset) {
...
+/*
+ * Import external memory.
+ */
+TEST_F(QnnGPUBackendTests, import_memory_basic) {
+  ComPtr<ID3D12Device> d3d12device;
+  HRESULT hr = D3D12CreateDevice(
+      nullptr,
+      D3D_FEATURE_LEVEL_12_0,
+      IID_PPV_ARGS(&d3d12device));
+  ASSERT_FALSE(FAILED(hr) || d3d12device == nullptr);
```
qti-mattsinc, line 3302: nit. It would be better to `GTEST_SKIP` here if DX12 device creation fails.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -3236,6 +3238,345 @@ TEST_F(QnnGPUBackendTests, DISABLED_io_binding_qnn_gpu_shared_offset) {
...
+  // input
+  const std::array<int64_t, 2> x_shape = {3, 2};
+  std::array<float, 3 * 2> x_values = {1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f};
+
+  ComPtr<ID3D12Device> d3d12device;
+  HRESULT hr = D3D12CreateDevice(
+      nullptr,
+      D3D_FEATURE_LEVEL_12_0,
+      IID_PPV_ARGS(&d3d12device));
+  ASSERT_FALSE(FAILED(hr) || d3d12device == nullptr);
```
qti-mattsinc, line 3370: nit. It would be better to `GTEST_SKIP` here if DX12 device creation fails.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -3236,6 +3238,345 @@ TEST_F(QnnGPUBackendTests, DISABLED_io_binding_qnn_gpu_shared_offset) {
...
+  input_tensor_desc.shape = x_shape.data();
+  input_tensor_desc.rank = x_shape.size();
+  input_tensor_desc.offset_bytes = 0;
+  OrtValue* x_tensor = nullptr;
+  Ort::GetInteropApi().CreateTensorFromMemory(importer, input_mem_handle, &input_tensor_desc, &x_tensor);
+  ASSERT_FALSE(x_tensor == nullptr);
+
+  // Real use case imported memory would already contain input data from the gpu.
+  // Since ours doesn't we need to copy input data to it from the cpu.
+  // However it seems D3D12_HEAP_FLAG_SHARED & D3D12_HEAP_TYPE_UPLOAD does not seem to work.
```
qti-mattsinc, line 3430: Why use `D3D12_HEAP_FLAG_SHARED` here if it complicates the buffer initialization? Why not just use `D3D12_HEAP_FLAG_NONE`?

onnxruntime/core/providers/qnn/qnn_external_resource_importer.h
johnpaultaken: The full framework for the resource importer is included in this change. The semaphore framework is populated where it is uniform with the memory framework. The semaphore body will be populated once QNN-GPU has the changes required.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
johnpaultaken: This was causing compile failures when the new headers were introduced. Can you check that on your system as well ?

onnxruntime/test/providers/qnn/qnn_basic_test.cc
johnpaultaken: D3D12_HEAP_FLAG_SHARED  is required for CreateSharedHandle

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: This semaphore code is provided so that it is uniform with the memory code. It will be enabled with the semaphore change.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+  // Create and return the derived handle
+  std::unique_ptr<QnnExternalMemoryHandle> handle{
+      new (std::nothrow) QnnExternalMemoryHandle()};
+  if (!handle) {
+    return impl.ort_api_.CreateStatus(ORT_FAIL, "Failed to allocate external memory handle");
+  }
+
+  switch (desc->handle_type) {
+    case ORT_EXTERNAL_MEMORY_HANDLE_TYPE_D3D12_RESOURCE: {
+      HRESULT hr = impl.d3d12_device_->OpenSharedHandle(
```
vjatoth-qti, line 139: No validation that the imported D3D12 resource's actual size (ID3D12Resource::GetDesc().Width) matches/exceeds the caller-supplied desc->size_bytes, and no check that the resource dimension is D3D12_RESOURCE_DIMENSION_BUFFER. CreateTensorFromMemoryImpl later trusts this unchecked size_bytes for its bounds check, so a caller under-reporting size would let a tensor be created larger than the real buffer — an OOB GPU read/write. Worth calling GetDesc() after OpenSharedHandle and validating against the descriptor.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -0,0 +1,455 @@
...
+}
+
+void* ORT_API_CALL QnnSyncStreamImpl::GetHandleImpl(
+    _In_ OrtSyncStreamImpl* this_ptr) noexcept {
+  auto& impl = *static_cast<QnnSyncStreamImpl*>(this_ptr);
+
+  // Currently not implemented.
+  return impl.ort_api_.CreateStatus(ORT_NOT_IMPLEMENTED, "Not implemented");
+
+  return static_cast<void*>(nullptr);
```
vjatoth-qti, line 399: Unreachable return statement.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.h
qti-mattsinc: The EP needs to be made stream-aware (implement `OrtEpFactory::IsStreamAware` and `OrtEp::CreateSyncStreamForDevice`) for the semaphore importing part of the framework to work. I do not see that in this PR. It would make more sense to put phase 2 (semaphore importing) in a separate PR, because as it stands only half the necessary phase 2 changes are being scaffolded out here, and it's just introducing dead code.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
qti-mattsinc: It's probably because of `interface` being defined as a macro when you include Windows headers. You can try `#undef interface` after the new Windows includes in qnn_test_utils.h

onnxruntime/test/providers/qnn/qnn_basic_test.cc
qti-mattsinc: Got it. But can we create an upload buffer, map+write to it, and copy it to the shared buffer that we import? So that we can properly initialize it with data for the test.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
qti-mattsinc: To elaborate on this, I am suggesting that we do this to get the memory info:

```suggestion
  const OrtMemoryInfo* memory_info = impl.ort_api_.EpDevice_MemoryInfo(impl.ep_device_, OrtDeviceMemoryType_DEFAULT);
```
(Where we add `const OrtEpDevice* ep_device_` as a member of `QnnExternalResourceImporterImpl` instead of `device_id_`.) For this to work, we need to make sure we add the DEFAULT memory info to the GPU device in `QnnEpFactory::CreateEpImpl`, using `EpDevice_AddAllocatorInfo`.

We probably also need to implement `OrtDataTransferImpl::CopyTensors` to conform to the ORT API expectations.

See https://github.com/microsoft/onnxruntime/issues/29602#issuecomment-5124881468 and the surrounding discussion:

> I believe that's correct. If the EP can import the memory it's saying it can use it, which requires the EP to provide the CopyTensors implementation for (2). that implementation is also required for ORT to use the EP in general so we can move data to/from its default device.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
johnpaultaken: That would require creation of gpu command queue etc which unnecessarily complicates the test. The required functionality of veryfying the tensor contents is already tested in the next test without introducing the GPU command queue complexity. This test is only a progressive step towards the third test.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
johnpaultaken: It seems to me that both are not the same thing.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
johnpaultaken: Both are not the same thing. If new fails due to allocation failure we are ok with a nullptr. But if it fails due to QnnExternalResourceImporterImpl constructor throwing then we want to know about it. This is because the expectation is that it doesn't throw as implied by the noexcept . I think we need to change the rest of the code to match this because that is a bad practise.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
johnpaultaken: Thanks for carefully reading the FR and bringing it to my attention. However the same guy who wrote the FR does the opposite in his sample implementation. FYI there are many things that changed from the  FR by the time the implementation was done in ORT and EPs. The FR was not updated to reflect the changes. 
So I want to know if there is any existing client code with the mentioned use case. Please let me know if you find any.
That aside, if the call was named GetCapability then the above assertion makes sense. However since the call is called CreateExternalResourceImporterForDevice it makes sense to indicate failure in the status code. We can discuss this further if needed.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
johnpaultaken: It is a good suggestion and we can discuss this. I need some more information. 1) Do we expect tests to be run on DX11 driver machines?  2) If so, what happens to the allocator tests when it is run on such a machine?

onnxruntime/core/providers/qnn/qnn_external_resource_importer.h
johnpaultaken: The semaphore importing part (body) can be added only once QNN-GPU has the changes required. As stated above this change includes all the required framework for the FR.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: The same comment is addressed at a different location.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: I think I can remove this case here because user is supposed to call CanImport before.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: The memory is external , it is not memory from this EpDevice. So the memory info would need to be independent of the EpDevice. Something like "Imported".  If CreateMemoryInfo_V2 allows this I can try that.

Thanks for providing the discussion link, I'll take a look. But here are my thoughts on the subject.
This whole FR is about having gpu buffers sharable for use by D3D and Ort without having to involve cpu for copy or synchronization. The tensors created by such imported gpu buffer are not required to be io tensors which might require a CopyTensor to the cpu address space. Preferably I do not want to provide a cpu mapping to make sure there is the expected zero copy hardware efficiency and avoid other potential pitfalls that includes cpu cache invalidation etc. If there is a need for imported memory to be an io buffer later we can add CopyTensor at that time. The current usage scenarios outlined by the FR does not require it.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: Thanks for pointing this out. It is a valid point.
If the desc can be obtained from the OpenSharedHandle then the function shouldn't be providing it as an input. This might be an oversight at the api design stage which we cannot change now.
The design goal of C++ gives more weightage to efficiency than safety. This means doing all possible safety checks are discouraged and the onus is on the user to use the api as defined, and if not, the behavior is left undefined.  It is likely that we are going to remove the size checks in CeateTensor to account for future data types anyway. (This seems to be what the cuda ep is doing as well.) Please let me know if you disagree.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-mattsinc: They express the same thing. From the docs:

```
   * \note Implementation of this function is optional.
   *       An EP factory should only implement this if it supports external resource import.
   *       If not implemented or not supported, return ORT_NOT_IMPLEMENTED or set out_importer to nullptr.
```

See also the ORT core code that calls this function at [onnxruntime/core/session/interop_api.cc:44](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/interop_api.cc#L44).

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-mattsinc: >  But if it fails due to QnnExternalResourceImporterImpl constructor throwing then we want to know about it.

Using `new (std::nothrow)` like this would cause the EP to call `std::terminate` if the `QnnExternalResourceImporterImpl` constructor throws, since this function is noexcept.

It's better to follow the existing pattern so that the EP returns an error gracefully instead of possibly calling `std::terminate`. If you want to handle an allocation failure separately, than you can add a catch branch for `std::bad_alloc`.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
qti-mattsinc: > 1) Do we expect tests to be run on DX11 driver machines?

Normally a QC WoS device has both a UMD for DX11 and DX12. So we don't expect these tests (which are already Windows only) to run on a device without DX12 support.

>  2) If so, what happens to the allocator tests when it is run on such a machine?

They use GTEST_SKIP.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-mattsinc: FYI the quote in my initial comment is from the docs in the ORT api header, not from the FR description.

> So I want to know if there is any existing client code with the mentioned use case. Please let me know if you find any.

The only client code I am aware of is in WebNN: https://chromium-review.googlesource.com/c/chromium/src/+/7833769. But they do not use the aforementioned pattern.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
qti-mattsinc: We can skip the implementation of CopyTensors for now. It seems to me that there is actually quite a lot that can be improved in the EP with regard to allocators and memory info (esp. now that ORT has https://github.com/microsoft/onnxruntime/pull/28037 in 1.28), so I may have a go at some of that in a follow-up.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: Update. The current thinking is to keep the size checks in Tensor creation and also add the desc cross check as suggested by @vjatoth-qti   because it is only a one time cost.

johnpaultaken: Since the GetDesc returns alignment, pixel format, there would need to be a utility to convert the various format to size in bytes etc to do the comparison verification. All that extra code unnecessarily complicates things.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: adding currently supported and expected future support packed types

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: changed name from DML to D3D12_RESOURCE_IMPORT

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
johnpaultaken: One doesn't call CreateExternalResourceImporterForDevice at all, whereas the other does.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
johnpaultaken: QnnExternalResourceImporterImpl constructor can throw. Also this is the C interface layer. So your argument is valid and addressed.

onnxruntime/test/providers/qnn/qnn_basic_test.cc
johnpaultaken: I am more worried about the test silently skipping due to driver issues etc when it should run on a D3D12 device. Hence GTEST_SKIP is not used.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
johnpaultaken: typo, reverted

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
```diff
@@ -244,11 +243,19 @@ OrtStatus* ORT_API_CALL QnnExternalResourceImporterImpl::CreateTensorFromMemoryI
...
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT2:
+    case ONNX_TENSOR_ELEMENT_DATA_TYPE_INT2:
+      element_size = 0.25;
+      break;
     default:
       return impl.ort_api_.CreateStatus(ORT_INVALID_ARGUMENT, "Unsupported tensor element type");
   }
 
-  size_t tensor_size_bytes = element_count * element_size;
+  size_t tensor_size_bytes = size_t(element_count * element_size);
```
qti-mattsinc: This computes `tensor_size_bytes` incorrectly if `element_count * element_size` has a remainder. e.g. if you had a tensor with 17 UINT4 elements, then `tensor_size_bytes` would be 8 instead of 9. So, we should round up here instead of truncating. 

<details open>
  <summary>nit.</summary>
I also think it would make more sense to put this in a helper function like

```
size_t GetTensorFootprint(ONNXTensorElementDataType element_type, const int64_t* shape, size_t rank);
```

Note that this function has nothing qnn-specific, so we could see if the ORT maintainers would support adding this as an API function in ORT core. That way it's ORT core's responsibility to update it when new types come around. But for now it could just be a helper in qnn_utils.cc. (Also note: I can see that we already have a `GetElementSizeByType` function in the utils, but it seems to assume that the packed values have been unpacked to int8 as apparently is done for htp, see https://github.com/onnxruntime/onnxruntime-qnn/pull/356#discussion_r3372997379)
</details>

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
qti-mattsinc: Right, and that is invisible to the user. But since this is a nit, I'm fine with not simplifying it. Just beware that it makes lines [547-548](https://github.com/onnxruntime/onnxruntime-qnn/pull/692/changes/142bbd33e498ab9f4996602bd631efdba2aa2c86..f9b7484ef47129bccf14774fdbd82456dec997cd#diff-076fc19d6a7a68ab829f45e20af67c04b1a86d560afa8de2a0cc78c2cc666e2cR547-R548) dead code.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
qti-mattsinc: This `new (std::nothrow)` here has not been fixed

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: good catch, will fix.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
johnpaultaken: having a double guard doesn't hurt, just in case if it is called later from outside the current location without ifdef.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
johnpaultaken: Unlike the other case QnnExternalMemoryHandle constructor cannot throw. The reason why it was changed in the other case is stated in the last comment.

qti-mattsinc: 1. This goes against the current convention and isn't defensive against changes to the constructor
2. `mem_handle_registry_.insert(raw_handle)` on line 162 in this function can throw and isn't in a try-catch

Probably best to wrap the function in a try/catch.

In the future, we should really add `QNN_EP_API_IMPL_BEGIN` / `QNN_EP_API_IMPL_END` macros. Looking around the code base, I see other locations where noexcept functions can throw. I think I'll raise that in another PR and see what other folks think.

johnpaultaken: There is no current agreed convention.
There is no defense against future code changes currently, such defense can only be had with entire function inside try catch.
Currently there is only try catch wrap where current code might throw.
Anyway, a struct constructor where fields are initialized is not likely to throw in the future either.

qti-mattsinc (approved):

tirupath-qti (approved):

qti-mbadnara:
Hi @johnpaultaken 

The GPU tests added as part of this PR seem to be flaky and failing CI. Eg: https://github.com/onnxruntime/onnxruntime-qnn/actions/runs/33367721710/job/99543025486?pr=784

Would it be possible to raise a fix or disable them till a resolution is achieved?

qti-mattsinc:
> The GPU tests added as part of this PR seem to be flaky and failing CI

@johnpaultaken, this is probably because some of the CI machines are on older GFX drivers. `QnnExternalResourceImporterImpl::CanImportMemoryImpl` probably needs to detect this and return false if drivers are too old, which could be done by calling `QnnBackendManager::IsDx12SharedMemoryAllocatorSupported` (and maybe renaming that function to `IsDx12SharedMemorySupported`). Then, you use GTEST_SKIP when CanImportMemory fails. That would be the proper fix but you can also disable the tests in the interim.

johnpaultaken:
> > The GPU tests added as part of this PR seem to be flaky and failing CI
> 
> @johnpaultaken, this is probably because some of the CI machines are on older GFX drivers. `QnnExternalResourceImporterImpl::CanImportMemoryImpl` probably needs to detect this and return false if drivers are too old, which could be done by calling `QnnBackendManager::IsDx12SharedMemoryAllocatorSupported` (and maybe renaming that function to `IsDx12SharedMemorySupported`). Then, you use GTEST_SKIP when CanImportMemory fails. That would be the proper fix but you can also disable the tests in the interim.

The GPU test assume all test machines have DX12 drivers. If that is not true we would need to put a GTEST skip. But I prefer updating the graphics drivers on target machines to the latest.

PR #694, opened 2026-08-04
Add matmul+add+relu QDQ node group

qti-shubham:
### Description
**Pattern:** DQ(input) + DQ(weight) -> MatMul -> Add(DQ(int32 bias)) -> Relu -> Q 
**Before:** quantized_reshape → quantized_FC → float_reshape → float_Relu → quantized_Q
**After:** quantized_reshape → quantized_FC(with bias) → quantized_reshape -> Quant_Relu (if encoding min !=0)


onnxruntime/test/providers/qnn/gemm_test.cc
```diff
@@ -773,6 +773,116 @@ TEST_F(QnnHTPBackendTests, DISABLED_GemmBQ_U16Int2_TransB0_BlockSize16) {
...
+
+}  // namespace
+
+// U16 activation, U8 weight, no Relu: baseline shape validated by
+// the default v68 HTP validator, no soc_model override needed.
+TEST_F(QnnHTPBackendTests, GemmMatMulAddFusion_U16Act_U8Weight_NoRelu) {
+  ProviderOptions opts;
+  opts["backend_type"] = "htp";
+  opts["offload_graph_io_quantization"] = "0";
+  RunQnnModelTest(BuildMatMulAddFusionQDQTestCase<uint16_t, uint8_t>(
```
qti-yuduo: **The `NoRelu` test never exercises this feature, and none of the three tests can prove absorption fired.**

I instrumented `GetOrtQDQSelection` with a stderr probe and ran all three tests locally on HTP V73. Results:

```
NoRelu:   Gemm consumers=1 kinds=[QuantizeLinear,]   <- absorption never fires
WithRelu: Gemm consumers=1 kinds=[Reshape,]
          absorbed output_reshape_node, clip=1
```

For `NoRelu` the Gemm's only consumer is already a `QuantizeLinear` by the time `GetCapability` runs, so `output_reshape_node` stays null and this test passes without touching the new code (see the note on `qnn_ep_utils.cc` for why).

Separately, `ExpectedEPNodeAssignment::All` cannot detect whether absorption happened in *any* of the three: a standalone Reshape/Relu is itself QNN-supported, so `All` holds either way. All three tests would still pass with this PR reverted.

Suggest asserting on the emitted QNN graph instead (FC output rank-2 + a following Reshape, or a node-count check) so a regression in the selector actually fails a test.

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1481,15 +1481,102 @@ std::optional<OrtNodeGroup> GetOrtQDQSelection(const OrtGraph* graph, const OrtA
     }
   }
 
-  // Find Q nodes that consume from this node or the clip node
+  // Detect the MatMulAddFusion post-Gemm Reshape: for a rank-2 Gemm sandwiched by rank-adapter
+  // Reshapes, absorb the trailing Reshape so its downstream Q's encoding attaches to the group.
+  // Supported downstream chains:
+  //   Gemm -> Reshape -> Q                     (Relu already dropped by ORT because Q.zp==0)
```
qti-yuduo: **This first chain is unreachable under ORT's L1 pipeline, and the parenthetical explains it backwards.**

`QDQPropagationTransformer` (L1, registered right after `MatMulAddFusion` in `graph_transformer_utils.cc`) inserts a Q/DQ pair *across* Reshape — `CanNodePropagate()` lists Reshape, and `PropagateQBackward()` walks back through it inserting Q->DQ on each edge. So a bare `Gemm -> Reshape -> Q` never survives to QNN EP; the Gemm's consumer is always a `QuantizeLinear` already. Probe output on the `NoRelu` test confirms this: `Gemm consumers=1 kinds=[QuantizeLinear,]`.

The reason the second chain works isn't that "the Relu survives" incidentally — it's that the Relu is *not* propagatable, so it blocks the Q/DQ insertion that would otherwise split the Gemm->Reshape edge. The Relu surviving is a precondition, not a tolerated edge case.

Concretely this means the `next_op == "QuantizeLinear"` branch below (L1517-1518) is dead code today, and it's the only path that can set `output_reshape_node` with `clip_node == nullptr`. That distinction matters because `OrtNodeUnit::GetOutputNodes()` in `ort_api.cc` resolves its output producer as `redundant_clip_node_ ? redundant_clip_node_ : target_node_` and was not updated for `output_reshape_node_` — harmless *only* because the reachable path always sets both. Worth either dropping the dead branch or noting the invariant, since as written it reads as live and makes the `GetOutputNodes()` omission look like a latent partitioning bug.

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
```diff
@@ -464,6 +464,63 @@ Ort::Status GemmOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
...
+    // weight [K, N] (transB=0 asserted for MatMulAddFusion pattern).
+    TensorInfo act_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[0], act_info));
+    TensorInfo weight_info = {};
+    RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(node_unit.Inputs()[1], weight_info));
+    RETURN_IF_NOT(act_info.shape.size() == 2, "QNN EP: absorbed-Reshape Gemm activation must be rank-2 [M, K].");
+    RETURN_IF_NOT(weight_info.shape.size() == 2, "QNN EP: absorbed-Reshape Gemm weight must be rank-2 [K, N].");
+    const int64_t trans_b = node_helper.Get("transB", static_cast<int64_t>(0));
+    RETURN_IF_NOT(trans_b == 0, "QNN EP: absorbed-Reshape Gemm only supports transB=0.");
+    std::vector<uint32_t> fc_output_shape{act_info.shape[0], weight_info.shape[1]};
```
qti-yuduo, line 489: **`transA` is never checked, but `fc_output_shape` assumes `transA == 0`.**

Per the ONNX Gemm spec, A is `[M, K]` when `transA=0` and `[K, M]` when `transA=1`. `act_info.shape[0]` is only M in the former case, so with `transA=1` this derives `{K, N}` instead of `{M, N}`. `transA` isn't inspected here, in `ExplictOpCheck`, or in the selector's structural detection.

I checked the blast radius: `MatMulAddFusion` never emits `transA`, so reaching this needs a user-authored Gemm or `GemmTransposeFusion`. I built such a case (explicit `transA=1`, sandwiching Reshapes, plus a Relu) and absorption *does* fire — but QNN validation then rejects the graph (4 of 12 nodes assigned), so the Gemm falls back to CPU and results stay correct. So this is a fail-fast/clarity issue rather than a miscompute, and it's consistent with the `transB` guard one line up:

```suggestion
    const int64_t trans_b = node_helper.Get("transB", static_cast<int64_t>(0));
    RETURN_IF_NOT(trans_b == 0, "QNN EP: absorbed-Reshape Gemm only supports transB=0.");
    const int64_t trans_a = node_helper.Get("transA", static_cast<int64_t>(0));
    RETURN_IF_NOT(trans_a == 0, "QNN EP: absorbed-Reshape Gemm only supports transA=0.");
    std::vector<uint32_t> fc_output_shape{act_info.shape[0], weight_info.shape[1]};
```

onnxruntime/core/providers/qnn/qnn_ep_utils.cc
```diff
@@ -1481,15 +1481,102 @@ std::optional<OrtNodeGroup> GetOrtQDQSelection(const OrtGraph* graph, const OrtA
...
+                          int64_t qmin = 0, qmax = 0;
+                          if (qnn::utils::GetQminQmax(qnn_dt, qmin, qmax).IsOK()) {
+                            float encoding_min = scale_val * static_cast<float>(qmin - zero_point);
+                            float encoding_max = scale_val * static_cast<float>(qmax - zero_point);
+                            float activation_min = 0.0f;
+                            float activation_max = std::numeric_limits<float>::max();
+                            if (next_op == "Clip") {
+                              GetClipMinMax(graph, ort_api, candidate_next, activation_min, activation_max);
+                            }
+                            safe = (encoding_min >= activation_min && encoding_max <= activation_max);
```
qti-yuduo: Minor: this encoding-safety fold check (`GetQNodeScaleAndZeroPoint` -> `GetQminQmax` -> `GetClipMinMax`, then comparing `encoding_min/max` against `activation_min/max`) is duplicated near-verbatim from the direct Relu/Clip fold at ~L1404-1456. Worth extracting a helper (e.g. `IsClipFoldEncodingSafe(graph, ort_api, q_node, clip_node)`) so a future correction to the bounds comparison only has to land once.

qti-yuduo (commented):
Reviewed the absorbed-Reshape path and validated each concern empirically on Windows ARM64 / HTP V73 (QAIRT 2.46) by instrumenting `GetOrtQDQSelection` and running the new tests. The design is sound and mirrors the existing `redundant_clip_node_` pattern; no correctness bug survived verification. Four inline notes follow, ordered by how much I'd want them addressed. The main one is test efficacy: none of the three new tests can currently fail if the feature were reverted.

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
```diff
@@ -464,6 +464,66 @@ Ort::Status GemmOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
     return Ort::Status();
   }
 
+  // MatMulAddFusion post-Gemm Reshape absorbed by the QDQ selector: FC (rank-2, encoded) followed
+  // by a QNN Reshape (rank-N, same encoding). node_unit.Outputs()[0] already reflects the Reshape's
+  // rank-N shape and the terminal Q's encoding.
+  const OrtNode* absorbed_reshape = node_unit.GetOutputReshapeNode();
```
qti-yuduo, line 470: IsBQGemmWeight accepts transB=0 and IsGemmSafeForAbsorbedReshape never inspects weight quant params, so a blocked-quantized Gemm forms an absorbed group but hits the pre-existing BQ branch at first, where node_unit.Outputs()[0] is now rank-N. Corrected from the original review: this is a CPU fallback via IsOpSupported, not a build failure, and not a regression. But likely missing fusion?

qti-yuduo (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
qti-shubham: let me explicitly gate it out: have IsGemmSafeForAbsorbedReshape reject BQ weights and the pre-existing BQ path (with Reshape as a separate QNN node) should handle it as before

PR #697, opened 2026-08-04
[QNN EP] Introduce QNN_ARCH_ARM64 arch-detection macro

huaychou:
### Description

Introduce a canonical `QNN_ARCH_ARM64` macro in `ort_api.h` that covers all ARM64 targets: native ARM64 (`__aarch64__`, `_M_ARM64`) **and** the ARM64EC half of a Windows arm64x fat binary (`_M_ARM64EC`).

This is a foundational change only — the macro is defined but not yet consumed. Each feature will be adopted in follow-up PRs.

### Motivation and Context

On a Windows-on-ARM device, an **amd64 (x64) Python process** loading an **arm64x** QNN EP artifact executes the module's **ARM64EC half**. MSVC compiles that half with `_M_ARM64EC` defined but **not** `_M_ARM64`. Throughout the QNN EP, on-device feature gates are keyed on `defined(__aarch64__) || defined(_M_ARM64)`, which causes the ARM64EC half to misidentify itself as an x86 host.

> **Note:** `GetPlatformInfo`, multi-SoC EP context guard, and `ValidateCompiledModelCompatibilityInfo` were already fixed by #701.

### DIFF
--- a/onnxruntime/core/providers/qnn/ort_api.h
+++ b/onnxruntime/core/providers/qnn/ort_api.h
@@ -38,6 +38,16 @@
 
 namespace onnxruntime {
 
+// True for any ARM64 target: native ARM64 (Windows/Linux) AND the ARM64EC half of a
+// Windows arm64x fat binary. ARM64EC must be included because an amd64 process on an
+// arm64 device executes the ARM64EC half of an arm64x module, which defines _M_ARM64EC
+// (not _M_ARM64). Omitting it makes on-device code misdetect itself as an x86 host.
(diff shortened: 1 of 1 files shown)


minfhong-qti (commented):
When I modify my part in #701, I did not know there will be a macro here.
Should it be revised to use the macro defined here in this PR?

qti-chuteng (approved):

huaychou:
> When I modify my part in #701, I did not know there will be a macro here. Should it be revised to use the macro defined here in this PR?

I think you can fix it after this PR merged.

PR #698, opened 2026-08-04
[QNN EP] Gate Num Prepare Threads for HTP Graph Splitting

qti-mbadnara:
PR #645 added support for Graph Splitting for HTP backend. It introduced a provider option `num_prepare_threads` to specify the number of thread used to prepare the sub graphs. This option will be available from `QAIRT 2.50`. Without this change, the build will fail when we uplevel QAIRT to `2.49`.

This PR adds `CMake check_struct_has_member` detection for `QnnHtpContext_GraphSplit_t.numPrepareThreads` in `onnxruntime_providers_qnn.cmake`. When the field is detected, defines `QNN_HTP_GRAPH_SPLIT_HAS_NUM_PREPARE_THREADS` as a compile definition, which gates the `numPrepareThreads` assignment in `qnn_backend_manager.cc`. A parse-time warning is emitted when the user sets `htp_graphsplitter_num_prepare_threads` but the SDK build does not expose the field.

### DIFF
--- a/cmake/onnxruntime_providers_qnn.cmake
+++ b/cmake/onnxruntime_providers_qnn.cmake
@@ -90,6 +90,21 @@
       QNN_SDK_VERSION_MINOR=${QNN_SDK_VERSION_MINOR})
   endif()
 
+  # Detect whether QnnHtpContext_GraphSplit_t contains numPrepareThreads.
+  include(CheckStructHasMember)
+  set(CMAKE_REQUIRED_INCLUDES "${onnxruntime_QNN_HOME}/include/QNN")
+  check_struct_has_member("QnnHtpContext_GraphSplit_t" "numPrepareThreads"
+                          "HTP/QnnHtpContext.h"
+                          QNN_HTP_GRAPH_SPLIT_HAS_NUM_PREPARE_THREADS
+                          LANGUAGE CXX)
+  unset(CMAKE_REQUIRED_INCLUDES)
+  if(QNN_HTP_GRAPH_SPLIT_HAS_NUM_PREPARE_THREADS)
+    target_compile_definitions(onnxruntime_providers_qnn PRIVATE QNN_HTP_GRAPH_SPLIT_HAS_NUM_PREPARE_THREADS)
+    message(STATUS "QNN: QnnHtpContext_GraphSplit_t.numPrepareThreads detected — graph splitting thread count will be wired in.")
+  else()
+    message(STATUS "QNN: QnnHtpContext_GraphSplit_t.numPrepareThreads not found — graph splitting thread count will be ignored.")
+  endif()
+
   # Set linker flags for function(s) exported by EP DLL
   if(UNIX)
     if(ENABLE_COVERAGE AND NOT APPLE AND CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -1590,7 +1590,7 @@ Ort::Status QnnBackendManager::CreateContext(bool enable_htp_weight_sharing,
     context_config_graph_splitting.option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
     context_config_graph_splitting.customConfig = &graph_splitting_custom_config;
   }
-#else  // QNN_HTP_GRAPH_SPLITTING_AVAILABLE
+#else   // QNN_HTP_GRAPH_SPLITTING_AVAILABLE
   ORT_UNUSED_PARAMETER(enable_htp_graph_splitting);
   ORT_UNUSED_PARAMETER(graphsplitter_num_prepare_threads);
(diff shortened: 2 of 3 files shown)


yath1:
If this option is not available until 2.50, I guess we have not tested this feature yet? If so, why not remove this field altogether from QNN-EP for now instead of guarding it all over? Once 2.50 is available, you could test it against a few LLMs or non-LLMs and introduce this option if the results look good. 
PS: This wont block any user because this is a perf change and not functional.

qti-mbadnara:
> If this option is not available until 2.50, I guess we have not tested this feature yet? If so, why not remove this field altogether from QNN-EP for now instead of guarding it all over? Once 2.50 is available, you could test it against a few LLMs or non-LLMs and introduce this option if the results look good.
> PS: This wont block any user because this is a perf change and not functional.

This was tried and tested with the Engineering SDK shared by the HTP team. The feature was supposed to go in for 2.49 but got pushed to 2.50

yath1:
> > If this option is not available until 2.50, I guess we have not tested this feature yet? If so, why not remove this field altogether from QNN-EP for now instead of guarding it all over? Once 2.50 is available, you could test it against a few LLMs or non-LLMs and introduce this option if the results look good.
> > PS: This wont block any user because this is a perf change and not functional.
> 
> This was tried and tested with the Engineering SDK shared by the HTP team. The feature was supposed to go in for 2.49 but got pushed to 2.50

Okay, cool. Things can change between engineering sdk and release sdk. Why not remove this flag until its official in 2.50? I am not a big fan of introducing a provider option now which is not supported in any versions of QNN.

cmake/onnxruntime_providers_qnn.cmake
```diff
@@ -90,6 +90,21 @@
       QNN_SDK_VERSION_MINOR=${QNN_SDK_VERSION_MINOR})
   endif()
 
+  # Detect whether QnnHtpContext_GraphSplit_t contains numPrepareThreads.
+  include(CheckStructHasMember)
```
minfhong-qti, line 94: Why use this instead of version check by macro?
If this is due to nightly or RC haven't bumped the version yet, I still think we should use macro version check as those versions will be bumped in final release.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1305,13 +1305,21 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
 #ifdef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
   QnnContext_Config_t context_config_graph_splitting_vtcm = QNN_CONTEXT_CONFIG_INIT;
   QnnHtpContext_CustomConfig_t graph_splitting_custom_config_vtcm;
+  ORT_UNUSED_PARAMETER(graphsplitter_num_prepare_threads);
   if (enable_htp_graph_splitting) {
     graph_splitting_custom_config_vtcm.option = QNN_HTP_CONTEXT_CONFIG_OPTION_GRAPH_SPLITTING_CONFIGS;
     graph_splitting_custom_config_vtcm.graphSplittingConfigs.graphSplittingEnabled = true;
+#ifdef QNN_HTP_GRAPH_SPLIT_HAS_NUM_PREPARE_THREADS
     graph_splitting_custom_config_vtcm.graphSplittingConfigs.numPrepareThreads = graphsplitter_num_prepare_threads;
+#endif
```
```suggestion
#endif  // QNN_HTP_GRAPH_SPLIT_HAS_NUM_PREPARE_THREADS
```
Probably worth comment which `#ifdef` that `#else` end `#endif` belong to. As well for others.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1305,13 +1305,21 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
 #ifdef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
   QnnContext_Config_t context_config_graph_splitting_vtcm = QNN_CONTEXT_CONFIG_INIT;
   QnnHtpContext_CustomConfig_t graph_splitting_custom_config_vtcm;
+  ORT_UNUSED_PARAMETER(graphsplitter_num_prepare_threads);
```
minfhong-qti: Use `ORT_UNUSED_PARAMETER` only in `#else`, and as well for below.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Yup fixed!

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Yup fixed!

qti-mbadnara:
Based on internal discussion with @yath1, @tirupath-qti and @qti-ashwshan will raise a new PR to remove the provider option. This can be exposed when it is exposed in QAIRT 2.50

Closing this PR

PR #482, opened 2026-06-04
[FCB] Enable Flexible Context Binary (FCB) support via QNN DLC

minfhong-qti:
### Description

- Add `QnnBackendSystemDlcPlugin` (new class) that wraps QNN System DLC APIs: create/release DLC, add contexts, retrieve binary buffers, query graph info, and compute max spill-fill buffer sizes. Requires QAIRT 2.48+ (QNN API >= 2.37).
- Update `QnnBackendManager` to expose `AddContextToDlc()`, `GetDlcBinaryBuffer()`, `GetDlcBinaryInfo()`, and `GetDlcMaxSpillFillBufferSize()` delegating to the new plugin, and plumb the system-context handle needed by DLC introspection.
- Add multi-SoC compilation path in `QnnExecutionProvider`:
  - New session option `htp_arch_per_soc` (comma-separated) specifies the target HTP arches to compile for; validated against `soc_model_per_soc` and per-SoC option vectors.
  - `GetCapability` iterates over each specified arch, calling `SetupDeviceAndContext` per arch to collect supported nodes.
  - New `CompileOnnxModelMultiSoc` iterates over each arch, overrides per-SoC HTP backend configs, compiles via `CompileOnnxModel`, and calls `AddContextToDlc` after each context is built. The final DLC binary is serialised into the EP context ONNX model node.
  - `enable_multi_soc_ep_context_` flag gates the new path; falls back to the existing single-SoC path when not set.
- Refactor `QnnCacheCompatibilityManager` out of `builder/` into a dedicated `cache_compatibility/` sub-directory. Split the flat `QnnCompatibilityInfo` struct into versioned variants (`V1` / `V2`) backed by `std::variant` in `qnn_cache_compatibility_info.h`, and add matching serialise/deserialise/validate paths for each version. V2 records a list of HTP arches to support multi-SoC FCB validation.

Doc:
   - Update `docs/execution_providers/QNN-ExecutionProvider.md`: add a "Flexible Context Binary (FCB) / multi-SoC EP context" section, and document the comma-separated list forms of `htp_arch` / `soc_model` and the new `enable_htp_ref_weight_sharing` option.

### Motivation and Context

Introduces the Flexible Context Binary (FCB) feature, which packages one or more QNN context binaries into a single DLC for cross-SoC deployment.
Testcases actually preparing multi-SoC binary will be re-enabled once CI uplevels to QAIRT 2.48.

PR #701, opened 2026-08-05
[QNN] Fix on-device macro guard to consider ARM64EC.

minfhong-qti:
### Description

Fix on-device macro guard to correctly consider ARM64EC use case by adding `_M_ARM64EC`.
This change only targets on those guards introduced when enabling FCB feature, especially V2 model compatibility.

### Motivation and Context

Currently, macros `__aarch64__` and `_M_ARM64` are used to check whether is on-device runtime. However, this combination lacks to consider ARM64EC use case (e.g., Python AMD64 on WoS), and causing unexpected behavior.
Considering a use case that EP context model is prepared on-device through Python AMD64 and then run with Python ARM64, the preparation is unable to acquire correct HTP arch and thus records default value 0 in compatibility info. During inference, when HTP arch is correctly acquired, model compatibility will report unsupported as cache HTP arch < 73 but runtime HTP arch >= 73.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -3142,7 +3142,7 @@ Ort::Status QnnBackendManager::GetPlatformInfo() {
     return Ort::Status();
   }
 
-#if defined(__aarch64__) || defined(_M_ARM64)
+#if defined(__aarch64__) || defined(_M_ARM64) || (defined(_M_ARM64EC))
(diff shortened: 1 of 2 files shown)


qti-chuteng (commented):
Please also check [qnn_model_compatibility_test.cc‎](https://github.com/onnxruntime/onnxruntime-qnn/pull/482/changes#diff-150bcaecae57f885878a32246c4eaac99df72cdd290a40e1b6729493d7b5a7a9) and the other macro guard in that files

qti-chuteng (approved):

PR #702, opened 2026-08-05
[QNN EP] Remove htp_graphsplitter_num_prepare_threads and htp_graph_splitting_kway_partitions provider options

qti-mbadnara:
## Summary

- Removes the `htp_graphsplitter_num_prepare_threads` provider option
- Removes the `htp_graph_splitting_kway_partitions` provider option
- Removes all `GPE_KWAY_PARTITIONS` environment variable manipulation from within the EP
- Removes 3 tests that covered the removed options
- Documents `GPE_KWAY_PARTITIONS` as a process-level environment variable

## Provider Options After This Change

| Option | Status |
|--------|--------|
| `enable_htp_graph_splitting` | ✅ Kept — enables graph splitting |
| `htp_graphsplitter_num_prepare_threads` | ❌ Removed |
| `htp_graph_splitting_kway_partitions` | ❌ Removed, set `GPE_KWAY_PARTITIONS` env var before process launch instead |


qti-yuduo:
Won't this cause backward compatibility issues? Are we sure no one depend on them?

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -335,15 +335,12 @@ For more information, see the [Parallel Graph Preparation](#parallel-graph-prepa
...
 
-|`"htp_graphsplitter_num_prepare_threads"`|Description|
+|`"GPE_KWAY_PARTITIONS"`|Description|
 |---|---|
-|Integer string ≥ 1, default `"8"`|Number of threads used to prepare sub-graphs in parallel. Only effective when `enable_htp_graph_splitting=1`. `"0"` means auto (uses `min(hardware_concurrency, number_of_splits)`). Higher values reduce prepare time at the cost of peak CPU/memory usage during preparation.|
-
-|`"htp_graph_splitting_kway_partitions"`|Description|
-|---|---|
-|Integer string, default `"4"`|Number of sub-graphs (k-way partitions) the HTP backend splits the model into during context creation. `"0"` leaves the decision to the SDK. Only effective when `enable_htp_graph_splitting=1`. Equivalent to setting the `GPE_KWAY_PARTITIONS` environment variable; setting this provider option is preferred since it applies per-session without process-wide side effects (note: implementation sets the env var immediately before context creation).|
+| `8` | Default |
```
yath1, line 342: When you say the default value is 8, does that mean if a user enables using `enable_htp_graph_splitting` provider option but has not set the `GPE_KWAY_PARTITIONS` env variable, by default 8 would be used?

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -335,15 +335,12 @@ For more information, see the [Parallel Graph Preparation](#parallel-graph-prepa
...
+| `N` | Split into N sub-graphs (must be ≥ 2) |
```
yath1, line 343: what happens in the following cases: 
1. GPE_KWAY_PARTITIONS is set but is empty
2. GPE_KWAY_PARTITIONS is set to 0
3. GPE_KWAY_PARTITIONS is random value (Example: -1, "blah blah", etc)

yath1:
Ideally if we can point to the docs of this env variable from HTP team, that would be great. They manage it, they can change the behavior and we don't have to update on our end to keep up with those changes.

qti-mbadnara:
> Won't this cause backward compatibility issues? Are we sure no one depend on them?

Hey @qti-yuduo, it wont cause any backward compatibility issue. Graph Splitting as a feature was not exposed to any ORT QNN EP consumers.

docs/execution_providers/QNN-ExecutionProvider.md
qti-mbadnara: Yes thats correct

docs/execution_providers/QNN-ExecutionProvider.md
qti-mbadnara: Any numeric value < 2, it will default to 8 according to HTP

yath1: Please try to get answer for the third point as well. Not a blocker though.
Also, I think its not an ideal design from HTP team. If GPE_KWAY_PARTITIONS==1, why do they not honor it for example?

yath1 (approved):
Although for now, this looks okay to begin with. After 2.50, the API might not look very smooth. 

For example: 
Case 1: Just to enable this feature, user has to do: 
provider_options["enable_htp_graph_splitting"] = 1
export GPE_KWAY_PARTITIONS=2

Case 2: To enable this feature and to set threads, user has to do: 
...
export GPE_KWAY_PARTITIONS=2
provider_options["enable_htp_graph_splitting"] = 1
provider_options["htp_graphsplitter_num_prepare_threads"] = 4
...

If you start taking other provider_options/env variables into account, this wouldn't look nice.

User would have to play with both env variables and provider options, which would easily cause confusion and frustrate them. I understand, we in QNN-EP cant control all these since they are coming from HTP team. I suggest brainstorm a clean API internally and propose to HTP team instead.

qti-mbadnara:
> Although for now, this looks okay to begin with. After 2.50, the API might not look very smooth.
> 
> For example: Case 1: Just to enable this feature, user has to do: provider_options["enable_htp_graph_splitting"] = 1 export GPE_KWAY_PARTITIONS=2
> 
> Case 2: To enable this feature and to set threads, user has to do: ... export GPE_KWAY_PARTITIONS=2 provider_options["enable_htp_graph_splitting"] = 1 provider_options["htp_graphsplitter_num_prepare_threads"] = 4 ...
> 
> If you start taking other provider_options/env variables into account, this wouldn't look nice.
> 
> User would have to play with both env variables and provider options, which would easily cause confusion and frustrate them. I understand, we in QNN-EP cant control all these since they are coming from HTP team. I suggest brainstorm a clean API internally and propose to HTP team instead.

Agreed, the environment variable needs to be fixed, will work with the stakeholders innvolved

PR #703, opened 2026-08-06
Add golden store publish scripts + nightly CI wiring

yuhuchua-qti:
## Summary

Stacked on PR1 (`migrate-clip-ut`, #399). Delivers the **producer** side of the QNN EP unit-test golden pipeline: scripts that regenerate goldens, keep only accuracy-passing op groups, version-stamp them, and publish to Artifactory — plus the nightly CI job that runs it and the retention policy that prunes old archives.

Pure additive. No production op-builder / snapshot / accuracy source is touched. The consumer (per-PR accuracy-routing gate) is a separate later PR.

## What's in it

**Version resolver — single source of truth**
- `qcom/scripts/linux/resolve_tool_versions.sh` — source-able *and* executable. Resolves QAIRT + ORT **runtime** version by reading `<bin_dir>/CMakeCache.txt`'s `onnxruntime_QNN_HOME` / `onnxruntime_ORT_HOME` entries (falling back to `_deps/ort_core-src/VERSION_NUMBER` when ORT is FetchContent-ed, which is what every current CI build does) — **not** env vars. No CI workflow ever exports `QAIRT_SDK_ROOT`/`QNN_SDK_ROOT`/`SNPE_ROOT`/`ORT_PREBUILT_ROOT`, so the original env-var-based resolver always failed with "undeterminable" in real CI; reading the build's own CMake record is a strict superset of reading an env var that's never set in practice. Deliberately does *not* read repo-root `VERSION_NUMBER` (that's the plugin/wheel version, not the ORT runtime). Undeterminable → exit 3 (graceful), so callers treat it as a safe signal. Writer and future gate stay consistent by sharing this *code*, never a process.

**Publisher — thin orchestrator**
- `qcom/scripts/linux/publish_goldens.sh` — regen goldens via `run_snapshot_accuracy.sh --generate-goldens` → read the accuracy report → package **only** groups whose every `QnnUnit_<Group>_Accuracy[_<Variant>]Test` case PASSED → write a version-stamped `manifest.json` → upload a write-once `archive/<date>-<sha>/` and the mutable `latest/` pointer. **Default is dry-run**; `--publish` is required to actually upload (it overwrites the shared `latest/` pointer — never implicit). `--keep-zip=<path>` (off by default) additionally copies the packaged `goldens.zip` to a known path for CI artifact upload, without affecting dry-run/publish behavior.
- `qcom/scripts/linux/filter_accuracy_pass_groups.py` — extracted pass-group selection (a group passes iff all its accuracy cases COMPLETED with no failures). Regex matches PR1's op-first suite naming (`QnnUnit_<Group>_Accuracy[_<Variant>]Test`).

**Nightly CI wiring**
- `.github/workflows/qualcomm-internal-golden-publish.yml` — reusable workflow (`workflow_call` + `workflow_dispatch`). Builds the coverage-instrumented binary, then runs `publish_goldens.sh`. `dry_run` defaults **true** so manual dispatch never clobbers the shared store; the scheduled producer opts into the real upload explicitly. `keep_zip_artifact` input (default `false`) wires `--keep-zip` and uploads `goldens.zip` as a 3-day-retention GitHub Actions artifact — for manual pre-merge inspection of dry-run output only.
- `.github/workflows/qualcomm-internal-release-nightly.yml` — adds a `golden-publish` job (9am UTC cron) passing `dry_run: false`.
- `qcom/build_and_test.py` — new `coverage_build_linux_x86_64` task (coverage build only, no HTML report — snapshot/accuracy tiers only compile under `--enable-coverage`).

**Golden store layout + retention**
- Store lives under `qnn-ep-test-store/ut-snapshot/{archive,latest}/`. The `qnn-ep-test-store/` parent is shared with future siblings (e.g. modelzoo goldens).
- `qcom/scripts/artifactory/aql/expired_golden_artifacts.json` — AQL that matches `qnn-ep-test-store/*/archive/*` older than 10d. Retention keys off the `archive/` segment only, so `latest/` pointers (and any permanent data outside `archive/`) are **never** matched — retention can never leave a store without a current golden.
- `.github/workflows/qualcomm-internal-clean-artifacts.yml` — new "Delete old golden archives" step (runs on the existing daily cleanup cron, 1h before the publish cron).

## Test plan
- [x] `pytest qcom/scripts/linux/tests` — 19 tests pass (resolver priority order incl. CMakeCache.txt fallback chain, sdk.yaml variants, undeterminable → exit 3).
- [x] `publish_goldens.sh` dry-run end-to-end against a real coverage build-dir: regen → parse accuracy report → stage passing groups → build `goldens.zip` + `manifest.json` → prints the two `jf rt upload` commands without executing.
- [x] `manifest.json` `qairt_version`/`ort_version` match `resolve_tool_versions.sh both`; `git_sha` matches `git rev-parse --short=10 HEAD`.
- [x] Empty-pass guard: zero passing groups → die, no zip/upload.
- [ ] CI dry-run verified end-to-end (push-triggered, `keep_zip_artifact: true`) — downloaded `goldens.zip` artifact inspected for correctness.
- [ ] Live Artifactory upload verified via nightly (post-merge).


.github/workflows/qualcomm-internal-snapshot-golden-publish.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Publish goldens
+        run: |
+          set -euo pipefail
+
+          build_dir="build/linux-x86_64"
+
+          args=(--build-dir="${build_dir}")
+
+          if [ -n "${{ inputs.filter }}" ]; then
```
kuanyul-qti: **[C-1] (Critical)** Workflow inputs are interpolated into shell source, enabling arbitrary commands on the self-hosted runner

`filter` and `repo_subpath` are user-controlled `workflow_dispatch` strings, but they are expanded directly into the multiline Bash source through `${{ inputs.* }}`. GitHub substitutes expressions before starting the shell, so input containing quotes, semicolons, and shell syntax can escape the intended string and execute arbitrary commands. This job runs on a self-hosted build runner, and publish mode configures JFrog credentials with Artifactory access.

**Impact:**A user allowed to dispatch the workflow, but not directly access the runner or Artifactory credentials, can use a crafted input to execute code on the privileged runner and read or exfiltrate credentials.

**Suggested fix:**

Pass every string input through step-level `env`, for example `FILTER_INPUT: ${{ inputs.filter }}` and `REPO_SUBPATH_INPUT: ${{ inputs.repo_subpath }}`, and reference only `"${FILTER_INPUT}"` and `"${REPO_SUBPATH_INPUT}"` from Bash. Also validate `repo_subpath` against an allowed character set and expected directory prefix.

.github/workflows/qualcomm-internal-snapshot-golden-publish.yml
```diff
@@ -0,0 +1,110 @@
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Initialize submodules
+        run: git submodule update --init --recursive
+
+      - name: Build coverage-instrumented test binary
+        if: ${{ !inputs.skip_regen }}
```
kuanyul-qti: **[M-1] (Major)** The skip_regen workflow mode has no build, goldens, or accuracy results to reuse

When `skip_regen=true`, the workflow skips both the coverage build and golden-directory preparation, while no download, cache restore, or other step supplies `onnxruntime_provider_test`, `accuracy_results.json`, or the golden tree to the fresh job. `actions/checkout` also cleans an existing workspace by default. The advertised "reuse existing goldens + accuracy results" contract therefore cannot be satisfied by this workflow.

**Impact:**The public manual/reusable workflow input fails deterministically when enabled. A clean-job reproduction exits with `--skip-regen but golden dir does not exist` before packaging or publishing.

**Suggested fix:**

On the skip path, explicitly download or restore previously produced build artifacts, the accuracy JSON, and the golden store, and validate their version/SHA. If there is no reliable persisted source, remove `skip_regen` from the workflow inputs and keep it only as a local-script option.

.github/workflows/qualcomm-internal-snapshot-golden-publish.yml
yuhuchua-qti: Thanks for catching this. I updated the workflow to avoid interpolating user-controlled workflow inputs directly into the Bash script. `filter` and `repo_subpath` are now passed through step-level environment variables, and `repo_subpath` is validated with an allowlist prefix/character check before being used.

.github/workflows/qualcomm-internal-snapshot-golden-publish.yml
yuhuchua-qti: Good point. I removed `skip_regen` from the workflow inputs since the CI job does not currently restore the required build outputs, accuracy report, or golden store. The `--skip-regen` option remains available only at the script level for local/debug reuse.

qcom/scripts/linux/publish_goldens.sh
```diff
@@ -0,0 +1,345 @@
...
+prove which groups pass (is QNN_EP_ACCURACY_UT enabled?). Refusing to publish \
+goldens with no pass evidence."
+fi
+log_info "Accuracy report: ${accuracy_json}"
+
+# ---------------------------------------------------------------------------
+# Select PASSING groups from the accuracy report.
+# ---------------------------------------------------------------------------
+filter_script="${REPO_ROOT}/qcom/scripts/linux/filter_accuracy_pass_groups.py"
+mapfile -t pass_groups < <(python3 "${filter_script}" "${accuracy_json}")
```
tzuhsuanwei: Non-blocking: mapfile does not propagate the exit status of the command in process substitution, so filter_exit here is normally 0 even when filter_accuracy_pass_groups.py exits 2 for malformed JSON.
This turns a report/parse failure into the less actionable “No group passed accuracy” error.
Could we capture the Python output in a temporary file (or command substitution), check its exit code first, then populate pass_groups?

qcom/scripts/linux/filter_accuracy_pass_groups.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# A group passes iff every QnnUnit_<Group>_Accuracy[_<Variant>]Test case has
+# result=COMPLETED and no failures. SKIPPED / NOTRUN cases count as not passed.
+
+import json
+import re
+import sys
+
+
+def filter_pass_groups(data):
```
tzuhsuanwei: Non-blocking: could we add focused tests for filter_pass_groups? This determines which goldens enter the shared store, but the new test file only covers the version resolver. At minimum, please cover a passing group, a failure, SKIPPED/NOTRUN, mixed variants in one group, and an unrelated suite.

qcom/scripts/linux/publish_goldens.sh
yuhuchua-qti: Thanks. 
I updated the publisher to capture the filter output through command substitution and check the Python exit code before populating pass_groups. Exit code 2 now reports an explicit accuracy-report read/parse failure, exit code 1 continues to report that no groups passed, and any other non-zero exit code fails with the actual code.

qcom/scripts/linux/filter_accuracy_pass_groups.py
yuhuchua-qti: Thanks. 
I added focused tests for filter_pass_groups, covering a passing group, a failed case, SKIPPED, NOTRUN, mixed variants within one group, and an unrelated suite. The fixtures also use the new QnnAcc_<Op>_Accuracy... naming introduced by PR #862.

tzuhsuanwei (approved):

PR #538, opened 2026-06-16
[QNN EP] Migrate activation ops to unified QNN_OP_ELEMENT_WISE_NEURON

qti-mbadnara:
### Description

Migrates the QNN Execution Provider from the per-activation QNN ops (`QNN_OP_RELU`, `QNN_OP_SIGMOID`, `QNN_OP_TANH`, `QNN_OP_GELU`, `QNN_OP_ELU`, `QNN_OP_RELU_MIN_MAX`) to the single unified `QNN_OP_ELEMENT_WISE_NEURON` op. The specific activation is now selected through the mandatory `QNN_OP_ELEMENT_WISE_NEURON_PARAM_OPERATION` scalar parameter (uint32) rather than being encoded in the QNN op type.

This covers the generic op-builder path (Elu, Gelu, Relu, Sigmoid, Tanh), the `Clip` op-builder
(ReluMinMax), the Gelu node-group fusion, and the QuickGelu decomposition's internal Sigmoid node.
Companion graph-assertion tests are updated to expect the new `ElementWiseNeuron` node name.

---

### Motivation & Context

The QNN SDK exposes `QNN_OP_ELEMENT_WISE_NEURON` as a single coarse operator covering 9 activation
functions via a mandatory `operation` enum parameter. Consolidating onto this op aligns the EP with
the SDK's preferred representation and collapses several dedicated-op code paths into one path keyed
by an enum value.

Because `operation` is **mandatory**, every site that previously emitted a dedicated activation op
must also attach the `operation` parameter (plus any operation-specific params such as `alpha`,
`min_value`/`max_value`), or QNN validation rejects the node.

`HardSigmoid`, `HardSwish`, `Softplus`, and `Selu` already used `QNN_OP_ELEMENT_WISE_NEURON` and
served as the reference pattern for this change.

---

### ONNX → operation enum mapping

| ONNX op | `QNN_OP_ELEMENT_WISE_NEURON_OPERATION_*` | Extra params |
|---|---|---|
| Elu | `_ELU` | `alpha` |
| Gelu | `_GELU` | — |
| Relu | `_RELU` | — |
| Sigmoid | `_SIGMOID` | — |
| Tanh | `_TANH` | — |
| Clip | `_RELU_MIN_MAX` | `min_value`, `max_value` |

Full neuron operation enum: ELU(0), GELU(1), HARD_SIGMOID(2), HARD_SWISH(3), RELU(4),
[...]

PR #531, opened 2026-06-15
[QNN EP] Migrate element-wise binary ops to unified QNN_OP_ELEMENT_WISE_BINARY

qti-mbadnara:
### Description

Migrates the QNN Execution Provider from the per-operation element-wise QNN ops (`QNN_OP_ELEMENT_WISE_ADD`, `QNN_OP_ELEMENT_WISE_MULTIPLY`, `QNN_OP_ELEMENT_WISE_SUBTRACT`, `QNN_OP_ELEMENT_WISE_DIVIDE`, …) to the single unified `QNN_OP_ELEMENT_WISE_BINARY` op. The specific operation is now selected through the mandatory
`QNN_OP_ELEMENT_WISE_BINARY_PARAM_OPERATION` scalar parameter (uint32) rather than being
encoded in the QNN op type itself.

This affects the generic op-builder path, the `Expand` op-builder (which implements broadcast via a binary op), and the two `DQ→integer bias-add` fusions. Companion graph-assertion tests are updated to expect the new `ElementWiseBinary` node name.

---

### Motivation & Context

The QNN SDK exposes `QNN_OP_ELEMENT_WISE_BINARY` as a single coarse operator that covers all
binary element-wise operations (ADD, SUBTRACT, MULTIPLY, DIVIDE, MIN, MAX, AND, OR, XOR, EQUAL,
GREATER, GREATER_EQUAL, LESS, LESS_EQUAL, POWER, …) via a mandatory `operation` enum parameter.

Because `operation` is **mandatory** for `QNN_OP_ELEMENT_WISE_BINARY`, every site that previously emitted a dedicated binary op must now also attach the `operation` parameter, or QNN validation rejects the node.

---

### ONNX → operation enum mapping

| ONNX op(s)                   | `QNN_OP_ELEMENT_WISE_BINARY_OPERATION_*` |
|------------------------------|------------------------------------------|
| Add, Sum                     | `_ADD`                                   |
| Sub                          | `_SUBTRACT`                              |
| Mul, Expand (numeric)        | `_MULTIPLY`                              |
| Div                          | `_DIVIDE`                                |
| Max                          | `_MAXIMUM`                               |
| Min                          | `_MINIMUM`                               |
| And, Expand (boolean)        | `_AND`                                   |
| Or                           | `_OR`                                    |
| Xor                          | `_XOR`                                   |
| Equal / NotEqual             | `_EQUAL`                                 |
| Greater                      | `_GREATER`                               |
[...]

PR #705, opened 2026-08-07
[QNN EP] Switch to Fine Grained Ops

qti-mbadnara:
The following PRs switched the op coverage to Coarse Ops

- #531
- #538 
- #541

After discussion with the stakeholders, decision was taken to move back to the original Fine Grained Ops. This PR implements that


qcom/packages.yml
```diff
@@ -148,9 +148,9 @@ nuget_windows_x86_64:
   md5: c51179f8b404bbf7f7b8e457240cc09d
   is_installer: false
 onnx_models:
-  version: 2026-06-23
+  version: 2025-10-02
```
qti-chuteng, line 151: May I know why this PR will fallback to old version?

qti-mbadnara: Hey @qti-chuteng 

The 2026-06-23 model zoo has relaxed tolerance for the coarse ops. Since we are moving back to the Fine-Grained Ops a stricter tolerance is used

tirupath-qti (approved):

PR #708, opened 2026-08-08
[QNN EP] Reject FP16/FP32 Tensors on HTP runtime for v68

qti-mbadnara:
### Description

HTP v68 does not have a hardware FP32 or FP16 execution path. Running a float model on a v68 device causes ops to fail inside the QNN SDK with opaque error code or silently fail at compile/runtime.

This PR fixes the problem at two levels:

**1. Fix HTP arch detection on multi-SOC devices (`qnn_backend_manager.cc`)**

`GetPlatformInfo` enumerates HTP devices to determine the silicon architecture. On multi-SOC devices like, the SDK reports devices as `QNN_HTP_DEVICE_TYPE_OFF_CHIP` from the primary chip's perspective. The old code only accepted `ON_CHIP` devices, so the loop found nothing, `htp_arch_internal_` stayed `NONE`, and the EP had no knowledge it was running on v68. The fix: also accept `OFF_CHIP` in the enumeration loop. The union has only one member (`onChipDevice`) regardless of device type, so the arch value is valid either way.

**2. Centralized FP32/FP16 rejection for v68 (`base_op_builder.cc`, `base_op_builder.h`)**

Rather than per-op manual guards, the rejection is placed in two existing chokepoints that together cover all op paths:

- **`ProcessDataTypes`** (NPU branch) — covers all simple ops (Relu, Add, Mul, etc.) and any op using `BaseOpBuilder::IsOpSupported`
- **`DataTypeCheckForCpuBackend`** — already called by all layout-sensitive ops (Conv, Pool, LRN, Resize, InstanceNorm, GroupNorm, LpPool) on the pre-layout-transform NCHW path (pass 1)

The `IsHtpV68Arch()` predicate in `base_op_builder.h` is available as a shared utility for any future op builder that needs it.

**Supporting changes**

- `qnn_model_wrapper.h` — adds `QnnHtpDevice_Arch_t htp_arch` to `ModelSettings` so the silicon architecture is visible inside op builders without threading the backend manager through
- `qnn_execution_provider.cc` — populates `model_settings_.htp_arch` immediately after `SetupBackend` resolves `htp_arch_internal_`

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
@@ -80,6 +80,14 @@ Ort::Status BaseOpBuilder::ProcessDataTypes(QnnModelWrapper& qnn_model_wrapper,
   if (IsCpuBackend(qnn_model_wrapper.GetQnnBackendType())) {
     return CheckCpuDataTypes(input_qnn_dtypes, output_qnn_dtypes);
   } else if (IsNpuBackend(qnn_model_wrapper.GetQnnBackendType())) {
+    // HTP v68 has no FP32/FP16 execution path. Reject early so nodes fall back
+    // to the CPU EP cleanly rather than failing deep inside the QNN SDK.
+    if (qnn_model_wrapper.GetModelSettings().htp_arch == QNN_HTP_DEVICE_ARCH_V68) {
+      for (const auto& qnn_dt : input_qnn_dtypes) {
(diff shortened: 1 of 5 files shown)


minfhong-qti:
Just curious for the motivation. If the op builder somehow can leverage backend validator to reject those FP16 in the first GetCapability call, would the motivation of this PR still stand?
I believe adding this "manual" guard is not scalable for increasing op builders and should leverage backend validators. Maybe we should try to explore how to avoid current manual check for those layout-sensitive ops in the builders.

PR #710, opened 2026-08-09
[QNN EP] Add support for ArrayFeatureExtractor Op

qti-mbadnara:
### Description

Adds support for the ONNX ML operator `ai.onnx.ml.ArrayFeatureExtractor` in the QNN Execution Provider, targeting HTP and CPU backends via a decomposition to the native `QNN_OP_GATHER` op.

---

### Motivation & Context

`ai.onnx.ml.ArrayFeatureExtractor(X, Y)` selects elements from tensor `X` using integer indices `Y` applied to the **last axis** only. It is the primary way scikit-learn feature-selection and column-transformer steps appear in ONNX ML pipelines.

ONNX ML spec: https://onnx.ai/onnx/operators/onnx_ml_ops__ArrayFeatureExtractor.html

The op has no attributes — axis is always `rank(X) - 1`. This makes it a direct, fixed-axis specialisation of `Gather`, so the implementation decomposes it:

```
ArrayFeatureExtractor(X, Y)  ≡  Gather(X, Y, axis=rank(X)-1)
```

Two additional concerns handled by the builder:

**`ai.onnx.ml` domain whitelist.** QNN EP previously filtered out all `ai.onnx.ml` nodes before `GetCapabilityImpl` was called. A new `kMLOnnxDomain` constant and a one-line addition to the domain whitelist in `qnn_ep_utils.cc` enable the EP to evaluate ML-domain ops.

**int64 data.** QNN HTP has no native int64 tensor support. When `X` is int64, a Cast to int32 is inserted before Gather and a Cast back to int64 is inserted after (for graph outputs only). `Y` indices are always cast from int64 → int32 since QNN Gather requires int32 indices.

---

### Supported Configuration

| Axis | Value |
|------|-------|
| Implicit axis | `rank(X) - 1` (last axis; fixed by ONNX spec — no attribute) |

| Input | Supported shapes |
|-------|-----------------|
| X     | Rank ≥ 1 |
| Y     | Rank ≥ 1 (1-D or N-D indices) |

---

### Unsupported Configuration (falls back to CPU EP)

- **Scalar Y (`rank(Y) == 0`)** — ORT does not infer the output shape for `ai.onnx.ml` ops with rank-0 indices. `QnnModel::SetGraphInputOutputInfo` throws before the op builder runs, so there is no EP-level workaround without an ORT-core shape inference change.
- **`double` X** — QNN has no float64 support.
- **`string` X** — QNN has no string tensor support.

---

### Data Types

| Backend | X dtypes              | Y dtypes |
|---------|-----------------------|----------|
| HTP     | float32, int32, int64 | int64 (int32 handled defensively) |
| CPU     | float32, int32, int64 | int64 (int32 handled defensively) |

QDQ is not applicable — `ai.onnx.ml.ArrayFeatureExtractor` is a non-quantised op.

---

### Files Changed

| File | Change |
|------|--------|
| `core/providers/qnn/common/qnn_graph_utils.h` | Added `kMLOnnxDomain = "ai.onnx.ml"` constant |
| `core/providers/qnn/qnn_ep_utils.cc` | Added `kMLOnnxDomain` to domain whitelist so EP evaluates `ai.onnx.ml` nodes |
| `core/providers/qnn/builder/op_builder_factory.h` | Added `CreateArrayFeatureExtractorOpBuilder` declaration |
| `core/providers/qnn/builder/op_builder_factory.cc` | Registered `CreateArrayFeatureExtractorOpBuilder("ArrayFeatureExtractor", *this)` |
| `core/providers/qnn/builder/opbuilder/array_feature_extractor_op_builder.cc` | **New** — `ArrayFeatureExtractorOpBuilder` |
| `test/providers/qnn/array_feature_extractor_test.cc` | **New** — HTP tests: float32 (2-D, 3-D), int32, int64, scalar Y (CPU fallback), dynamic int64 indices |

---


onnxruntime/core/providers/qnn/builder/opbuilder/array_feature_extractor_op_builder.cc
```diff
@@ -0,0 +1,309 @@
...
+
+// ---------------------------------------------------------------------------
+// IsOpSupported
+// ---------------------------------------------------------------------------
+Ort::Status ArrayFeatureExtractorOpBuilder::IsOpSupported(QnnModelWrapper& qnn_model_wrapper,
+                                                          const OrtNodeUnit& node_unit,
+                                                          const Ort::Logger& logger) const {
+  const auto& inputs = node_unit.Inputs();
+  RETURN_IF(inputs.size() != 2,
+            "QNN EP: ArrayFeatureExtractor must have exactly two inputs (X, Y).");
```
qti-kromero: do we need these QNN EP tags that seem to have been added to the logging messages? longer term do we want to just wrap these macros with our own macros that do this for us if it's not already accounted for in the existing logger?

onnxruntime/test/providers/qnn/array_feature_extractor_test.cc
```diff
@@ -0,0 +1,140 @@
...
+#if !defined(__linux__) || defined(__aarch64__)
+
+// 2D float32: X[3, 4], Y = [0, 2] selects columns 0 and 2.
+TEST_F(QnnHTPBackendTests, ArrayFeatureExtractor_Float32_2D_Basic) {
+  RunArrayFeatureExtractorTest<float, int64_t>(
+      TestInputDef<float>({3, 4}, false,
+                          {1.0f, 2.0f, 3.0f, 4.0f,
+                           5.0f, 6.0f, 7.0f, 8.0f,
+                           9.0f, 10.0f, 11.0f, 12.0f}),
+      TestInputDef<int64_t>({2}, true, {0, 2}),
```
qti-kromero, line 130: some negative axis index test would be good

qti-kromero (approved):

PR #716, opened 2026-08-12
Fix flaky archive-testdata CI: retry on IncompleteRead + persist ort_core zip cache

kuanyul-qti:
## Motivation and Context

The `archive-testdata` CI job fails intermittently when downloading the ~278 MiB `ort_core` source zip from GitHub's codeload endpoint: `http.client.IncompleteRead: IncompleteRead(0 bytes read)`

The download takes ~7 minutes on the self-hosted runners and codeload drops large transfers mid-stream under throttling. The previous code called `urlretrieve()` once with no retry, so a single dropped connection failed the whole job. It also cached the zip under `build/testdata-stage/`, which `git clean -ffdx` wipes on every checkout — so every run re-downloaded from scratch even though the build's `fetch_cmake_deps.py` already downloads the identical file into the persistent package cache.

The artifact itself is fine — URL, file, and SHA1 verified end-to-end (292 MB, hash matches `cmake/deps.txt`). This is purely a delivery-path problem.

## Changes

**`qcom/scripts/all/archive_testdata.py`**

- **Retry with backoff**: `download_and_verify` retries up to 3× with exponential backoff, deleting the partial file between attempts. The `except` catches `http.client.HTTPException` — the actual class of `IncompleteRead`, which is **not** an `OSError` and would slip past a naive `except OSError`.

- **Reuse the package cache**: new `ort_core_cache_path()` helper computes the cache location using the *same layout* as `package_manager.py`'s `FileCache` — `<root>/ort_core/<url-basename>` (e.g. `<root>/ort_core/v1.27.0.zip`) under `ORT_BUILD_PACKAGE_CACHE_PATH`. On a persistent runner where `fetch_cmake_deps.py` already downloaded that exact file, `archive_testdata` now sees a cache hit (SHA1 matches) and **skips the download entirely**. This root lives outside `build/`, so it also survives `git clean`. New `--cache-dir` arg overrides the root.

The script stays **stdlib-only** — the `archive_testdata` task has no `@depends(["create_venv"])` and runs on the bare interpreter, so it can't `import FileCache` directly (which pulls in `certifi`/`tqdm`/`yaml`); the helper replicates just the path computation.

**`qcom/scripts/all/tests/test_archive_testdata.py`**

- Added `test_ort_core_cache_path_matches_filecache_layout` locking the `<root>/ort_core/<url-basename>` path (the reuse guarantee).
- Updated the SHA-mismatch test (now retries then raises `RuntimeError` instead of fast-failing with `ValueError`).
- Added: `IncompleteRead` → retry → success (the exact CI failure), and a persistent-cache-hit-skips-download test.

### DIFF
--- a/qcom/scripts/all/tests/test_archive_testdata.py
+++ b/qcom/scripts/all/tests/test_archive_testdata.py
@@ -2,15 +2,27 @@
 # SPDX-License-Identifier: MIT
 
 import hashlib
+import http.client
 import shutil
 import tarfile
 import zipfile
 from pathlib import Path
 
 import pytest
-from archive_testdata import OrtCoreDep, download_and_verify, parse_deps_txt, stage_sources, write_archives
+from archive_testdata import (
+    OrtCoreDep,
+    download_and_verify,
+    ort_core_cache_path,
+    parse_deps_txt,
+    stage_sources,
+    write_archives,
+)
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
+
+
+def _sha1_of_file(path: Path) -> str:
+    return hashlib.sha1(path.read_bytes()).hexdigest()
 
 
 def test_parse_deps_txt_returns_url_and_sha1(tmp_path):
@@ -32,6 +44,13 @@ def test_parse_deps_txt_raises_when_ort_core_missing(tmp_path):
     deps_file.write_text("abseil_cpp;https://example.com/abseil.zip;deadbeef\n")
     with pytest.raises(ValueError, match="ort_core"):
         parse_deps_txt(deps_file)
+
+
+def test_ort_core_cache_path_matches_filecache_layout(tmp_path):
+    """The cache path must match package_manager.py's FileCache (<root>/ort_core/<url-basename>) so a
+    zip already fetched by fetch_cmake_deps.py is reused rather than re-downloaded."""
+    url = "https://github.com/microsoft/onnxruntime/archive/refs/tags/v1.27.0.zip"
+    assert ort_core_cache_path(tmp_path, url) == tmp_path / "ort_core" / "v1.27.0.zip"
 
 
 def _make_fake_zip(path: Path) -> str:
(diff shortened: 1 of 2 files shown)


qcom/scripts/all/tests/test_archive_testdata.py
```diff
@@ -34,6 +46,13 @@ def test_parse_deps_txt_raises_when_ort_core_missing(tmp_path):
         parse_deps_txt(deps_file)
 
 
+def test_ort_core_cache_path_matches_filecache_layout(tmp_path):
+    """The cache path must match package_manager.py's FileCache (<root>/ort_core/<url-basename>) so a
+    zip already fetched by fetch_cmake_deps.py is reused rather than re-downloaded."""
+    url = "https://github.com/microsoft/onnxruntime/archive/refs/tags/v1.27.0.zip"
```
qti-mbadnara, line 52: Can we populate this from the package.yml / deps.txt file?

qcom/scripts/all/tests/test_archive_testdata.py
```diff
@@ -34,6 +46,13 @@ def test_parse_deps_txt_raises_when_ort_core_missing(tmp_path):
...
+    assert ort_core_cache_path(tmp_path, url) == tmp_path / "ort_core" / "v1.27.0.zip"
```
qti-mbadnara, line 53: Same comment as above

qti-mbadnara (commented):
Minor nit, rest LGTM

qcom/scripts/all/tests/test_archive_testdata.py
kuanyul-qti: The URL here is a hardcoded structural fixture — it tests that `ort_core_cache_path()` produces `<root>/ort_core/<url-basename>`, not a live URL. `packages.yml` also has no `ort_core` entry; the real URL lives in `cmake/deps.txt` and is read at runtime via `parse_deps_txt()`. So there's nothing to populate from `packages.yml` here.

qcom/scripts/all/tests/test_archive_testdata.py
kuanyul-qti: Same as above — this asserts the expected path `<root>/ort_core/v1.27.0.zip`, which directly verifies the FileCache-compatible layout. The version in the fixture doesn't need to track the current `cmake/deps.txt` value since we're testing the path formula, not the current version.

qcom/scripts/all/tests/test_archive_testdata.py
qti-mbadnara: Thanks for clarifying!

qti-mbadnara (approved):

qti-chuteng (approved):

PR #717, opened 2026-08-12
[QNN EP] Fix archive-testdata CI flake by retrying on transient download errors

huaychou:
### Description

- Wrap the download in a retry loop with exponential backoff (default 3 attempts, 2s base), catching OSError, HTTPException (IncompleteRead is not an OSError), and SHA1 mismatches of a completed-but-truncated body.
- Policy mirrors package_manager.py and is overridable via ORT_BUILD_DOWNLOAD_ATTEMPTS / ORT_BUILD_DOWNLOAD_BACKOFF_SECONDS env vars.

### Motivation and Context

- GitHub's archive CDN intermittently drops connections mid-body with http.client.IncompleteRead, causing a bare single-shot urlretrieve call to fail with no recovery.

### DIFF
--- a/qcom/scripts/all/archive_testdata.py
+++ b/qcom/scripts/all/archive_testdata.py
@@ -18,10 +18,13 @@ expected by run_tests.{ps1,sh} and the test binaries.
 
 import argparse
 import hashlib
+import http.client
 import logging
+import os
 import re
 import shutil
 import tarfile
+import time
 import zipfile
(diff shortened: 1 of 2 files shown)


kuanyul-qti:
Duplicated with https://github.com/onnxruntime/onnxruntime-qnn/pull/716

huaychou:
Duplicated.

PR #718, opened 2026-08-13
[QNN EP] Apply fp16_clamp_overflow via graphSetConfig on EPContext binary-load path

huaychou:
### Description

- FP16_CLAMP_OVERFLOW is a runtime graph config settable via graphSetConfig on a finalized graph. Add QnnModel::ApplyRuntimeGraphConfigs taking HtpGraphConfigs_t& (extensible for future runtime options) and call it
from both branches of CompileContextModel and after SSR recovery.
- Add EPContext_Fp16ClampOverflow_RoundTrip unit test.

### Motivation and Context
- On the context-binary path, enable_htp_fp16_clamp_overflow was silently ignored because QnnModelContext.graph_configs is always nullptr there, causing NaN output on HTP v79+.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model.cc
@@ -9,9 +9,11 @@
 #include <thread>
 
 #include "HTP/QnnHtpContext.h"
+#include "HTP/QnnHtpGraph.h"
 #include "QnnOpDef.h"
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
+#include "core/providers/qnn/builder/qnn_configs_helper.h"
 #include "core/providers/qnn/builder/qnn_node_group/qnn_node_group.h"
 #include "core/providers/qnn/builder/op_tracing/qnn_op_tracing.h"
 #include "core/providers/qnn/builder/qnn_profile_serializer.h"
@@ -440,6 +442,60 @@ Ort::Status QnnModel::SetupQnnInputOutput(const Ort::Logger& logger) {
   return Ort::Status();
 }
 
+Ort::Status QnnModel::ApplyRuntimeGraphConfigs(const HtpGraphConfigs_t& configs,
+                                               const Ort::Logger& logger) {
+  // Record for re-application after an SSR event re-retrieves the graph handle.
+  // Contract: intended to be called once per QnnModel at session creation; RecoverFromSSR
+  // relies on this cached value, so re-calling with different configs would silently change
+  // post-SSR behaviour.
+  runtime_graph_configs_ = configs;
+
+  if (qnn_backend_type_ != QnnBackendType::HTP || graph_info_ == nullptr) {
+    return Ort::Status();
+  }
+
+  // Build the runtime-settable subset of graph configs using the same builder pattern as
+  // QnnEp::InitQnnHtpGraphConfigs. Only add options confirmed settable on a finalized graph;
+  // a compile-time-only option here would come back as QNN_GRAPH_ERROR_GRAPH_FINALIZED.
+  QnnConfigsBuilder<QnnGraph_Config_t, QnnHtpGraph_CustomConfig_t> builder(
+      QNN_GRAPH_CONFIG_INIT, QNN_HTP_GRAPH_CUSTOM_CONFIG_INIT);
+
+#ifdef QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
+  if (configs.enable_htp_fp16_clamp_overflow) {
+    gsl::not_null<QnnHtpGraph_CustomConfig_t*> cc = builder.PushCustomConfig();
+    cc->option = QNN_HTP_GRAPH_CONFIG_OPTION_FP16_CLAMP_OVERFLOW;
+    cc->fp16ClampOverflow = true;
+    gsl::not_null<QnnGraph_Config_t*> gc = builder.PushConfig();
+    gc->option = QNN_GRAPH_CONFIG_OPTION_CUSTOM;
+    gc->customConfig = cc;
+  }
+  // NOTE: future runtime-settable options append their own guarded block here.
+#endif
+
+  const QnnGraph_Config_t** graph_configs = builder.GetQnnConfigs();
+  if (graph_configs == nullptr) {
+    return Ort::Status();  // Nothing runtime-applicable was requested.
+  }
+
+  const auto& qnn_interface = qnn_backend_manager_->GetQnnInterface();
+  RETURN_IF(nullptr == qnn_interface.graphSetConfig,
+            "Invalid function pointer for graphSetConfig; cannot apply runtime graph configs.");
+  Qnn_ErrorHandle_t rt = qnn_interface.graphSetConfig(graph_info_->Graph(), graph_configs);
+  if (QNN_SUCCESS != rt) {
+    // QNN_GRAPH_ERROR_GRAPH_FINALIZED here means an option in the set is not runtime-settable
+    // on this SDK and must move back to compile-time wiring in InitQnnHtpGraphConfigs.
+    const std::string message = "Failed to apply runtime graph configs for graph: " +
+                                graph_info_->Name() + ". " +
+                                utils::FormatQnnError(qnn_interface, rt);
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_ERROR, message.c_str());
+    return MAKE_EP_FAIL(message.c_str());
+  }
+
+  ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
+              ("Applied runtime graph configs for graph: " + graph_info_->Name()).c_str());
+  return Ort::Status();
+}
+
 static Ort::Status BindQnnTensorMemoryToOrtValueMemory(const OrtApi& ort_api,
                                                        const Ort::Logger& logger,
                                                        QnnBackendManager& qnn_backend_manager,
@@ -551,7 +607,10 @@ Ort::Status QnnModel::RecoverFromSSR(const Ort::Logger& logger) {
   graph_info_->ResetHandles(new_graph, new_context);
 
   // Re-build the tensor I/O metadata against the new graph handles.
-  return SetupQnnInputOutput(logger);
+  RETURN_IF_ERROR(SetupQnnInputOutput(logger));
+
+  // The freshly retrieved graph handle does not carry runtime graph configs, so re-apply them.
+  return ApplyRuntimeGraphConfigs(runtime_graph_configs_, logger);
 }
 
 Ort::Status QnnModel::BindAndExecuteGraph(OrtKernelContext* context,
--- a/onnxruntime/core/providers/qnn/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model.cc
@@ -444,10 +444,10 @@ Ort::Status QnnModel::SetupQnnInputOutput(const Ort::Logger& logger) {
 
 Ort::Status QnnModel::ApplyRuntimeGraphConfigs(const HtpGraphConfigs_t& configs,
                                                const Ort::Logger& logger) {
-  // Record for re-application after an SSR event re-retrieves the graph handle.
-  // Contract: intended to be called once per QnnModel at session creation; RecoverFromSSR
-  // relies on this cached value, so re-calling with different configs would silently change
-  // post-SSR behaviour.
+  // Caches configs for re-application after an SSR event re-retrieves the graph handle.
+  // Each call overwrites runtime_graph_configs_; RecoverFromSSR always re-applies the most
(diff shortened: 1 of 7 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model.h
```diff
@@ -61,6 +61,12 @@ class QnnModel {
 
   Ort::Status SetupQnnInputOutput(const Ort::Logger& logger);
 
+  // Apply runtime-settable HTP graph configs (currently fp16 clamp overflow) on an already
+  // finalized graph loaded from a context binary, via QnnGraph_setConfig. Only options confirmed
+  // settable on a finalized graph belong here; compile-time-only options must stay in
+  // QnnEp::InitQnnHtpGraphConfigs. No-op when the backend is not HTP or nothing applies.
+  Ort::Status ApplyRuntimeGraphConfigs(const HtpGraphConfigs_t& configs, const Ort::Logger& logger);
```
minfhong-qti, line 68: Remember to add corresponding testcase in `unit/qnn_model_test.cc`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2560,6 +2560,8 @@ OrtStatus* QnnEp::CompileContextModel(const OrtGraph** graphs,
             /*json_qnn_graph_path=*/{}};
         RETURN_IF_NOT_OK(qnn_model_shared->SetGraphInputOutputInfo(context));
         RETURN_IF_NOT_OK(qnn_model_shared->SetupQnnInputOutput(logger_));
+        RETURN_IF_NOT_OK(qnn_model_shared->ApplyRuntimeGraphConfigs(
+            htp_graph_configs_, logger_));
```
```suggestion
        RETURN_IF_NOT_OK(qnn_model_shared->ApplyRuntimeGraphConfigs(htp_graph_configs_, logger_));
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2662,6 +2664,8 @@ OrtStatus* QnnEp::CompileContextModel(const OrtGraph** graphs,
         /*json_qnn_graph_path=*/{}};
     RETURN_IF_NOT_OK(qnn_model->SetGraphInputOutputInfo(context));
     RETURN_IF_NOT_OK(qnn_model->SetupQnnInputOutput(logger_));
+    RETURN_IF_NOT_OK(qnn_model->ApplyRuntimeGraphConfigs(
+        htp_graph_configs_, logger_));
```
minfhong-qti: Ditto

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -440,6 +442,60 @@ Ort::Status QnnModel::SetupQnnInputOutput(const Ort::Logger& logger) {
   return Ort::Status();
 }
 
+Ort::Status QnnModel::ApplyRuntimeGraphConfigs(const HtpGraphConfigs_t& configs,
+                                               const Ort::Logger& logger) {
+  // Record for re-application after an SSR event re-retrieves the graph handle.
+  // Contract: intended to be called once per QnnModel at session creation; RecoverFromSSR
+  // relies on this cached value, so re-calling with different configs would silently change
+  // post-SSR behaviour.
+  runtime_graph_configs_ = configs;
```
minfhong-qti, line 451: It could be a little bit weird that the config is set again during SSR recovery even though the set options are in fact the same with the previous ones.
May not be a real problem.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou: Fixed

onnxruntime/core/providers/qnn/builder/qnn_model.cc
huaychou: Add corresponding unit tests in qnn_ssr_test.cc

onnxruntime/core/providers/qnn/builder/qnn_model.h
huaychou: Add corresponding unit tests in qnn_model_test.cc

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
```diff
@@ -10,6 +10,8 @@
 #include "onnxruntime_cxx_api.h"
 #include "onnxruntime_session_options_config_keys.h"
 
+#include "core/providers/qnn/builder/qnn_def.h"
```
huaychou: **[M-1] (Major)** qnn_ssr_test.cc pulls in a private EP header, violating the file's public-API-only design

The PR adds at line 13:

```
#include "core/providers/qnn/builder/qnn_def.h"
```

However, `qnn_ssr_test.cc` was explicitly designed to avoid private EP headers. Lines 23–24 (now 25–26) of the original file state:

```
// Must match QNN_SSR_UNRECOVERABLE_ERROR_CODE in core/providers/qnn/ort_api.h; duplicated
// to avoid pulling that EP-private header into a public-API-only test.
```

`qnn_def.h` is a private EP header that depends on `QnnInterface.h`, `core/providers/qnn/builder/qnn_quant_params_wrapper.h`, and `core/providers/qnn/ort_api.h`. Unlike `qnn_model_test.cc` (which has a `QNN_EP_INTERNAL_SYMBOL_ACCESS` guard), this file has no such guard. Including this header directly contradicts the file's documented design intent.

**Impact:** Breaks the SSR test's public-API boundary. If this pattern is not addressed, future non-SDK-path builds and test refactoring become harder. Directly contradicts the existing design rationale in the same file.

**Suggested fix:**

Two options:

(A) If the new test only needs the `QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE` feature-detection macro, mirror the existing pattern for `QNN_SSR_UNRECOVERABLE_ERROR_CODE` — duplicate the macro definition locally with a comment citing its source (remove the `#include`):

```
// QNN_HTP_GRAPH_CONFIG_OPTION_FP16_CLAMP_OVERFLOW is available from QNN API 2.38 (QAIRT 2.49).
// Duplicated from core/providers/qnn/builder/qnn_def.h to avoid pulling that EP-private header.
#if QNN_API_VERSION_MAJOR > 2 || \
    (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 38)
#define QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
#endif
```

(B) If this test is already an integration test that legitimately needs SDK access, wrap the new test block in a `QNN_EP_INTERNAL_SYMBOL_ACCESS` guard and update comments to document the design intent shift.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -440,6 +442,60 @@ Ort::Status QnnModel::SetupQnnInputOutput(const Ort::Logger& logger) {
   return Ort::Status();
 }
 
+Ort::Status QnnModel::ApplyRuntimeGraphConfigs(const HtpGraphConfigs_t& configs,
+                                               const Ort::Logger& logger) {
+  // Record for re-application after an SSR event re-retrieves the graph handle.
+  // Contract: intended to be called once per QnnModel at session creation; RecoverFromSSR
```
huaychou: **[N-1] (Minor)** Contract comment says 'called once' but the unit test validates multiple calls — semantic mismatch

Lines 448–450 state:

```
// Contract: intended to be called once per QnnModel at session creation; RecoverFromSSR
// relies on this cached value, so re-calling with different configs would silently change
// post-SSR behaviour.
```

However, the `ApplyRuntimeGraphConfigs_CachesConfigsForSsr` unit test deliberately calls the function three times (on → off → on) to verify caching behavior. This is functionally correct, but contradicts the 'called once' wording. Furthermore, the phrase 're-calling with *different* configs would silently change post-SSR behaviour' implies multiple calls are valid (just with side effects), which is inconsistent with 'intended to be called once'.

**Impact:** A future maintainer who takes the single-call contract literally and adds an assert will break the SSR re-application path.

**Suggested fix:**

Rewrite the contract comment to describe actual behavior, e.g.:

```
// Caches configs for re-application after SSR re-retrieves the graph handle.
// Each call overwrites runtime_graph_configs_; RecoverFromSSR always re-applies
// the most recently cached value. Normal usage calls this once at session creation,
// but correctness does not depend on that.
```

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
```diff
@@ -669,6 +671,139 @@ TEST_F(QnnMockSSRBackendTests, DISABLED_SSRGraphExecuteEpContextWeightSharing) {
 }
 
 #endif  // defined(_WIN32) && (defined(_M_ARM64) || defined(_M_ARM64EC))
+
+// Test that ApplyRuntimeGraphConfigs is re-applied after an SSR event on the
+// context-binary path. The mock SSR fires on the first graphExecute, triggering
+// RecoverFromSSR which calls ApplyRuntimeGraphConfigs(runtime_graph_configs_).
+// With inputs/weights of 300.0f the Conv accumulator overflows fp16 max (~65504):
+// if the clamp config is not re-applied after recovery the output would be NaN.
+//
```
huaychou: **[N-2] (Minor)** fp16 Conv model builder is duplicated across two test files

`qnn_ep_context_test.cc` adds a static helper `BuildFp16ConvModel()` (returning `GetTestModelFn`), while `qnn_ssr_test.cc` adds a functionally identical lambda `build_fp16_conv` using the same input shape (`{1,2,4,4}`), weight shape (`{2,2,2,2}`), fill value (`300.0f`), and Conv attributes.

**Impact:** Future changes to the model definition (shape, fill value) must be applied in two places, risking inconsistency.

**Suggested fix:**

Move `BuildFp16ConvModel` (or an equivalent helper) into `qnn_test_utils.h` or a shared test header so both files can reuse it. If moving is too costly, at minimum add a comment at both sites noting they must stay in sync.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -3453,6 +3454,129 @@ TEST_F(QnnHTPBackendTests, QnnContextGenHtpBackendNoGpuConfig) {
...
+    std::vector<float> input_value(32, 300.0f);
+    std::vector<int64_t> input_dim{1, 2, 4, 4};
+    Ort::MemoryInfo info("Cpu", OrtDeviceAllocator, 0, OrtMemTypeDefault);
+    std::vector<Ort::Value> ort_inputs;
+    ort_inputs.push_back(Ort::Value::CreateTensor(info, input_value.data(), input_value.size(),
+                                                  input_dim.data(), input_dim.size()));
+    const char* input_names_c[] = {"input"};
+    const char* output_names_c[] = {"output"};
+
+    auto ort_outputs = scoped.session().Run(Ort::RunOptions{}, input_names_c, ort_inputs.data(), ort_inputs.size(),
```
huaychou: **[T-1] (Nit)** Context file not cleaned up if Phase 1 ASSERT fires early in the EPContext test

`CleanUpCtxFile` is called only at the end of the test. If any Phase 1 `ASSERT_TRUE` fires early, the context file will be left on disk. This is consistent with the cleanup pattern in other context cache tests in this file; impact is minor.

**Suggested fix:**

Consider adding a RAII cleanup guard (e.g., `absl::Cleanup`) to ensure file removal on any exit path. Alternatively, keep the current pattern to stay consistent with other context cache tests in the file.

minfhong-qti (approved):

onnxruntime/test/providers/qnn/ssr/qnn_ssr_test.cc
```diff
@@ -669,6 +669,116 @@ TEST_F(QnnMockSSRBackendTests, DISABLED_SSRGraphExecuteEpContextWeightSharing) {
...
+// Gated on QNN_TEST_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE (QNN API >= 2.38 / QAIRT >= 2.49)
+// and HTP arch > V75 since fp16 overflow→NaN only manifests on v79+.
+#if defined(_WIN32) && (defined(_M_ARM64) || defined(_M_ARM64EC)) && \
```
qti-chuteng, line 681: Is that fitable to use QNN_ARCH_ARM64?

huaychou: Yes. However, I think there are lots of unit tests need to be updated with `QNN_ARCH_ARM64`, we should create another ticket to cleanup them.

qti-chuteng (approved):

PR #722, opened 2026-08-13
[ORT-QNNEP-LORA] Add Support for Lora in ORT-QNNEP flow

qti-akshjame:
- Use the tensor names specified in lora_tensor_names.txt
  to mark the corresponding tensors as updatable in the
  QNN graph created after the ORT-QNN EP flow
- Modified the QNN EP workflow to use the original ONNX
  output names instead of ORT's internally modified names,
  eliminating tensor name mismatch issues
- Modified the QNN EP workflow to skip the QnnNodeGroup fusions
  like HardSigmoid+Mul, Scale+Softmax, Gelu
- Disabled all ORT level optimization/transformation### Description


qti-akshjame:
Closing in favor of new PR rebased on AISW-185935 branch with only related changes.

PR #723, opened 2026-08-13
[QNNEP-LORA] Add support for Lora in ORT-QNNEP-LORA flow

qti-akshjame:
- Use the tensor names specified in lora_tensor_names.txt
  to mark the corresponding tensors as updatable in the
  QNN graph created after the ORT-QNN EP flow
- Modified the QNN EP workflow to use the original ONNX
  output names instead of ORT's internally modified names,
  eliminating tensor name mismatch issues
- Modified the QNN EP workflow to skip the QnnNodeGroup fusions
  like HardSigmoid+Mul, Scale+Softmax, Gelu
- Disabled all ORT level optimization/transformation

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
@@ -273,7 +273,9 @@ Status BaseOpBuilder::ProcessOutputs(QnnModelWrapper& qnn_model_wrapper,
   }
   const auto output_count = GetOutputCountQnnRequired(node_unit);
   for (size_t output_i = 0; output_i < output_count; ++output_i) {
-    const auto& output_name = outputs[output_i].node_arg.Name();
+    // Use original ONNX tensor name — ORT may have stripped _output/_output_N suffix
+    const auto& output_name = qnn_model_wrapper.GetOriginalOnnxName(
+        outputs[output_i].node_arg.Name());
 
     TensorInfo output_info = {};
     ORT_RETURN_IF_ERROR(qnn_model_wrapper.GetTensorInfo(outputs[output_i], output_info));
@@ -329,7 +331,7 @@ Status BaseOpBuilder::ProcessOutputs(QnnModelWrapper& qnn_model_wrapper,
       output_info.qnn_data_type = supported_qnn_data_type;
       output_names.push_back(output_name);
     }
-    Qnn_TensorType_t tensor_type = is_graph_output ? QNN_TENSOR_TYPE_APP_READ : QNN_TENSOR_TYPE_NATIVE;
+    Qnn_TensorType_t tensor_type = qnn_model_wrapper.GetTensorType(output_name);
     QnnTensorWrapper output_tensorwrapper(output_name,
                                           tensor_type,
                                           output_info.qnn_data_type,
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -4,9 +4,11 @@
 #include "qnn_execution_provider.h"
 
 #include <filesystem>
+#include <fstream>
 #include <optional>
 #include <string_view>
 #include <thread>
+#include "onnx/onnx_pb.h"
 #include <unordered_set>
 
 #include "IR/QnnIrGraph.h"
@@ -380,6 +382,26 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
   }
 
   std::unique_ptr<qnn::QnnSerializerConfig> qnn_serializer_config = ParseSerializerBackendOptions(provider_options_map);
+
+  // Parse lora_tensor_names_path — used to mark tensors as UPDATEABLE in the QNN graph
+  static const std::string LORA_TENSOR_NAMES_PATH = "lora_tensor_names_path";
+  auto lora_names_path_pos = provider_options_map.find(LORA_TENSOR_NAMES_PATH);
+  if (lora_names_path_pos != provider_options_map.end()) {
+    std::ifstream lora_file(lora_names_path_pos->second);
+    std::string line;
+    while (std::getline(lora_file, line)) {
+      if (!line.empty()) lora_updatable_tensors_.insert(line);
+    }
+    LOGS_DEFAULT(INFO) << "Loaded " << lora_updatable_tensors_.size()
+                       << " lora updatable tensor names from " << lora_names_path_pos->second;
+  }
+
+  // Parse qnn_direct_dlc_mode — serialize DLC directly via IrGraphBuilder, skip EPContext
+  static const std::string QNN_DIRECT_DLC_MODE = "qnn_direct_dlc_mode";
+  direct_dlc_mode_ = ParseBoolOption(QNN_DIRECT_DLC_MODE, false, provider_options_map);
+  if (direct_dlc_mode_) {
+    LOGS_DEFAULT(INFO) << "direct_dlc_mode enabled: DLC will be written directly via IrGraphBuilder";
+  }
 
   std::string profiling_file_path;
   static const std::string PROFILING_LEVEL = "profiling_level";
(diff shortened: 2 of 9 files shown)


qti-chuteng:
This PR doesn't follow the policy of the development, could you re-create a new PR which target to merge into main branch?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1166,6 +1215,11 @@ QNNExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_viewer
 }
 
 DataLayout QNNExecutionProvider::GetPreferredLayout() const {
+  // In direct_dlc_mode, skip layout transforms to preserve original ONNX tensor names.
+  // This ensures QNN graph tensor names match lora_tensor_names.txt exactly.
+  if (direct_dlc_mode_) {
+    return DataLayout::NCHW;
```
kuanyul-qti: **[C-1] (Critical)** Direct DLC mode disables NHWC layout conversion while existing op builders still interpret tensor shapes as NHWC

`GetPreferredLayout()` returns `NCHW` when `direct_dlc_mode_` is enabled, so the second capability/compile pass no longer converts layout-sensitive nodes to the internal NHWC domain. Existing builders explicitly assume that compile-time tensors are already NHWC; for example, `ConvOpBuilder::GetInputChannelNumber()` reads `input_shape.back()` as the channel dimension and documents that assumption.

**Impact:** For an NCHW Conv, the builder treats W as C, corrupting depthwise/group detection, output-channel calculations, and QNN tensor shapes/parameters. Other layout-sensitive builders that rely on the QNN NHWC contract can likewise produce a DLC that fails finalization or, worse, executes with incorrect results.

**Suggested fix:**

Do not globally disable layout conversion through `GetPreferredLayout()`. Preserve the QNN EP NHWC compile contract and handle original ONNX names in serializer/name metadata. If direct mode truly requires an NCHW QNN graph, update every layout-sensitive builder and add Conv/Pool/Resize regression coverage first.

onnxruntime/core/providers/qnn/builder/opbuilder/base_op_builder.cc
```diff
@@ -273,7 +273,9 @@ Status BaseOpBuilder::ProcessOutputs(QnnModelWrapper& qnn_model_wrapper,
   }
   const auto output_count = GetOutputCountQnnRequired(node_unit);
   for (size_t output_i = 0; output_i < output_count; ++output_i) {
-    const auto& output_name = outputs[output_i].node_arg.Name();
+    // Use original ONNX tensor name — ORT may have stripped _output/_output_N suffix
+    const auto& output_name = qnn_model_wrapper.GetOriginalOnnxName(
```
kuanyul-qti: **[C-2] (Critical)** Renaming only producer outputs disconnects producer/consumer edges and stops graph outputs from being APP_READ

This changes the output wrapper and QNN node output to the original ONNX name, but `ProcessInputs()`/`MakeTensorWrapper()` still create consumer inputs from the filtered graph's ORT `NodeArg` name. When a mapping matches, the producer writes the original name while the consumer reads the ORT name, so the model wrapper treats them as different tensors. The same mismatch affects graph I/O: `model_output_index_map_` remains keyed by the fused node's ORT names, so `IsGraphOutput(output_name)`/`GetTensorType(output_name)` classify a renamed output as NATIVE instead of APP_READ.

**Impact:** Internal edges become consumer tensors with no matching producer; graph outputs are not added to `model_output_names_`, and `SetupTensors()` cannot find a renamed name in `outputs_info_`. Any partition that actually exercises the rename can fail during compose/setup instead of producing a usable DLC.

**Suggested fix:**

Introduce one canonical-name/alias mechanism used consistently by inputs, outputs, initializers, graph I/O index maps, and runtime lookup; do not apply local substitutions in only two output builders. Add tests with two connected nodes and a renamed graph output to verify both the QNN edge and APP_READ registration.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -960,6 +982,33 @@ QNNExecutionProvider::GetCapability(const onnxruntime::GraphViewer& graph_viewer
...
+      if (model_file) {
+        ONNX_NAMESPACE::ModelProto model_proto;
+        if (model_proto.ParseFromIstream(&model_file)) {
+          for (const auto& node : model_proto.graph().node()) {
+            if (node.name().empty()) continue;
+            for (int i = 0; i < node.output_size(); i++) {
+              const std::string& onnx_out = node.output(i);
+              if (!onnx_out.empty() && onnx_out != node.name()) {
+                // node.name() is what ORT uses; onnx_out is the original
+                onnx_tensor_name_map_[node.name()] = onnx_out;
```
kuanyul-qti: **[M-1] (Major)** The tensor-name map uses node names as keys and cannot reliably represent ONNX value names

ONNX `NodeProto.name` is an optional node identifier, not an output tensor name, and is not guaranteed to be unique. This loop writes every output to the same `onnx_tensor_name_map_[node.name()]` entry: multi-output nodes retain only the last output, unnamed nodes are skipped, and duplicate node names collide. It also reopens `GraphViewer::ModelPath()`, which provides no source for memory-loaded models and may not represent the optimized/filtered graph currently being compiled.

**Impact:** The map can be missing or point to the wrong tensor, so original value names from the LoRA list cannot be mapped reliably to the compile graph. Memory-loaded models silently get an empty map and therefore fail to mark the intended tensors updateable.

**Suggested fix:**

Build value-name aliases from the current `GraphViewer`/`NodeArg` plus explicit ORT rename metadata. The key must be the actual ORT tensor name and the value its corresponding original ONNX value name, handled one output at a time. Do not reparse the model file.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -381,6 +383,26 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
 
   std::unique_ptr<qnn::QnnSerializerConfig> qnn_serializer_config = ParseSerializerBackendOptions(provider_options_map);
 
+  // Parse lora_tensor_names_path — used to mark tensors as UPDATEABLE in the QNN graph
+  static const std::string LORA_TENSOR_NAMES_PATH = "lora_tensor_names_path";
+  auto lora_names_path_pos = provider_options_map.find(LORA_TENSOR_NAMES_PATH);
+  if (lora_names_path_pos != provider_options_map.end()) {
+    std::ifstream lora_file(lora_names_path_pos->second);
```
kuanyul-qti: **[M-2] (Major)** An unreadable LoRA tensor list is reported as successfully loaded with an empty set

The code does not check `is_open()`/`fail()` after constructing the `std::ifstream`. A missing path, permission error, or read failure simply ends the loop and still logs `Loaded 0 ... from <path>`. Lines are also not normalized for CRLF or surrounding whitespace, leaving `\r` or spaces in names that will never match.

**Impact:** Users can receive an apparently successful DLC while every LoRA tensor remains STATIC/NATIVE, making runtime adapter updates ineffective with no configuration error in the log.

**Suggested fix:**

When the provider option is supplied, fail with `ORT_THROW` including the path on open/read errors. Trim CRLF and surrounding whitespace, ignore whitespace-only lines, and either reject an empty file or emit an explicit warning.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -381,6 +383,26 @@ QNNExecutionProvider::QNNExecutionProvider(const ProviderOptions& provider_optio
...
+    while (std::getline(lora_file, line)) {
+      if (!line.empty()) lora_updatable_tensors_.insert(line);
+    }
+    LOGS_DEFAULT(INFO) << "Loaded " << lora_updatable_tensors_.size()
+                       << " lora updatable tensor names from " << lora_names_path_pos->second;
+  }
+
+  // Parse qnn_direct_dlc_mode — serialize DLC directly via IrGraphBuilder, skip EPContext
+  static const std::string QNN_DIRECT_DLC_MODE = "qnn_direct_dlc_mode";
+  direct_dlc_mode_ = ParseBoolOption(QNN_DIRECT_DLC_MODE, false, provider_options_map);
```
kuanyul-qti: **[M-3] (Major)** Direct DLC mode does not require an IR serializer and can skip both DLC and EPContext output

`direct_dlc_mode_` can be enabled independently, but a serializer is only created when `dump_qnn_ir_dlc=1` or a saver path is provided. If a user supplies only the new `qnn_direct_dlc_mode=1` option, even with context caching enabled, `CompileFromOrtGraph()` has no serializer configuration to write a DLC while `Compile()` skips EPContext generation because direct mode is set.

**Impact:** The log says that a DLC will be written and session compilation may succeed, but no DLC appears and the normal EPContext artifact is also suppressed. This is a silent configuration failure.

**Suggested fix:**

After parsing provider options, require direct mode to have an IR serializer, not Saver, and a valid output directory; reject incompatible combinations. Alternatively, remove the duplicate option and let the existing `dump_qnn_ir_dlc` flow select direct serialization.

PR #729, opened 2026-08-13
[QNN EP] Sign-extend sub-byte weights when pre-dequantizing a per-channel integer-op weight

qti-yuduo:
## Summary

Defensive hardening of `PreDequantizePerChannelWeight()`:

- Pass the ONNX weight type, sign-extend unpacked INT2/INT4 values, preserve 8-bit and unsigned behavior, and reject unsupported wider types.
- Add seven focused helper tests and share the existing mock-wrapper fixture.

## Review status

This is not needed for reachable production behavior. ONNX restricts `MatMulInteger` and `ConvInteger` inputs to INT8/UINT8, ORT rejects an INT4 graph as `INVALID_GRAPH`, and both QNN fusion gates reject non-8-bit weights before this helper. The Qwen W4A16 issue is already fixed by #728.

**Recommendation: close without merging.** The patch only hardens direct/helper invocation of an otherwise unreachable sub-byte path.

## Validation

- Coverage-symbol provider tests: 11/11 passed on `tetra18-linux` (7 new tests plus 4 fixture consumers).
- Qwen3-0.6B legacy converter vs EP: all jobs passed; fresh PSNR metrics exactly match the prior accepted run; EP PPL 208.7230 vs baseline 223.9822.
- Standard CI passed. The coverage job compiled the new test, then failed during wheel packaging with `No space left on device`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_node_group/dq_conv_integer_fusion.cc
@@ -503,7 +503,7 @@ Ort::Status DQConvIntegerFusion::CreateOrValidateOnQnn(QnnModelWrapper& qmw, boo
   } else {
     std::vector<uint8_t> float_bytes;
     RETURN_IF_ERROR(PreDequantizePerChannelWeight(qmw, *b_scale_iodef_, b_zp_iodef,
-                                                  is_signed_weight,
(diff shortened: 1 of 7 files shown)


qti-yuduo:
valid MatMulInteger/ConvInteger graphs only allow INT8/UINT8, and QNN rejects sub-byte weights before reaching it. So #729 was defensive hardening based on a hypothetical path, not a required Qwen fix

PR #733, opened 2026-08-14
[QNN EP] Fix FP16 Reciprocal divisor serialization

qti-yuduo:
## Summary
- Serialize the static `1.0` divisor emitted for `Reciprocal` using the matching QNN tensor element type.
- Add QNN HTP FP16 `Reciprocal` coverage.

## Root Cause
For FP16 input, the divisor is declared as `QNN_DATATYPE_FLOAT_16`, so QNN expects a one-element static buffer of 2 bytes. The builder always allocated and copied an FP32 `float`, producing a 4-byte client buffer and the reported static-tensor data-length mismatch.

The fix serializes `Ort::Float16_t(1.0f)` for FP16 and the equivalent BF16 scalar for the analogous two-byte type. The FP32 and quantized paths are unchanged.

### DIFF
--- a/onnxruntime/test/providers/qnn/simple_op_test.cc
+++ b/onnxruntime/test/providers/qnn/simple_op_test.cc
@@ -1274,6 +1274,14 @@ TEST_F(QnnHTPBackendTests, Reciprocal_Basic_FLOAT) {
                    ExpectedEPNodeAssignment::All);
 }
 
+TEST_F(QnnHTPBackendTests, Reciprocal_Basic_FP16) {
+  RunFP16OpTest("Reciprocal",
+                {TestInputDef<float>({2, 2}, false, {1.0f, 2.0f, 0.5f, 4.0f})},
+                {},  // No attributes
+                13,
+                ExpectedEPNodeAssignment::All);
+}
+
 TEST_F(QnnHTPBackendTests, Reciprocal_QU8) {
   RunQDQOpTest<uint8_t>("Reciprocal",
                         {TestInputDef<float>({2, 2}, false, GetFloatDataInRange(1.0f, 5.0f, 4))},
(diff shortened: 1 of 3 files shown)


onnxruntime/test/providers/qnn/simple_op_test.cc
```diff
@@ -1274,6 +1274,14 @@ TEST_F(QnnHTPBackendTests, Reciprocal_Basic_FLOAT) {
                    ExpectedEPNodeAssignment::All);
 }
 
+TEST_F(QnnHTPBackendTests, Reciprocal_Basic_FP16) {
```
yath1, line 1277: Minor: Not a blocker - If there's a way to add a BF16 test, please add that as well

yath1 (approved):
lgtm.

onnxruntime/test/providers/qnn/simple_op_test.cc
qti-yuduo: Added `DISABLED_BF16_Reciprocal` in `onnxruntime/test/providers/qnn/bf16_test.cc` (dbeee05), following the convention of the 15 existing BF16 tests there -- all are `DISABLED_` because BF16 mode requires `soc_model >= 88` (SM8850), which no CI runner or dev box has yet.

Two things worth flagging that I verified while doing this:

1. **The test does cover the divisor tensor, but not the new `QNN_DATATYPE_BFLOAT_16` branch.** Ran it locally with `--gtest_also_run_disabled_tests` on a V73 box; it gets as far as graph composition and fails on `reciprocal_node_divisor_bf16_intermediate` (`CreateQnnInputOutputTensors`, error 1000) -- i.e. the divisor is built as FP32 and then a `Cast` to BF16 is inserted. That is expected: with `htp_bf16_enable`, op builders always see `QNN_DATATYPE_FLOAT_32`, and BF16 conversion is a post-pass (`ProcessBF16InputConversion`, `qnn_model_wrapper.cc:320-337`) that adds Cast ops for builder-created STATIC tensors.

2. **The `QNN_DATATYPE_BFLOAT_16` branch in `reciprocal_op_builder.cc` is currently unreachable.** `input_info.qnn_data_type` comes from `GetTensorInfo` -> `GetQnnDataType`, and the ONNX->QNN type maps in `qnn_utils.cc` (`CreateMap` / `CreateMapQuantize`) have no `ONNX_TENSOR_ELEMENT_DATA_TYPE_BFLOAT16` entry, so a bfloat16 ONNX model cannot reach the builder at all. I kept the branch for parity with `selu_op_builder.cc:84`, which has the identical (also unreachable) branch -- happy to drop it if you'd rather not carry dead code.

Existing `Reciprocal_Basic_FLOAT` / `_FP16` / `_QU8` all still pass locally on V73.

yath1: On your second point, I think its okay to keep this part of the code (even though its not actively tested) since we eventually want to support bf16 models.

PR #734, opened 2026-08-14
Ensure unique QNN graph node names

qti-yuduo:
## Summary
- QNN graph node names must be unique across the fully composed graph. `UniqueNameGenerator` only counts repeated base-plus-suffix requests; it cannot reserve names requested later or raw ONNX names.
- Reserve each emitted QNN node name centrally in `QnnModelWrapper`; validation remains non-mutating. BF16-generated Cast operations use the same allocator.
- Use the established `_<N>` suffix convention and cover both the real collision shape and collisions with literal suffixed names.

Tracks https://github.com/qcom-ai-hub/tetracode/issues/20855.

## Validation
- Built the touched ARM64 QNN provider and provider-test translation units.
- Replayed the original ZeoDepth production ONNX: the unpatched provider fails with `EP_FAIL: Failed to compose Qnn graph` / QNN `6005` and emits no DLC; the patched provider emits one `1,907,062,452`-byte DLC.
- The focused wrapper test verifies `duplicate`, `duplicate_2`, and a later literal `duplicate_2` resolving to `duplicate_2_2`.
- `git diff --check` passed.

The full `onnxruntime_provider_test` target cannot complete in this local prebuilt-ORT configuration because the supplied headers omit `cpu_provider_factory.h`. The repository lint hook is also blocked while installing pinned `lintrunner==0.12.7` because the published package metadata is rejected by the current `uv` parser.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
@@ -306,7 +306,9 @@ bool QnnModelWrapper::ProcessBF16InputConversion(const std::string& qnn_node_nam
                     ORT_LOGGING_LEVEL_VERBOSE,
                     ("BF16: Adding Cast op " + input_name + " -> " + cast_output_name).c_str());
 
-        QnnOpProperty cast_op(cast_output_name, QNN_OP_PACKAGE_NAME_QTI_AISW, QNN_OP_CAST,
+        QnnOpProperty cast_op(MakeUniqueQnnNodeName(cast_output_name),
+                              QNN_OP_PACKAGE_NAME_QTI_AISW,
+                              QNN_OP_CAST,
                               std::vector<std::string>{input_name},
                               std::vector<std::string>{cast_output_name},
                               std::vector<std::string>{});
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -602,6 +607,21 @@ bool QnnModelWrapper::ProcessBF16Conversions(std::vector<QnnOpProperty>& final_o
   return true;
 }
 
+std::string QnnModelWrapper::MakeUniqueQnnNodeName(const std::string& requested_name) {
+  if (qnn_node_names_.insert(requested_name).second) {
+    return requested_name;
+  }
+
+  for (size_t suffix = 2;; ++suffix) {
+    std::string unique_name = requested_name + "_qnn_ep_" + std::to_string(suffix);
```
yath1: I think an exit with failure for a sufficiently large value for suffix would help instead of an unbounded for loop

yath1: Why add `_qnn_ep_` token here? First entry would not have this extra string added but subsequent ones would have them.

yath1: Would using a unordered_map and storing the counter help avoid the loop though?

qti-yuduo: Bounded the search to suffixes `[2, qnn_node_names_.size() + 1]` and added an explicit failure if that range is exhausted. With `M` reserved names, `requested_name` occupies one entry, so at most `M - 1` of those `M` candidates can already be occupied.

A counter map alone is not sufficient: generating `duplicate_2` for a repeated `duplicate` can collide with a later literal `duplicate_2`. The unit test now covers that sequence and expects the literal request to resolve to `duplicate_2_2`. A map plus the reservation set could cache the next probe, but adds state for a non-hot path without eliminating collision checks.

qti-yuduo: Removed `_qnn_ep_`. Duplicate names now use the existing `_<N>` convention.

yath1 (approved):
lgtm.

PR #737, opened 2026-08-15
[QNN EP] Fix Linux x86_64 wheel packaging failure due to stale patchelf

qti-mbadnara:
### Motivation and Context:

The Linux `x86_64 Ubuntu 22.04` CI build was failing during the Python wheel packaging step with:

  `ValueError: patchelf 0.14.3 found. auditwheel repair requires patchelf >= 0.14.5`

`auditwheel` (unpinned, resolved to 6.8.1) raised its minimum `patchelf` version floor, but the Docker image's system
`patchelf` (installed via apt from Ubuntu 22.04's package index) is fixed at `0.14.3` and cannot be updated without
rebuilding the image.

The fix moves ownership of patchelf from the Docker image (apt) to the Python virtual environment (pip). The patchelf pip package bundles a current patchelf binary that is installed into the venv's bin/ at build time, shadowing the stale system binary when auditwheel repair runs inside the activated venv. This also makes patchelf version management consistent with auditwheel both resolved by uv at build time from the same requirements.txt.

### DIFF
--- a/tools/ci_build/github/linux/python/requirements.txt
+++ b/tools/ci_build/github/linux/python/requirements.txt
@@ -5,6 +5,7 @@ pytest
 setuptools>=68.2.2
 wheel
 auditwheel
+patchelf
 protobuf==6.33.0
 sympy==1.14
 flatbuffers
--- a/qcom/scripts/linux/ubuntu_22_04/Dockerfile
+++ b/qcom/scripts/linux/ubuntu_22_04/Dockerfile
@@ -18,7 +18,6 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     libc++1 \
     libc++abi1 \
-    patchelf \
     && add-apt-repository ppa:deadsnakes/ppa \
     && apt-get update \
     && apt-get install -y \


huaychou (approved):
LGTM. Just wondering, should we pin the auditwheel and patchelf versions?

qti-mbadnara:
> LGTM. Just wondering, should we pin the auditwheel and patchelf versions?

I contemplated the same but it might introduce a maintenance burden to update the pins down the line

tirupath-qti (approved):

tools/ci_build/github/linux/python/requirements.txt
```diff
@@ -5,6 +5,7 @@ pytest
 setuptools>=68.2.2
 wheel
 auditwheel
+patchelf
```
huaychou: **[N-1] (Minor)** Newly added patchelf package is not version-pinned

The root cause of this PR is exactly that `auditwheel` was left unpinned and drifted to 6.8.1, which raised its minimum `patchelf` requirement from 0.14.3 to 0.14.5 and broke packaging. The newly added `patchelf` entry is also unpinned — the same pattern that caused this failure. Per **[P29]**, CI packaging tools such as `auditwheel` and `patchelf` must be version-pinned in `requirements.txt`, with an inline comment naming the ceiling and why. Without a pin, the next `patchelf` pip release can cause the same version-drift failure.

**Impact:** A future `patchelf` pip release that introduces a breaking change or raises its own dependency floor can silently break the Linux x86_64 packaging CI again, which is hard to diagnose.

**Suggested fix:**

Pin `patchelf` to the currently known-working version and add an inline comment. It is also recommended to pin `auditwheel` (known-working: 6.8.1) at the same time to eliminate version-drift risk entirely:

tools/ci_build/github/linux/python/requirements.txt
```diff
@@ -5,6 +5,7 @@ pytest
 setuptools>=68.2.2
 wheel
 auditwheel
```
huaychou: **[T-1] (Nit)** auditwheel is still unpinned (pre-existing; good time to fix)

`auditwheel` is the direct root cause of this PR (unpinned → resolved to 6.8.1), but the fix does not pin it. This is a pre-existing gap; since the PR description already identifies `auditwheel 6.8.1` as a known-working version, this is a good time to pin it together with `patchelf`.

**Suggested fix:**

Address together with N-1: change `auditwheel` to `auditwheel==6.8.1` with an explanatory comment.

PR #741, opened 2026-08-17
[QNN EP] Handle broadcast Gemm C with FC and Add

qti-yuduo:
## Summary
- Lower valid Gemm C broadcasts that cannot be represented as a QNN FullyConnected bias to FullyConnected + ElementWiseAdd.
- Share the FC-compatible bias-shape rule and keep the absorbed-Reshape selector aligned with the builder.
- Add direct, MatMulAddFusion, and selector regression coverage.

## Root cause
ORT MatMulAddFusion can produce Gemm with C=[M,1] from MatMul followed by a column-broadcast Add. ONNX permits C to broadcast to [M,N], but QNN FullyConnected can consume only one bias value per output channel. Validation rejected C before the existing FC + Add lowering could run.

The fix keeps scalar C for N=1, [N], and [1,N] on the direct FC path. Other valid broadcasts use a separate Add; block-quantized Gemm remains strict.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
@@ -41,6 +41,31 @@ bool IsBQGemmWeight(const QnnModelWrapper& qnn_model_wrapper, const OrtNodeUnitI
   return bq::IsBQScale(scale_shape, weight_shape, block_axis);
 }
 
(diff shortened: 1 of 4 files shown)


quic-muchhsu (approved):

qti-ashwshan (approved):
LGTM! we are correctly gating ExplictOpCheck to let decomposable bias shapes flow through to the existing FC+Add path instead of rejecting them outright.

PR #743, opened 2026-08-18
[QNN EP] Fix flaky ScatterElementsSharedNegativeIndicesDifferentAxes test

huaychou:
### Description

- Fix by using distinct indices {-1,-2,-3} so each update lands on a unique cell for both axes, making expected outputs deterministic. The _qnn_idx per-axis non-aliasing path the test was designed to cover is still exercised since the two axes rewrite the shared initializer to different byte sequences.
- Remove DISABLED_ prefix and update comment accordingly.

### Motivation and Context

- The test used indices {-1,-1,-1} with axis=0, causing all three updates to collide at the same destination cell (row 2). ONNX ScatterElements with reduction='none' and duplicate indices is order-undefined; CPU reference EP picks last-write-wins (30) while HTP's write order varies, producing 10/20/30 intermittently.

### DIFF
--- a/onnxruntime/test/providers/qnn/scatterelements_test.cc
+++ b/onnxruntime/test/providers/qnn/scatterelements_test.cc
@@ -189,11 +189,12 @@ TEST_F(QnnHTPBackendTests, ScatterElementsSharedNegativeIndicesInitializer) {
                   EPVerificationParams{ExpectedEPNodeAssignment::All});
 }
 
-// Different axis bounds -- indices `[-1]` resolves to `kRows-1` for scatterA
-// and `kCols-1` for scatterB. The `_qnn_idx` rename prevents the two
-// per-axis rewrites from aliasing.
-// Disabling this test as it is flaky with QAIRT 2.49.40
-TEST_F(QnnHTPBackendTests, DISABLED_ScatterElementsSharedNegativeIndicesDifferentAxes) {
+// Different axis bounds -- the shared indices initializer {-1,-2,-3} rewrites to
+// rows {2,1,0} for scatterA (axis=0, dim=kRows) and cols {6,5,4} for scatterB
+// (axis=1, dim=kCols). Distinct per-row destinations keep the result deterministic
+// (no duplicate-index collision); the `_qnn_idx` rename prevents the two per-axis
+// rewrites from aliasing.
+TEST_F(QnnHTPBackendTests, ScatterElementsSharedNegativeIndicesDifferentAxes) {
   constexpr int64_t kRows = 3;
   constexpr int64_t kCols = 7;  // != kRows so rewritten bytes differ.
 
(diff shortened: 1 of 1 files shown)


onnxruntime/test/providers/qnn/scatterelements_test.cc
```diff
@@ -203,7 +204,7 @@ TEST_F(QnnHTPBackendTests, DISABLED_ScatterElementsSharedNegativeIndicesDifferen
     builder.MakeInput<int32_t>("dataA", {kRows, kCols}, data_a);
     builder.MakeInput<int32_t>("dataB", {kRows, kCols}, data_b);
 
-    std::vector<int64_t> indices = {-1, -1, -1};
+    std::vector<int64_t> indices = {-1, -2, -3};  // distinct: no duplicate-index collision
```
huaychou: **[N-1] (Minor)** Kill-test fence verification not documented in PR description

The PR fixes the flakiness root cause (duplicate-index write ordering is non-deterministic) and re-enables the test. Per the `[G02]` kill-test fence convention: when re-enabling a correctness test, the PR description should record the observation that "temporarily reverting the fix causes the test to actually fail (CI red)", proving the fence is meaningful.

The current PR description explains why the test was disabled and how the fix works, but does not explicitly record a kill-test observation (e.g., confirming that `{-1, -1, -1}` produces a failure or non-deterministic result).

**Impact:** The effectiveness of the fence cannot be confirmed from PR history. If someone later needs to verify this test's validity, they must re-run the kill-test themselves.

**Suggested fix:**

Add a line in the PR description (or commit message) such as: "Kill-test: reverting indices to {-1,-1,-1} caused the test to fail non-deterministically in M of N runs." No code change is needed.

onnxruntime/test/providers/qnn/scatterelements_test.cc
```diff
@@ -189,11 +189,12 @@ TEST_F(QnnHTPBackendTests, ScatterElementsSharedNegativeIndicesInitializer) {
...
 
-// Different axis bounds -- indices `[-1]` resolves to `kRows-1` for scatterA
-// and `kCols-1` for scatterB. The `_qnn_idx` rename prevents the two
-// per-axis rewrites from aliasing.
-// Disabling this test as it is flaky with QAIRT 2.49.40
-TEST_F(QnnHTPBackendTests, DISABLED_ScatterElementsSharedNegativeIndicesDifferentAxes) {
+// Different axis bounds -- the shared indices initializer {-1,-2,-3} rewrites to
+// rows {2,1,0} for scatterA (axis=0, dim=kRows) and cols {6,5,4} for scatterB
+// (axis=1, dim=kCols). Distinct per-row destinations keep the result deterministic
+// (no duplicate-index collision); the `_qnn_idx` rename prevents the two per-axis
```
huaychou: **[T-1] (Nit)** Comment phrase "per-row destinations" is slightly ambiguous for scatterB (axis=1)

Lines 192–196 state: "Distinct per-row destinations keep the result deterministic" — technically accurate (each row of the indices tensor targets a distinct destination), but in the context of `scatterB` (`axis=1`) the destinations are columns (`{6,5,4}`). Using "per-row" to describe a column-axis destination is slightly misleading; readers may interpret it as "different rows write to different rows" rather than "each index tensor row targets a distinct column".

**Suggested fix:**

Optionally rephrase to "Distinct scatter destinations (per-element of the indices tensor) keep the result deterministic", or clarify with axis context: "per-row-of-indices, distinct across the target axis". This is an optional polish — no behavior change.

onnxruntime/test/providers/qnn/scatterelements_test.cc
qti-mbadnara, line 207: Thanks for the fix @huaychou

qti-mbadnara (approved):
LGTM

qti-chuteng (approved):

PR #744, opened 2026-08-19
[QNN EP] Fix cmake build crashes in static_lib mode and ort_core prebuilt path

qti-mbadnara:
## Description

Fixes two cmake defects that crash the build when using `--use_qnn static_lib` via the ort_core prebuilt path: an unguarded `cmake -E copy` on an empty `QNN_LIB_FILES` list, and a missing `--skip_submodule_sync` flag that causes `build.py` to run `git submodule sync` on a non-git source tree.

---

## Motivation & Context

Both issues manifest only in the `onnxruntime_prebuilt.cmake` ExternalProject build path and were worked around locally by commenting out code and manually patching the build command — fragile fixes invisible to CI.

**Issue 1 — empty `QNN_LIB_FILES` copy**

`QNN_LIB_FILES` is populated by a `file(GLOB ...)` inside an `if(MSVC OR Linux)` block in `microsoft/onnxruntime cmake/CMakeLists.txt`. On Android / static_lib builds that block is skipped, leaving the variable undefined (empty list).

In `cmake/onnxruntime_unittests.cmake` (upstream), the block that copies QNN libraries to the Java native-test directory is guarded only by `if(onnxruntime_USE_QNN)`. With `QNN_LIB_FILES` empty, CMake generates:

```
cmake -E copy <destination>
```

`cmake -E copy` requires at least one source argument before the destination. With zero sources it prints its usage banner and exits non-zero, crashing the build.

**Issue 2 — missing `--skip_submodule_sync`**

ort_core is fetched as a URL zip archive (`microsoft/onnxruntime v1.27.0`) — the extracted tree has no `.git` directory.
`build.py` unconditionally runs `git submodule sync --recursive` regardless, crashing with:

```
fatal: not a git repository (or any of the parent directories): .git
```

`build.py` exposes `--skip_submodule_sync` exactly for this scenario, but it was never appended to `ORT_BUILD_COMMAND` in this repo.

---

### DIFF
--- a/cmake/external/onnxruntime_prebuilt.cmake
+++ b/cmake/external/onnxruntime_prebuilt.cmake
@@ -83,6 +83,12 @@ else()
         --qnn_home "${onnxruntime_QNN_HOME}"
         --no_kleidiai
     )
+    # ort_core is fetched as a URL archive (zip), not a git clone, so there is no
+    # .git directory. Without this flag, build.py unconditionally runs
+    # "git submodule sync --recursive" and crashes with "fatal: not a git repository".
+    # NOTE: if ort_core is ever switched to GIT_REPOSITORY/GIT_TAG, remove this flag
+    # so that submodule sync runs correctly.
+    list(APPEND ORT_BUILD_COMMAND --skip_submodule_sync)
     if(onnxruntime_BUILD_CACHE)
         list(APPEND ORT_BUILD_COMMAND "--use_cache")
     endif()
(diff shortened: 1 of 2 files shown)


minfhong-qti:
I thought QNN EP  is always built as shared lib. Why `--use_qnn static_lib` is ever used?

qti-mbadnara:
> I thought QNN EP is always built as shared lib. Why `--use_qnn static_lib` is ever used?

It is used for Android builds.

qti-yuduo (approved):

PR #746, opened 2026-08-19
[QNN EP] 2.6.0 RC2 Changes

qti-mbadnara:
This PR reverts the following commit from the `mainline` as part of `2.6.0 RC2 Changes`

- #671


qti-yuduo:
Every revert should describe the reason root cause analysis and how to re-enable.

qti-mbadnara:
> Every revert should describe the reason root cause analysis and how to re-enable.

Yup creating an internal ticket for tracking. Also investigating if the fix can be made and the revert can be avoided

tirupath-qti (approved):

PR #749, opened 2026-08-20
[QNN EP] Use QNN API Version instead of QNN SDK version

qti-mbadnara:
### Description

This PR updates the `QNN SDK Version` based gating to `QNN API Version` based gating across the codebase.

### Motivation and Context:

Several feature gates in the QNN EP were keyed off `QNN_SDK_VERSION_MAJOR/QNN_SDK_VERSION_MINOR` a macro derived from `sdk.yaml`, instead of the QNN Core API version (`QNN_API_VERSION_MAJOR/MINOR` from `QnnCommon.h`). This was inconsistent with how every other version-gated feature in the codebase is handled.

This change replaces the `QNN_SDK_VERSION_*` checks with equivalent `QNN_API_VERSION_*` checks in the two production feature gates and their corresponding test, aligning all three with the `QAIRT 2.49 / QNN API 2.38` boundary confirmed against the actual SDK headers.


qti-ashwshan:
It would be better if we use the QAIRT SDK version instead of the QNN API version for simplicity. this would maintain the uniformity and also just using one unified QAIRT SDK version gate.

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
```diff
@@ -193,15 +193,13 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
     scale_info.shape = *squeezed_shape;
   }
 
-#if !defined(QNN_SDK_VERSION_MINOR) || (QNN_SDK_VERSION_MAJOR == 2 && QNN_SDK_VERSION_MINOR < 49)
-  // QNN SDK < 2.49 requires an explicit beta/bias input for QNN_OP_RMS_NORM on NPU.
-  // SDK 2.49+ accepts beta as optional, so the dummy tensor is only needed for older SDKs.
-  // Note: SDK 2.47 and 2.48 share the same QNN API version (2.36), so QNN_SDK_VERSION_MINOR
-  // derived from CMake is used here instead of QNN_API_VERSION_MINOR.
+#if !(QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 38))
```
huaychou: **[Minor] N-1** The comment that explained *why* the API-version threshold is correct was removed without replacement.

The deleted lines included:
```
// Note: SDK 2.47 and 2.48 share the same QNN API version (2.36), so QNN_SDK_VERSION_MINOR
// derived from CMake is used here instead of QNN_API_VERSION_MINOR.
```
This was the key rationale for the original design choice. The new guard using `QNN_API_VERSION_MINOR >= 38` is semantically correct (SDK 2.47 = API 2.36, SDK 2.48 = API 2.36, both < 2.38), but the reasoning is no longer documented in the code.

Suggestion: add a comment above the `#if` guard, e.g.:
```cpp
// QNN API 2.38 corresponds to QAIRT SDK 2.49. SDK 2.47 and 2.48 both shipped with
// QNN API 2.36 (< 2.38), so this guard covers them correctly without needing the
// CMake-injected QNN_SDK_VERSION_MINOR.
```

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -3302,9 +3302,8 @@ TEST_F(QnnHTPBackendTests, PrepareOnly_RunReturnsError) {
 // On SDK 2.48 the Graph Splittling config block is compiled out; the test still passes because
```
huaychou, line 3302: Nit: update the comment to QNN API version

huaychou (approved):

qti-mbadnara:
Closing this PR, will need to migrate the QNN SDK Version instead of QNN API Version

PR #752, opened 2026-08-23
[QNN EP] Support htp_reused_io_limit_mb provider option for HTP AOT context load

chunghow-qti:
### Description

* Adds a new `htp_reused_io_limit_mb` provider option for the HTP backend, wired through to
  `QNN_HTP_CONTEXT_CONFIG_OPTION_REUSED_IO_LIMIT` when loading a cached QNN context binary (`LoadCachedQnnContextFromBuffer`).
* Usage: set `ep.qnnexecutionprovider.reused_io_limit_mb`  value in MB. Default is `0`, which leaves the SDK's default (unbounded) behavior unchanged.
* This option allow client specify the effective I/O buffer size used by a context instead of QNN HTP default estimation. QNN HTP uses this value for memory estimation and DSP PD placement decisions when it loads a context binary.

### Motivation and Context

QNN HTP assumes no I/O reuse and estimates the context's memory using the total I/O size of all graphs in the context. If you actually share (reuse) one or more I/O buffers across multiple graphs in a context, or across multiple contexts, that default estimate can overshoot what the context really needs.

This overestimation matters because HTP loads each context into one of several process domains (PDs), each with a limited memory budget. An inflated estimate can cause a context to be placed in its own PD instead of sharing one with others (which is slower, since contexts on different PDs pay extra cost to talk to each other) — or it can cause the context to fail to load at all (QNN error 1002, "Failed to find available PD") even though it would actually fit. 

By explicitly specifying the reused I/O size, you let QNN HTP use a smaller, more accurate estimate instead.

### Test plan
- [x]  Context that doesn't fit in a PD → fits with the option
Take a context binary that fails to load by default (QNN error 1002, "Failed to find available PD"). Load it 
    - (a) without htp_reused_io_limit_mb → expect 1002; 
    - (b) with a known-safe value → expect load success and correct inference output.
- [x] Verify `htp_reused_io_limit_mb` is applied on both AOT context-load paths
    - `contextCreateFromBinary`: load four static-1024 contexts with
      `htp_reused_io_limit_mb=2`; confirm QNN verbose log reports
      `Setting Reused IO memory estimation to 2 MB` and all contexts load successfully.
    - `contextCreateFromBinaryListAsync`: run
      `QnnHTPBackendTests.HtpSharedResourceOptimization_HtpReusedIoLimitMb_LoadsSucceeds`;
      confirm QNN verbose log reports
      `Setting Reused IO memory estimation to 128 MB`.


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1353,6 +1330,28 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+  GetSessionConfigEntryOrDefault(ort_api,
+                                 session_options_,
+                                 FormatEPConfigKey("htp_reused_io_limit_mb"),
+                                 "0",
+                                 reused_io_limit_mb_str);
+  if (!reused_io_limit_mb_str.empty() && reused_io_limit_mb_str != "0") {
+    uint64_t parsed_reused_io_limit_mb = 0;
+    const char* begin = reused_io_limit_mb_str.data();
+    const char* end = begin + reused_io_limit_mb_str.size();
+    auto [ptr, ec] = std::from_chars(begin, end, parsed_reused_io_limit_mb);
```
minfhong-qti: Maybe worth defining a dedicated function for parsing int values.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1421,6 +1420,8 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     }
   }
 
+  qnn_backend_manager_->SetReusedIoLimitMb(reused_io_limit_mb_);
```
minfhong-qti: Prefer passing during `::Create`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1854,7 +1853,30 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
                     ORT_LOGGING_LEVEL_VERBOSE,
                     ("Max spill fill buffer size: " + std::to_string(max_spill_fill_size)).c_str());
 
-    const QnnContext_Config_t* context_configs[] = {&qnn_context_config, spill_fill_config_pointer, nullptr};
+    QnnContext_Config_t reused_io_limit_config = QNN_CONTEXT_CONFIG_INIT;
```
minfhong-qti: A refactor suggestion.
Since we need to write two places of below config setting everytime we have a new option, how about define a function in QnnBackendManager, accepting parameters (or possibly a struct) and returns vector of context configs? And this function can be reused by SSR recovery in QnnModel.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
minfhong-qti, line 2042: Define a `ContextRecoveryInfo` struct to collect all necessary configs instead of plainly record in QnnModel.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc

onnxruntime/core/providers/qnn/qnn_execution_provider.cc

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1353,6 +1354,13 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                    false,
                                                    logger_);
 
+  // Specifies the effective I/O buffer size used by a context across contextCreateFromBinary calls (AOT context load).
+  reused_io_limit_mb_ = ParseUInt64Option(ort_api,
```
kuanyul-qti: Do we need to update all the other integer config parsing to stay consistent?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
chunghow-qti: Done. Defined a `BuildContextConfigs` function for `LoadCachedQnnContextFromBuffer` and `RecoverFromSSR`

I also noticed that `contextCreateFromBinaryListAsync` can take `QNN_HTP_CONTEXT_CONFIG_OPTION_REUSED_IO_LIMIT`, so I updated that path as well.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
chunghow-qti: done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: Currently, seems there is no other uint64 config. we could create another PR as follow-up.
Such as `ParseUint32Option`

onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -218,6 +218,8 @@ class QnnEp : public OrtEp, public ApiPtrs {
   bool context_cache_enabled_ = false;
   bool share_ep_contexts_ = false;
   int htp_share_resource_optimization_ = -1;
+  // Caps the reused IO buffer size at context load; mitigates NPU memory exhaustion. 0 = SDK default.
+  uint64_t reused_io_limit_mb_ = 0;
```
minfhong-qti: No need to set as QnnEp's member now.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -346,6 +347,30 @@ static bool ParseBoolOption(const OrtApi& ort_api,
...
+  std::string value_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options, key, std::to_string(default_value), value_str);
+  if (value_str.empty()) {
+    return default_value;
+  }
+
+  uint64_t result = default_value;
+  const char* begin = value_str.data();
+  const char* end = begin + value_str.size();
+  auto [ptr, ec] = std::from_chars(begin, end, result);
```
minfhong-qti: Can we make this function a template? Then probably the output value should be in parameter instead of return value.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1285,76 +1284,71 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
...
-  // Set share resource optimization type to SEQUENTIAL_WITHOUT_VA_OPTIMIZATION
-  // This is controlled by the htp_share_resource_optimization option
-  QnnHtpContext_CustomConfig_t context_config_resource_sharing_opt_type;
-  context_config_resource_sharing_opt_type.option = QNN_HTP_CONTEXT_CONFIG_OPTION_SHARE_RESOURCES_OPTIMIZATION_TYPE;
-  context_config_resource_sharing_opt_type.shareResOptType = SEQUENTIAL_WITHOUT_VA_OPTIMIZATION;
-  QnnContext_Config_t resource_sharing_opt_type_config;
-  resource_sharing_opt_type_config.option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
-  resource_sharing_opt_type_config.customConfig = &context_config_resource_sharing_opt_type;
+  // Group-level listConfig for contextCreateFromBinaryListAsync (applies to all contexts in the list).
+  qnn::QnnConfigsBuilder<QnnContext_Config_t, QnnHtpContext_CustomConfig_t> configs_builder(
```
```suggestion
  QnnConfigsBuilder<QnnContext_Config_t, QnnHtpContext_CustomConfig_t> configs_builder(
```
Check for all the others as well.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -196,6 +201,28 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
       const qnn::EpContextIoDispatch& io_dispatch,
       bool is_multi_soc_buffer = false);
 
+  // Context configs shared by the initial AOT load (LoadCachedQnnContextFromBuffer)
+  // and SSR recovery (QnnModel::RecoverFromSSR), keeping the two call sites in sync.
+  struct ContextConfigOptions {
```
minfhong-qti: I think this struct is duplicate with `ContextRecoveryInfo` in `QnnModel`. There are two options:
1. Remove `ContextConfigOptions` and explicitly list parameters in `BuildContextConfigs`.
2. Reuse `ContextConfigOptions` in `QnnModel`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -196,6 +201,28 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
+    ContextPriority context_priority;
+    int64_t max_spill_fill_size = 0;
+    // Spill-fill group to join; 0x0 starts a new group (e.g. sole context after SSR).
+    Qnn_ContextHandle_t spill_fill_group_handle = 0x0;
+  };
+
+  // Builds the QnnContext_Config_t list (priority + optional spill-fill + optional reused-IO-limit)
+  // into configs_builder, which must outlive the subsequent contextCreateFromBinary call.
+  Ort::Status BuildContextConfigs(
```
minfhong-qti:
```suggestion
  Ort::Status BuildContextFromBinConfigs(
```
The naming should point out it is only for loading context binaries. As well for the above struct if kept.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1285,76 +1284,71 @@ Ort::Status QnnBackendManager::CreateContextVtcmBackupBufferSharingEnabled(
...
-  configs_vec.push_back(&context_config_weight_sharing);
-#endif
-#ifdef QNN_HTP_GRAPH_SPLITTING_AVAILABLE
-  if (enable_htp_graph_splitting) {
-    configs_vec.push_back(&context_config_graph_splitting_vtcm);
-  }
-#endif
-  configs_vec.push_back(nullptr);
+  // reusedIoLimitMb is a group-level property here (shared by all contexts in the list).
+  AppendReusedIoLimitConfig(configs_builder, *logger_ptr_);
```
minfhong-qti: IMO, `AppendReusedIoLimitConfig` is too dedicated, and thus I prefer to explicitly append the config here.
And since no need of `QnnConfigsBuilder`, probably consider leave other parts in this function untouched.
In addition, `reused_io_limit_mb_` should be in the struct.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1329,6 +1354,13 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                    false,
                                                    logger_);
 
+  // Caps the reused IO buffer size at context load. See docs; 0 = SDK default.
+  reused_io_limit_mb_ = ParseUInt64Option(ort_api,
+                                          session_options_,
+                                          FormatEPConfigKey("htp_reused_io_limit_mb"),
+                                          0,
+                                          logger_);
```
huaychou: [M-2 Major] After `reused_io_limit_mb_` is assigned via `ParseUInt64Option`, there is no VERBOSE log of the effective value. Compare `htp_share_resource_optimization` at line 887:
```cpp
            ("htp_share_resource_optimization: " + std::to_string(htp_share_resource_optimization_)).c_str());
```
In a debug session with VERBOSE logging, the user has no way to confirm from EP-init output that this option is active. The backend log (`Applying reused_io_limit_mb: N`) only fires at context-load time.

**Fix**: add after `ParseUInt64Option` returns:
```cpp
            ("htp_reused_io_limit_mb: " + std::to_string(reused_io_limit_mb_)).c_str());
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1329,6 +1354,13 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                    false,
                                                    logger_);
 
+  // Caps the reused IO buffer size at context load. See docs; 0 = SDK default.
+  reused_io_limit_mb_ = ParseUInt64Option(ort_api,
+                                          session_options_,
+                                          FormatEPConfigKey("htp_reused_io_limit_mb"),
```
huaychou: [N-1 Minor] The string `"htp_reused_io_limit_mb"` is used as an inline literal in `FormatEPConfigKey(...)`. Other HTP options are defined as `static constexpr const char*` constants (see `kEnableHtpMonolithicLstm` at line 968, `kEnableHtpFp16ClampOverflow` at line 980). An inline literal is typo-prone: a misspelling (e.g., `"htp_reused_io_limit_mb_"`) silently disables the feature with no compile-time or runtime error.

**Fix**: add alongside the other `static constexpr` constants:
```cpp
static constexpr const char* kHtpReusedIoLimitMb = "htp_reused_io_limit_mb";
```
and replace the inline literals with `kHtpReusedIoLimitMb`.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -146,6 +146,10 @@ Alternatively to setting profiling_level at compile time, profiling can be enabl
 |'0'|Default. Disabled.|
 |'1'|Enable VTCM backup buffer sharing across sessions. Requires QNN API version >= 2.26. Conflicts with `ep.context_embed_mode`.|
 
+|`"htp_reused_io_limit_mb"`|Description|
```
huaychou, line 150: [N-2 Minor] The `htp_reused_io_limit_mb` doc entry does not state a minimum QAIRT SDK version requirement. Other options with SDK prerequisites include that annotation (e.g., `enable_vtcm_backup_buffer_sharing`: "Requires QNN API version >= 2.26"; `optrace`: "Requires QAIRT 2.39 or later"). The underlying enum `QNN_HTP_CONTEXT_CONFIG_OPTION_REUSED_IO_LIMIT` (value 17) is confirmed present in SDK 2.45; if it was introduced after the repo minimum (2.29), a compile-time guard and runtime warning are also needed (follow the `htp_share_resource_optimization` pattern for the version guard).

**Fix**: Confirm the earliest SDK version that includes this enum, then add a "Requires QAIRT SDK X.X or later" note to the doc table entry; add a `#if QNN_API_VERSION_MINOR >= X` guard in `qnn_backend_manager.cc` if the version is > 2.29.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -1690,6 +1690,86 @@ TEST_F(QnnHTPBackendTests, QnnContextBinaryCacheNonEmbedModeTest) {
...
+                       BuildQDQOpTestCase<uint8_t>(op_type + "_node", op_type, {input_def}, {}, {}),
+                       provider_options,
+                       14,
+                       ExpectedEPNodeAssignment::All,
+                       QDQTolerance(),
+                       OrtLoggingLevel::ORT_LOGGING_LEVEL_ERROR,
+                       context_binary_file,
+                       session_option_pairs2);
+
+  std::remove(context_binary_file.c_str());
```
huaychou: [T-1 Nit] Post-test cleanup uses bare `std::remove(...)` without checking the return value. Other tests in this file use `ASSERT_EQ(std::remove(...), 0)` (e.g., lines 1291–1292). A failed removal silently leaves stale test artifacts.

**Fix**:
```cpp
ASSERT_EQ(std::remove(context_binary_file.c_str()), 0);
ASSERT_EQ(std::remove(qnn_ctx_bin.c_str()), 0);
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
huaychou: Just wondering, what is the usage of QnnConfigsBuilder?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1735,6 +1729,51 @@ Ort::Status QnnBackendManager::GetMaxSpillFillBufferSize(unsigned char* buffer,
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::BuildContextConfigs(
```
huaychou: Why do we need this function?

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -390,6 +369,20 @@ The `op_affinity` option points at a JSON config file that pins ONNX op types to
...
+
+**Multi-context / weight-sharing scenarios**: when `htp_share_resource_optimization` is enabled (multiple context binaries loaded together with resource/weight sharing), this value is applied as a group-level property — it is the shared IO buffer size for the whole group of contexts, not a per-context limit. The same option value is used in both cases; only its scope differs.
+
+**Warning**: this value is also a hard runtime cap on reused IO buffer size — the underlying QAIRT SDK does not guarantee correct behavior if actual reused IO traffic at runtime exceeds it. When using this option purely to work around a context load failure (rather than from a known buffer-reuse budget), the safe value is model-dependent and has not been validated across all models; a value verified safe for one model is not guaranteed safe for another. Verify empirically for your model before relying on a specific value in production.
```
huaychou: Is it possible to check the I/O buffer size in QNN EP to avoid error 1002?

huaychou (commented):
Do we need to add htp_reused_io_limit_mb in onnxruntime_perf_test?

onnxruntime/core/providers/qnn/qnn_execution_provider.h
chunghow-qti: done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
chunghow-qti: done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
chunghow-qti: it's suggested by @minfhong-qti , since both `LoadCachedQnnContextFromBuffer` and `RecoverFromSSR` can use the function.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc

onnxruntime/core/providers/qnn/qnn_execution_provider.cc

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
chunghow-qti: done

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
chunghow-qti: updated with options 1.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: done.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
chunghow-qti: Removed `AppendReusedIoLimitConfig`, and explicitly append the config here.
Leave the other part in this function untouched.

And since I removed the `ContextConfigOptions`, so still read `reused_io_limit_mb_` as a member here.

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: It's qairt 2.43. Updated.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
chunghow-qti: it's defined in `onnxruntime\core\providers\qnn\builder\qnn_configs_
it's the unique_ptr-backed builder that keeps base+custom config objects at stable addresses for the null-terminated QnnContext_Config_t*[] QNN expects (fixes the fragile "stack config lives to end-of-function" pattern).

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1735,6 +1752,53 @@ Ort::Status QnnBackendManager::GetMaxSpillFillBufferSize(unsigned char* buffer,
...
+    gsl::not_null<QnnHtpContext_CustomConfig_t*> spill_fill_custom_config = configs_builder.PushCustomConfig();
+    spill_fill_custom_config->option = QNN_HTP_CONTEXT_CONFIG_OPTION_REGISTER_MULTI_CONTEXTS;
+    spill_fill_custom_config->groupRegistration.firstGroupHandle = spill_fill_group_handle;
+    spill_fill_custom_config->groupRegistration.maxSpillFillBuffer = max_spill_fill_size;
+
+    gsl::not_null<QnnContext_Config_t*> spill_fill_config = configs_builder.PushConfig();
+    spill_fill_config->option = QNN_CONTEXT_CONFIG_OPTION_CUSTOM;
+    spill_fill_config->customConfig = spill_fill_custom_config;
+  }
+#endif
```
qti-hungjuiw: Since QNN-EP ABI plugin is introduced after QAIRT 2.42. Do we still need this check?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1735,6 +1752,53 @@ Ort::Status QnnBackendManager::GetMaxSpillFillBufferSize(unsigned char* buffer,
   return Ort::Status();
 }
 
+Ort::Status QnnBackendManager::BuildContextFromBinConfigs(
```
qti-hungjuiw: FYI. This `QnnBackendManager::BuildContextFromBinConfigs` has similar role with `QnnBackendManager::ReloadContextForSSR` in https://github.com/onnxruntime/onnxruntime-qnn/pull/677

chunghow-qti: In this case, do we still need this refactor? or we can just add the configs in two place manually. @minfhong-qti , could you share your thoughts

minfhong-qti: My concern is that the required modification for adding a new option should be minimized. Adding the same config in two places (even in the same QnnBackendManager) is not scalable. The boundary should be whether these places are in fact the same path of calling QNN API (that's why I allow `CreateVtcmBackup...` explicitly setting as that is another path.

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: As discussed with the owner, this requires the user to enable verbose logging, which will show output similar to:
```
2026-08-27 15:50:26.3436313 [V:onnxruntime:Default, qnn_backend_manager.cc:519 onnxruntime::qnn::QnnLogging]  <V> Context 1 graph QNNExecutionProvider_QNNExecutionProvider_4809872835210383736_1_0 estimated PD size ~3491.61MB, including nonSharedWeight 1908408320 B I/O 1662533632 B runlist 59602944 B spillfill 22282240 B 
```
The user can then check the I/O buffer size from this log.

chunghow-qti:
will rebase on PR#778 and PR#677
Also updated the doc from owners info.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -346,6 +347,30 @@ static bool ParseBoolOption(const OrtApi& ort_api,
...
+static void ParseIntOption(const OrtApi& ort_api,
+                           const OrtSessionOptions& session_options,
+                           const std::string& key,
+                           T default_value,
+                           T& out,
+                           const Ort::Logger& logger) {
+  out = default_value;
+  std::string value_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options, key, std::to_string(default_value), value_str);
+  if (value_str.empty()) {
```
huaychou, line 361: **[T-1] (Nit)** Dead early-return in ParseIntOption when a non-empty default is always provided

**Issue**: `ParseIntOption` lines 360–362 contain an `if (value_str.empty()) return;` early-return. However, `GetSessionConfigEntryOrDefault` is called with `std::to_string(default_value)` as the default — `std::to_string(uint64_t{0})` = `"0"` (non-empty). When the key is absent, the function receives `"0"` and proceeds to `from_chars`; the early-return is never triggered. Additionally, `ParseIntOption` is a template but is only called once with `uint64_t`.

**Suggested fix:**

Remove the dead `if (value_str.empty()) return;` (or retain it with a comment that the guard is only effective when the caller passes an empty string as the default). If there is currently only one call site, consider making the template a concrete non-template function taking `uint64_t`, or add a TODO noting that additional type instantiations are expected.

chunghow-qti: done, and kept the template for future reuse.

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -58,6 +58,13 @@ namespace qnn {
 #define QNN_HTP_GROUP_QUERY_ATTENTION_AVAILABLE
 #endif
 
+// QNN_HTP_CONTEXT_CONFIG_OPTION_REUSED_IO_LIMIT is available from QAIRT 2.45
```
minfhong-qti, line 61: QAIRT 2.45 is rather old comparing to current one. Do we really need this?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -371,6 +374,8 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   void SetQnnAllocatorType(QnnAllocatorType allocator_type) { qnn_allocator_type_ = allocator_type; }
   QnnAllocatorType GetQnnAllocatorType() const { return qnn_allocator_type_; }
 
+  uint64_t GetReusedIoLimitMb() const { return reused_io_limit_mb_; }
```
minfhong-qti: This function can be removed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -346,6 +347,31 @@ static bool ParseBoolOption(const OrtApi& ort_api,
...
+  std::string value_str;
+  GetSessionConfigEntryOrDefault(ort_api, session_options, key, std::to_string(default_value), value_str);
+  // An explicitly provided option may have an empty value. Treat it as default.
+  if (value_str.empty()) {
+    return;
+  }
+
+  const char* begin = value_str.data();
+  const char* end = begin + value_str.size();
+  auto [ptr, ec] = std::from_chars(begin, end, out);
```
huaychou, line 367: Just wondering, why not use std::stoi?

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -2386,18 +2468,14 @@ TEST_F(QnnHTPBackendTests, DISABLED_VTCMBackupBufferSharing) {
...
 
-TEST_F(QnnHTPBackendTests, FileMapping_Off) {
-#if (defined(__aarch64__) || defined(_M_ARM64)) && \
-    !(QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 34))
-  GTEST_SKIP() << "HTP weight sharing on ARM64 requires QNN API version >= 2.34.";
-#elif defined(__ANDROID__)
-  GTEST_SKIP() << "Weight sharing on Android devices is disabled";
-#else
-
+static void RunFileMappingOffTest(const char* htp_reused_io_limit_mb = nullptr) {
```
huaychou: Looks weird to pass `htp_reused_io_limit_mb` to `RunFileMappingOffTest`. Could we rename the function?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
chunghow-qti: done

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
chunghow-qti: done

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -1690,6 +1690,89 @@ TEST_F(QnnHTPBackendTests, QnnContextBinaryCacheNonEmbedModeTest) {
...
 }
 
+// htp_reused_io_limit_mb: a valid numeric value is accepted and does not block AOT context load.
+TEST_F(QnnHTPBackendTests, QnnContextBinary_HtpReusedIoLimitMbValid_LoadsSucceeds) {
+#if QNN_API_VERSION_MAJOR < 2 || \
+    (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR < 34)
+  GTEST_SKIP() << "htp_reused_io_limit_mb requires QAIRT 2.45 or later (QNN API >= 2.34).";
+#elif defined(__linux__) && !defined(__aarch64__)
+  GTEST_SKIP() << "htp_reused_io_limit_mb is not supported by the x86_64 HTP emulator.";
+#else
```
minfhong-qti:
```suggestion
#end
```
It can be ended here.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -1690,6 +1690,89 @@ TEST_F(QnnHTPBackendTests, QnnContextBinaryCacheNonEmbedModeTest) {
...
+
+  std::unordered_map<std::string, std::string> session_option_pairs;
+  session_option_pairs.emplace(kOrtSessionOptionEpContextEnable, "1");
+  session_option_pairs.emplace(kOrtSessionOptionEpContextFilePath, context_binary_file);
+  session_option_pairs.emplace(kOrtSessionOptionEpContextEmbedMode, "0");
+
+  const TestInputDef<float> input_def({1, 2, 3}, false, -10.0f, 10.0f);
+  const std::string op_type = "Atan";
+
+  // 1st run generates the Onnx skeleton file + Qnn context cache binary file with the
```
minfhong-qti: Can we leverage `enable_htp_prepare_and_load` to test in one session?

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -2507,6 +2586,31 @@ TEST_F(QnnHTPBackendTests, FileMapping_Off) {
...
+#elif defined(__ANDROID__)
+  GTEST_SKIP() << "Weight sharing on Android devices is disabled";
+#else
+  RunSharedContextWithFileMappingDisabledTest();
+#endif
+}
+
+// Verifies that htp_reused_io_limit_mb is accepted as a group-level config by
+// contextCreateFromBinaryListAsync. File mapping is disabled to select the non-callback list API.
+TEST_F(QnnHTPBackendTests, CreateFromBinaryListAsync_HtpReusedIoLimitMb_LoadsSucceeds) {
```
minfhong-qti: I recall that `CreateContextVtcmBackupBufferSharingEnabled` is only enabled when `htp_share_resource_optimization=1` and not necessary related to file mapping or `CreateFromBinaryListAsync`.

onnxruntime/core/providers/qnn/builder/qnn_def.h
minfhong-qti: Leave for @qti-chuteng to decide.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
chunghow-qti: done

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
chunghow-qti: done

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
chunghow-qti: updated the test name. btw, here the `htp_share_resource_optimization=1` is enabled through `enable_vtcm_backup_buffer_sharing`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: it supports both signed and unsigned integer option types. For reused io limit mb it is instantiated as uint64_t, so std::stoi would be narrower and require exception handling.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -393,6 +397,34 @@ The `op_affinity` option points at a JSON config file that pins ONNX op types to
...
+
+```
+... estimated PD size ~3491.61MB, including nonSharedWeight 1908408320 B I/O 1662533632 B runlist 59602944 B spillfill 22282240 B
+```
+
+The `I/O` field is that graph's I/O size. Sum or take the max over the graphs your app uses concurrently, per the rules above.
+
+**Per-context vs. group scope** — the same intended peak maps to a *different* value depending on which context-creation path is used, because QNN only sees one context at a time on the default path:
+
+- **Default path (`contextCreateFromBinary`, one config per context):** QNN adds up the per-context limits, so set each context's value to `peak / number_of_contexts`. Example: 4 contexts, each 1 graph of 100 MB, app runs one graph at a time (real peak 100 MB) → set `25` on each context so QNN totals 100 MB.
```
huaychou: Does it mean that if the I/O is 100 MB, we need to set htp_reused_io_limit_mb as 25 for QNN EP?

chunghow-qti: Yes, but Only for the default per-context `contextCreateFromBinary` path, and only when four contexts have a combined actual I/O **peak** of 100 MB.
For the group `contextCreateFromBinaryListAsync` path (by enable `htp_share_resource_optimization`), set 100 MB once for the group. The value represents the actual peak registered I/O, not simply the I/O size of an individual graph.

huaychou: Do you mean that the four contexts in 1 graph have the same I/O size? Another question is, how does user know the number of contexts?

chunghow-qti: Clarified on the documentation. One context binary can contain multiple QNN graphs; `graphs` here specifically means `QNN graphs`.

The example has four independently loaded contexts, each containing one QNN graph. Each QNN graph is assumed to have 100 MB of I/O only to simplify the calculation.

If the application maps the I/O for only one QNN graph at a time, the actual peak registered I/O across all four contexts is 100 MB. With the traditional `contextCreateFromBinary` path, QNN sees only one context at a time and sums the limit configured for all independently loaded
  contexts. Therefore, setting 25 MB for each context makes QNN estimate 100 MB in total.

With the async list API (`contextCreateFromBinaryListAsync`) (`htp_shared_resource_optimization` path), QNN sees the whole context group, so the group-level value should be set directly to 100 MB.

Users can find the number of QNN graphs in a context from the VERBOSE load log:
  `Graph count from QNN context: N`.
  The number of contexts is the number of context binaries / EP-context models that the application loads independently.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -393,6 +397,40 @@ The `op_affinity` option points at a JSON config file that pins ONNX op types to
...
+
+```
+... Graph count from QNN context: 4
+```
+
+The per-QNN-graph estimates show the I/O sizes; the `I/O` field is that QNN graph's I/O size. Sum or take the max over the QNN graphs your app uses concurrently, per the rules above.
+
+**Per-context vs. group scope** — the same intended peak maps to a *different* value depending on which context-creation path is used, because QNN only sees one context at a time on the default path:
+
+- **Default path (`contextCreateFromBinary`, one config per context):** QNN adds up the limits configured on independently loaded contexts. The app knows this context count from the context binaries / EP-context models it loads. Set the per-context values so their sum is the app's actual peak. Example: 4 contexts, each containing 1 QNN graph with 100 MB I/O; if the app maps the I/O for only 1 QNN graph at a time (real peak 100 MB), set `25` on each context so QNN totals 100 MB. Contexts with different I/O sizes or mapping lifetimes do not need equal values.
```
huaychou: Setting 25 MB on each context doesn't mean that each context has 25 MB I/O size. It is used to make the total I/O size become 100 MB. Is this correct? If so, is it possible to set the total I/O size directly for default path?

chunghow-qti: Yes, I believe this behavior is required by QNN's internal per-context configuration mechanism, although it looks a little bit unintuitive. 

The default `contextCreateFromBinary` API only accepts a config for one context at a time; it has no group-wide scope, so there is no API to pass the total directly on that path. Passing 100 MB directly is supported by the `contextCreateFromBinaryListAsync` group path, where QNN sees all contexts together.

The appropriate value depends on how the user app use it. We cannot assume the total I/O size is the actual peak memory usage in every case. Therefore, I think proving the guideline/example is more appropriate.

huaychou: Acknowledge

huaychou (approved):

chunghow-qti:
Update the doc to use `htp_share_resource_optimization` to clarify default path or not, which is more clear to user instead of using QNN API explaination.

onnxruntime/core/providers/qnn/builder/qnn_def.h
qti-chuteng: Because the backward-compatibility policy is not clear now, so we can have this for now and clean up later.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -393,6 +397,41 @@ The `op_affinity` option points at a JSON config file that pins ONNX op types to
...
+
+```
+... Graph count from QNN context: 4
+```
+
+The per-QNN-graph estimates show the I/O sizes; the `I/O` field is that QNN graph's I/O size. Sum or take the max over the QNN graphs your app uses concurrently, per the rules above.
+
+**Per-context vs. group scope** — choose the value based on whether `htp_share_resource_optimization` is enabled:
+
+- **Default (`htp_share_resource_optimization` disabled):** QNN adds up the limits configured on independently loaded contexts. The app knows this context count from the context binaries / EP-context models it loads. Set the per-context values so their sum is the app's actual peak. Example: 4 contexts, each containing 1 QNN graph with 100 MB I/O; if the app maps the I/O for only 1 QNN graph at a time across all 4 contexts (real peak 100 MB), set `25` on each context so QNN totals 100 MB. Contexts with different I/O sizes or mapping lifetimes do not need equal values.
```
qti-chuteng: Put the example above default bulletpoint

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -346,6 +347,31 @@ static bool ParseBoolOption(const OrtApi& ort_api,
...
+  if (ec != std::errc{} || ptr != end) {
```
qti-chuteng, line 368: will it fail when the number is floating, like 1.1?
It should report the error when the datatype mismatched.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -1690,6 +1690,61 @@ TEST_F(QnnHTPBackendTests, QnnContextBinaryCacheNonEmbedModeTest) {
   ASSERT_EQ(std::remove(qnn_ctx_bin.c_str()), 0);
 }
 
+// htp_reused_io_limit_mb: a valid numeric value is accepted when preparing and loading an AOT context.
+TEST_F(QnnHTPBackendTests, QnnContextBinary_HtpReusedIoLimitMbValid_LoadsSucceeds) {
+#if QNN_API_VERSION_MAJOR < 2 || \
```
qti-chuteng, line 1695: Could we reuse QNN_HTP_REUSED_IO_LIMIT_AVAILABLE here?

chunghow-qti: I don't think so, since it's define in qnn_def.h and not good to include it at test.
So I just keep simple version check followed other ut tradition like graph splitting.

docs/execution_providers/QNN-ExecutionProvider.md
chunghow-qti: done

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
chunghow-qti: yes, update for reporting the error

qti-chuteng (approved):

PR #754, opened 2026-08-24
[OpBuilder] Allow op builder acquiring HTP arch.

minfhong-qti:
### Description

Revise QnnModelWrapper to set QnnBackendManager as member and define a getter function which can be called in op builder to query HTP arch.

Update the unit tests for the new constructor. Add StubBackendManager to qnn_unit_test_utils.h: it owns a QnnBackendManager and exposes the QNN interface, backend handles, backend type and HTP arch as mutable references, so tests can keep stubbing them now that they are no longer constructor arguments. Point the four wrapper fixtures at it.

Add coverage for the new behaviour: GetQnnBackendManager, the op-builder HTP arch query, backend type being read live from the manager rather than snapshotted, and the arch handling in SetupDeviceAndContext / ReleaseDeviceAndContext.

### Motivation and Context

Due to the requirement of determining op translation according to HTP arch, op builder must be able to query HTP arch.


onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -52,40 +52,34 @@ class QnnModelWrapper {
...
   }
   ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(QnnModelWrapper);
 
   ~QnnModelWrapper() = default;
 
+  const QnnBackendManager* GetQnnBackendManager() const { return qnn_backend_manager_; }
+
   const ModelSettings& GetModelSettings() const { return model_settings_; }
 
   bool CreateQnnGraph(const Qnn_ContextHandle_t& context,
```
huaychou, line 72: **[D-1] 🟠 Significant Concern — `GetQnnBackendManager()` exposes the full manager to op-builder consumers**

The PR's goal is to let op builders query `GetHtpArch()`. However, `GetQnnBackendManager()` returns the entire `const QnnBackendManager*`, giving every caller access to backend creation, context release, power-config management, device setup, and the full manager API.

This is a layer violation: op builders (per-op/per-fusion knowledge) should not reach through the wrapper layer into the lifecycle manager. `const` qualification prevents mutation but does not narrow the API surface. Additionally, the accessor has **zero production callers today** — it is promoted to the public interface before any call site exists.

**Suggested fix:** Add a thin forwarding method directly on `QnnModelWrapper`, projecting only what op builders need:
```cpp
QnnHtpDevice_Arch_t GetHtpArch() const {
  return qnn_backend_manager_->GetHtpArch();
}
```
Op builders call `wrapper.GetHtpArch()` without ever holding the manager pointer. `GetQnnBackendManager()` can be removed (or made private) until a concrete production caller exists. The existing `GetQnnBackendType()` forwarder on the wrapper (line 387) is the established pattern.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -52,40 +52,34 @@ class QnnModelWrapper {
...
         api_ptrs_(ApiPtrs{api_ptrs.ort_api, api_ptrs.ep_api, api_ptrs.model_editor_api}),
         tensor_name_overrides_(tensor_name_overrides),
         op_trace_collector_(op_trace_collector),
         is_post_layout_transform_(is_post_layout_transform) {
     // Invariant: validator interface and handle must both be set or both be null.
     // They are populated together by QnnBackendManager::LoadQnnSerializerBackend() (QnnIr flow).
-    assert((validator_backend_handle == nullptr) ==
-           (qnn_validator_interface.backendValidateOpConfig == nullptr));
+    assert((qnn_backend_manager_->GetQnnValidatorBackendHandle() == nullptr) ==
+           (qnn_backend_manager_->GetQnnValidatorInterface().backendValidateOpConfig == nullptr));
```
huaychou: **[D-2] 🟠 Significant Concern — Null dereference before the constructor `assert`; raw-pointer lifetime contract unencoded**

The constructor stores `const QnnBackendManager*` as a raw pointer but dereferences it inside the `assert()` (`qnn_backend_manager_->GetQnnValidatorBackendHandle()`) before any null check. In an optimized (Release) build, `assert()` is silent, so a null pointer becomes immediate UB.

The deeper issue is the lifetime contract: `QnnModel` holds the manager as `std::shared_ptr<QnnBackendManager>` (explicit shared ownership), but `QnnModelWrapper` switches to a raw pointer with no type or static-analysis enforcement that the manager outlives the wrapper.

**Suggested fix (pick one):**
- **Option 1 (minimal):** Add `ORT_ENFORCE(qnn_backend_manager != nullptr, "QnnBackendManager must not be null")` at the top of the constructor, before any dereference. Add a comment encoding the lifetime invariant.
- **Option 2 (align with QnnModel):** Change the member to `std::shared_ptr<const QnnBackendManager>`, encoding ownership at the type level.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2256,7 +2256,7 @@ Ort::Status QnnBackendManager::SetupDeviceAndContext(QnnHtpDevice_Arch_t htp_arc
   }
 
   // Override cached values.
-  htp_arch_ = htp_arch;
+  htp_arch_ = htp_arch_internal_ = htp_arch;
   soc_model_ = soc_model;
```
huaychou, line 2263: **[D-4] 🔵 Improvement — `htp_arch_internal_` double-write lacks an explanatory comment**

The line `htp_arch_ = htp_arch_internal_ = htp_arch` writes both fields simultaneously. The semantic distinction is valid (`htp_arch_` drives QNN device creation; `htp_arch_internal_` is exposed to op builders via `GetHtpArch()`), but why both are updated together in the multi-SoC path is not self-evident.

**Suggested fix:** Add a brief comment, e.g.:
```cpp
// Update both: htp_arch_ drives QNN device creation; htp_arch_internal_
// is the op-builder-visible value that GetHtpArch() returns. In the
// multi-SoC path the caller explicitly provides the arch, overriding
// any previous GetPlatformInfo() result.
```

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
```diff
@@ -167,6 +168,125 @@ struct OrtApiStubContext {
...
+//     type. Do NOT call it from a scope that declares a member of the same name:
+//     ordinary unqualified lookup finding a class member suppresses ADL
+//     ([basic.lookup.argdep]/1) and the call fails to compile.
+//   - Prefer the public API where one exists. QnnBackendManager::SetQnnBackendType()
+//     is the public path when you hold a QNN backend *id*; BackendType() below
+//     exists because the fixtures take a qnn::QnnBackendType directly (and
+//     QnnBackendType::HTP_FP16 has no corresponding backend id).
+
+template <typename Tag, typename Tag::type Member>
+struct PrivateMember {
```
huaychou: **[D-5] 🔵 Improvement — `StubBackendManager` explicit-instantiation definitions in a header have IFNDR risk**

The `PrivateMember<Tag, &Member>` explicit instantiation definitions appear in a header included by multiple translation units. The C++ standard ([temp.spec]/6) requires that an explicit instantiation definition of the same specialization appear in only one TU; having multiple TUs include this header makes the program IFNDR. GCC/Clang accept this via comdat/weak symbols, but MSVC behavior differs and may produce linker errors or silent symbol collisions. The README acknowledges this risk but only says to re-verify on other toolchains.

**Suggested fix:** Move the six explicit instantiation definitions (`template struct PrivateMember<...>`) into a single `qnn_unit_test_utils.cc`. Keep only the tag struct declarations in the header. This is the standard-conforming approach and eliminates the IFNDR concern without changing any test logic.

onnxruntime/test/providers/qnn/unit/qnn_backend_manager_test.cc
```diff
@@ -235,6 +236,118 @@ TEST(QnnUnit_BackendManagerTest, ParseLoraConfig_ValidFormatNoContext_ReturnsErr
...
+  EXPECT_EQ(const_manager.GetQnnValidatorBackendHandle(), nullptr);
+  EXPECT_EQ(const_manager.GetQnnBackendType(), qnn::QnnBackendType::CPU);
+  EXPECT_EQ(const_manager.GetHtpArch(), QNN_HTP_DEVICE_ARCH_NONE);
+  EXPECT_EQ(const_manager.GetQnnInterface().backendCreate,
+            const_manager.GetQnnValidatorInterface().backendCreate);
+}
+
 // ===========================================================================
 // Real-HTP-backend tests
 //
```
huaychou, line 359: **[D-3] 🟠 Significant Concern — `htp_arch_internal_` ARM64 real-device path has no unit test; x86 SERIALIZER path also uncovered**

`htp_arch_internal_` has two population paths:
1. **x86**: `GetPlatformInfo()`'s `#else` branch copies from `htp_arch_` if non-NONE
2. **ARM64**: queries real hardware via `QnnDevice_PlatformInfo`

The ARM64 path is guarded by `#if defined(__aarch64__)` and cannot be reached by x86 unit tests. Since `htp_arch_internal_` is the sole source for `GetHtpArch()` — the field this entire PR is built around — its correctness on real hardware has no unit-test validation.

Additionally, the **x86 SERIALIZER-backend path** through `GetPlatformInfo` (which is the primary host-side path for op builders) also has no dedicated host-only test. The existing `GetHtpArch_UserProvidedArchBeforeSetup_ReturnsNone` test validates only the pre-setup state.

**[D-6] is the actionable fix for the x86 path:**
Add a host-only test: create a manager with `htp_arch = QNN_HTP_DEVICE_ARCH_V75`, call `SetupBackend` with a SERIALIZER backend, and assert `GetHtpArch() == V75`. This directly exercises the x86 `GetPlatformInfo` else-branch without any real device.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -270,9 +270,9 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
 
   void RemovePerThreadHtpPowerConfigMapping(const std::thread::id& thread_id);
 
-  const QNN_INTERFACE_VER_TYPE& GetQnnInterface() { return qnn_interface_; }
+  const QNN_INTERFACE_VER_TYPE& GetQnnInterface() const { return qnn_interface_; }
 
-  const QNN_INTERFACE_VER_TYPE& GetQnnValidatorInterface() { return qnn_validator_interface_; }
+  const QNN_INTERFACE_VER_TYPE& GetQnnValidatorInterface() const { return qnn_validator_interface_; }
 
   const QNN_SYSTEM_INTERFACE_VER_TYPE& GetQnnSystemInterface() { return qnn_sys_interface_; }
```
huaychou, line 278: **[D-7] 🟢 Nit — `GetQnnSystemInterface()` is not const-qualified, inconsistent with the other getters**

The PR const-qualifies five getters, but `GetQnnSystemInterface()` remains non-const. Any op builder that tries to call it through the `const QnnBackendManager*` stored in `QnnModelWrapper` will get a compile error. This is an incomplete const API surface.

**Suggested fix:** Add `const` to the declaration of `GetQnnSystemInterface()` (mechanical change, no semantic impact).

huaychou (commented):
## Design Review — PR #754

> **Design verdict**: ❌ Needs rework  
> **Findings**: 0 Design Blocker · 3 Significant Concern · 3 Improvement · 2 Consideration

---

### Headline Finding

`GetQnnBackendManager()` exposes the entire `QnnBackendManager` object to op-builder consumers — granting access to backend creation, context release, and the full manager lifecycle API — far beyond the stated goal of letting op builders query `GetHtpArch()`. A thin `GetHtpArch()` forwarder on `QnnModelWrapper` achieves the intent while preserving the layer boundary. This is the highest-leverage change before merge.

---

### What's Good

The core refactor direction (collapsing 4 constructor args into 1 manager pointer) is correct and valuable. The `htp_arch_internal_` / `htp_arch_` split is semantically justified (x86 `QnnDevice_getPlatformInfo` always returns a dummy arch). The `const`-correctness upgrades to five getters are clean. Test coverage for the new APIs is strong (6+ new test cases). `StubBackendManager` is well-documented and proportionate to its 4-file usage.

---

### Summary of Issues

| ID | Severity | File | Topic |
|---|---|---|---|
| D-1 | 🟠 Significant | `qnn_model_wrapper.h:84` | `GetQnnBackendManager()` crosses wrapper-to-builder layer boundary; zero production callers |
| D-2 | 🟠 Significant | `qnn_model_wrapper.h:74` | Null dereference before `assert`; raw-pointer lifetime contract unencoded |
| D-3+D-6 | 🟠 Significant | `qnn_backend_manager_test.cc` | `htp_arch_internal_` x86 SERIALIZER path has no host-only test |
| D-4 | 🔵 Improvement | `qnn_backend_manager.cc:2259` | Double-write lacks explanatory comment |
| D-5 | 🔵 Improvement | `qnn_unit_test_utils.h:211` | Explicit instantiation definitions in header: IFNDR on non-GCC/Clang |
| D-7 | 🟢 Nit | `qnn_backend_manager.h:277` | `GetQnnSystemInterface()` not const — incomplete const API surface |
| D-8 | 🟢 Nit | `qnn_model_wrapper.h:84` | `GetVtcmSize()` not exposed alongside `GetHtpArch()` |

---

**After D-1 and D-2 are addressed, the PR can be merged.** D-3/D-6 (host-only SERIALIZER test) is the next priority; D-4/D-5 are low-cost clean-ups recommended before merge. D-7/D-8 are nits you can decide on.

<sub>Generated by deep-review skill</sub>

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
minfhong-qti: Added `QnnModelWrapper::GetHtpArch`.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
minfhong-qti: Added `ORT_CXX_API_THROW` if null.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
minfhong-qti: Added comment.

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
minfhong-qti: Moved to a new .cc file.

onnxruntime/test/providers/qnn/unit/qnn_backend_manager_test.cc
minfhong-qti: Not entirely correct. But indeed there is no explicit check for HTP arch with SetupBackend. Revised existing testcase to `SetupBackend_HTP_WithHtpArch_ReportsConfiguredArch` to check.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
minfhong-qti: Added `const` quailifier and corresponding testcase.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -10,6 +10,9 @@
 #include <utility>
 #include <vector>
 
+#include "QnnCommon.h"
```
huaychou: [N-1 Minor] `#include "QnnCommon.h"` is redundant here. `QnnInterface.h` (line 14) already begins with `#include "QnnCommon.h"`, and `qnn_model_wrapper.h` now includes `qnn_backend_manager.h` which also includes it. Suggest removing line 13 and keeping only `#include "QnnInterface.h"`.

onnxruntime/test/providers/qnn/unit/qnn_backend_manager_test.cc
```diff
@@ -235,6 +236,123 @@ TEST(QnnUnit_BackendManagerTest, ParseLoraConfig_ValidFormatNoContext_ReturnsErr
   std::filesystem::remove(bin_file);
 }
 
+// ---------------------------------------------------------------------------
+// Group 6: HTP arch / SoC model holders, and const-qualified getters
+// ---------------------------------------------------------------------------
+//
+// GetHtpArch() returns htp_arch_internal_, the holder that op builders read through
+// QnnModelWrapper::GetQnnBackendManager(). It is deliberately distinct from the
```
huaychou: [N-2 Minor] The comment reads "op builders read through `QnnModelWrapper::GetQnnBackendManager()`" but no such method exists on `QnnModelWrapper`. Op builders obtain the HTP arch through `QnnModelWrapper::GetHtpArch()`, which internally delegates to `qnn_backend_manager_->GetHtpArch()`. Suggest replacing `GetQnnBackendManager()` with `GetHtpArch()` in this comment.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2257,7 +2257,7 @@ Ort::Status QnnBackendManager::SetupDeviceAndContext(QnnHtpDevice_Arch_t htp_arc
   }
 
   // Override cached values.
-  htp_arch_ = htp_arch;
+  htp_arch_ = htp_arch_internal_ = htp_arch;  // Update both to make sure GetHtpArch returning the correct value.
```
huaychou: [T-1 Nit] The inline comment is imprecise in two ways: (1) grammar: `returning` → `returns`; (2) it says `GetHtpArch` reads both fields, but `GetHtpArch()` only reads `htp_arch_internal_` — `htp_arch_` is what `GetPlatformInfo()` uses on x86 in the `SetupBackend` path. Suggested rewrite:
```
// Update htp_arch_ (used by GetPlatformInfo on x86) and htp_arch_internal_
// (returned by GetHtpArch) for the multi-SoC EP-context flow.
```

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
```diff
@@ -167,6 +168,87 @@ struct OrtApiStubContext {
...
+    cfg.skip_qnn_version_check = true;
+    manager_ = qnn::QnnBackendManager::Create(cfg, api_ptrs, logger);
+  }
+
+  StubBackendManager(const StubBackendManager&) = delete;
+  StubBackendManager& operator=(const StubBackendManager&) = delete;
+  StubBackendManager(StubBackendManager&&) = delete;
+  StubBackendManager& operator=(StubBackendManager&&) = delete;
+
+  qnn::QnnBackendManager* Get() const { return manager_.get(); }
```
huaychou: [T-3 Nit] `StubBackendManager::Get()` returns `qnn::QnnBackendManager*` (non-const), while `QnnModelWrapper`'s constructor takes `const QnnBackendManager*`. Current callers are all mutable so the implicit conversion compiles, but a future `const StubBackendManager` context would require a `const`-qualified overload. Consider adding:
```cpp
const qnn::QnnBackendManager* Get() const { return manager_.get(); }
```
or replacing the existing one with a `const`-returning version.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -287,9 +287,9 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
     return contexts_.size();
   }
 
-  const Qnn_BackendHandle_t& GetQnnBackendHandle() { return backend_handle_; }
+  const Qnn_BackendHandle_t& GetQnnBackendHandle() const { return backend_handle_; }
 
-  const Qnn_BackendHandle_t& GetQnnValidatorBackendHandle() { return validator_backend_handle_; }
+  const Qnn_BackendHandle_t& GetQnnValidatorBackendHandle() const { return validator_backend_handle_; }
 
   const Qnn_DeviceHandle_t& GetQnnDeviceHandle() { return device_handle_; }
```
huaychou: Do we need to add const for `GetQnnDeviceHandle`?

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -52,42 +52,41 @@ class QnnModelWrapper {
...
+    // QnnBackendManager is assumed to be non-nullptr.
+    if (qnn_backend_manager_ == nullptr) {
+      ORT_CXX_API_THROW("Unexpected nullptr QnnBackendManager.", ORT_EP_FAIL);
+    }
+
     // Invariant: validator interface and handle must both be set or both be null.
     // They are populated together by QnnBackendManager::LoadQnnSerializerBackend() (QnnIr flow).
-    assert((validator_backend_handle == nullptr) ==
-           (qnn_validator_interface.backendValidateOpConfig == nullptr));
+    assert((qnn_backend_manager_->GetQnnValidatorBackendHandle() == nullptr) ==
```
huaychou: Should we define a `Get*()` function for all the `qnn_backend_manager_->Get*()` functions?

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -549,8 +548,9 @@ class QnnModelWrapper {
 
   // BF16 conversion helper methods
   bool IsBF16ConversionEnabled() const {
+    QnnBackendType qnn_backend_type = qnn_backend_manager_->GetQnnBackendType();
```
huaychou:
```suggestion
    QnnBackendType qnn_backend_type = GetQnnBackendType();
```

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
minfhong-qti: Although I think we should include what we use, I removed it as suggested.

onnxruntime/test/providers/qnn/unit/qnn_backend_manager_test.cc
minfhong-qti: Removed.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
minfhong-qti: Revised.

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
minfhong-qti: Added.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
minfhong-qti: Since not all `qnn_backend_manager_->Get*()` would be required by `qnn_model_wrapper`'s user, I think they can be added per request.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
minfhong-qti: Updated.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -585,10 +585,7 @@ class QnnModelWrapper {
 
   const OrtGraph& ort_graph_;
   const Ort::Logger& logger_;
-  const QNN_INTERFACE_VER_TYPE& qnn_interface_;
-  const Qnn_BackendHandle_t& backend_handle_;
-  const QNN_INTERFACE_VER_TYPE& qnn_validator_interface_;
-  const Qnn_BackendHandle_t& validator_backend_handle_;
+  const QnnBackendManager* qnn_backend_manager_;
```
kuanyul-qti: `QnnModelWrapper` requires a valid `QnnBackendManager` for its entire lifetime and `nullptr` is treated as an invalid state. Would it make sense to store a `const QnnBackendManager&` instead of a raw pointer? This removes the runtime `nullptr` check.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
minfhong-qti: Added.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
minfhong-qti: Revised as suggested.

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
```diff
@@ -167,6 +168,87 @@ struct OrtApiStubContext {
...
+    cfg.htp_arch = QNN_HTP_DEVICE_ARCH_NONE;
+    cfg.soc_model = QNN_SOC_MODEL_UNKNOWN;
+    cfg.skip_qnn_version_check = true;
+    manager_ = qnn::QnnBackendManager::Create(cfg, api_ptrs, logger);
+  }
+
+  StubBackendManager(const StubBackendManager&) = delete;
+  StubBackendManager& operator=(const StubBackendManager&) = delete;
+  StubBackendManager(StubBackendManager&&) = delete;
+  StubBackendManager& operator=(StubBackendManager&&) = delete;
```
huaychou: **[N-1] (Minor)** StubBackendManager bypasses the QNN EP non-copyability macro

`StubBackendManager` owns a `std::shared_ptr<QnnBackendManager>`, while its construction-time `ApiPtrs` and logger have lifetime constraints, making it an explicit owning helper. Lines 233–236 manually spell out four deleted special members instead of using the QNN EP-standard `ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE` macro.

**Impact**: There is no functional difference, but it bypasses the established ownership convention and makes the helper's non-copyable/non-movable contract less immediately recognizable alongside other owning QNN classes.

**Suggested fix:**

Replace the four manual deleted declarations with `ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(StubBackendManager);`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2257,7 +2257,8 @@ Ort::Status QnnBackendManager::SetupDeviceAndContext(QnnHtpDevice_Arch_t htp_arc
   }
 
   // Override cached values.
-  htp_arch_ = htp_arch;
+  // Update htp_arch_internal_ as well to make sure GetHtpArch returns correct value.
```
huaychou: **[T-1] (Nit)** HTP arch synchronization comment omits why htp_arch_ is updated

The new comment explains synchronizing `htp_arch_internal_` for `GetHtpArch()`, but the same statement also updates `htp_arch_`. That field is not read by the getter; it is the user-provided source for x86 `GetPlatformInfo()`. Describing the distinct roles of the two holders would make the multi-SoC lifecycle easier to maintain.

**Suggested fix:**

Consider: `// Update htp_arch_ (the x86 GetPlatformInfo source) and htp_arch_internal_ (returned by GetHtpArch) for the multi-SoC flow.`

kuanyul-qti (approved):

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -52,42 +52,36 @@ class QnnModelWrapper {
...
         model_settings_(model_settings),
         api_ptrs_(ApiPtrs{api_ptrs.ort_api, api_ptrs.ep_api, api_ptrs.model_editor_api}),
         tensor_name_overrides_(tensor_name_overrides),
         op_trace_collector_(op_trace_collector),
         is_post_layout_transform_(is_post_layout_transform) {
     // Invariant: validator interface and handle must both be set or both be null.
     // They are populated together by QnnBackendManager::LoadQnnSerializerBackend() (QnnIr flow).
-    assert((validator_backend_handle == nullptr) ==
-           (qnn_validator_interface.backendValidateOpConfig == nullptr));
+    assert((qnn_backend_manager_.GetQnnValidatorBackendHandle() == nullptr) ==
```
huaychou: Just wondering, what is the use case that both art nullptr?

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
```diff
@@ -882,7 +901,7 @@ Ort::Status QnnModelWrapper::GetTensorInfo(const OrtNodeUnitIODef& tensor, Tenso
   RETURN_IF_ERROR(utils::GetQnnDataType(tensor.quant_param.has_value(),
                                         tensor.type,
                                         tensor_info.qnn_data_type,
-                                        qnn_backend_type_));
+                                        qnn_backend_manager_.GetQnnBackendType()));
```
huaychou:
```suggestion
                                        GetQnnBackendType()));
```

huaychou (approved):
Some minor issues, other LGTM.

onnxruntime/test/providers/qnn/unit/qnn_unit_test_utils.h
minfhong-qti: Revised.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
minfhong-qti: The usage of `qnn_interface` and `qnn_validator_interface` is for dumping DLC. In this case, `qnn_interface` is from `QnnIr.dll` and `qnn_validator_interface` is for `QnnHtp.dll`. In normal usage, these two are redundant and thus can be null.

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.cc
minfhong-qti: Revised.

qti-chuteng (approved):

onnxruntime/core/providers/qnn/builder/qnn_model_wrapper.h
```diff
@@ -10,8 +10,8 @@
 #include <vector>
 
 #include "nlohmann/json.hpp"
-#include "QnnInterface.h"
 
+#include "core/providers/qnn/builder/qnn_backend_manager.h"
```
qti-kromero: I believe this include now causes all op builders to include qnn_backend_manager.h which may cause a significant amount of rebuilding on incremental builds when changing things in qnn_backend_manager.h

we could not include it and forward declare QnnBackendManager as a class inside of this include to satisfy the compiler, then we can remove qnn_backend_manager.h from this file and not deal with the consequences on build times of the above

minfhong-qti: Revised as suggested.

PR #755, opened 2026-08-24
[QNN EP] Refactor LPBQ lowering helpers across op builders

qti-ashimaj:
### Description
Refactors the block-quantized (BQ) and LPBQ weight lowering logic across the QNN EP op builders (Conv, MatMul, MatMulNBits) to eliminate code duplication and consolidate shared helpers.

__`qnn_bq_utils`__ :

- `RegisterWeightAsConv1x1Filter`: registers an LPBQ weight `[K, N]` as a `[1, 1, K, N]` HWCN 1×1 Conv2D filter, updating the LPBQ quant axis.
- `AddConv2DNodeforBQLowering`: creates a QNN Conv2D node (stride=1, pad=0, dilation=1, group=1) shared by both LPBQ and BwFloatBlock lowering paths.

__`qnn_utils`__ :

- `ProcessBiasForQuantizedOp`: handles bias processing for quantized ops, requantizes INT32 bias if scales mismatch or quantizes float bias using `activation_scale * weight_scale[c]`.

__`conv_op_builder`__ : removed duplicated helpers (`GetActivationScale`, `BuildBiasQuantParams`, `AddStaticBiasTensor`, `ProcessRequantizeBias`, `ProcessFloatBias`).

__`matmul_op_builder` / `matmulnbits_op_builder`__ : updated to use the new shared helpers.

__`matmul_test`__ : enabled `MatMulOp_QDQ_BlockQuant` on Windows in addition to Linux.

### Motivation and Context
The LPBQ lowering code was duplicated across various op builders like `conv_op_builder`, `matmul_op_builder`, and `matmulnbits_op_builder`. This refactor centralizes the shared logic into `qnn_bq_utils` and `qnn_utils`, making future changes (e.g., adding LPBQ support to new op builders) require only a single implementation.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
@@ -2104,6 +2104,150 @@ Ort::Status DequantizeInt32BiasToFp16(gsl::span<const uint8_t> raw_int32_bytes,
   return Ort::Status();
 }
 
+Ort::Status RequantizeBiasIfNeeded(QnnModelWrapper& qnn_model_wrapper,
+                                   const Ort::Logger& logger,
+                                   const OrtNodeUnitIODef& bias_def,
+                                   const TensorInfo& bias_info,
+                                   gsl::span<const float> weights_scales,
+                                   float activation_scale,
+                                   std::vector<std::string>& input_names,
+                                   bool& was_requantized) {
+  was_requantized = false;
+  int32_t bias_quant_axis = 0;
+  std::vector<float> current_scales;
+  std::vector<int32_t> current_offsets;
+  RETURN_IF_ERROR(GetBiasQuantScalesAndOffsets(bias_info.quant_param, current_scales,
+                                               current_offsets, bias_quant_axis));
+
+  const size_t num_channels = current_scales.size();
+  bool needs_requantization = false;
+  for (size_t i = 0; i < num_channels && !needs_requantization; ++i) {
+    const float w = (i < weights_scales.size()) ? weights_scales[i] : weights_scales[0];
+    if (current_offsets[i] != 0 ||
+        !CheckBiasScaleMatch(current_scales[i], w, activation_scale, 1e-5f)) {
+      needs_requantization = true;
+    }
+  }
+
+  if (!needs_requantization) {
+    return Ort::Status();  // Scales already match; caller processes bias normally.
+  }
+
+  ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, ("Requantizing bias " + bias_def.name).c_str());
+
+  std::vector<uint8_t> original_bias_data;
+  RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(bias_info.initializer_tensor, original_bias_data));
+
+  std::vector<uint8_t> new_bias_data;
+  std::vector<float> new_scales;
+  std::vector<int32_t> new_offsets;
+  const auto axis_opt = (num_channels > 1) ? std::optional<int64_t>(bias_quant_axis) : std::nullopt;
+  RETURN_IF_ERROR(RequantizeBiasTensor(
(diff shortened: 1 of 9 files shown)


onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -2104,6 +2104,170 @@ Ort::Status DequantizeInt32BiasToFp16(gsl::span<const uint8_t> raw_int32_bytes,
...
+  RETURN_IF_ERROR(RequantizeBiasTensor(
+      original_bias_data, bias_info.shape, current_scales, current_offsets,
+      weights_scales, activation_scale, bias_info.qnn_data_type,
+      new_bias_data, new_scales, new_offsets, axis_opt));
+
+  QnnQuantParamsWrapper new_quant_params = (new_scales.size() == 1)
+                                               ? QnnQuantParamsWrapper::PerTensor(new_scales[0], new_offsets[0])
+                                               : QnnQuantParamsWrapper::PerChannel(new_scales, new_offsets, bias_quant_axis);
+
+  const std::string rq_bias_name = bias_def.name + "_rq";
```
huaychou: **[M-1] (Major)** Rewritten bias tensor names are no longer unique and collide for shared initializers

Description

This refactor replaces the previous `UniqueNameGenerator().New(...)` names with fixed `bias_def.name + "_rq"`, `bias_def.name + "_q"`, and `bias_def.name + "_fp16"` names. When two Conv/Gemm consumers share one initializer, each consumer attempts to register the same QNN static-tensor name again; the rewritten bytes may also differ because activation/weight scales are consumer-specific. This violates the [C14]/[C53] shared-initializer rule and can either make the second `AddTensorWrapper` fail or make one consumer reuse another consumer's bias.

**Impact**: Valid shared-bias graphs may fail during graph construction/finalization or produce incorrect quantized results.

**Suggested fix:**

Restore `UniqueNameGenerator` with a deterministic suffix that distinguishes the consumer/context. For an unchanged bias that is safe to share, use the existing idempotent tensor-registration path. Add a regression test with two quantized Conv/Gemm consumers sharing one bias initializer while using different activation/weight scales.

onnxruntime/test/providers/qnn/matmul_test.cc
```diff
@@ -847,12 +847,8 @@ TEST_F(QnnHTPBackendTests, MatMulOp_QDQ_Regression_uint16_static_weight) {
         provider_options, 21, ExpectedEPNodeAssignment::All, QDQTolerance());
   }
 }
-#endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
-
-#if defined(__linux__)
 
 // Tests MatMul with ONNX block-quantized (BQ) weight using the BQ -> QNN LPBQ conversion path.
-// Currently BQ -> LPBQ conversion is only supported on Linux. It will be later enabled for windows as well.
 TEST_F(QnnHTPBackendTests, MatMulOp_QDQ_BlockQuant) {
```
huaychou: **[M-3] (Major)** Rank-2 bias and shared-helper behavior lack targeted regression coverage

The latest commit explicitly fixes rank-2 bias handling, while the shared helpers change tensor registration and bias quantization across Conv, MatMul, and MatMulNBits. However, the test diff only removes the Linux-only guard so the existing block-quant MatMul cases—documented as having no bias—also run on Windows ARM64. It adds no rank-2 `[1,N]` bias, matched/mismatched bias-scale, float-bias, or shared-initializer case.

**Impact**: The central behavior change and latest bug fix have no direct regression coverage, and existing tests cannot detect failures such as M-1.

**Suggested fix:**

Add MatMul/Gemm tests that exercise LPBQ Conv2D lowering with rank-2 bias, covering at least matched quantized scales, requantization, float bias, and two consumers sharing a bias initializer; verify both EP assignment and output accuracy.

minfhong-qti:
Could you help rebase on mainline and resolve the conflicts?

PR #759, opened 2026-08-24
[CI] Redirect docker container TMPDIR to repo volume to fix ENOSPC

huaychou:
### Description

- The repo is already mounted at /ort (host: /local/mnt, ~20 GB free). Set TMPDIR/TEMP/TMP inside DockerBuildAndTestTask to /ort/build/_docker_tmp and create that directory on the host before the container runs, so all temp churn lands on the roomy volume instead of /var.

### Motivation and Context

- Wheel packaging (bdist_wheel) calls Python's tempfile.mkdtemp() which defaults to /tmp inside the container.  The container's /tmp is backed by docker's overlay on /var, which is only ~1 GB free on CI runners. This causes OSError: [Errno 28] No space left on device when building the Python wheel.

### DIFF
--- a/qcom/ep_build/tasks/docker.py
+++ b/qcom/ep_build/tasks/docker.py
@@ -9,7 +9,7 @@ from typing import Any
 
 from ..task import RunExecutablesTask
 from ..typing import TargetPyVersionT
-from ..util import DEFAULT_PYTHON_LINUX
+from ..util import DEFAULT_PYTHON_LINUX, REPO_ROOT
 
 DOCKER_BUILD_USER = "ortqnnep"
 DOCKER_REPO_ROOT = Path("/ort")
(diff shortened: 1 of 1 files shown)


qcom/ep_build/tasks/docker.py
```diff
@@ -116,6 +123,9 @@ def __init__(
             "ORT_BUILD_PACKAGE_CACHE_PATH": "/ort_caches/packages",
             "ORT_BUILD_TOOLS_PATH": "/ort_caches/tools",
             "ORT_BUILD_PRUNE_PACKAGES": "0",
+            "TMPDIR": docker_tmp,
+            "TEMP": docker_tmp,
```
huaychou: [T-1 Nit] `TEMP` and `TMP` are Windows-convention variables. On Linux (including the `linux/aarch64` manylinux container), Python's `tempfile` module and Unix tools only honour `TMPDIR`; `TEMP`/`TMP` are silently ignored. These two entries have no effect and add noise — a future reader may mistake them for intentional cross-platform coverage.

**Fix**: remove the `"TEMP": docker_tmp` and `"TMP": docker_tmp` entries; `"TMPDIR": docker_tmp` alone is sufficient.

qcom/ep_build/tasks/docker.py
```diff
@@ -97,6 +97,13 @@ def __init__(
         # deferred to avoid circular import
         from ..tools import get_package_dir, get_tools_dir  # noqa:PLC0415
 
+        # Redirect Python's tempfile to the repo volume (/local/mnt) so wheel packaging
+        # doesn't write to the container's /tmp, which lands on docker's data-root on /var
+        # (only ~1 GB free on CI runners).
+        host_docker_tmp = REPO_ROOT / "build" / "_docker_tmp"
+        host_docker_tmp.mkdir(parents=True, exist_ok=True)
```
huaychou: [T-2 Nit] `host_docker_tmp.mkdir(parents=True, exist_ok=True)` creates the directory but `REPO_ROOT/build/_docker_tmp` is never purged after the task completes or fails. While the Docker container is ephemeral (`--rm`), this host-side directory persists across builds. If wheel packaging produces large temp files, repeated builds can grow this directory until the workaround itself becomes a source of disk pressure on CI runners.

**Fix**: either wipe stale contents at startup (`shutil.rmtree(host_docker_tmp, ignore_errors=True)` before the `mkdir`), add a CI post-step to clean up, or add a brief comment if `build/` is already covered by an existing CI teardown step.

huaychou:
No issue found.

PR #761, opened 2026-08-24
Prevent inf limits in clip op

qti-mattsinc:
* The GPU backend doesn't support min/max bounds to clip explicitly provided as +/- infinity. Work around this by clamping to the min/max finite floats instead.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc
@@ -196,6 +196,7 @@ Ort::Status ClipOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
   float min_value = std::numeric_limits<float>::lowest();
   if (num_inputs > 1 && !inputs[1].name.empty()) {
     RETURN_IF_ERROR(ProcessClipMinMax(qnn_model_wrapper, inputs[1], min_value));
+    min_value = std::max(min_value, std::numeric_limits<float>::lowest());
   }
   RETURN_IF_ERROR(AddQnnScalar<float>(qnn_model_wrapper, node_unit.Index(), node_unit.Name(), min_value,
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc
```diff
@@ -196,6 +196,7 @@ Ort::Status ClipOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mode
   float min_value = std::numeric_limits<float>::lowest();
   if (num_inputs > 1 && !inputs[1].name.empty()) {
     RETURN_IF_ERROR(ProcessClipMinMax(qnn_model_wrapper, inputs[1], min_value));
```
qti-yuduo: Curious is this something should be inside ProcessClipMinMax?

qti-yuduo:
Minor question otherwise LGTM.

onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc
qti-mattsinc: It could be, but right now `ProcessClipMinMax` doesn't know whether it's doing min or max. So in order to decide whether to do  `max(min, -FLT_MAX)` or `min(max, FLT_MAX)`  it would need an arg like `is_max`. I thought it was simpler just to leave it outside the process fn.

qti-yuduo: I see, it just looks incomplete. Adding a flag might be slightly less confusing. But fine either way.

qti-yuduo (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/clip_op_builder.cc
qti-mattsinc: I see what you mean. I have moved it inside `ProcessClipMinMax` with a flag in the latest commit. Also added an explanatory comment there. Thanks for reviewing!

PR #764, opened 2026-08-26
[QNN EP] Re-encode shifted INT16 RMSNorm scale for HTP

1duo:
## Problem
HTP rejects `SFIXED_POINT_16` RMSNorm gamma tensors with a nonzero QNN offset. AIMET can produce this encoding for nonnegative INT16 weights, blocking Qwen compilation.

## Fix
For static per-tensor INT16 gamma on HTP, re-encode the same values as UINT16 (`q_u = q_s + 32768`, `offset_u = offset_s - 32768`). The represented real values are unchanged. Other backends, dynamic gamma, symmetric INT16, and non-per-tensor encodings retain existing behavior.

Adds a focused HTP regression test for shifted INT16 gamma.

## Validation
- `git diff --check`
- `clang-format` on changed C++ files
- Full Qwen QDQ -> QNN DLC -> linked executable run on Snapdragon 8 Elite; CPU and device tokens both `[353, 1052, 11, 358]`

Local pre-commit lint bootstrap was unavailable because upstream `lintrunner==0.12.7` metadata is rejected by the current `uv`; remaining pre-commit checks passed.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
@@ -3,12 +3,14 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cstring>
 #include <optional>
 #include <string>
 #include <vector>
 
 #include "core/providers/qnn/builder/op_builder_factory.h"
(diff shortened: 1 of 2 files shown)


1duo:
Superseded by #773, which keeps the same validated re-encoding but removes the helper API, mutable out-parameters, and unnecessary folded-constant dependency.

PR #765, opened 2026-08-26
[QNN EP] Preserve symmetric U16 MatMul input encoding

1duo:
## Problem
QNN stores a per-tensor offset as the negated ONNX zero point. The dynamic U16 MatMul workaround compared that offset with `+32768`, so an already symmetric input (`zero_point = 32768`, QNN `offset = -32768`) was treated as asymmetric.

The inserted `Convert` is not semantically neutral: it derives a new symmetric encoding and can change the scale. This produced a measurable device-output difference.

## Fix
Compare with the correct QNN symmetric offset, `-32768`. Keep the existing `Convert` for asymmetric dynamic input 1.

The regression test asserts both branches in the dumped QNN graph:
- symmetric U16 input: zero `Convert` nodes
- asymmetric U16 input: one `Convert` node

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -378,8 +378,9 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnMatMul(QnnModelWrapper& qnn_mode
       input_1_shape = {input_info_1.shape[0], 1};
     }
     if (!input_info_1.is_initializer) {
-      // Only insert Convert for asymmetric quantization (i.e., offset != 2^(16-1)).
-      if (quant_param.scaleOffsetEncoding.offset != 32768) {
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
qti-mattsinc, line 347: Not really related to your change, but should this workaround be gated by `IsNpuBackend(qnn_model_wrapper.GetQnnBackendType())`? Some of the assumptions here (asym->sym Convert is needed; per-tensor quantization is required for the input) seem like they come from restrictions imposed by the HTP backend and not by the QNN API.

qti-yuduo: Sounds like a good follow up! -- to make thing easier to review, I'd keep it separately.

qti-mattsinc (approved):
LGTM but someone from the HTP side should also review

minfhong-qti (approved):

PR #766, opened 2026-08-26
Allow quantized uint8/int8 Conv on QNN CPU backend

mantpand:
## Summary

- Allow Conv operators with uint8/int8 inputs on the QNN CPU backend.
- Replace the previous float-only validation path for Conv with a Conv-specific datatype check.
- Keep supported quantized Conv nodes within QNN partitions.

## Motivation

The QNN CPU backend supports uint8/int8 Conv execution. The previous validation logic rejected non-float Conv inputs on the CPU backend, causing quantized Conv nodes to fall back to ORT's CPUExecutionProvider.

This resulted in excessive graph fragmentation and additional execution boundaries between QNN and CPU partitions. Allowing supported quantized Conv nodes to remain on QNN reduces unnecessary fallback and improves partitioning behavior.

## Scope

- Conv only
- QNN CPU backend only
- No changes to datatype validation behavior for other operators


qti-chuteng:
I think we can use this to determine whether an operator's inputs can successfully pass through qnn_model_wrapper.CreateQnnNode(). Please refer to the corresponding logic in base_op_builder.cc.

For example, see the implementation in conv_op_builder.cc:

https://github.com/onnxruntime/onnxruntime-qnn/blob/dc63106dd1fc7ecfafa5f7ab71d60bd2179a9c4a/onnxruntime/core/providers/qnn/builder/opbuilder/conv_op_builder.cc#L1274

PR #767, opened 2026-08-26
[QNN EP] Add ORT profiling timeline support

kuanyul-qti:
### Motivation and Context

QNN QAIRT profiling currently writes provider-specific CSV, `_qnn.log`, and ETW output only. ORT session and run profiling therefore cannot show or correlate QNN backend activity in the ORT JSON timeline.

This change integrates QNN QAIRT profiling with the ORT 1.25 plugin EP profiling API. It does not modify ORT core or add provider options.

### Description

- Add `QnnEpProfiler` to implement the plugin EP profiling lifecycle for session and run profiling.
- Add `QnnBackendProfilingManager` to centrally own QAIRT profile-handle lifetime, extraction routing, synchronization, and framework-op trace lookup.
- Emit eligible QNN compose, serial finalize, context_load, and execute QAIRT events into ORT JSON, with operation, graph, timing-source, event metadata, and parent ORT event information.
- Lazily create a BASIC QAIRT profile handle for ORT-only profiling when `profiling_level` is `off`.
- Preserve existing QNN provider profiling behavior, including CSV, `_qnn.log`, ETW, and framework-op trace output.
- Keep profiled shared-handle access serialized and defer execute extraction to `StopEvent`, so HTP detailed descendants are available and concurrent profiled runs remain isolated.
- Exclude SSR context recreation from provider and ORT profiling output; successful retry execution remains profiled.
- Add HTP coverage for ORT-only profiling, CSV coexistence, session setup attribution, AOT context loading, concurrent runs, default-off behavior, and DETAILED/OPTRACE output.
- Document ORT profiling timeline behavior, scope attribution, and the concurrency trade-off.

### Notes

- Concurrent profiled QNN runs serialize access to the single shared QAIRT profile handle by design.
- With neither ORT profiling nor QNN provider profiling enabled, QNN EP does not create a QAIRT profile handle or perform profiling extraction.


onnxruntime/core/providers/qnn/qnn_execution_provider.h
```diff
@@ -62,6 +62,10 @@ class QnnEp : public OrtEp, public ApiPtrs {
 
  private:
   static const char* ORT_API_CALL GetNameImpl(const OrtEp* this_ptr) noexcept;
+#if QNN_ORT_EP_PROFILING_API_ENABLED
+  static OrtStatus* ORT_API_CALL CreateProfilerImpl(_In_ OrtEp* this_ptr,
```
minfhong-qti: Prefer move to after `GetCompiledModelCompatibilityInfoImpl`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -27,6 +27,9 @@
 
 #include "core/providers/qnn/common/qnn_graph_utils.h"
 #include "core/providers/qnn/ort_api.h"
+#if QNN_ORT_EP_PROFILING_API_ENABLED
```
minfhong-qti: Is this macro necessary considering `qnn_ep_profiler.h` is already guarded? And please check for others.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1579,6 +1585,30 @@ const char* ORT_API_CALL QnnEp::GetNameImpl(const OrtEp* this_ptr) noexcept {
...
+/*static*/
+OrtStatus* ORT_API_CALL QnnEp::CreateProfilerImpl(OrtEp* this_ptr,
+                                                  OrtEpProfilerImpl** profiler) noexcept {
+  *profiler = nullptr;
+  auto* ep = static_cast<QnnEp*>(this_ptr);
+  if (!ep->qnn_backend_manager_) {
+    return nullptr;
+  }
+
+  try {
```
minfhong-qti, line 1729: FYI. This try-catch will be unified in #776. I think we may either skip for now and let #776 handle. As well for API impls in `qnn_ep_profiler.cc`.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
```diff
@@ -0,0 +1,130 @@
...
+#include <mutex>
+#include <string>
+#include <thread>
+#include <utility>
+#include <vector>
+
+#include <QnnInterface.h>
+#include <QnnProfile.h>
+
+#include "core/providers/qnn/builder/qnn_backend_manager.h"
```
minfhong-qti: Can as well forward declare without including the header here.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
```diff
@@ -0,0 +1,130 @@
...
+    void Activate() noexcept;
+    Ort::Status Reset();
+
+   private:
+    QnnBackendManager& backend_manager_;
+    bool active_ = false;
+  };
+
+  const OrtEpApi& ep_api_;
+  QnnEp& ep_;
```
minfhong-qti: It seems `ep_` is only used for acquiring `OrtApi`. Why not pass it directly?

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
```diff
@@ -0,0 +1,130 @@
...
+                                               uint64_t ort_event_correlation_id,
+                                               const OrtProfilingEvent* ort_event) noexcept;
+  static OrtStatus* ORT_API_CALL EndProfilingImpl(OrtEpProfilerImpl* this_ptr,
+                                                  OrtProfilingEventsContainer* events_container) noexcept;
+
+  Ort::Status DrainExecuteProfilingExtractions(std::vector<profile::ProfilingInfo> profiling_infos);
+  class OrtProfilingConsumer {
+   public:
+    explicit OrtProfilingConsumer(QnnBackendManager& backend_manager) : backend_manager_(backend_manager) {}
+    ~OrtProfilingConsumer() { ORT_UNUSED_PARAMETER(Reset()); }
```
minfhong-qti: Why wrap `ORT_UNUSED_PARAMETER`? For ignoring returning value? If so, probably better to have another macro with more intuitive naming.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
```diff
@@ -0,0 +1,130 @@
...
+
+    OrtProfilingConsumer(const OrtProfilingConsumer&) = delete;
```
minfhong-qti: Leverage `ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE`?

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,521 @@
...
+    self->ep_profiling_start_time_us_ = qnn::utils::GetTimeStampInUs();
+    // ORT calls StartProfiling before QNN EP initialization has necessarily reached GetCapability(),
+    // where SetupBackend() creates the backend and QAIRT interface. SetupBackend() creates an
+    // ORT-only BASIC handle for session-initialization events; if the backend was already set up,
+    // create it here instead. Initializing before either point would call profileCreate on a null backend.
+    if (self->backend_manager_.IsBackendSetup() && !self->backend_manager_.ProfilingEnabled()) {
+      Ort::Status status = self->backend_manager_.InitializeProfiling(ProfilingLevel::BASIC);
+      if (!status.IsOK()) {
+        ORT_UNUSED_PARAMETER(self->ort_profiling_consumer_.Reset());
+        return self->ep_.ort_api.CreateStatus(ORT_FAIL, status.GetErrorMessage().c_str());
```
minfhong-qti:
```suggestion
        return status.release();
```

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    OrtEventStart event_start = std::move(tls_ort_event_start_indices.back());
+    tls_ort_event_start_indices.pop_back();
+
+    if (Ort::Status status = self->DrainExecuteProfilingExtractions(
+            std::move(event_start.pending_execute_extractions));
+        !status.IsOK()) {
+      return self->ep_.ort_api.CreateStatus(ORT_FAIL, status.GetErrorMessage().c_str());
+    }
```
minfhong-qti:
```suggestion
    RETURN_IF_NOT_OK(self->DrainExecuteProfilingExtractions(std::move(event_start.pending_execute_extractions)));
```

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,521 @@
...
+/*static*/
+OrtStatus* ORT_API_CALL QnnEpProfiler::EndProfilingImpl(OrtEpProfilerImpl* this_ptr,
+                                                        OrtProfilingEventsContainer* events_container) noexcept {
+  auto* self = static_cast<QnnEpProfiler*>(this_ptr);
+  try {
+    auto release_ort_only_handle = [self]() -> OrtStatus* {
+      Ort::Status status = self->ort_profiling_consumer_.Reset();
+      if (!status.IsOK()) {
+        return self->ep_.ort_api.CreateStatus(ORT_FAIL, status.GetErrorMessage().c_str());
+      }
```
minfhong-qti:
```suggestion
      RETURN_IF_NOT_OK(self->ort_profiling_consumer_.Reset());
```

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -935,13 +939,24 @@ Ort::Status QnnBackendManager::ReleaseValidatorDevice() {
...
+  std::lock_guard<std::recursive_mutex> lock(profile_handle_mutex_);
+
+  // Idempotent: a handle already exists, so don't leak it with a second profileCreate.
+  if (profile_backend_handle_ != nullptr) {
+    return Ort::Status();
+  }
+
+  // Use a local effective level to avoid mutating profiling_level_ (would corrupt the ETW escalation guard).
+  const ProfilingLevel effective_level =
+      (override_level != ProfilingLevel::INVALID) ? override_level : profiling_level_;
```
minfhong-qti: Does this mean user-specified `profiling_level_` is not used when ORT profiling is enabled?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -995,18 +1011,128 @@ Ort::Status QnnBackendManager::InitializeProfiling() {
...
+  const auto& provider = QnnTelemetry::Instance();
+  return provider.IsEnabled() &&
+         (provider.Keyword() & static_cast<uint64_t>(qnn::ORTTraceLoggingKeyword::Profiling)) != 0 &&
+         provider.Level() >= 5;
+#else
+  return false;
+#endif
+}
+
+Ort::Status QnnBackendManager::GetProfileEventsLocked(const QnnProfile_EventId_t*& profile_events,
```
minfhong-qti: Just curious why extract as separate function but it seems not used elsewhere other than original place.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,521 @@
...
+  auto* self = static_cast<QnnEpProfiler*>(this_ptr);
+  self->ort_profiling_consumer_.Activate();
+  try {
+    self->ep_profiling_start_offset_ns_ = ep_profiling_start_offset_ns;
+    self->ep_profiling_start_time_us_ = qnn::utils::GetTimeStampInUs();
+    // ORT calls StartProfiling before QNN EP initialization has necessarily reached GetCapability(),
+    // where SetupBackend() creates the backend and QAIRT interface. SetupBackend() creates an
+    // ORT-only BASIC handle for session-initialization events; if the backend was already set up,
+    // create it here instead. Initializing before either point would call profileCreate on a null backend.
+    if (self->backend_manager_.IsBackendSetup() && !self->backend_manager_.ProfilingEnabled()) {
```
minfhong-qti: Based on the description above, would ever go into this branch as it seems backend would not never be setup when `StartProfiling` called?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -387,11 +390,44 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
   // Resets the context priority to the session default as defined by context_priority_
   Ort::Status ResetContextPriority();
 
-#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
-  bool ProfilingEnabled() { return profiling_enabled_; }
-#endif
+  bool ProfilingEnabled() const { return profiling_enabled_.load(std::memory_order_acquire); }
+  bool ProviderProfilingActive() const {
```
minfhong-qti: Please add some description explaining what these functions are for. For example, it's hard to understand what "Provider" and "Output" below mean.

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
```diff
@@ -26,16 +25,30 @@
...
 
+const std::string& GetUnitString(QnnProfile_EventUnit_t unit_type);
+std::string GetEventTypeString(QnnProfile_EventType_t event_type);
+
 struct ProfilingInfo {
   std::string graph_name = "";
   std::string csv_output_filepath = "";
+  // Host steady-clock timestamp captured immediately before the profiled QNN operation.
+  // It anchors QAIRT root/sub-event offsets to the EP clock used by OrtEpProfilerImpl.
+  uint64_t operation_start_time_us = 0;
```
minfhong-qti, line 40: What's the difference between `operation_start_time` and `start_time`? It seems they are both set to the same value.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
```diff
@@ -0,0 +1,130 @@
...
+                                       bool use_extended_event_data,
+                                       const profile::ProfilingInfo& profiling_info);
+  void QueueExecuteProfilingExtraction(profile::ProfilingInfo profiling_info);
+
+  size_t MarkPendingExecuteProfilingExtractions();
+  void DiscardPendingExecuteProfilingExtractionsSince(size_t mark);
+
+ private:
+  static void ORT_API_CALL ReleaseImpl(OrtEpProfilerImpl* this_ptr) noexcept;
+  static OrtStatus* ORT_API_CALL StartProfilingImpl(OrtEpProfilerImpl* this_ptr,
```
minfhong-qti, line 97: Since it's hard to follow how these functions interact, how about add a high level call flow explanation and example for how these functions are called, including where threading must take into consideration.

onnxruntime/core/providers/qnn/builder/qnn_profile_serializer.h
kuanyul-qti: Updated in the code.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
kuanyul-qti: Added.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Acknowledged.

onnxruntime/core/providers/qnn/qnn_execution_provider.h
kuanyul-qti: Done.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
kuanyul-qti: Added.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: This branch can happen when run profiling: the backend is already established during session initialization, but the run profiler starts afterward. The code explains both sequences and avoids a null backend `profileCreate`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
kuanyul-qti: The extracted extraction is now provided by the profiling manager, and is used in multiple call points such as compose, finalize, execute, context load, and OnRunStart.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -347,13 +348,32 @@ Ort::Status QnnModel::ComposeGraph(const QnnModelContext& context) {
   const bool build_json_graph = !context.json_qnn_graph_path.empty();
   RETURN_IF_NOT(qnn_model_wrapper.ComposeQnnGraph(build_json_graph), "Failed to compose Qnn graph.");
 
+#if QNN_ORT_EP_PROFILING_API_ENABLED
```
minfhong-qti: Just curious. Would including intermediate `AddToModelBuilder` parts affect the profiling results?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -201,12 +203,37 @@ Ort::Status QnnBackendManager::ParseLoraConfig(std::string lora_config_path) {
 
           graph_retrieve_success = true;
 
+          // OnRunStart has no thread-local ORT event scope. The manager only supplies a QAIRT
```
minfhong-qti: Just curious. Why add profiling here?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -1964,26 +1904,45 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
+    if (GetProfilingManager().OrtProfilingActive()) {
+      profiling_info.ort_profiler = QnnEpProfiler::Current();
+    }
+    if (profiling_info.ort_profiler != nullptr) {
+      profiling_info.graph_name = node_name;
+      profiling_info.operation_start_time_us = qnn::utils::GetTimeStampInUs();
+      profiling_info.ort_profiling_operation = "context_load";
+    }
+#endif
+    auto profiling_scope = GetProfilingManager().AcquireProfilingScope(profiling_info.ort_profiler != nullptr);
```
minfhong-qti: Since it is already named as scope, can we leverage RAII to wrap all profiling info setting work in it?
As for `GetProfilingManager().ExtractBackendProfilingInfo`, I think the check can be replaced by checking `profiling_info.graph_name` empty or not.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,518 @@
...
+        static_cast<const QnnEpProfiler*>(current));
+    if (profiler != nullptr && IsLiveProfiler(current)) {
+      return profiler;
+    }
+    tls_ort_event_start_indices.pop_back();
+  }
+  return nullptr;
+}
+
+Ort::Status QnnEpProfiler::AppendNewQnnEventRecords(QNN_INTERFACE_VER_TYPE qnn_interface,
```
minfhong-qti: Prefer any QNN API related calls are all in `qnn_backend_profiling_manager`. Possibly `qnn_backend_profiling_manager` should collect necessary info through QNN API and then pass in here.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.h
```diff
@@ -0,0 +1,176 @@
...
+  Ort::Status ExtractProfilingEventExtended(QnnProfile_EventId_t profile_event_id,
+                                            const std::string& event_level,
+                                            profile::Serializer& profile_writer);
+  static const char* QnnProfileErrorToString(QnnProfile_Error_t error);
+
+  QNN_INTERFACE_VER_TYPE& qnn_interface_;
+  Qnn_BackendHandle_t& backend_handle_;
+  QNN_SYSTEM_INTERFACE_VER_TYPE& qnn_system_interface_;
+  const QnnSerializerConfig* qnn_serializer_config_;
+  const Ort::Logger* const& logger_;
```
minfhong-qti: Prefer to pass logger whenever needed instead of cache one here.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -177,6 +175,19 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
         htp_power_config_manager_(power::HtpPowerConfigManager()),
         api_ptrs_(api_ptrs),
         logger_ptr_(&logger) {
+    profiling_manager_ = std::make_unique<QnnBackendProfilingManager>(
```
minfhong-qti: Prefer not to initialize here and it's a bit weird that below used members are in fact not initialized yet.
Considering that `QnnEpProfiler` in fact checks whether backend setup, I have two ideas:
1. Pass `QnnBackendManager` to `QnnEpProfiler` directly.
2. Pass reference of pointer.

Not sure which one is better or whether there is other options. Please consult with agent.

kuanyul-qti: Applied with a different solution. I did not pass the full `QnnBackendManager`, nor a reference to a nullable pointer, into `QnnEpProfiler`. That would broaden coupling or require nullable lifetime handling where the profiler needs a stable manager.

`QnnBackendManager::Create()` now constructs the backend manager first, then calls private `InitializeProfilingManager(config)`. Therefore, every referenced backend-manager member is fully initialized before the profiling manager is created, without adding any new `QnnBackendManager` data members.

`QnnEpProfiler` keeps the narrower `QnnBackendProfilingManager&` dependency. QAIRT profile creation remains gated on backend setup, preventing `profileCreate` against an uninitialized backend.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
kuanyul-qti: Removed.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
kuanyul-qti: Applied partially. `CreateScopedProfilingInfo()` now centralizes profiling-info initialization, and `QnnProfilingScope::Complete()` centralizes the corresponding end-time fields.
Synchronous compose/finalize/context-load extraction now checks `!profiling_info.graph_name.empty()`.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.h
kuanyul-qti: Applied. The profiling-manager dependency struct no longer stores a logger. Manager APIs receive const `Ort::Logger&` explicitly.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: Yes. The `compose` scope starts before QNN graph creation, includes `AddToModelBuilder`, and ends only after `ComposeQnnGraph()`. Excluding AddToModelBuilder would omit QNN graph-add work and fragment one synchronous compose operation. The scope is deliberately the complete QNN graph composition, not only the final compose call.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: Applied. QAIRT event traversal and reads (`profileGetEventData`, `profileGetExtendedEventData`, `profileGetSubEvents`) are now owned by `QnnBackendProfilingManager`. `QnnEpProfiler` only receives normalized records and appends ORT timeline events.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -691,13 +692,27 @@ Ort::Status QnnModel::BindAndExecuteGraph(OrtKernelContext* context,
...
 #ifdef QNN_SYSTEM_PROFILE_API_ENABLED
-  if (qnn_backend_manager_->ProfilingEnabled()) {
+  if (profiling_scope.Active()) {
     profiling_info.start_time = qnn::utils::GetTimeStampInUs();
   }
 #endif
-  auto profile_backend_handle = qnn_backend_manager_->GetQnnProfileHandle();
+  auto profile_backend_handle = profiling_scope.Handle();
 
   auto thread_id = std::this_thread::get_id();
```
huaychou, line 707: should we use CurrentThreadId?

huaychou (commented):
Do we need to add corresponding component tests under unit?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: I kept `std::this_thread::get_id()` intentionally. This value is used as the key for `SetPerThreadHtpPowerConfigs()`, whose API and backing map are both keyed by `std::thread::id`. `CurrentThreadId()` returns an `int32_t` hash for ORT timeline records and can collide, so it is not appropriate for the power-config mapping.

kuanyul-qti:
> Do we need to add corresponding component tests under unit?

Added.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.h
```diff
@@ -0,0 +1,195 @@
...
+  }
+  bool OrtProfilingActive() const { return HasActiveOrtProfilingConsumer(); }
+  Ort::Status ReleaseOrtProfilingHandleIfUnused();
+
+  // ORT profiling is correlated through a thread-local profiler scope. Work without that scope
+  // must not be appended to ORT; parallel finalization is intentionally provider-output only.
+  QnnProfilingScope AcquireProfilingScope(bool current_operation_has_ort_profiler = false);
+  QnnProfilingScope CreateScopedProfilingInfo(profile::ProfilingInfo& profiling_info,
+                                              const std::string& graph_name,
+                                              const char* ort_profiling_operation
```
minfhong-qti: Define similar enum for ORT profiling like `ProfilingMethodType`, possibly in `qnn_ep_profiler.h`.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
```diff
@@ -0,0 +1,624 @@
...
+
+  std::unique_lock<std::recursive_mutex> lock(profile_handle_mutex_);
+  return QnnProfilingScope(std::move(lock), profile_handle_);
+}
+
+QnnProfilingScope QnnBackendProfilingManager::CreateScopedProfilingInfo(
+    profile::ProfilingInfo& profiling_info,
+    const std::string& graph_name,
+    const char* ort_profiling_operation
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
```
minfhong-qti: I think `ProfilingMethodType` can always be given and only guarded below instead of in caller side.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
```diff
@@ -0,0 +1,624 @@
...
+    profiling_info.operation_end_time_us = qnn::utils::GetTimeStampInUs();
+  }
+#endif
+
+#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
+  if (Active()) {
+    profiling_info.stop_time = qnn::utils::GetTimeStampInUs();
+  }
+#else
+  ORT_UNUSED_PARAMETER(profiling_info);
```
minfhong-qti: Not sure but is it better to use `[[maybe unsused]]`` as these `ORT_UNUSED_PARAMETER` are getting annoyed.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2016,26 +1948,27 @@ Ort::Status QnnBackendManager::LoadCachedQnnContextFromBuffer(
...
+                                                  io_dispatch, profiling_scope.Handle(), context));
 
-#ifdef QNN_SYSTEM_PROFILE_API_ENABLED
-    if (ProfilingEnabled()) {
-      profiling_info.stop_time = qnn::utils::GetTimeStampInUs();
-      profiling_info.method_type = ProfilingMethodType::CREATE_FROM_BINARY;
-      profiling_info.graph_name = node_name;
-    }
-#endif
+    profiling_scope.Complete(profiling_info);
```
minfhong-qti, line 1963: Is it reasonable to include `ExtractBackendProfilingInfo` in `Complete`?

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -691,13 +692,27 @@ Ort::Status QnnModel::BindAndExecuteGraph(OrtKernelContext* context,
 
   ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE, ("Start execute QNN graph:" + graph_info_->Name()).c_str());
 
+  QnnProfilingScope profiling_scope;
+  RETURN_IF_ERROR(qnn_backend_manager_->GetProfilingManager().CreateGraphProfilingScope(
```
minfhong-qti, line 697: Just curious why `CreateScopedProfilingInfo` not applicable here?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
kuanyul-qti: I do not think we should do that.
`Complete()` only finalizes timing metadata and currently has no error return path, while `ExtractBackendProfilingInfo()` can fail because it calls QNN profiling APIs and provider output serialization. Keeping extraction explicit in the caller preserves correct error handling and caller-specific ordering.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
kuanyul-qti: Fixed. `CreateScopedProfilingInfo()` now always takes `ProfilingMethodType`, and the `QNN_SYSTEM_PROFILE_API_ENABLED` guard remains only around the system-profile assignment inside the helper. The caller-side `#ifdef` blocks were removed.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.h
kuanyul-qti: Added `OrtProfilingOperation` in `qnn_def.h` and changed `ProfilingInfo::ort_profiling_operation` from a string to that enum.
`OrtProfilingOperation` is stored in `ProfilingInfo`, and `ProfilingInfo` is used outside `qnn_ep_profiler.h`. `qnn_ep_profiler.h` is guarded by `QNN_ORT_EP_PROFILING_API_ENABLED`, so putting the enum there would make non-ORT-profiling builds harder to compile cleanly. `qnn_def.h` already contains shared profiling enums like `ProfilingMethodType`, so it is the simplest shared place.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: It is not directly applicable to execute.
The execute path needs `CreateGraphProfilingScope()` because ORT-only execute profiling may need to lazily recreate the QAIRT profile handle after a previous extraction released it, and that operation can fail, so the helper must return `Ort::Status`. Execute also queues profiling extraction until ORT `StopEvent` so the shared QAIRT handle stays isolated for the active ORT event scope.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.h
kuanyul-qti: Moved the enum to `qnn_ep_profiler.h` and placed it outside the `QNN_ORT_EP_PROFILING_API_ENABLED` guard.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: `CreateGraphProfilingScope` now also handles the `profiling_info` setting.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -800,17 +804,26 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
     ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_ERROR,
                 "NPU crashed. SSR detected during QNN graph execute.");
     if (!context_bin_filepath_.empty()) {
       ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING, "Attempting SSR recovery.");
       RETURN_IF_ERROR(RecoverFromSSR(logger, io_dispatch));
 
       // Retry once with fresh context and re-bound tensors.
-      RETURN_IF_ERROR(BindAndExecuteGraph(context, logger, execute_status));
+      RETURN_IF_ERROR(BindAndExecuteGraph(context, logger, execute_status, ort_profiler));
       if (QNN_COMMON_ERROR_SYSTEM_COMMUNICATION == execute_status) {
```
minfhong-qti, line 831: Should here `DiscardPendingExecuteProfilingExtractionsSince` again?

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,446 @@
...
+}
+
+void QnnEpProfiler::QueueExecuteProfilingExtraction(profile::ProfilingInfo profiling_info) {
+  if (!tls_ort_event_start_indices.empty() && tls_ort_event_start_indices.back().profiler == this) {
+    auto& event_start = tls_ort_event_start_indices.back();
+    if (!event_start.profile_handle_lock.owns_lock()) {
+      // The caller already holds the same recursive lock through graphExecute. Retain an
+      // additional ownership level until StopEvent drains this run's QAIRT events, preventing a
+      // concurrent profiled run from appending events to this profiler's extraction.
+      event_start.profile_handle_lock = profiling_manager_.AcquireProfileHandleLock();
```
minfhong-qti, line 135: Just to confirm whether `StopEvent` is always called to make sure the lock is always released expectedly.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1503,6 +1505,32 @@ Additionally, if the profiling_level is set to "detailed" or "optrace", addition
...
+
+If `profiling_level` is not set, enabling ORT profiling activates QNN profiling at the `basic` level. Run profiling records only `execute` events. Session profiling additionally records synchronous graph `compose`, serial graph `finalize`, and synchronous EPContext `context_load` events under ORT's `session_initialization` event. QAIRT events use `qnn_timing_source=BACKEND`; when QAIRT does not report a setup operation, QNN emits one explicitly marked `HOST_OPERATION` event with `qnn_timing_source=HOST`. Parallel graph finalization is excluded because it runs on QNN worker threads outside ORT's thread-local profiling scope. SSR context recreation is retry work and is not profiled, so it is excluded from provider output and the ORT JSON timeline. Existing `profiling_file_path` CSV, `_qnn.log`, and ETW outputs remain available and can be used alongside the ORT timeline.
```
minfhong-qti: How parallel finalization is omitted in profiling as I did not see in code?

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
```diff
@@ -0,0 +1,640 @@
...
+}
+
+Ort::Status QnnBackendProfilingManager::ReleaseOrtProfilingHandleIfUnused() {
+  if (ProviderProfilingActive() || HasActiveOrtProfilingConsumer()) {
+    return Ort::Status();
+  }
+  return ReleaseProfileHandle();
+}
+
+QnnProfilingScope QnnBackendProfilingManager::AcquireProfilingScope(
```
minfhong-qti, line 211: The logic here is greatly duplicated in `CreateGraphProfilingScope`. Please consult with Agent to unify the logic.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,446 @@
...
+}
+
+Ort::Status QnnEpProfiler::DrainExecuteProfilingExtractions(std::vector<profile::ProfilingInfo> profiling_infos) {
+  for (auto& profiling_info : profiling_infos) {
+    RETURN_IF_ERROR(profiling_manager_.ExtractBackendProfilingInfo(profiling_info, logger_));
+  }
+  return Ort::Status();
+}
+
+QnnEpProfiler* QnnEpProfiler::Current() noexcept {
```
minfhong-qti, line 148: `Current()`'s liveness check keys a global registry by raw pointer address, not object identity. If a `QnnEpProfiler` is freed and the allocator reuses its address for a new session's profiler before a stale thread-local entry elsewhere is popped, `Current()` can hand a caller the wrong session's profiler pointer; the check-then-use pattern is also not protected against a concurrent `Release`. Replace raw-address identity with a generation counter or a `weak_ptr`-based liveness check.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
```diff
@@ -0,0 +1,640 @@
...
+#include "Saver/QnnSaver.h"
+
+#include "core/providers/qnn/builder/qnn_utils.h"
+#include "core/providers/qnn/qnn_ep_profiler.h"
+#include "core/providers/qnn/qnn_telemetry.h"
+
+namespace {
+
+bool IsQnnProfileSuccess(Qnn_ErrorHandle_t status) noexcept {
+  return static_cast<QnnProfile_Error_t>(status & 0xFFFF) == QNN_PROFILE_NO_ERROR;
```
minfhong-qti: Define a function for `static_cast<QnnProfile_Error_t>(status & 0xFFFF)` and resue.

docs/execution_providers/QNN-ExecutionProvider.md
kuanyul-qti: Parallel finalization is submitted to worker threads. Those threads have no initialization-thread TLS ORT profiler scope. For ORT-only profiling, scope acquisition returns inactive; `ProfilingInfo::graph_name` remains empty and `FinalizeGraphs()` skips extraction. When CSV/ETW is configured, the same worker work remains provider-output-only.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
kuanyul-qti: `GetQnnProfileError()` now performs the `status & 0xFFFF` conversion once and is used by the success helper plus the basic, extended, and sub-event extraction paths.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.cc
kuanyul-qti: The common acquisition policy is now shared by `AcquireProfilingScope()`.

A complete merge is not suitable because setup and graph execute have different contracts:

- QAIRT-only graph execution must keep its direct unlocked path and initialize provider metadata. Routing it through an additional helper would add work to the QAIRT-only inference path.
- Background work with an active ORT consumer but no current TLS ORT scope must return without graph metadata when no CSV/ETW sink exists. This prevents it from being attributed to a later ORT JSON event.
- Current-ORT graph execution must lazily create the QAIRT handle when needed, propagate initialization failures, and retain the lock through `StopEvent()`. Setup does not have this lifetime requirement.

Therefore, the remaining graph branch is not duplicated acquisition logic; it selects the required metadata and lock-lifetime behavior after the shared acquisition policy. This keeps profiling-off and QAIRT-only inference paths unchanged.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: Added.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: The ORT guarantees one `StopEvent()` for every `StartEvent()` with the same correlation ID, and also guarantees that `StartEvent()`, `Compute()`, and `StopEvent()` for an event run on the same CPU thread. The retained `unique_lock` is owned by the TLS event entry; `StopEvent()` moves that entry to a local object, so it is released both on the normal path and on an early error return during drain.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: The TLS stack no longer identifies a profiler by a raw address registered globally. Each profiler owns a distinct `ProfilerLifetimeToken`; TLS retains that token, and `Current()` reads its atomic profiler pointer. A destroyed profiler clears its token, so a stale TLS entry cannot become associated with a new profiler that happens to reuse the same address.

minfhong-qti (approved):

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1445,6 +1445,8 @@ Profiling data is available with the HTP backend. Enabling QNN profiling will ge
 
 If onnxruntime is compiled with a more recent QAIRT SDK (2.39 or later), then a _qnn.log file will also be generated alongside the .csv file. This .log file is parsable by [qnn-profile-viewer](https://docs.qualcomm.com/doc/80-63442-10/topic/general_tools.html#qnn-profile-viewer), which is provided in the SDK.
 
+ORT profiling is supported on CPU and HTP backends through the unified JSON timeline described below. It is independent of provider CSV output.
```
qti-chuteng: Do we not support GPU?

kuanyul-qti: The implementation is backend-agnostic at the QNN EP layer: ORT profiling and provider QAIRT profiling both use the same `QnnBackendProfilingManager` / QAIRT profile-handle path regardless of whether the selected backend is CPU, HTP, or GPU.
Updated the word to avoid limiting ORT profiling to CPU/HTP.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
```diff
@@ -0,0 +1,147 @@
...
+  OrtProfilingEventCategory category = OrtProfilingEventCategory_KERNEL;
+  int32_t process_id = 0;
+  int32_t thread_id = 0;
+  std::string name;
+  std::vector<std::pair<std::string, std::string>> args;
+
+  uint64_t qairt_ts_us = 0;       // device-clock timestamp (0 for basic events)
+  int64_t qairt_duration_us = 0;  // duration for MICROSEC events; 0 otherwise
+
+  int64_t ort_event_start_us = -1;  // set by StopEvent; -1 means unanchored
```
qti-chuteng: Can it be replaced by below statement?

bool parent_ort_event_assigned = false;

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
```diff
@@ -0,0 +1,147 @@
...
+namespace onnxruntime {
+namespace qnn {
+
+class QnnBackendProfilingManager;
+struct ProfilerLifetimeToken;
+namespace profile {
+struct ProfilingInfo;
+}  // namespace profile
+
+inline int32_t CurrentThreadId() noexcept {
```
qti-chuteng: This function only be used in qnn_ep_profiler.cc, we can move it to .cc

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
```diff
@@ -0,0 +1,427 @@
...
+    active_ = true;
+  }
+}
+
+Ort::Status QnnEpProfiler::OrtProfilingConsumer::Reset() {
+  if (!active_) {
+    return Ort::Status();
+  }
+
+  profiling_manager_.ReleaseOrtProfilingConsumer();
```
qti-chuteng: ### Potential Risk

  The current shutdown path performs two separate operations:

  ReleaseOrtProfilingConsumer();          // Decrement count
  ReleaseOrtProfilingHandleIfUnused();    // Check count, then free handle

  A new profiler can start between these calls:

  Thread A: count 1 → 0; decides to free handle H1
  Thread B: count 0 → 1; sees H1 enabled and skips initialization
  Thread A: frees H1

  Result:
    active consumer count = 1
    profile_handle_ = nullptr

  This inconsistent state can silently omit compose, finalize, or context_load events. Execute profiling may later recover through
  lazy initialization, making the race difficult to detect.

  ### Suggested Solution

  Move consumer registration and handle management into transactional APIs:

  Ort::Status AcquireOrtProfilingConsumer(const Ort::Logger& logger);
  Ort::Status ReleaseOrtProfilingConsumer();

  Both operations must hold profile_handle_mutex_ while updating the count and handle:

  // Acquire
  lock();
  ++active_count;

  if (backend_setup && profile_handle_ == nullptr) {
    InitializeProfilingForCurrentConsumers(logger);
  }

  unlock();

  // Release
  lock();
  --active_count;

  if (active_count == 0 && !ProviderProfilingActive()) {
    ReleaseProfileHandle();
  }

  unlock();

  Additionally:

  - Remove ReleaseOrtProfilingHandleIfUnused() so callers cannot split the transaction.
  - Make OrtProfilingConsumer::Activate() return Ort::Status.
  - Roll back the consumer count if handle initialization fails.
  - Add a barrier-controlled concurrency test and verify:

onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -800,18 +808,32 @@ Ort::Status QnnModel::ExecuteGraph(OrtKernelContext* context,
...
+#if QNN_ORT_EP_PROFILING_API_ENABLED
+  const size_t pending_extraction_mark =
+      ort_profiler ? ort_profiler->MarkPendingExecuteProfilingExtractions() : 0;
+#endif
+  RETURN_IF_ERROR(BindAndExecuteGraph(context, logger, execute_status, ort_profiler));
 
   if (QNN_COMMON_ERROR_SYSTEM_COMMUNICATION == execute_status) {
+#if QNN_ORT_EP_PROFILING_API_ENABLED
+    if (ort_profiler) {
+      ort_profiler->DiscardPendingExecuteProfilingExtractionsSince(pending_extraction_mark);
```
qti-chuteng, line 820: ### Potential Risk

  The SSR path currently discards only queued extraction metadata:

  const size_t mark =
      profiler->MarkPendingExecuteProfilingExtractions();

  BindAndExecuteGraph(...);  // Returns SYSTEM_COMMUNICATION

  profiler->DiscardPendingExecuteProfilingExtractionsSince(mark);

  RecoverFromSSR(...);
  BindAndExecuteGraph(...);  // Successful retry

  The failed graphExecute() may already have written events into the shared QAIRT profile handle. Removing its ProfilingInfo does not clear those backend events.

  At StopEvent(), the retry extraction may therefore return:

  failed-attempt events
  +
  successful-retry events

  Both sets may be attributed to the successful ORT parent event, causing duplicate root events, inflated durations, or misleading operator traces.

  ### Suggested Solution

  First confirm whether QAIRT guarantees that QNN_COMMON_ERROR_SYSTEM_COMMUNICATION leaves no readable profiling events.

  If stale events may remain, add a manager API:

  Ort::Status ResetProfilingAfterFailedExecute(
      const Ort::Logger& logger);

  Under profile_handle_mutex_, it should:

  1. Discard, or extract and ignore, failed-attempt events.
  2. Release or invalidate the old profile handle.
  3. Recreate a clean handle if an ORT or provider consumer remains active.
  4. Run the retry using only the new handle.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.h
```diff
@@ -0,0 +1,209 @@
...
+ public:
+  QnnBackendProfilingManager(QnnBackendProfilingManagerDependencies dependencies,
+                             ProfilingLevel provider_profiling_level,
+                             ProfilingLevel etw_profiling_level,
+                             std::string profiling_file_path,
+                             bool enable_framework_op_trace);
+
+  ORT_DISALLOW_COPY_ASSIGNMENT_AND_MOVE(QnnBackendProfilingManager);
+
+  bool HasProfileHandle() const { return profile_handle_ != nullptr; }
```
qti-chuteng: bool HasProfileHandle() const {
  std::lock_guard<std::recursive_mutex> lock(
      profile_handle_mutex_);
  return profile_handle_ != nullptr;
}

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
qti-chuteng: ReleaseOrtProfilingConsumer() / ReleaseOrtProfilingHandleIfUnused() can be merged into one atomic function.

onnxruntime/core/providers/qnn/builder/qnn_backend_profiling_manager.h
kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/builder/qnn_model.cc
kuanyul-qti: The original mark is enough for queued `ProfilingInfo`. The retry execute reuses the handle, and QAIRT documents that reuse resets the handle, so failed-attempt backend events should not be extracted with the retry.

onnxruntime/core/providers/qnn/qnn_ep_profiler.cc
kuanyul-qti: Fixed.

kuanyul-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
kuanyul-qti: Moved.

onnxruntime/core/providers/qnn/qnn_ep_profiler.h
kuanyul-qti: Done.

qti-chuteng (approved):

PR #768, opened 2026-08-26
[Docs] Split Compiled Model Encryption into its own top-level section

tzuhsuanwei:
## Summary
- Promotes the at-rest context-binary encryption/decryption docs from a subsection under "QNN context binary cache feature" into an independent, linkable top-level section — same pattern as the LoRA support section, with its own Contents entry.
- Adds a reference to the upstream ORT core feature this maps to: [microsoft/onnxruntime#28624](https://github.com/microsoft/onnxruntime/pull/28624) and its [design doc](https://github.com/microsoft/onnxruntime/blob/main/docs/design/Compiled_Model_Encryption.md).
- Keeps the existing content (encrypt/decrypt C++ callback examples, scope/fallback notes) unchanged, just restructured.

Docs-only change; no code touched.

### DIFF
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -21,6 +21,7 @@ ONNX Runtime QNN EP can be used on Windows devices with Qualcomm Snapdragon SOC'
 - [Running a model with QNN EP's GPU backend](#running-a-model-with-qnn-eps-gpu-backend)
 - [Running an LLM model with QNN EP's Genie backend](#running-an-llm-model-with-qnn-eps-genie-backend)
 - [QNN context binary cache feature](#qnn-context-binary-cache-feature)
+- [Compiled Model Encryption](#compiled-model-encryption)
 - [QNN EP Framework Op Tracing](#qnn-ep-framework-op-tracing)
 - [QNN EP Input Graph Dump](#qnn-ep-input-graph-dump)
 - [QNN EP Profiling](#qnn-ep-profiling)
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
 options.add_session_config_entry("ep.context_embed_mode", "1")
 ```
 
-### At-rest encryption of the context binary (ORT API v28+)
+## Compiled Model Encryption
 
 By default the QNN context binary is written to / read from disk in plaintext (as an external
 file when `ep.context_embed_mode` is `"0"`, or embedded in the EPContext model when `"1"`). ORT
 API v28 adds a write/read callback pair so an application can encrypt the binary before it is
 persisted and decrypt it before QNN consumes it, without either version of ONNX Runtime having
-any built-in cipher.
+any built-in cipher. This maps to the ORT core feature added in
+[microsoft/onnxruntime#28624](https://github.com/microsoft/onnxruntime/pull/28624) — see its
+[design doc](https://github.com/microsoft/onnxruntime/blob/main/docs/design/Compiled_Model_Encryption.md)
+for the full API rationale.
 
-The callback-based encryption/decryption path described here applies only when
-`ep.context_embed_mode` is `"0"`. When `ep.context_embed_mode` is `"1"` (embedded), the context
-binary is stored within the EPContext model and applications should use ONNX Runtime model
-protection mechanisms (for example, `SetOutputModelWriteFunc`) to protect the generated ONNX
-model. If no callback is registered while `ep.context_embed_mode` is `"0"`, the EP uses the
-legacy plaintext behavior.
+**Scope: these callbacks only apply when `ep.context_embed_mode` is `"0"`.** When
+`ep.context_embed_mode` is `"1"` (embedded), the context binary is stored inside the EPContext
+ONNX model itself, and there is no separate context-binary file for these callbacks to intercept.
+To protect the model in that mode, use ONNX Runtime's model protection mechanisms instead (for
+example, `SetOutputModelWriteFunc`) to encrypt the generated ONNX model. **The rest of this
+section assumes `ep.context_embed_mode` is `"0"`.**
 
-- **Write callback** (`SetEpContextDataWriteFunc`) is set on `Ort::ModelCompilationOptions`
-  during a compile session. The EP hands the plaintext context bytes to the callback instead of
-  writing them to disk; the callback is responsible for encrypting and persisting them itself.
-- **Read callback** (`SetEpContextDataReadFunc`) is set on `Ort::SessionOptions` for a later
-  inference session. The EP asks the callback for the plaintext bytes (by name) instead of
-  reading the on-disk file directly; the callback decrypts and returns them.
-- Registering **no callback** is fully backward compatible — the EP falls back to the original
-  plaintext disk read/write, unchanged.
+**Encrypt (write callback)** — set on `Ort::ModelCompilationOptions` during compile. ORT hands
+the plaintext context bytes to the callback instead of writing them to disk. The example below
+uses a single-byte XOR so the callback mechanism is easy to follow — replace the body of
+`WriteCb` with your own cipher (AES-GCM, a KMS/TEE-backed cipher, etc.); everything else about
+how ORT calls the callback stays the same:
+
+```cpp
+// C++
+constexpr uint8_t g_key = 0x5A;  // must match the key ReadCb uses below
+
+OrtStatus* WriteCb(void* state, const char* name, const void* buffer, size_t n) noexcept {
+  // >>> Replace this block with your own encryption. <<<
+  std::ofstream out(name, std::ios::binary);
+  for (size_t i = 0; i < n; ++i) {
+    out.put(static_cast<const uint8_t*>(buffer)[i] ^ g_key);
+  }
+  return out ? nullptr : Ort::GetApi().CreateStatus(ORT_FAIL, "write failed");
+}
+
+Ort::ModelCompilationOptions compile_options(env, session_options);
+compile_options.SetEpContextEmbedMode(false);
+Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(
+    &Ort::GetApi())(compile_options, WriteCb, /*state=*/nullptr);
+Ort::CompileModel(env, compile_options);
+```
+
+**Decrypt (read callback)** — set on `Ort::SessionOptions` before creating the inference session.
+ORT asks the callback for the plaintext bytes (by name) instead of reading the file directly.
+Replace the body of `ReadCb` with the matching decryption for whatever cipher `WriteCb` used:
+
+```cpp
+// C++
+OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
+                  void** buffer, size_t* size) noexcept {
+  // >>> Replace this block with your own decryption. <<<
+  std::ifstream in(name, std::ios::binary | std::ios::ate);
+  if (!in) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
+  std::streamoff n_signed = in.tellg();
+  if (n_signed < 0) return Ort::GetApi().CreateStatus(ORT_FAIL, "tellg failed");
+  size_t n = static_cast<size_t>(n_signed);
+  in.seekg(0);
+  if (!in) return Ort::GetApi().CreateStatus(ORT_FAIL, "seekg failed");
+  void* mem = allocator->Alloc(allocator, n);
+  if (mem == nullptr) return Ort::GetApi().CreateStatus(ORT_FAIL, "alloc failed");
+  in.read(static_cast<char*>(mem), static_cast<std::streamsize>(n));
+  if (static_cast<size_t>(in.gcount()) != n) {
+    allocator->Free(allocator, mem);
+    return Ort::GetApi().CreateStatus(ORT_FAIL, "read failed");
+  }
+  for (size_t i = 0; i < n; ++i) static_cast<uint8_t*>(mem)[i] ^= g_key;
+  *buffer = mem;
+  *size = n;
+  return nullptr;
+}
+
+Ort::SessionOptions session_options;
+Ort::Experimental::Get_OrtApi_SessionOptions_SetEpContextDataReadFunc_SinceV28_Fn(
+    &Ort::GetApi())(session_options, ReadCb, /*state=*/nullptr);
+Ort::Session session(env, ctx_model_path, session_options);
+```
+
+If no callback is registered, the EP uses the legacy plaintext behavior.
+
 - File-mapped weights and this feature are mutually exclusive for a given session: registering a
   read callback disables file mapping for that session, since file mapping requires reading the
   on-disk bytes directly.
 
-**Scope — what is (and isn't) covered:**
+**What is (and isn't) covered:**
 
 - Only the standard EPContext artifact is encrypted: the external `_qnn.bin` (or the
   `EP_CACHE_CONTEXT`-referenced buffer) and the context-binary-list buffers used by multi-SoC /
--- a/docs/execution_providers/QNN-ExecutionProvider.md
+++ b/docs/execution_providers/QNN-ExecutionProvider.md
@@ -21,7 +21,7 @@ ONNX Runtime QNN EP can be used on Windows devices with Qualcomm Snapdragon SOC'
 - [Running a model with QNN EP's GPU backend](#running-a-model-with-qnn-eps-gpu-backend)
 - [Running an LLM model with QNN EP's Genie backend](#running-an-llm-model-with-qnn-eps-genie-backend)
 - [QNN context binary cache feature](#qnn-context-binary-cache-feature)
-- [Compiled Model Encryption](#compiled-model-encryption)
+- [Compiled Model Encryption (ORT API v28+)](#compiled-model-encryption-ort-api-v28)
 - [QNN EP Framework Op Tracing](#qnn-ep-framework-op-tracing)
 - [QNN EP Input Graph Dump](#qnn-ep-input-graph-dump)
 - [QNN EP Profiling](#qnn-ep-profiling)
@@ -1194,7 +1194,7 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
 options.add_session_config_entry("ep.context_embed_mode", "1")
 ```
 
-## Compiled Model Encryption
+## Compiled Model Encryption (ORT API v28+)
 
 By default the QNN context binary is written to / read from disk in plaintext (as an external
 file when `ep.context_embed_mode` is `"0"`, or embedded in the EPContext model when `"1"`). ORT
@@ -1225,6 +1225,7 @@ constexpr uint8_t g_key = 0x5A;  // must match the key ReadCb uses below
 OrtStatus* WriteCb(void* state, const char* name, const void* buffer, size_t n) noexcept {
   // >>> Replace this block with your own encryption. <<<
   std::ofstream out(name, std::ios::binary);
+  if (!out.is_open()) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
   for (size_t i = 0; i < n; ++i) {
     out.put(static_cast<const uint8_t*>(buffer)[i] ^ g_key);
   }
@@ -1233,8 +1234,13 @@ OrtStatus* WriteCb(void* state, const char* name, const void* buffer, size_t n) 
 
 Ort::ModelCompilationOptions compile_options(env, session_options);
 compile_options.SetEpContextEmbedMode(false);
-Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(
-    &Ort::GetApi())(compile_options, WriteCb, /*state=*/nullptr);
+auto write_fn = Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(
+    &Ort::GetApi());
+if (write_fn == nullptr) {
+  // ORT core predates API v28; this feature is unavailable.
+  return;
+}
+Ort::ThrowOnError(write_fn(compile_options, WriteCb, /*state=*/nullptr));
 Ort::CompileModel(env, compile_options);
 ```
 
@@ -1244,6 +1250,8 @@ Replace the body of `ReadCb` with the matching decryption for whatever cipher `W
 
 ```cpp
 // C++
+constexpr uint8_t g_key = 0x5A;  // must match the key WriteCb used above
+
 OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
                   void** buffer, size_t* size) noexcept {
   // >>> Replace this block with your own decryption. <<<
@@ -1267,10 +1275,15 @@ OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
   return nullptr;
 }
 
-Ort::SessionOptions session_options;
-Ort::Experimental::Get_OrtApi_SessionOptions_SetEpContextDataReadFunc_SinceV28_Fn(
-    &Ort::GetApi())(session_options, ReadCb, /*state=*/nullptr);
-Ort::Session session(env, ctx_model_path, session_options);
+Ort::SessionOptions inference_session_options;
+auto read_fn = Ort::Experimental::Get_OrtApi_SessionOptions_SetEpContextDataReadFunc_SinceV28_Fn(
+    &Ort::GetApi());
+if (read_fn == nullptr) {
+  // ORT core predates API v28; this feature is unavailable.
+  return;
+}
+Ort::ThrowOnError(read_fn(inference_session_options, ReadCb, /*state=*/nullptr));
+Ort::Session session(env, ctx_model_path, inference_session_options);
 ```
 
 If no callback is registered, the EP uses the legacy plaintext behavior.


minfhong-qti (approved):

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,86 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+ORT asks the callback for the plaintext bytes (by name) instead of reading the file directly.
+Replace the body of `ReadCb` with the matching decryption for whatever cipher `WriteCb` used:
+
+```cpp
+// C++
+OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
+                  void** buffer, size_t* size) noexcept {
+  // >>> Replace this block with your own decryption. <<<
+  std::ifstream in(name, std::ios::binary | std::ios::ate);
+  size_t n = static_cast<size_t>(in.tellg());
```
huaychou: **[M-1] ReadCb: `tellg()` return not validated before cast to `size_t`**

If the file cannot be opened (path missing, permission denied, etc.), `std::ifstream` enters a fail state and `in.tellg()` returns `-1`. The cast `static_cast<size_t>(-1)` produces `SIZE_MAX`, so `allocator->Alloc(allocator, SIZE_MAX)` either crashes with OOM or returns null — after which `in.read(nullptr, n)` is UB. The function then returns `nullptr` (success) unconditionally.

Suggested fix:
```cpp
OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
                  void** buffer, size_t* size) noexcept {
  std::ifstream in(name, std::ios::binary | std::ios::ate);
  if (!in)
    return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
  std::streamoff n_signed = in.tellg();
  if (n_signed < 0)
    return Ort::GetApi().CreateStatus(ORT_FAIL, "tellg failed");
  size_t n = static_cast<size_t>(n_signed);
  in.seekg(0);
  // ... rest unchanged
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,86 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+  in.seekg(0);
+  void* mem = allocator->Alloc(allocator, n);
```
huaychou, line 1265: **[Minor] `allocator->Alloc` return not checked for null**

If the allocator returns null (OOM or any allocator error), `in.read(nullptr, n)` on the next line is undefined behaviour, and `*buffer = nullptr` is returned to the caller with a success status (`return nullptr`).

Add a null check:
```cpp
void* mem = allocator->Alloc(allocator, n);
if (!mem)
  return Ort::GetApi().CreateStatus(ORT_FAIL, "alloc failed");
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,86 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+Replace the body of `ReadCb` with the matching decryption for whatever cipher `WriteCb` used:
+
+```cpp
+// C++
+OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
+                  void** buffer, size_t* size) noexcept {
+  // >>> Replace this block with your own decryption. <<<
+  std::ifstream in(name, std::ios::binary | std::ios::ate);
+  size_t n = static_cast<size_t>(in.tellg());
+  in.seekg(0);
```
huaychou, line 1263: **[Minor] `in.seekg(0)` failure not checked**

If `seekg` fails (e.g., interrupted I/O), the stream enters a bad state. The subsequent `in.read` then reads 0 bytes, yet `*size = n` and `return nullptr` report success — the caller receives a size-n buffer of uninitialized allocator memory.

Add a stream-state check after `seekg`:
```cpp
in.seekg(0);
if (!in)
  return Ort::GetApi().CreateStatus(ORT_FAIL, "seekg failed");
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,86 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+  void* mem = allocator->Alloc(allocator, n);
+  in.read(static_cast<char*>(mem), static_cast<std::streamsize>(n));
+  for (size_t i = 0; i < n; ++i) static_cast<uint8_t*>(mem)[i] ^= g_key;
+  *buffer = mem;
+  *size = n;
```
huaychou, line 1274: **[Minor] `in.read()` bytes actually read not validated via `gcount()`**

After `in.read(mem, n)`, there is no check that `in.gcount() == n`. On a truncated file or mid-read I/O error, fewer than `n` bytes are read, but `*size = n` still reports the full allocated size — QNN receives a partially-filled buffer with garbage in the tail, potentially causing silent context-binary corruption.

Add after the read:
```cpp
if (static_cast<size_t>(in.gcount()) != n) {
  allocator->Free(allocator, mem);
  return Ort::GetApi().CreateStatus(ORT_FAIL, "read failed");
}
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,86 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+
+**Encrypt (write callback)** — set on `Ort::ModelCompilationOptions` during compile. ORT hands
+the plaintext context bytes to the callback instead of writing them to disk. The example below
+uses a single-byte XOR so the callback mechanism is easy to follow — replace the body of
+`WriteCb` with your own cipher (AES-GCM, a KMS/TEE-backed cipher, etc.); everything else about
+how ORT calls the callback stays the same:
+
+```cpp
+// C++
+uint8_t g_key = 0x5A;  // must match the key ReadCb uses below
```
huaychou: **[Nit] `g_key` should be `const` (or `constexpr`)**

`uint8_t g_key = 0x5A;` is a mutable file-scope global. The callbacks only read it, so declaring it `const` (or `constexpr`) makes the intent explicit, prevents accidental mutation, and is safer in multi-session scenarios where ORT may invoke callbacks concurrently.

```cpp
constexpr uint8_t g_key = 0x5A;  // must match the key ReadCb uses below
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,86 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+}
+
+Ort::SessionOptions session_options;
+Ort::Experimental::Get_OrtApi_SessionOptions_SetEpContextDataReadFunc_SinceV28_Fn(
+    &Ort::GetApi())(session_options, ReadCb, /*state=*/nullptr);
+Ort::Session session(env, ctx_model_path, session_options);
+```
+
+This callback-based path applies only when `ep.context_embed_mode` is `"0"`. If no callback is
+registered while `ep.context_embed_mode` is `"0"`, the EP uses the legacy plaintext behavior.
```
huaychou: **[Nit] Redundant scope caveat**

The restriction "This callback-based path applies only when `ep.context_embed_mode` is `"0"`" is already stated clearly in the bold paragraph at the section opening (~line 1207). This sentence at line 1268 repeats it a third time after two complete code examples. Consider removing or trimming it to just: "If no callback is registered, the EP uses the legacy plaintext behavior."

huaychou (commented):
Just wondering, are WriteCb and ReadCb in the unit tests?

tzuhsuanwei:
Fixed all 6: added `tellg()`/`seekg()`/`Alloc()`/`gcount()` validation in the `ReadCb` example, made `g_key` `constexpr`, and trimmed the redundant embed_mode sentence at the end (scope is now stated once, up front). Pushed.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+
+```cpp
+// C++
+OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
+                  void** buffer, size_t* size) noexcept {
+  // >>> Replace this block with your own decryption. <<<
+  std::ifstream in(name, std::ios::binary | std::ios::ate);
+  if (!in) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
+  std::streamoff n_signed = in.tellg();
+  if (n_signed < 0) return Ort::GetApi().CreateStatus(ORT_FAIL, "tellg failed");
```
huaychou: **[M-1] (Major)** ReadCb example: tellg() returns -1 on open failure, unsafe cast to size_t passes SIZE_MAX to allocator->Alloc

`std::ifstream in(name, std::ios::binary | std::ios::ate)` sets `in` to a fail state when the file cannot be opened. In that state, `in.tellg()` returns `-1`, and `static_cast<size_t>(-1)` produces `SIZE_MAX`, so the subsequent `allocator->Alloc(allocator, SIZE_MAX)` attempts to allocate nearly the entire virtual address space — OOM crash or undefined behaviour.

**Impact:** Users who copy this example will crash rather than receive a descriptive `OrtStatus*` error when the target file is absent.

**✅ Fixed in this push**: the code now checks `if (!in) return ... "open failed"` before `seekg`, and validates the return value with `if (n_signed < 0) return ... "tellg failed"` before the cast.

**Suggested fix:**

(Already applied) Add an open-success check and validate the tellg() return value is non-negative before casting — the current code already does this.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+  for (size_t i = 0; i < n; ++i) {
+    out.put(static_cast<const uint8_t*>(buffer)[i] ^ g_key);
+  }
+  return out ? nullptr : Ort::GetApi().CreateStatus(ORT_FAIL, "write failed");
+}
+
+Ort::ModelCompilationOptions compile_options(env, session_options);
+compile_options.SetEpContextEmbedMode(false);
+Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(
+    &Ort::GetApi())(compile_options, WriteCb, /*state=*/nullptr);
```
huaychou: **[M-3] (Major)** Callback registration calls through the versioned-API resolver without a null check, and discards the returned OrtStatus*

Both new registration snippets call the resolver inline and immediately invoke the returned function pointer:

```
Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(
    &Ort::GetApi())(compile_options, WriteCb, /*state=*/nullptr);
```

`Get_..._SinceV28_Fn` resolvers look up a function pointer that only exists in ORT API v28+; if the linked ORT core predates v28 (this repo's own vendored build currently reports `ORT_API_VERSION` 24), the resolver returns `nullptr`. The example calls through the result without checking for `nullptr` first — calling through a null function pointer is undefined behaviour / a crash. It also discards the `OrtStatus*` returned by the setter call itself: if the setter fails, the error is silently swallowed and the returned status object leaks (per ORT convention, any non-null `OrtStatus*` must be released via `Ort::GetApi().ReleaseStatus()`). The ReadCb registration call (lines 1271-1272) has the identical issue.

Notably, this repo's own test file `onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc` (e.g. lines 369-373, 411-415) guards both of these exact resolver calls with `ASSERT_NE(set_fn, nullptr)` before calling, and releases the returned status via `Ort::GetApi().ReleaseStatus(st)`. The doc example should mirror this established defensive pattern rather than omitting it.

**Impact:** A reader who copy-pastes this example verbatim and runs it against an ORT core older than v28 will crash on a null function-pointer call instead of getting a clear "unsupported" error. Even on a v28+ build, any setup failure is silently dropped. This is the primary onboarding example for a security feature, so the missing error handling is particularly worth fixing.

**Suggested fix:**

Check the resolver for null before calling, and check/release the returned status, e.g.:

```cpp
auto* set_write_fn = Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(&Ort::GetApi());
if (set_write_fn == nullptr) {
  // ORT core predates API v28 — encryption callback not available.
} else if (OrtStatus* st = set_write_fn(compile_options, WriteCb, /*state=*/nullptr)) {
  Ort::GetApi().ReleaseStatus(st);  // handle/report the error
}
```

Apply the same pattern to the ReadCb registration call. See the existing pattern in `onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc` for reference.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+`WriteCb` with your own cipher (AES-GCM, a KMS/TEE-backed cipher, etc.); everything else about
+how ORT calls the callback stays the same:
+
+```cpp
+// C++
+constexpr uint8_t g_key = 0x5A;  // must match the key ReadCb uses below
+
+OrtStatus* WriteCb(void* state, const char* name, const void* buffer, size_t n) noexcept {
+  // >>> Replace this block with your own encryption. <<<
+  std::ofstream out(name, std::ios::binary);
```
huaychou, line 1227: **[N-1] (Minor)** WriteCb example does not check ofstream open success before the write loop

`std::ofstream out(name, std::ios::binary)` leaves `out.is_open()` false if the file cannot be opened, but the write loop still runs (each `out.put()` silently no-ops). The final `return out ? nullptr : Ort::GetApi().CreateStatus(ORT_FAIL, "write failed")` does catch the failed stream state, but users who replace only the loop body per the "Replace this block" comment and omit the final check will have a silent failure path. This was already flagged in the v1 review; the author added four defensive checks to ReadCb in this push but this one for WriteCb remains unaddressed.

**Impact:** The example does not demonstrate best practice; a copied-and-modified version may silently drop the error.

**Suggested fix:**

Add an open check before the loop:

```cpp
std::ofstream out(name, std::ios::binary);
if (!out)
  return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
// >>> Replace this block with your own encryption. <<<
for (size_t i = 0; i < n; ++i) { ... }
return nullptr;  // out failure is now covered by the early-return above
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+  if (!in) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
+  std::streamoff n_signed = in.tellg();
+  if (n_signed < 0) return Ort::GetApi().CreateStatus(ORT_FAIL, "tellg failed");
+  size_t n = static_cast<size_t>(n_signed);
+  in.seekg(0);
+  if (!in) return Ort::GetApi().CreateStatus(ORT_FAIL, "seekg failed");
+  void* mem = allocator->Alloc(allocator, n);
+  if (mem == nullptr) return Ort::GetApi().CreateStatus(ORT_FAIL, "alloc failed");
+  in.read(static_cast<char*>(mem), static_cast<std::streamsize>(n));
+  if (static_cast<size_t>(in.gcount()) != n) {
```
huaychou: **[N-2] (Minor)** ReadCb example does not validate that actual bytes read equals expected size

After `in.read(static_cast<char*>(mem), static_cast<std::streamsize>(n))`, there was no `in.gcount()` check to validate that the number of bytes actually read equals `n`. On a partial read, `*size = n` would still report the full allocated size, handing QNN a partially-filled buffer.

**Impact:** On a partial read, QNN receives a corrupted context binary, potentially causing graph finalization failure or silent accuracy issues.

**✅ Fixed in this push**: the code now adds `if (static_cast<size_t>(in.gcount()) != n) { allocator->Free(...); return ...; }` right after `in.read(...)`.

**Suggested fix:**

(Already applied) Add a gcount validation after the read and free the allocated buffer on mismatch — the current code already does this.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
 options.add_session_config_entry("ep.context_embed_mode", "1")
 ```
 
-### At-rest encryption of the context binary (ORT API v28+)
+## Compiled Model Encryption
```
huaychou: **[N-3] (Minor)** The h3→h2 heading promotion dropped the explicit '(ORT API v28+)' version qualifier from the old heading

The old heading was `### At-rest encryption of the context binary (ORT API v28+)` — a reader skimming just the headings (e.g. in a rendered table of contents) could immediately tell this feature requires ORT API v28+. The new heading `## Compiled Model Encryption` drops this qualifier; the version requirement is still mentioned in the first paragraph below the heading ("ORT API v28 adds a write/read callback pair..."), but a reader who only skims headings/ToC loses this at-a-glance signal.

**Impact:** A documentation-discoverability regression, not a functional bug — but unfriendly to readers who only scan section titles to check whether a feature applies to their ORT version.

**Suggested fix:**

Keep a short version qualifier in the new heading, e.g.:

```markdown
## Compiled Model Encryption (ORT API v28+)
```

and update the Contents entry text to match.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+    allocator->Free(allocator, mem);
+    return Ort::GetApi().CreateStatus(ORT_FAIL, "read failed");
+  }
+  for (size_t i = 0; i < n; ++i) static_cast<uint8_t*>(mem)[i] ^= g_key;
+  *buffer = mem;
+  *size = n;
+  return nullptr;
+}
+
+Ort::SessionOptions session_options;
```
huaychou: **[T-2] (Nit)** ReadCb snippet re-declares session_options, creating ambiguity with the identically-named variable used in the WriteCb snippet

The WriteCb snippet uses `Ort::ModelCompilationOptions compile_options(env, session_options)`, where `session_options` refers to compile-phase session options (created earlier, in the "Enable the embed mode" subsection). The ReadCb snippet then re-declares `Ort::SessionOptions session_options`, which could lead a reader to assume both snippets live in the same scope and share the same `session_options` object. In reality they belong to separate sessions (compile vs. inference). This was already flagged in the v1 review and remains unaddressed in this push.

**Suggested fix:**

Add a comment before the session_options declaration in the ReadCb snippet, e.g.:

```cpp
// This is the *inference* session's options — separate from the compile-phase session_options above.
Ort::SessionOptions session_options;
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,92 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+
+**Encrypt (write callback)** — set on `Ort::ModelCompilationOptions` during compile. ORT hands
+the plaintext context bytes to the callback instead of writing them to disk. The example below
+uses a single-byte XOR so the callback mechanism is easy to follow — replace the body of
+`WriteCb` with your own cipher (AES-GCM, a KMS/TEE-backed cipher, etc.); everything else about
+how ORT calls the callback stays the same:
+
+```cpp
+// C++
+constexpr uint8_t g_key = 0x5A;  // must match the key ReadCb uses below
```
huaychou, line 1223: **[T-3] (Nit)** g_key is declared only in the WriteCb code block; the ReadCb code block references it without its own declaration

`constexpr uint8_t g_key = 0x5A;` only appears inside the WriteCb C++ code block (line 1223). The ReadCb block (line 1264) references `g_key` directly in `static_cast<uint8_t*>(mem)[i] ^= g_key;` without declaring it in that block. The intro text ("must match the key ReadCb uses below") explains the two blocks share one global, which is clear when reading the whole section top-to-bottom, but a reader who copies only the ReadCb block into their own file gets an "undeclared identifier" compile error.

**Impact:** A minor documentation usability nit — doesn't affect readers who follow the section in order, but is confusing for anyone who copies only the ReadCb block in isolation.

**Suggested fix:**

Add a short reminder comment at the top of the ReadCb block, e.g.:

```cpp
// C++ — shares g_key with the WriteCb snippet above; declare it here too if copying this block in isolation.
```

docs/execution_providers/QNN-ExecutionProvider.md
tzuhsuanwei: Good catch, fixed — the resolver now checks for null (so it won't crash on a pre-v28 core), and the call result goes through ThrowOnError.

docs/execution_providers/QNN-ExecutionProvider.md
tzuhsuanwei: Added, thanks for flagging it.

docs/execution_providers/QNN-ExecutionProvider.md
tzuhsuanwei: Yep, added `(ORT API v28+)` back.

docs/execution_providers/QNN-ExecutionProvider.md
tzuhsuanwei: Fair point, renamed to `inference_session_options` so it's distinct from WriteCb's.

docs/execution_providers/QNN-ExecutionProvider.md
tzuhsuanwei: Added, so copying just the ReadCb block compiles on its own now.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,105 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+Ort::CompileModel(env, compile_options);
+```
+
+**Decrypt (read callback)** — set on `Ort::SessionOptions` before creating the inference session.
+ORT asks the callback for the plaintext bytes (by name) instead of reading the file directly.
+Replace the body of `ReadCb` with the matching decryption for whatever cipher `WriteCb` used:
+
+```cpp
+// C++
+constexpr uint8_t g_key = 0x5A;  // must match the key WriteCb used above
```
huaychou: **[M-1] (Major)** ReadCb example: tellg() returns -1 on open failure, unsafe cast to size_t passes SIZE_MAX to allocator->Alloc

`std::ifstream in(name, std::ios::binary | std::ios::ate)` sets `in` to a fail state when the file cannot be opened. In that state, `in.tellg()` returns `-1`, and `static_cast<size_t>(-1)` produces `SIZE_MAX`, so the subsequent `allocator->Alloc(allocator, SIZE_MAX)` attempts to allocate nearly the entire virtual address space — OOM crash or undefined behaviour.

**Impact:** Users who copy this example will crash rather than receive a descriptive `OrtStatus*` error when the target file is absent.

**✅ Fixed in this push**: the code now checks `if (!in) return ... "open failed"` before `seekg`, and validates the return value with `if (n_signed < 0) return ... "tellg failed"` before the cast.

**Suggested fix:**

(Already applied) Add an open-success check and validate the tellg() return value is non-negative before casting — the current code already does this.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,105 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+  if (!out.is_open()) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
+  for (size_t i = 0; i < n; ++i) {
+    out.put(static_cast<const uint8_t*>(buffer)[i] ^ g_key);
+  }
+  return out ? nullptr : Ort::GetApi().CreateStatus(ORT_FAIL, "write failed");
+}
+
+Ort::ModelCompilationOptions compile_options(env, session_options);
+compile_options.SetEpContextEmbedMode(false);
+auto write_fn = Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(
```
huaychou: **[M-3] (Major)** Callback registration calls through the versioned-API resolver without a null check, and discards the returned OrtStatus*

Both registration snippets (v2 state) called through the resolver inline without a null check and discarded the returned `OrtStatus*`:

```
Ort::Experimental::Get_OrtCompileApi_ModelCompilationOptions_SetEpContextDataWriteFunc_SinceV28_Fn(
    &Ort::GetApi())(compile_options, WriteCb, /*state=*/nullptr);
```

The `Get_..._SinceV28_Fn` resolver returns `nullptr` on a pre-v28 ORT core. Calling through a null function pointer is undefined behaviour. The setter's returned `OrtStatus*` was also silently discarded and leaked.

**✅ Fixed in v3 push**: both WriteCb and ReadCb registration sites now save the resolver result into `auto write_fn / read_fn`, guard with `if (write_fn == nullptr) { return; }`, and wrap the call with `Ort::ThrowOnError(write_fn(...))` so the returned status is not silently dropped.

**Suggested fix:**

(Already applied) Save the resolver result, guard with null check, and wrap the call with ThrowOnError — the current code already does this.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,105 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+how ORT calls the callback stays the same:
+
+```cpp
+// C++
+constexpr uint8_t g_key = 0x5A;  // must match the key ReadCb uses below
+
+OrtStatus* WriteCb(void* state, const char* name, const void* buffer, size_t n) noexcept {
+  // >>> Replace this block with your own encryption. <<<
+  std::ofstream out(name, std::ios::binary);
+  if (!out.is_open()) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
```
huaychou: **[N-1] (Minor)** WriteCb example does not check ofstream open success before the write loop

`std::ofstream out(name, std::ios::binary)` leaves `out.is_open()` false if the file cannot be opened, but the write loop still runs (each `out.put()` silently no-ops). The trailing `return out ? nullptr : ...` does catch the failed stream state, but users who replace only the loop body per the "Replace this block" comment and omit the final check will have a silent failure path.

**✅ Fixed in v3 push**: an explicit `if (!out.is_open()) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");` is now added before the loop.

**Suggested fix:**

(Already applied) Add an is_open() check before the loop — the current code already does this.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,105 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+```cpp
+// C++
+constexpr uint8_t g_key = 0x5A;  // must match the key WriteCb used above
+
+OrtStatus* ReadCb(void* state, const char* name, OrtAllocator* allocator,
+                  void** buffer, size_t* size) noexcept {
+  // >>> Replace this block with your own decryption. <<<
+  std::ifstream in(name, std::ios::binary | std::ios::ate);
+  if (!in) return Ort::GetApi().CreateStatus(ORT_FAIL, "open failed");
+  std::streamoff n_signed = in.tellg();
```
huaychou: **[N-2] (Minor)** ReadCb example does not validate that actual bytes read equals expected size

After `in.read(static_cast<char*>(mem), static_cast<std::streamsize>(n))`, there was no `in.gcount()` check to validate that the number of bytes actually read equals `n`. On a partial read, `*size = n` would still report the full allocated size, handing QNN a partially-filled buffer.

**Impact:** On a partial read, QNN receives a corrupted context binary, potentially causing graph finalization failure or silent accuracy issues.

**✅ Fixed in this push**: the code now adds `if (static_cast<size_t>(in.gcount()) != n) { allocator->Free(...); return ...; }` right after `in.read(...)`.

**Suggested fix:**

(Already applied) Add a gcount validation after the read and free the allocated buffer on mismatch — the current code already does this.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,105 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
 options.add_session_config_entry("ep.context_embed_mode", "1")
 ```
 
-### At-rest encryption of the context binary (ORT API v28+)
+## Compiled Model Encryption (ORT API v28+)
```
huaychou: **[N-3] (Minor)** The h3→h2 heading promotion dropped the explicit '(ORT API v28+)' version qualifier from the old heading

The old heading was `### At-rest encryption of the context binary (ORT API v28+)` — a reader skimming headings could immediately tell this feature requires ORT API v28+. The first v2 commit's new heading `## Compiled Model Encryption` dropped this qualifier, reducing at-a-glance discoverability.

**✅ Fixed in v3 push**: both the heading and the Contents entry have been restored to `## Compiled Model Encryption (ORT API v28+)`.

**Suggested fix:**

(Already applied) The heading and Contents entry now both include the '(ORT API v28+)' version qualifier — the current doc already does this.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1193,34 +1194,105 @@ g_ort->AddSessionConfigEntry(session_options, kOrtSessionOptionEpContextEmbedMod
...
+    allocator->Free(allocator, mem);
+    return Ort::GetApi().CreateStatus(ORT_FAIL, "read failed");
+  }
+  for (size_t i = 0; i < n; ++i) static_cast<uint8_t*>(mem)[i] ^= g_key;
+  *buffer = mem;
+  *size = n;
+  return nullptr;
+}
+
+Ort::SessionOptions inference_session_options;
```
huaychou: **[T-2] (Nit)** ReadCb snippet re-declares session_options, creating ambiguity with the identically-named variable used in the WriteCb snippet

The WriteCb snippet uses `Ort::ModelCompilationOptions compile_options(env, session_options)`, where `session_options` refers to the compile-phase session options. The ReadCb snippet then re-declared `Ort::SessionOptions session_options`, creating potential confusion about whether both snippets share the same object.

**✅ Fixed in v3 push**: the ReadCb snippet's variable has been renamed to `inference_session_options`, clearly distinguishing it from the compile-phase `session_options`.

**Suggested fix:**

(Already applied) The ReadCb session options variable has been renamed to `inference_session_options` — the current code already does this.

docs/execution_providers/QNN-ExecutionProvider.md
huaychou: **[T-3] (Nit)** g_key is declared only in the WriteCb code block; the ReadCb code block references it without its own declaration

`constexpr uint8_t g_key = 0x5A;` only appeared in the WriteCb C++ code block. The ReadCb block referenced `g_key` directly without declaring it, so a reader who copied only the ReadCb block would get a compile error.

**✅ Fixed in v3 push**: the ReadCb block now begins with `constexpr uint8_t g_key = 0x5A;  // must match the key WriteCb used above`, making it self-contained when copied in isolation.

**Suggested fix:**

(Already applied) The ReadCb block now includes its own g_key declaration — the current code already does this.

huaychou (approved):
LGTM

qti-chuteng (approved):

PR #770, opened 2026-08-27
[QNN EP] Implement automatic UDO custom-op domain registration via factory hooks

huaychou:
### Description

- Domain name and op-types are sourced from the new env var ORT_QNN_CUSTOM_OP_DOMAINS (format: domain:OpA,OpB;domain2:OpC), mirroring the TensorRT ORT_TENSORRT_EXTRA_PLUGIN_LIB_PATHS precedent. The factory parses the env var once at construction time and caches the result; all sessions in the process share the same domain list.

### Motivation and Context

- Implement GetNumCustomOpDomains / GetCustomOpDomains on QnnEpFactory so that UDO custom-op domains are registered automatically at session-options time, removing the requirement for users to manually construct Ort::CustomOpDomain.

### New files:
- qnn_custom_op.h: QnnUdoPlaceholderOp (variadic-IO no-op) and QnnUdoPlaceholderKernel (Compute returns error) for schema validation
- qnn_custom_op_domain_parser.h/.cc: ParseCustomOpDomains() following the same pattern as op_package_parser for testability and warning behavior

### Changes:
- qnn_provider_factory.h/.cc: wire factory hooks, read env var, build and own domain + op objects
- cmake/onnxruntime_unittests.cmake: add parser source to test build
- docs/QNN-ExecutionProvider.md: document ORT_QNN_CUSTOM_OP_DOMAINS in UDO Step 5
- qnn_basic_test.cc: 7 ParseCustomOpDomains unit tests
- udo_op_test.cc: 2 QnnUdoPlaceholderOp unit tests


onnxruntime/core/providers/qnn/qnn_custom_op.h
```diff
@@ -0,0 +1,95 @@
...
+
+// A variadic-input/variadic-output placeholder OrtCustomOp used by QnnEpFactory to
+// register a UDO op schema so ORT can load and validate ONNX models that contain
+// custom-domain nodes.  The factory builds one of these per op-type entry found in
+// ORT_QNN_CUSTOM_OP_DOMAINS and keeps it alive for the factory's lifetime.
+struct QnnUdoPlaceholderOp
+    : Ort::CustomOpBase<QnnUdoPlaceholderOp, QnnUdoPlaceholderKernel, /*WithStatus=*/true> {
+  QnnUdoPlaceholderOp(std::string op_type, std::string ep_type)
+      : op_type_(std::move(op_type)), ep_type_(std::move(ep_type)) {}
+  ORT_DISALLOW_COPY_AND_ASSIGNMENT(QnnUdoPlaceholderOp);
```
huaychou: **[M-1] (Major)** QnnUdoPlaceholderOp missing ORT_DISALLOW_COPY_AND_ASSIGNMENT (B5 violation)

**Issue:** QNN EP mandatory convention B5 requires every resource-owning class to declare `ORT_DISALLOW_COPY_AND_ASSIGNMENT(ClassName)` in its `public:` section. `QnnUdoPlaceholderOp` extends `Ort::CustomOpBase` (non-copyable semantics) but lacked this declaration.

**Impact:** While actual copying was prevented by the `std::unique_ptr` container, the absence of an explicit non-copyable declaration violated a mandatory convention and left a latent refactoring hazard.

**Suggested fix:**

✅ Fixed in this commit (b1a46eb): added `ORT_DISALLOW_COPY_AND_ASSIGNMENT(QnnUdoPlaceholderOp);` after the constructor.

onnxruntime/test/providers/qnn/udo_op_test.cc
```diff
@@ -180,6 +198,68 @@ TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd) {
...
 
+// Verifies the full factory-level custom-op domain registration path:
+//   ORT_QNN_CUSTOM_OP_DOMAINS env var → QnnEpFactory construction → GetNumCustomOpDomains /
+//   GetCustomOpDomains callbacks → ORT registers domain → ONNX model containing a
+//   udo_domain::MyAdd node loads successfully with no manual Ort::CustomOpDomain.
+// We bypass the RunQnnModelTest helper (which also creates a CPU-EP reference session that
+// needs the domain) and directly assert that Ort::Session construction succeeds — confirming
+// the factory hook delivers the domain. An "Unknown domain" exception indicates the hook
+// did NOT fire, which would be a regression.
+TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd_AutoDomainFromEnvVar) {
```
huaychou: **[N-1] (Minor)** Missing end-to-end integration test: factory callback to ORT domain receipt to model load

**Issue:** v1 review noted the absence of an end-to-end integration test verifying the complete path: `ORT_QNN_CUSTOM_OP_DOMAINS` env var to `QnnEpFactory` construction to `GetNumCustomOpDomains` / `GetCustomOpDomains` callbacks to ORT receiving the domain to a UDO-containing ONNX model loading successfully (with no manual `Ort::CustomOpDomain`).

**Impact:** Without an integration test, the correctness of the factory callback path cannot be automatically verified.

**Suggested fix:**

✅ Fixed in this commit (b1a46eb): added `UDO_Op_MyAdd_AutoDomainFromEnvVar` test (L209–261), fully verifying the env var to factory construction to `GetCustomOpDomains` callback to `Ort::Session` construction success path.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -504,6 +547,30 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetHardwareDeviceIncompatibilityDetailsImp
...
+  *num_domains = factory->custom_op_domains_.size();
+  return nullptr;
+}
+
+OrtStatus* ORT_API_CALL QnnEpFactory::GetCustomOpDomainsImpl(
+    _In_ OrtEpFactory* this_ptr,
+    _Out_writes_all_(num_domains) OrtCustomOpDomain** domains,
+    _In_ size_t num_domains) noexcept {
+  const auto* factory = static_cast<const QnnEpFactory*>(this_ptr);
+  if (num_domains > factory->custom_op_domains_.size()) {
```
huaychou: **[N-2] (Minor)** GetCustomOpDomainsImpl lacks defensive bounds check

**Issue:** v1 review noted that `GetCustomOpDomainsImpl`'s loop did not validate whether `num_domains` exceeds `custom_op_domains_.size()`, creating an out-of-bounds write risk.

**Impact:** If ORT passed a `num_domains` value larger than the factory's actual domain count (e.g., due to API misuse), the loop would access `custom_op_domains_` out of bounds, causing undefined behavior or a crash.

**Suggested fix:**

✅ Fixed in this commit (b1a46eb): added an early-return guard before the loop — returns `ORT_INVALID_ARGUMENT` when `num_domains > custom_op_domains_.size()`.

onnxruntime/core/providers/qnn/qnn_custom_op.h
```diff
@@ -0,0 +1,95 @@
...
+  // Propagate the first input's type/shape to all outputs so ORT's type-inference pass
+  // succeeds during model load. Without this, UNDEFINED output type causes an inference
+  // error even though the node will be fused+compiled by QNN EP and the placeholder kernel
+  // is never executed.
+  static OrtStatusPtr InferOutputShape(Ort::ShapeInferContext& ctx) {
+    if (ctx.GetInputCount() == 0) {
+      return nullptr;
+    }
+    const auto& input_shape = ctx.GetInputShape(0);
+    // SetOutputShape's type defaults to FLOAT, which matches the primary UDO use case.
```
huaychou: **[N-3] (Minor)** InferOutputShape hardcodes output type to FLOAT, breaking non-FLOAT-output UDOs

**Issue:** `InferOutputShape` calls `ctx.SetOutputShape(i, input_shape)`, and `SetOutputShape`'s type defaults to FLOAT (`ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT`). The function is intended to let ORT's type-inference pass succeed during model load, but for UDOs whose output type is not FLOAT (e.g., an index-type UDO outputting INT32), ORT will infer FLOAT for the output — causing a type mismatch against downstream nodes and making `Ort::Session` construction fail instead.

**Impact:** UDOs with non-FLOAT output types will fail to load due to type mismatch. This does not surface with the current `MyAdd` (FP32 in/out) example but will be triggered by INT32/INT64-output UDOs, contradicting this PR's goal of eliminating manual domain management for all UDOs.

**Suggested fix:**

Consider having `InferOutputShape` propagate shape without overwriting the output type, or add a `std::vector<ONNXTensorElementDataType> output_types` parameter to `QnnUdoPlaceholderOp`'s constructor to support per-output types. If the `Ort::ShapeInferContext` API does not support omitting the type, another option is to remove `InferOutputShape` entirely and verify whether ORT can still pass model-load validation without shape inference for placeholder ops.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1705,11 +1705,45 @@ make -C <output_dir>/MyAddOpPackage htp_x86
...
+```bash
+export ORT_QNN_CUSTOM_OP_DOMAINS="udo_domain:MyAdd;other_domain:OpA,OpB"
 ```
 
-For the whole pipeline, refer [udo unit test](../../cmake/onnxruntime_unittests_udo.cmake)
+**Then run inference as usual:**
+
+```bash
+./onnxruntime_plugin_ep_onnx_test -v -e qnn -j 1 \
+  -i "backend_path|./libQnnCpu.so op_packages|<op_type>:<op_package_path>:<interface_symbol_name>[:<target>]" \
```
huaychou: **[T-1] (Nit)** Documentation placeholder name inconsistency: interface_symbol vs interface_symbol_name

**Issue:** The same documentation command-line example used two inconsistent placeholder names in the `op_packages` field: `<interface_symbol>` (older `onnx_test_runner` example) and `<interface_symbol_name>` (newer `onnxruntime_plugin_ep_onnx_test` example).

**Impact:** Users could mistakenly assume the two names map to different fields or carry different semantics.

**Suggested fix:**

✅ Fixed in this commit (b1a46eb): unified the older `<interface_symbol>` placeholder to `<interface_symbol_name>`.

onnxruntime/core/providers/qnn/qnn_custom_op_domain_parser.cc
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    if (op_types_csv.empty()) {
+      ORT_CXX_LOG_PTR(logger_ptr, ORT_LOGGING_LEVEL_WARNING,
+                      ("ORT_QNN_CUSTOM_OP_DOMAINS entry has empty op-types list for domain '" + domain +
+                       "'. Skipping.")
+                          .c_str());
+      continue;
+    }
+
```
huaychou: **[T-2] (Nit)** ParseCustomOpDomains uses std::vector instead of InlinedVector (B3)

**Issue:** `qnn_custom_op_domain_parser.cc` L55 uses `std::vector<std::string> op_types` to accumulate the op-type list for each domain. QNN EP convention B3 requires performance-sensitive paths to use `onnxruntime::InlinedVector<T, N>` instead of `std::vector`.

**Impact:** UDO domains typically have only a small number of op-types (1–5); triggering a heap allocation on every parse is unnecessary overhead and violates convention B3.

**Suggested fix:**

Change `std::vector<std::string> op_types;` to `onnxruntime::InlinedVector<std::string, 4> op_types;`.

onnxruntime/test/providers/qnn/udo_op_test.cc
```diff
@@ -180,6 +198,68 @@ TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd) {
...
+  std::filesystem::path path = getLibPath("cpu");
+  if (!std::filesystem::exists(path)) {
+    GTEST_SKIP() << "UDO CPU op package not found: " << path;
+  }
+
+  // Set ORT_QNN_CUSTOM_OP_DOMAINS before RegisterExecutionProviderLibrary so the factory
+  // reads it during CreateEpFactories() (called from RegisterExecutionProviderLibrary).
```
huaychou: **[T-3] (Nit)** Integration test comment implies a test-ordering dependency

**Issue:** The comment in `UDO_Op_MyAdd_AutoDomainFromEnvVar` (L216–217) states "The previous test unregisters the QNN library on teardown, so a fresh factory is constructed for this test," implying an ordering dependency (must run after `UDO_Op_MyAdd`). GTest does not guarantee execution order under filter or shuffle scenarios.

**Impact:** The comment is inaccurate and could mislead future maintainers. The test itself should work correctly given the `RegisterQnnEpLibrary` call, but the precondition described in the comment does not hold when the test runs in isolation.

**Suggested fix:**

Rewrite the comment to describe the test's own behavior: "This test registers QNN EP directly via RegisterQnnEpLibrary, constructing a fresh factory with ORT_QNN_CUSTOM_OP_DOMAINS set. The env var is set before registration and restored on exit to avoid cross-test leakage."

onnxruntime/test/providers/qnn/qnn_basic_test.cc
```diff
@@ -3553,5 +3554,78 @@ TEST_F(QnnCPUBackendTests, DumpQnnEpInputGraph_DisabledByDefault) {
 
 #endif  // !defined(ORT_MINIMAL_BUILD)
 
+// ── ParseCustomOpDomains unit tests ─────────────────────────────────────────
```
minfhong-qti: Should these testcases be under `unit`?

onnxruntime/core/providers/qnn/qnn_custom_op_domain_parser.cc
```diff
@@ -0,0 +1,79 @@
...
+#include "core/providers/qnn/ort_api.h"  // Ort::Logger full definition + ORT_CXX_LOG_PTR
+
+namespace onnxruntime {
+
+void ParseCustomOpDomains(const std::string& spec,
+                          std::vector<CustomOpDomainSpec>& out,
+                          const Ort::Logger& logger) {
+  // Use ORT_CXX_LOG_PTR (instead of ORT_CXX_LOG) so unit tests can pass a default-constructed
+  // Ort::Logger() without crashing. ORT_CXX_LOG dereferences the null OrtLogger* internally.
+  const Ort::Logger* logger_ptr = &logger;
```
minfhong-qti: Prefer not to be affected how UT would test. Directly use `const Ort::Logger& logger` instead.
I believe there are ways to workaround, and probably refer to existing UTs.

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
minfhong-qti, line 1: Create a `custom_op` folder for these files?

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -86,6 +103,32 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
...
+
+  // Build custom-op domains from ORT_QNN_CUSTOM_OP_DOMAINS env var.
+  // GetCustomOpDomains is called at SessionOptionsAppendExecutionProvider_V2 time (before CreateEp),
+  // so we parse once here at factory construction and cache the result for all sessions.
+  const std::string custom_op_domains_spec = ReadCustomOpDomainsEnvVar();
+  if (!custom_op_domains_spec.empty()) {
+    // Use the default logger if available; otherwise a default-constructed Ort::Logger() is safe
+    // because ParseCustomOpDomains uses ORT_CXX_LOG_PTR which tolerates a null OrtLogger*.
+    Ort::Logger logger{};
+    if (OrtLoggingManager::HasDefaultLogger()) {
```
minfhong-qti: Default logger is set after factory constructed so this will always be null.
How about passing in the `const OrtLogger*` and use it by `Ort::Logger`? I think it can be trusted that the logger would never be null.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -86,6 +103,32 @@ QnnEpFactory::QnnEpFactory(const char* ep_name,
   IsStreamAware = IsStreamAwareImpl;
   ValidateCompiledModelCompatibilityInfo = ValidateCompiledModelCompatibilityInfoImpl;
   GetHardwareDeviceIncompatibilityDetails = GetHardwareDeviceIncompatibilityDetailsImpl;
+  GetNumCustomOpDomains = GetNumCustomOpDomainsImpl;
+  GetCustomOpDomains = GetCustomOpDomainsImpl;
+
+  // Build custom-op domains from ORT_QNN_CUSTOM_OP_DOMAINS env var.
+  // GetCustomOpDomains is called at SessionOptionsAppendExecutionProvider_V2 time (before CreateEp),
+  // so we parse once here at factory construction and cache the result for all sessions.
+  const std::string custom_op_domains_spec = ReadCustomOpDomainsEnvVar();
```
minfhong-qti: I think this entire block can be implemented in `qnn_custom_op_xxx.h`, including `ReadCustomOpDomainsEnvVar`, and factory only needs to invoke a function.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -504,6 +547,30 @@ OrtStatus* ORT_API_CALL QnnEpFactory::GetHardwareDeviceIncompatibilityDetailsImp
   return nullptr;
 }
 
+OrtStatus* ORT_API_CALL QnnEpFactory::GetNumCustomOpDomainsImpl(
```
minfhong-qti, line 562: Just a heads up for #776 which is wrapping all noexcept ORT API with proper try-catch. Depends on which PR merges first then the other one should revise.

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
```diff
@@ -0,0 +1,104 @@
...
+
+// A variadic-input/variadic-output placeholder OrtCustomOp used by QnnEpFactory to
+// register a UDO op schema so ORT can load and validate ONNX models that contain
+// custom-domain nodes.  The factory builds one of these per op-type entry found in
+// ORT_QNN_CUSTOM_OP_DOMAINS and keeps it alive for the factory's lifetime.
+struct QnnUdoPlaceholderOp
+    : Ort::CustomOpBase<QnnUdoPlaceholderOp, QnnUdoPlaceholderKernel, /*WithStatus=*/true> {
+  QnnUdoPlaceholderOp(std::string op_type, std::string ep_type)
+      : op_type_(std::move(op_type)), ep_type_(std::move(ep_type)) {}
+  ORT_DISALLOW_COPY_AND_ASSIGNMENT(QnnUdoPlaceholderOp);
```
huaychou: **[M-1] (Major)** QnnUdoPlaceholderOp missing ORT_DISALLOW_COPY_AND_ASSIGNMENT (B5 violation)

**Issue (carried from v2):** QNN EP mandatory convention B5 requires every resource-owning class to declare `ORT_DISALLOW_COPY_AND_ASSIGNMENT(ClassName)` in its `public:` section.

**Suggested fix:**

✅ Fixed, and preserved through the file's move to `custom_op/qnn_custom_op.h`: `ORT_DISALLOW_COPY_AND_ASSIGNMENT(QnnUdoPlaceholderOp);` is declared right after the constructor (L37).

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op_domain_parser.cc
```diff
@@ -0,0 +1,75 @@
...
+    return;
+  }
+
+  // Format: domain_name:OpType1[,OpType2][;domain_name2:OpType3]
+  for (const auto& entry : qnn::utils::SplitString(spec, ";", /*keep_empty=*/false)) {
+    // Split on the first ':' to separate domain name from op-types CSV.
+    // string_view::find returns npos if ':' is absent.
+    const auto colon_pos = entry.find(':');
+    if (colon_pos == std::string_view::npos) {
+      ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_WARNING,
```
huaychou: **[M-2] (Major)** ParseCustomOpDomains's ORT_CXX_LOG_PTR null-safety guard was regressed during this restructure (confirmed)

**Issue (confirmed as a regression via git history):** diffing the old commit `b1a46eb` (the PR head at v2 review time, still present in the local git object store) against the current `7f78b2f` for the same file: the old version declared `const Ort::Logger* logger_ptr = &logger;` and used `ORT_CXX_LOG_PTR(logger_ptr, ...)` at all 5 log sites, with an explicit comment: "Use ORT_CXX_LOG_PTR (instead of ORT_CXX_LOG) so unit tests can pass a default-constructed Ort::Logger() without crashing." The current version has all 5 sites calling plain `ORT_CXX_LOG(logger, ...)` instead — both the `logger_ptr` variable and that comment are gone. This is inconsistent with the sibling `ParseOpPackages` (`op_package_parser.cc`, which still correctly uses `ORT_CXX_LOG_PTR`) and is an actual regression introduced during this round's file restructure, not merely a style difference. This is the root cause of why the new unit test file `unit/qnn_custom_op_test.cc` has to duplicate a local `MakeNullLogger()` helper (memcpy'ing a fake FATAL severity into `Ort::Logger`'s private memory layout).

**Impact:** the production path is unaffected (`BuildCustomOpDomainsFromEnv` always passes a non-null `default_logger`), but this is a confirmed code-quality regression — an existing null-safety guard and its explanatory comment were dropped. If any future caller passes an uninitialized `Ort::Logger` to this parser directly (e.g. a future test that forgets to route through `MakeNullLogger()`), it will crash dereferencing a null `OrtLogger*` inside `ORT_CXX_LOG`.

**Suggested fix:**

Restore the old pattern: add back `const Ort::Logger* logger_ptr = &logger;` (and its explanatory comment) at the top of `ParseCustomOpDomains`, and change all 5 `ORT_CXX_LOG(logger, ...)` call sites back to `ORT_CXX_LOG_PTR(logger_ptr, ...)`. `unit/qnn_custom_op_test.cc` can then drop its local `MakeNullLogger()` and just call the parser with `Ort::Logger(nullptr)`.

onnxruntime/test/providers/qnn/udo_op_test.cc
```diff
@@ -180,6 +196,68 @@ TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd) {
...
 
+// Verifies the full factory-level custom-op domain registration path:
+//   ORT_QNN_CUSTOM_OP_DOMAINS env var → QnnEpFactory construction → GetNumCustomOpDomains /
+//   GetCustomOpDomains callbacks → ORT registers domain → ONNX model containing a
+//   udo_domain::MyAdd node loads successfully with no manual Ort::CustomOpDomain.
+// We bypass the RunQnnModelTest helper (which also creates a CPU-EP reference session that
+// needs the domain) and directly assert that Ort::Session construction succeeds — confirming
+// the factory hook delivers the domain. An "Unknown domain" exception indicates the hook
+// did NOT fire, which would be a regression.
+TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd_AutoDomainFromEnvVar) {
```
huaychou: **[N-1] (Minor)** Missing end-to-end integration test: factory callback to ORT domain receipt to model load

**Issue (carried from v2):** the prior review noted the absence of an integration test verifying the full path: env var to factory construction to `GetCustomOpDomains` callback to ORT receiving the domain to successful model load.

**Suggested fix:**

✅ Fixed: added `UDO_Op_MyAdd_AutoDomainFromEnvVar` (starting L207), which intentionally registers no manual domain and directly asserts `Ort::Session` construction succeeds, confirming the factory hook actually fires.

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -549,6 +562,30 @@ OrtStatus* ORT_API_CALL QnnEpFactory::CreateExternalResourceImporterForDeviceImp
...
+  *num_domains = factory->custom_op_domains_.size();
+  return nullptr;
+}
+
+OrtStatus* ORT_API_CALL QnnEpFactory::GetCustomOpDomainsImpl(
+    _In_ OrtEpFactory* this_ptr,
+    _Out_writes_all_(num_domains) OrtCustomOpDomain** domains,
+    _In_ size_t num_domains) noexcept {
+  const auto* factory = static_cast<const QnnEpFactory*>(this_ptr);
+  if (num_domains > factory->custom_op_domains_.size()) {
```
huaychou: **[N-2] (Minor)** GetCustomOpDomainsImpl lacks defensive bounds check

**Issue (carried from v2):** the loop previously did not validate whether `num_domains` exceeds `custom_op_domains_.size()`, risking an out-of-bounds access.

**Suggested fix:**

✅ Fixed: added an early-return guard `if (num_domains > factory->custom_op_domains_.size())` (L578) before the loop, returning `ORT_INVALID_ARGUMENT`.

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
```diff
@@ -0,0 +1,104 @@
...
+  // because SetOutputShape unconditionally calls SetTensorElementType with this value. Per-op
+  // type configurability (e.g. an optional type suffix in ORT_QNN_CUSTOM_OP_DOMAINS) is
+  // deferred to a follow-up if a concrete non-FLOAT UDO use case arises.
+  static OrtStatusPtr InferOutputShape(Ort::ShapeInferContext& ctx) {
+    if (ctx.GetInputCount() == 0) {
+      return nullptr;
+    }
+    const auto& input_shape = ctx.GetInputShape(0);
+    for (size_t i = 0; i < 1 /* GetOutputCount not available; placeholder has 1 output */; ++i) {
+      ctx.SetOutputShape(i, input_shape);
```
huaychou: **[N-3] (Minor)** InferOutputShape still hardcodes output type to FLOAT, only mitigated via documentation

**Issue:** `InferOutputShape` (L87-95) calls `ctx.SetOutputShape(i, input_shape)`, whose type still defaults to FLOAT. Compared to v2, the underlying limitation itself is **not fixed**, but this round adds an explicit `KNOWN LIMITATION` code comment (L76-85) and a matching update to the user docs (see T-4), clarifying that a non-FLOAT-output UDO will now fail `Ort::Session` construction with a clear type-inference error instead of silently producing wrong results.

**Impact:** the functional limitation remains — a UDO whose output type isn't FLOAT (e.g. an INT32/INT64-output index-style UDO) still cannot load via the automatic `ORT_QNN_CUSTOM_OP_DOMAINS` path. However, the failure mode has clearly improved from v2's "could silently misbehave" to a fail-fast error with a clear message, so the risk level is downgraded accordingly.

**Suggested fix:**

Same suggestion as before: consider letting `QnnUdoPlaceholderOp`'s constructor accept an `output_type` parameter (or one per output), supplied by the caller (`BuildCustomOpDomainsFromEnv` / an extended `ORT_QNN_CUSTOM_OP_DOMAINS` syntax) instead of always assuming FLOAT. If there is no near-term real non-FLOAT UDO need, the current "documented limitation + fail-fast" approach is acceptable — recommend tracking it as a follow-up.

onnxruntime/test/providers/qnn/udo_op_test.cc
```diff
@@ -104,26 +104,42 @@ static GetTestQDQModelFn<InputQType> BuildUDOQDQTestCase(const std::string& op_t
...
+// ORT_QNN_CUSTOM_OP_DOMAINS. In that case the node must be fully assigned to QNN EP (no CPU
+// fallback) for the session to run correctly.
 static void RunOpTestOnCPU(const std::string& op_type,
                            const TestInputDef<float>& input_def,
                            const std::vector<ONNX_NAMESPACE::AttributeProto>& attrs,
                            const std::string& op_packages,
                            int opset_version,
-                           ExpectedEPNodeAssignment expected_ep_assignment) {
+                           ExpectedEPNodeAssignment expected_ep_assignment,
+                           bool register_domain_manually = true) {
```
huaychou: **[N-5] (Minor)** RunOpTestOnCPU's new register_domain_manually=false branch is never exercised by any test call site

**Issue:** `RunOpTestOnCPU` gained a `register_domain_manually` parameter (default `true`, L118) and a matching `else` branch (L134-142, skips manual domain registration when `register_domain_manually=false`). The only call site in the file (L191, `RunOpTestOnCPU("MyAdd", ...)`) does not pass this parameter, so it always uses the default `true` — the new `else` branch is dead code in this PR, never exercised by any test.

**Impact:** no runtime error, but it adds an unverified code path; if that branch's logic were ever wrong (e.g. overlapping/conflicting with the new `UDO_Op_MyAdd_AutoDomainFromEnvVar` test's verification), the test suite would not catch it.

**Suggested fix:**

Either (a) add a test case that calls `RunOpTestOnCPU` with `register_domain_manually=false` so the branch is actually exercised, or (b) if this parameterization isn't currently needed (the new test already uses its own inline session-construction approach), remove the `register_domain_manually` parameter and the `else` branch to reduce complexity.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1705,11 +1705,48 @@ make -C <output_dir>/MyAddOpPackage htp_x86
...
+
+```bash
+export ORT_QNN_CUSTOM_OP_DOMAINS="udo_domain:MyAdd;other_domain:OpA,OpB"
+```
+
+**Then run inference as usual:**
+
+```bash
+./onnxruntime_plugin_ep_onnx_test -v -e qnn -j 1 \
+  -i "backend_path|./libQnnCpu.so op_packages|<op_type>:<op_package_path>:<interface_symbol_name>[:<target>]" \
```
huaychou: **[T-1] (Nit)** Documentation placeholder name inconsistency: interface_symbol vs interface_symbol_name

**Issue (carried from v2):** the same documentation command-line example used two inconsistent placeholder names in the `op_packages` field.

**Suggested fix:**

✅ Fixed: this section was rewritten wholesale (L1708-1749) and now uses a single, consistent `<interface_symbol_name>` example (L1731), and correctly references the real `onnxruntime_plugin_ep_onnx_test` tool name (confirmed to exist in `cmake/external/onnxruntime_prebuilt.cmake`).

onnxruntime/test/providers/qnn/udo_op_test.cc
```diff
@@ -180,6 +196,68 @@ TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd) {
...
+// did NOT fire, which would be a regression.
+TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd_AutoDomainFromEnvVar) {
+  std::filesystem::path path = getLibPath("cpu");
+  if (!std::filesystem::exists(path)) {
+    GTEST_SKIP() << "UDO CPU op package not found: " << path;
+  }
+
+  // Set ORT_QNN_CUSTOM_OP_DOMAINS before RegisterExecutionProviderLibrary so the factory
+  // reads it during CreateEpFactories() (called from RegisterExecutionProviderLibrary).
+  // The previous test unregisters the QNN library on teardown, so a fresh factory is
```
huaychou: **[T-2] (Nit)** New integration test comment still implies a test-ordering dependency

**Issue (carried from v2, wording unchanged):** the new test's comment reads "The previous test unregisters the QNN library on teardown, so a fresh factory is constructed for this test." (L215). This actually describes something true for **every** test — the fixture's `TearDown` unregisters the QNN library regardless of run order (it holds even if this test is run alone via `--gtest_filter`) — but the phrase "the previous test" reads as implying a real inter-test ordering dependency.

**Suggested fix:**

Reword to something like: "Each test's fixture TearDown unregisters the QNN library, so a fresh factory is always constructed here regardless of run order." to avoid implying an ordering dependency.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1705,11 +1705,48 @@ make -C <output_dir>/MyAddOpPackage htp_x86
...
+> needed for model-load schema validation. The actual QNN HW kernel still comes from `op_packages` at
+> session time; the two config items are complementary. The registered placeholder schema currently
+> assumes FLOAT input/output element types (matching all supported UDO usage today); a UDO whose
+> ONNX-level type is genuinely non-FLOAT will fail session construction with a clear type-inference
+> error rather than silently misbehaving.
+
+> **Optional — manual domain registration for CPU fallback:** If you need a real CPU kernel to run
+> the op on the CPU EP (e.g., for accuracy comparison), you can still construct an
+> `Ort::CustomOpDomain` manually and register it via session options. The EP-registered placeholder
+> domain and a user-supplied domain for the same domain name are deduplicated by ORT — both
```
huaychou: **[T-3] (Nit)** Documentation's dual-domain dedup/coexistence claim is not covered by any test in this PR

**Issue:** the newly added doc paragraph claims: "The EP-registered placeholder domain and a user-supplied domain for the same domain name are deduplicated by ORT — both coexist safely." (L1746-1747). This is a specific technical claim about ORT core behavior, but this PR's new integration test `UDO_Op_MyAdd_AutoDomainFromEnvVar` deliberately registers **no** manual domain (it only exercises the EP-only path), and the other existing `udo_op_test.cc` tests only register manually without the env var set — so the "EP domain + user domain coexisting" dual-registration scenario is not actually exercised by any test in this PR.

**Impact:** if this dedup assumption turns out to be wrong (e.g. across ORT versions, or an edge case in the dedup logic), a user following the docs and using both simultaneously could hit an unexpected domain-conflict error that CI would not catch.

**Suggested fix:**

Suggest either (a) adding a test that sets `ORT_QNN_CUSTOM_OP_DOMAINS` AND manually registers a domain of the same name, verifying they do coexist; or (b) if unverified, soften the doc wording (e.g. note that this behavior is per ORT core and not covered by this project's test suite).

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
huaychou: Done

onnxruntime/test/providers/qnn/qnn_basic_test.cc
huaychou: Moved.

onnxruntime/core/providers/qnn/qnn_custom_op_domain_parser.cc
huaychou: Done

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
huaychou: Move default logger before factory constrution

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
huaychou: Done

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
huaychou: Acknowledge

qti-mattsinc: I'll wait until this PR is merged before merging #776

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op_domain_registry.cc
```diff
@@ -0,0 +1,56 @@
...
+                                 std::vector<std::unique_ptr<qnn::QnnUdoPlaceholderOp>>& out_ops) {
+  const std::string custom_op_domains_spec = ReadCustomOpDomainsEnvVar();
+  if (custom_op_domains_spec.empty()) {
+    return;
+  }
+
+  std::vector<CustomOpDomainSpec> specs;
+  ParseCustomOpDomains(custom_op_domains_spec, specs, logger);
+  for (const auto& spec : specs) {
+    Ort::CustomOpDomain domain{spec.domain.c_str()};
```
minfhong-qti, line 49: Since `spec.domain` is local variable, I think the `c_str` usage is UB outside this function.

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
minfhong-qti: What if the actual output shape is different from input?
If the custom op in model has already shapes inferred, will this function still be called?

docs/execution_providers/QNN-ExecutionProvider.md
kuanyul-qti: Should we keep `<op_type2>:<op_package_path2>:<interface_symbol_name2>[:<target2>]` for clear usage?

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
```diff
@@ -19,6 +19,7 @@
 #include "core/providers/qnn/ort_api.h"
 #include "core/providers/qnn/ort_api_version_parser.h"
 #include "core/providers/qnn/qnn_allocator.h"
+#include "core/providers/qnn/custom_op/qnn_custom_op_domain_registry.h"
```
kuanyul-qti: Re-order it.

onnxruntime/test/providers/qnn/udo_op_test.cc
```diff
@@ -180,6 +196,62 @@ TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd) {
                  ExpectedEPNodeAssignment::All);
 }
 
+// Verifies the full factory-level custom-op domain registration path with no manual
+// domain registration: ORT_QNN_CUSTOM_OP_DOMAINS env var → QnnEpFactory construction
+// → GetCustomOpDomains hook → ORT registers domain → model loads, node is assigned to
+// QNN EP, and inference runs.  verify_outputs is false (no CPU-side kernel for reference);
+// RunAndVerifyOutputsWithEP skips the CPU reference session when verify_outputs=false.
+TEST_F(QnnCPUBackendTests, UDO_Op_MyAdd_AutoDomainOnly) {
```
kuanyul-qti, line 208: Do we still need CPU backend tests as we no longer claim to support QNN CPU?

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op_domain_registry.cc
huaychou: Add comment for `spec.domain.c_str()`

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
huaychou: It always calls InferOutputShapeFn even already shape inferred.

onnxruntime/test/providers/qnn/udo_op_test.cc
huaychou: Yes, CPU backend is used as the test vehicle for x86 CI coverage.

docs/execution_providers/QNN-ExecutionProvider.md
huaychou: Updated

onnxruntime/core/providers/qnn/qnn_provider_factory.cc
huaychou: Fixed

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1705,11 +1705,48 @@ make -C <output_dir>/MyAddOpPackage htp_x86
...
+Multiple domains and op-types are separated by `;` and `,` respectively:
+
+```bash
+export ORT_QNN_CUSTOM_OP_DOMAINS="udo_domain:MyAdd;other_domain:OpA,OpB"
+```
+
+**Then run inference as usual:**
+
+```bash
+./onnxruntime_plugin_ep_onnx_test -v -e qnn -j 1 \
```
minfhong-qti: I think `-e` and `-i` should be replaced with plugin ones.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1705,11 +1705,48 @@ make -C <output_dir>/MyAddOpPackage htp_x86
...
+> ONNX-level type is genuinely non-FLOAT will fail session construction with a clear type-inference
+> error rather than silently misbehaving.
+
+> **Optional — manual domain registration for CPU fallback:** If you need a real CPU kernel to run
+> the op on the CPU EP (e.g., for accuracy comparison), you can still construct an
+> `Ort::CustomOpDomain` manually and register it via session options. The EP-registered placeholder
+> domain and a user-supplied domain for the same domain name are deduplicated by ORT — both
+> coexist safely.
+
+For the whole pipeline, refer to the [udo unit test](../../cmake/onnxruntime_unittests_udo.cmake)
```
minfhong-qti: I don't think cmake file is released with doc?

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op_domain_registry.cc
minfhong-qti: As offline discussed, please help double check again whether it really copies.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -1705,11 +1705,48 @@ make -C <output_dir>/MyAddOpPackage htp_x86
 
 #### **Step 5: Execute the Model with UDO**
 
+When a model contains nodes in a custom ONNX domain (e.g., `udo_domain::MyAdd`), ORT must have
```
minfhong-qti, line 1711: Can Python API use custom op?

huaychou: Yes. Update in the document.

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op_domain_registry.cc
huaychou: After confirmation, it copies the const char* into string in:
1. https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/onnxruntime_c_api.cc#L823
2. https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/framework/framework_provider_common.h#L8

docs/execution_providers/QNN-ExecutionProvider.md
huaychou: Updated

docs/execution_providers/QNN-ExecutionProvider.md
huaychou: Updated

onnxruntime/core/providers/qnn/custom_op/qnn_custom_op.h
huaychou: Remove the InferOutputShapeFn and require the source model to declare the custom-op node's output type and shape in its graph in document.

minfhong-qti (approved):

kuanyul-qti (approved):

qti-chuteng (approved):

PR #771, opened 2026-08-27
[QNN] Enable cross device preparation on WoS.

minfhong-qti:
### Description

Introduce a new provider option `enable_cross_device_prepare` to allow user to configure WoS as host mode. This option is then set to backend through QNN API `QnnGlobalConfig_set`.
Provide `QnnBackendManager::IsBackendHostMode` to allow querying whether current backend is configured to be host mode. For example, HNRD is always not adopted when backend in host mode even Stub/Skel libs not present.

Test:
- Add UT for qnn_execution_provider.cc and qnn_backend_manager.cc.
- Add compatibility testcase checking HNRD is not adopted in host mode.
- Allow FCB testcases to be tested on WoS CI by setting the new option.

### Motivation and Context

Starting from QAIRT 2.51, WoS can be configured to host mode by QNN API `QnnGlobalConfig_set` and setting machine type to host. With WoS in host mode, any Windows x86 host only features are expected to work on WoS as well (e.g., FCB).


onnxruntime/core/providers/qnn/builder/qnn_def.h
qti-mbadnara, line 81: Not related to this PR but Maybe we can use the updated Macro here?

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -64,6 +64,13 @@ namespace qnn {
 #endif
 #endif
 
+// Cross device prepare is only available on WoS starting from QAIRT 2.51 (or QNN API 2.40).
+#if defined(_WIN32) && QNN_ARCH_ARM64
+#if QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
```
qti-mbadnara, line 94: Please check with @chuteng-quic / @tirupath-qti to use QNN_SDK_VERSION or QNN_API_VERSION

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -2176,6 +2176,12 @@ inline HRESULT CreateD3D12Buffer(
 
 #endif  // _WIN32 && ...
 
+#if defined(_WIN32) && defined(_M_ARM64)
```
qti-mbadnara, line 2185: Duplicate definition of the macro?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2185,7 +2202,7 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
     return ep->ort_api.CreateStatus(ORT_EP_FAIL, message.c_str());
   }
 
-  if (qnn::IsNpuBackend(ep->qnn_backend_manager_->GetQnnBackendType())) {
+  if (qnn::IsNpuBackend(ep->qnn_backend_manager_->GetQnnBackendType()) && !ep->enable_multi_soc_ep_context_) {
```
qti-mbadnara, line 2360: Why was the second condition added??

qti-mbadnara: Maybe add a comment explaining why

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -3068,7 +3107,22 @@ Ort::Status QnnBackendManager::GetPlatformInfo() {
...
+#if QNN_ARCH_ARM64
+  if (IsBackendHostMode()) {
+    // Backend is configured to host mode and thus should adopt user-specified value like on x86 platform.
+#else
+  {
+#endif  // QNN_ARCH_ARM64
+    // QnnDevice_getPlatformInfo will always return HTP arch 68 and VTCM size 4 on x86 platform even if GetPlatformInfo
+    // is called after device is created. Thus, adopting user-specified value is the only option.
+    if (htp_arch_ != QNN_HTP_DEVICE_ARCH_NONE) {
+      htp_arch_internal_ = htp_arch_;
```
qti-mbadnara, line 3228: Do we need to specify the vtcm_size_internal_ value (default 4, should also be fine) here?

onnxruntime/test/providers/qnn/qnn_ep_context_multi_soc_test.cc
```diff
@@ -37,6 +37,9 @@ void CompileModelWithPerSocOptions(const ProviderOptions& per_soc_options,
   provider_options["backend_type"] = "htp";
   provider_options["offload_graph_io_quantization"] = "0";  // Avoid IO QDQ CPU fallback.
   provider_options["num_graph_prepare_threads"] = "1";
+#if defined(_WIN32) && defined(_M_ARM64) && defined(QNN_CROSS_DEVICE_PREPARE_AVAILABLE)
```
qti-mbadnara: I think QNN_CROSS_DEVICE_PREPARE_AVAILABLE will cover for the first two conditions, can remove defined(_WIN32) && defined(_M_ARM64)

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -358,6 +358,11 @@ The `enable_htp_prepare_and_load` option performs AOT compilation and context lo
 - Setting `enable_htp_prepare_and_load=1` with `ep.context_enable=0` AND an explicit `ep.context_file_path` raises an error (contradictory: "don't persist" + "here's where to persist").
 - If the input model is already a pre-compiled context model (`_ctx.onnx`), `enable_htp_prepare_and_load` is silently ignored with a warning — the model loads directly via the existing AOT path.
 
+|`"enable_cross_device_prepare"`|Description|
+|---|---|
+|'0'|Default. Disabled.|
+|'1'|Enable cross device prepare on WoS, allowing WoS to behave like Windows x86 host. Features originally restricted to Windows x86 host (e.g., Flexible Context Binary) are now supported on WoS when setting this option.|
```
qti-mbadnara: Please add the min QAIRT version or QNN API version needed for this feature

onnxruntime/test/providers/qnn/qnn_model_compatibility_test.cc
```diff
@@ -406,6 +409,80 @@ TEST_F(QnnHTPBackendTests, ModelCompatibility_GetCompatibility) {
...
+TEST_F(QnnHTPBackendTests, ModelCompatibility_GetCompatibility_HostModeNoHnrd) {
+#ifndef QNN_CROSS_DEVICE_PREPARE_AVAILABLE
+  GTEST_SKIP() << "Skip as cross device prepare is not available in this build.";
+#endif
+
+  QNN_SKIP_TEST_IF_NO_PLATFORM_ATTRS();
+  auto platform_attrs = QnnHTPBackendTests::GetPlatformAttributes();
+  const uint32_t htp_arch = static_cast<uint32_t>(platform_attrs.htp_arch);
+  // Host mode is not affected by missing Stub/Skel libraries.
+  auto hnrd_test_handle = std::make_unique<HnrdTestHandle>(htp_arch, /*skip_preapre_lib*/ true);
```
qti-mbadnara:
```suggestion
  auto hnrd_test_handle = std::make_unique<HnrdTestHandle>(htp_arch, /*skip_prepare_lib*/ true);
```

onnxruntime/core/providers/qnn/builder/qnn_def.h
minfhong-qti: Revise with the updated macro is the original owner's responsibility and I believe Hua-Yu has notified each owner.

onnxruntime/test/providers/qnn/qnn_test_utils.h
minfhong-qti: This is intentional as we do not directly include headers from `onnxruntime/core/providers/qnn` here except "real UT" under `onnxruntime/test/providers/qnn/unit`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: This is in fact accidentally removed by #38. Although it is functionally ok without this condition, a misleading error message would keep printed as backend is not setup.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
minfhong-qti: I deliberately not using default 4 as I found out some SoC has default 2 (e.g., I remember it's one of Android in our CI), and thus setting to 4 would cause problem.

onnxruntime/test/providers/qnn/qnn_ep_context_multi_soc_test.cc
minfhong-qti: Removed.

docs/execution_providers/QNN-ExecutionProvider.md
minfhong-qti: Updated.

onnxruntime/test/providers/qnn/qnn_model_compatibility_test.cc
minfhong-qti: Reivsed.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -4130,5 +4130,42 @@ TEST_F(QnnHTPBackendTests, PrepareAndLoad_EmbedModeRespected) {
...
+
+  QNN_SKIP_TEST_IF_NO_PLATFORM_ATTRS();
+  auto platform_attrs = QnnHTPBackendTests::GetPlatformAttributes();
+  const uint32_t htp_arch = static_cast<uint32_t>(platform_attrs.htp_arch);
+
+  const ORTCHAR_T* input_model_file = ORT_MODEL_FOLDER "mul_1.onnx";
+  const std::string output_model_file = "mul_1_ctx.onnx";
+  std::filesystem::remove(output_model_file);
+
+  ProviderOptions provider_options = {{"backend_type", "htp"},
```
yuhuchua-qti: The new CrossDevicePrepare test in qnn_ep_context_test.cc does not seem to set enable_`cross_device_prepare=1` in provider_options. As written, it looks like the test can pass without exercising the new host-mode path added by this PR. Should this test explicitly set {"enable_cross_device_prepare", "1"}?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -470,6 +471,30 @@ Ort::Status QnnBackendManager::LoadQnnSystemLib() {
...
+    configs.push_back(&machine_type_config);
+  }
+#endif
+
+  if (!configs.empty()) {
+    RETURN_IF(qnn_interface_.globalConfigSet == nullptr,
+              "Failed to set global config without QnnGlobalConfig_set API.");
+
+    configs.push_back(nullptr);
+    Qnn_ErrorHandle_t result = qnn_interface_.globalConfigSet(configs.data());
```
yuhuchua-qti: Since this appears to be the first use of QnnGlobalConfig_set in QNN EP, could we clarify the lifetime/scope of this machine type setting? If this API configures QNN library/process-global state, `enable_cross_device_prepare` may not behave like an independent per-session option.

For example, in the same process:
- Session A is created with `enable_cross_device_prepare=1`, so `QnnGlobalConfig_set(machineType=HOST)` is called and manager A reports `IsBackendHostMode()==true`.
- Session B is later created with `enable_cross_device_prepare=0`, so it does not call `QnnGlobalConfig_set` and manager B reports `IsBackendHostMode()==false`.
- However, if the QNN global machine type remains HOST from session A, manager B's local `IsBackendHostMode()` value may no longer reflect the effective QNN backend state.

This could affect code paths that rely on `IsBackendHostMode()`, such as platform-info selection, HNRD detection, and compatibility metadata. Should we document/prohibit mixing `enable_cross_device_prepare=0` and 1 within the same process, or track the effective host-mode state at a shared/global level?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1026,16 +1026,32 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                              logger_);
 #endif
 
+  // Option to enable cross device prepare. Requires QAIRT >= 2.51.
+  static const std::string ENABLE_CROSS_DEVICE_PREPARE = "enable_cross_device_prepare";
```
huaychou: **[M-1] (Major)** Session config key missing htp_ prefix

**Issue**

The new session config key is named `enable_cross_device_prepare`, but all HTP-specific options must carry the `enable_htp_` prefix. Existing analogous keys: `enable_htp_prepare_only`, `enable_htp_prepare_and_load`, `enable_htp_fp16_precision`, `enable_htp_extended_udma_mode`, `enable_htp_graph_splitting`.

**Impact**

The missing `htp_` prefix breaks the established naming convention for HTP-specific provider options. Users cannot tell from the key name that the option is HTP-only. Renaming later would be an API breaking change. Checklist item [C60] explicitly states: reviewers should reject names that drop the `htp_` qualifier — the option only applies to the HTP backend and the name should reflect that.

**Suggested fix:**

Rename the key to `enable_htp_cross_device_prepare` and update all references: the docs table in QNN-ExecutionProvider.md, the `ENABLE_CROSS_DEVICE_PREPARE` constant in qnn_execution_provider.cc, and all test files that set this option.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
huaychou, line 4151: **[M-2] (Major)** CrossDevicePrepare test does not set enable_cross_device_prepare=1

**Issue**

The new `TEST_F(QnnHTPBackendTests, CrossDevicePrepare)` correctly specifies a different `htp_arch` from the device, but the `provider_options` are missing `{"enable_cross_device_prepare", "1"}`. On WoS, cross-arch compilation for a different HTP arch requires host mode (the SDK does not allow it without). As a result, `configure_host_mode_` stays `false`, and `SetGlobalConfig()` never calls `globalConfigSet`. The test does not actually exercise the cross device prepare code path.

**Impact**

The test name implies it validates the cross device prepare feature, but it executes without the feature enabled. This is a fake fence — if the feature path is broken, this test will not catch it.

**Suggested fix:**

Add `{"enable_cross_device_prepare", "1"}` to `provider_options`. This option is silently disabled with a warning on non-ARM64 builds, so it has no impact on other platforms. Alternatively, if the intent is to test that cross-arch compilation fails without the feature, rename the test and add a negative assertion (e.g., expect an error on session creation).

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1026,16 +1026,32 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+  auto enable_cross_device_prepare = ParseBoolOption(ort_api,
+                                                     session_options,
+                                                     FormatEPConfigKey(ENABLE_CROSS_DEVICE_PREPARE),
+                                                     false,
+                                                     logger_);
+#ifndef QNN_CROSS_DEVICE_PREPARE_AVAILABLE
+  if (enable_cross_device_prepare) {
+    ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING, "Cross device prepare is not available in current build.");
+    enable_cross_device_prepare = false;
+  }
```
huaychou, line 1071: **[N-1] (Minor)** Missing effective-state log for enable_cross_device_prepare

**Issue**

After parsing `enable_cross_device_prepare`, the EP logs neither the enabled state nor the fallback-disabled state (when the platform does not support it). Comparable options such as `enable_htp_prepare_only` log explicitly: `"User specified option - enable_htp_prepare_only: ..."`. The `enable_htp_prepare_and_load` option also emits a verbose log.

**Impact**

Users cannot confirm whether the option took effect when reviewing verbose logs, reducing debuggability.

**Suggested fix:**

After the `#ifndef QNN_CROSS_DEVICE_PREPARE_AVAILABLE` block and before passing `enable_cross_device_prepare` to `QnnBackendManagerConfig`, add a verbose log line, e.g.: `ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE, ("enable_cross_device_prepare effective value: " + std::to_string(enable_cross_device_prepare)).c_str())`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -434,6 +436,9 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
     return file_mapped_weights_enabled_;
   }
```
huaychou: **[N-2] (Minor)** IsBackendHostMode() should be declared const

**Issue**

`bool IsBackendHostMode() { return configure_host_mode_; }` is a pure getter that does not modify any member, but it is not declared `const`.

**Impact**

Cannot be called on a `const QnnBackendManager` reference; violates the C++ const-correctness convention.

**Suggested fix:**

Change to `bool IsBackendHostMode() const { return configure_host_mode_; }`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2294,7 +2311,7 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
     return ep->ort_api.CreateStatus(ORT_EP_FAIL, message.c_str());
   }
 
-  if (qnn::IsNpuBackend(ep->qnn_backend_manager_->GetQnnBackendType())) {
+  if (qnn::IsNpuBackend(ep->qnn_backend_manager_->GetQnnBackendType()) && !ep->enable_multi_soc_ep_context_) {
```
huaychou: **[N-3] (Minor)** GetCapabilityImpl skips HTP power config without explanatory comment

**Issue**

In `GetCapabilityImpl`, the new `&& !ep->enable_multi_soc_ep_context_` condition skips HTP power config ID creation, but there is no comment explaining why.

**Impact**

Readers must infer that the multi-SoC path uses `SetupBackendExceptDeviceAndContext()` and does not create a QNN device, so there is no HTP device to associate with a power config. Without a comment, this non-obvious constraint will confuse future maintainers.

**Suggested fix:**

Add a comment before the condition explaining that the multi-SoC path uses `SetupBackendExceptDeviceAndContext` and does not create a QNN device, so HTP power config and HNRD checks do not apply.

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -371,6 +371,11 @@ The `enable_htp_prepare_and_load` option performs AOT compilation and context lo
 - Setting `enable_htp_prepare_and_load=1` with `ep.context_enable=0` AND an explicit `ep.context_file_path` raises an error (contradictory: "don't persist" + "here's where to persist").
 - If the input model is already a pre-compiled context model (`_ctx.onnx`), `enable_htp_prepare_and_load` is silently ignored with a warning — the model loads directly via the existing AOT path.
 
+|`"enable_cross_device_prepare"`|Description|
+|---|---|
+|'0'|Default. Disabled.|
+|'1'|Enable cross device prepare on WoS, allowing WoS to behave like Windows x86 host. Features originally restricted to Windows x86 host (e.g., Flexible Context Binary) are now supported on WoS when setting this option. Requires at least QAIRT 2.51 SDK for this feature; specifying this option with older SDK would not take effect.|
```
huaychou: **[T-1] (Nit)** Doc description phrasing can be more precise

**Issue**

The `'1'` description says "specifying this option with older SDK would not take effect" — this is ambiguous. Does the runtime silently ignore the option or return an error? Looking at the code, when `QNN_CROSS_DEVICE_PREPARE_AVAILABLE` is not defined (older SDK build), the option is warned and forced to `false` (silent downgrade).

**Suggested fix:**

Suggest phrasing: "Requires QAIRT 2.51+ SDK (QNN API 2.40+). On older SDK builds, this option is silently ignored with a warning."

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -3177,7 +3216,22 @@ Ort::Status QnnBackendManager::GetPlatformInfo() {
...
   }
 
-#if defined(__aarch64__) || defined(_M_ARM64) || (defined(_M_ARM64EC))
+#if QNN_ARCH_ARM64
+  if (IsBackendHostMode()) {
+    // Backend is configured to host mode and thus should adopt user-specified value like on x86 platform.
+#else
+  {
+#endif  // QNN_ARCH_ARM64
+    // QnnDevice_getPlatformInfo will always return HTP arch 68 and VTCM size 4 on x86 platform even if GetPlatformInfo
```
huaychou, line 3225: Just wondering, is this the expected behavior?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: Added `htp_` prefix.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
minfhong-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: Added log.

docs/execution_providers/QNN-ExecutionProvider.md
minfhong-qti: Revised.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
minfhong-qti: On x86 platform, yes. The only way to acquire specified SoC through `GetPlatformInfo` on x86 is through `QnnGlobalConfig_set` by setting SoC upfront. However, that API only accepts deprecated `SocModel_t` which hinders the usage.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
kuanyul-qti, line 1544: `enable_htp_cross_device_prepare` is parsed per session, but it is only applied when creating a new QnnBackendManager.
When `share_ep_contexts` or `htp_share_resource_optimization=1` reuses an existing manager, this code only resets the logger. The requested host mode is neither applied nor checked. Since the manager's global QNN config is set only during its first setup, the first session silently determines host mode for all later shared sessions.
You can either reject conflicting host-mode requests when reusing a manager, or document host mode as manager/process-scoped and add a two-session shared-manager test.

onnxruntime/test/providers/qnn/component/qnn_execution_provider_test.cc
```diff
@@ -667,6 +667,35 @@ TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpFp16ClampOverflowTrueOnUnsupported
 }
 #endif  // !QNN_HTP_FP16_CLAMP_OVERFLOW_AVAILABLE
 
+// Covers the ctor WARNING that fires when enable_htp_cross_device_prepare=1 but the
+// current build lacks cross-device-prepare support (QNN_CROSS_DEVICE_PREPARE_AVAILABLE
+// requires QNN_ARCH_ARM64, which is never true on this unit-test host). The guard
+// mirrors the source's own #ifndef so the test self-disables if a future arm64 unit-test
+// tier defines the macro (see qnn_def.h for the version/arch gate).
+#ifndef QNN_CROSS_DEVICE_PREPARE_AVAILABLE
```
kuanyul-qti, line 675: Should it be `QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE`?

huaychou: **[M-1] (Major)** Wrong guard macro name in test: QNN_CROSS_DEVICE_PREPARE_AVAILABLE is never defined

The test `Ctor_EnableCrossDevicePrepareOnUnsupportedBuild_LogsWarning` uses `#ifndef QNN_CROSS_DEVICE_PREPARE_AVAILABLE` as a guard, but the actual macro defined in `qnn_def.h` and `qnn_test_utils.h` is `QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE` (with the `HTP_` infix). Because `QNN_CROSS_DEVICE_PREPARE_AVAILABLE` is never defined anywhere, this guard is always true — the test always runs on every platform and SDK version, including ARM64 WoS with QAIRT >= 2.51.

**Impact**: On WoS builds with SDK >= 2.51, `enable_htp_cross_device_prepare=1` does not trigger the warning in production code (because `QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE` is defined and the `#ifndef` branch does not execute), but the test still runs and asserts that it expects to see the warning, causing a test failure.

**Suggested fix:**

Change #ifndef QNN_CROSS_DEVICE_PREPARE_AVAILABLE to #ifndef QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE (add the HTP_ infix) to match the actual macro name in qnn_def.h. Similarly update the closing comment #endif // !QNN_CROSS_DEVICE_PREPARE_AVAILABLE to #endif // !QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
```diff
@@ -434,6 +436,9 @@ class QnnBackendManager : public std::enable_shared_from_this<QnnBackendManager>
...
 
+  // Note that whether in host mode is meaningless if not on device.
+  bool IsBackendHostMode() { return configure_host_mode_; }
```
huaychou: **[N-1] (Minor)** IsBackendHostMode() should be declared const

`IsBackendHostMode()` is a pure query method that only reads `configure_host_mode_` with no side effects, but is declared as non-const: `bool IsBackendHostMode() { return configure_host_mode_; }`. QNN EP convention is that pure query getter methods are marked `const` (see `GetQnnBackendType() const`, `GetHtpArch() const`, and similar accessor methods).

**Impact**: No compilation error today, but violates QNN EP const-correctness conventions and will cause a problem if the method is ever called in a const context.

**Suggested fix:**

Change the declaration to `bool IsBackendHostMode() const { return configure_host_mode_; }`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2336,7 +2357,7 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
...
```
huaychou: **[N-2] (Minor)** GetCapabilityImpl power config skip for multi-SoC context lacks explanatory comment

In `GetCapabilityImpl`, the HTP power config creation condition was changed from `if (qnn::IsNpuBackend(...))` to `if (qnn::IsNpuBackend(...) && !ep->enable_multi_soc_ep_context_)`, but there is no comment explaining why HTP power config should be skipped in multi-SoC EP context (cross-device-prepare) mode.

**Impact**: Code readability. Future maintainers may remove this condition without understanding its purpose, leading to unnecessary power config attempts in cross-device-prepare mode.

**Suggested fix:**

Add an explanatory comment above the condition, for example:
```
// In cross-device-prepare (host) mode, the WoS session targets a different
// SoC architecture for offline compilation; setting up a local HTP power
// config id on the device is unnecessary and should be skipped.
```

docs/execution_providers/QNN-ExecutionProvider.md
```diff
@@ -371,6 +371,11 @@ The `enable_htp_prepare_and_load` option performs AOT compilation and context lo
 - Setting `enable_htp_prepare_and_load=1` with `ep.context_enable=0` AND an explicit `ep.context_file_path` raises an error (contradictory: "don't persist" + "here's where to persist").
 - If the input model is already a pre-compiled context model (`_ctx.onnx`), `enable_htp_prepare_and_load` is silently ignored with a warning — the model loads directly via the existing AOT path.
 
+|`"enable_htp_cross_device_prepare"`|Description|
```
huaychou: **[T-1] (Nit)** Doc only mentions QAIRT version; consider also noting the QNN API version

The documentation says "Requires at least QAIRT 2.51 SDK for this feature", but the comment in `qnn_def.h` notes both SDK and QNN API versions: "QAIRT 2.51 (or QNN API 2.40)". The convention for other feature macro comments is to list both (e.g., `// HTP native BQ support starts from QAIRT 2.51 (QNN API 2.40).`).

**Suggested fix:**

Consider changing the version note in documentation to "Requires at least QAIRT 2.51 SDK (QNN API 2.40)" to be consistent with the code comment.

onnxruntime/test/providers/qnn/component/qnn_execution_provider_test.cc
minfhong-qti: Fixed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: I think it's a common issue for several other options. We may fix it later.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -4133,5 +4133,43 @@ TEST_F(QnnHTPBackendTests, PrepareAndLoad_EmbedModeRespected) {
 
 #endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
 
+#if defined(_WIN32) && defined(_M_ARM64)
+
+TEST_F(QnnHTPBackendTests, CrossDevicePrepare) {
```
huaychou: Is it possible to test cross device prepare with execution?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.h
minfhong-qti: Revised.

docs/execution_providers/QNN-ExecutionProvider.md
minfhong-qti: Added.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
minfhong-qti: I think it is not possible in current CI.

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -2181,6 +2181,13 @@ inline HRESULT CreateD3D12Buffer(
 #define QNN_HTP_NATIVE_BQ_AVAILABLE
 #endif  // QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
 
+// Cross device prepare is available on WoS starting from QAIRT 2.51 (QNN API 2.40).
```
huaychou: **[N-1] (Minor)** Test helper macro guard missing _M_ARM64EC, inconsistent with production code

**Problem**

In `qnn_test_utils.h`, the guard for `QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE` is:

```
#if defined(_WIN32) && defined(_M_ARM64)
```

But the production code in `qnn_def.h` uses:

```
#if defined(_WIN32) && QNN_ARCH_ARM64
```

where `QNN_ARCH_ARM64` (defined in `ort_api.h`) expands to `__aarch64__ || _M_ARM64 || _M_ARM64EC`. On ARM64EC builds — where an x64 process runs the ARM64EC half of an arm64x fat binary — the production macro is defined but the test macro is not, so tests are silently excluded while production paths are active. The same issue exists in the new per-test guards in `qnn_ep_context_test.cc` and `qnn_model_compatibility_test.cc` (`#if defined(_WIN32) && defined(_M_ARM64)`).

**Impact**

On ARM64EC builds, `CrossDevicePrepare`, `ModelCompatibility_GetCompatibility_HostModeNoHnrd`, and related tests will not compile/run, even though the production host-mode paths (`SetGlobalConfig()`, `GetPlatformInfo()` early return) are active — creating a coverage gap.

**Suggested fix:**

Change the guard in `qnn_test_utils.h`, `qnn_ep_context_test.cc`, `qnn_model_compatibility_test.cc`, and `qnn_ep_context_multi_soc_test.cc` to:
`#if defined(_WIN32) && (defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC))`
This matches the semantics of `QNN_ARCH_ARM64` in `qnn_def.h` and eliminates the coverage gap on ARM64EC builds.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1057,16 +1057,36 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                              logger_);
 #endif
 
+  // Option to enable cross device prepare. Requires QAIRT >= 2.51.
+  static const std::string ENABLE_HTP_CROSS_DEVICE_PREPARE = "enable_htp_cross_device_prepare";
```
huaychou: **[T-1] (Nit)** Session config key constant should use constexpr const char* instead of static const std::string

**Problem**

The new session config key constant is declared as:

```
static const std::string ENABLE_HTP_CROSS_DEVICE_PREPARE = "enable_htp_cross_device_prepare";
```

Adjacent recently-added session config keys consistently use `static constexpr const char*`:

```
static constexpr const char* kEnableHtpMonolithicLstm = "enable_htp_monolithic_lstm";
static constexpr const char* kEnableHtpFp16ClampOverflow = "enable_htp_fp16_clamp_overflow";
```

The name also uses SCREAMING_SNAKE_CASE, while recent additions use camelCase with the `k` prefix.

**Suggested fix:**

Change to `static constexpr const char* kEnableHtpCrossDevicePrepare = "enable_htp_cross_device_prepare";`.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1057,16 +1057,36 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+  auto enable_htp_cross_device_prepare = ParseBoolOption(ort_api,
+                                                         session_options,
```
huaychou: **[T-2] (Nit)** ParseBoolOption call uses constructor parameter session_options instead of member session_options_

**Problem**

The new `ParseBoolOption` call passes `session_options` (the constructor parameter) as the second argument, while all other identical calls in the same constructor pass `session_options_` (the member variable), e.g. lines 1027 and 1042. Both refer to the same object (`session_options_{session_options}`), but the inconsistency is a minor readability issue.

**Suggested fix:**

Change `session_options` to `session_options_` to match all other `ParseBoolOption` calls in the constructor.

huaychou (commented):
Do we need to add document that enabling cross_device_prepare will introduce inference failure?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: Revised.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
minfhong-qti: Revised.

kuanyul-qti (approved):

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -2181,6 +2181,13 @@ inline HRESULT CreateD3D12Buffer(
...
+#if defined(_WIN32) && defined(_M_ARM64)
```
huaychou:
```suggestion
#if defined(_WIN32) && QNN_ARCH_ARM64
```

onnxruntime/test/providers/qnn/qnn_model_compatibility_test.cc
```diff
@@ -589,6 +580,61 @@ TEST_F(QnnHTPBackendTests, ModelCompatibility_GetCompatibility_MultiSoc_Override
 #endif  // !defined(_M_ARM64) && !defined(__aarch64__)
 
 #if defined(_WIN32) && defined(_M_ARM64)
```
huaychou:
```suggestion
#if defined(_WIN32) && QNN_ARCH_ARM64
```

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -4133,5 +4133,43 @@ TEST_F(QnnHTPBackendTests, PrepareAndLoad_EmbedModeRespected) {
 
 #endif  // defined(__aarch64__) || defined(_M_ARM64) || defined(__linux__)
 
+#if defined(_WIN32) && defined(_M_ARM64)
```
huaychou:
```suggestion
#if defined(_WIN32) && QNN_ARCH_ARM64
```

onnxruntime/test/providers/qnn/component/qnn_execution_provider_test.cc
```diff
@@ -667,6 +667,35 @@ TEST_F(QnnUnit_ExecutionProviderTest, Ctor_HtpFp16ClampOverflowTrueOnUnsupported
...
+
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_EnableCrossDevicePrepareDefault_NoWarningLogged_Succeeds) {
+  EpStubContext ctx;
+  ctx.log_severity = ORT_LOGGING_LEVEL_VERBOSE;
+  // enable_htp_cross_device_prepare defaults to false, so the "not available" warning
+  // (which only fires when the option was explicitly requested) must not be logged.
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+  for (const auto& rec : ctx.log_records) {
+    EXPECT_EQ(rec.message.find("Cross device prepare"), std::string::npos)
```
huaychou:
```suggestion
    EXPECT_EQ(rec.message.find("cross device prepare"), std::string::npos)
```

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -90,6 +90,13 @@ namespace qnn {
 #define QNN_HTP_NATIVE_BQ_AVAILABLE
 #endif  // QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
 
+// Cross device prepare is only available on WoS starting from QAIRT 2.51 (or QNN API 2.40).
+#if defined(_WIN32) && QNN_ARCH_ARM64
+#if QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
+#define QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE
+#endif  //  QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
```
huaychou:
```suggestion
#endif  // QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
```

onnxruntime/test/providers/qnn/qnn_test_utils.h
```diff
@@ -2181,6 +2181,13 @@ inline HRESULT CreateD3D12Buffer(
...
+#if QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
+#define QNN_HTP_CROSS_DEVICE_PREPARE_AVAILABLE
+#endif  //  QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
```
huaychou:
```suggestion
#endif  // QNN_API_VERSION_MAJOR > 2 || (QNN_API_VERSION_MAJOR == 2 && QNN_API_VERSION_MINOR >= 40)
```

huaychou (approved):
Other LGTM

onnxruntime/test/providers/qnn/qnn_test_utils.h
minfhong-qti: The macro cannot be used in test and since CI does not execute these UT in arm64ec scenario, skip now.

onnxruntime/test/providers/qnn/component/qnn_execution_provider_test.cc
minfhong-qti: Revised to the whole log message.

onnxruntime/core/providers/qnn/builder/qnn_def.h
minfhong-qti: Fixed.

onnxruntime/test/providers/qnn/qnn_test_utils.h
minfhong-qti: Fixed.

qti-chuteng (approved):

PR #773, opened 2026-08-27
[QNN EP] Re-encode shifted INT16 RMSNorm scale for HTP

1duo:
## Problem
HTP rejects `SFIXED_POINT_16` RMSNorm gamma with a nonzero QNN offset. Qwen's nonnegative INT16 gamma uses offset `32768`, so the RMSNorm QDQ group is not assigned to QNN.

## Fix
For static, per-tensor INT16 gamma on HTP, re-encode the same values as UINT16:

`q_u = q_s + 32768`, `offset_u = offset_s - 32768`

This preserves `(q + offset) * scale` exactly. Other backends, dynamic gamma, symmetric INT16, and non-per-tensor encodings keep existing behavior.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
@@ -149,43 +149,66 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
   bool reencode_scale = false;
   float scale_quant_scale = 0.0f;
   int32_t scale_quant_offset = 0;
+  // BW_SCALE_OFFSET is deliberately excluded: it only ever pairs with a 4-bit datatype, and its
+  // bitwidth has no equivalent in the SCALE_OFFSET encoding re-emitted below.
   if (IsNpuBackend(qnn_model_wrapper.GetQnnBackendType()) &&
       scale_info.qnn_data_type == QNN_DATATYPE_SFIXED_POINT_16 &&
-      scale_info.quant_param.IsPerTensor()) {
+      scale_info.quant_param.IsPerTensor(/*include_bw=*/false)) {
     RETURN_IF_ERROR(scale_info.quant_param.GetPerTensorScaleOffset(scale_quant_scale, scale_quant_offset));
     reencode_scale = scale_quant_offset != 0;
   }
 
   if (reencode_scale) {
+    // Distance between the INT16 and UINT16 representations of the same value.
+    constexpr int32_t kInt16ToUint16Shift = 32768;
+
     RETURN_IF_NOT(scale_info.is_initializer,
                   "QNN RMSNorm requires a static SFIXED_POINT_16 scale when its offset is nonzero");
-    RETURN_IF(scale_quant_offset < -32767 || scale_quant_offset > 32768,
+    // QNN offsets negate ONNX zero points, so an INT16 zero point maps to [-32767, 32768].
+    RETURN_IF(scale_quant_offset < 1 - kInt16ToUint16Shift || scale_quant_offset > kInt16ToUint16Shift,
               "QNN RMSNorm scale offset is outside the valid INT16 range");
 
     std::vector<uint8_t> signed_bytes;
     RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(scale_info.initializer_tensor, signed_bytes));
-    RETURN_IF(signed_bytes.size() % sizeof(int16_t) != 0,
-              "QNN RMSNorm INT16 scale data has an invalid byte size");
+    const size_t expected_num_bytes = utils::GetQnnTensorDataSizeInBytes(*squeezed_shape,
+                                                                         QNN_DATATYPE_SFIXED_POINT_16);
+    RETURN_IF(signed_bytes.size() != expected_num_bytes,
+              ("QNN RMSNorm INT16 scale holds " + std::to_string(signed_bytes.size()) +
+               " bytes of data but its shape implies " + std::to_string(expected_num_bytes) + ".")
+                  .c_str());
 
     std::vector<uint8_t> unsigned_bytes(signed_bytes.size());
     for (size_t i = 0; i < signed_bytes.size(); i += sizeof(int16_t)) {
       int16_t signed_value = 0;
       std::memcpy(&signed_value, signed_bytes.data() + i, sizeof(signed_value));
-      const uint16_t unsigned_value = static_cast<uint16_t>(static_cast<int32_t>(signed_value) + 32768);
+      const uint16_t unsigned_value = static_cast<uint16_t>(static_cast<int32_t>(signed_value) +
+                                                            kInt16ToUint16Shift);
       std::memcpy(unsigned_bytes.data() + i, &unsigned_value, sizeof(unsigned_value));
     }
 
     // q_s + offset_s == (q_s + 32768) + (offset_s - 32768).
+    const int32_t unsigned_quant_offset = scale_quant_offset - kInt16ToUint16Shift;
+
+    ORT_CXX_LOG(logger, ORT_LOGGING_LEVEL_VERBOSE,
+                ("RMSNorm node " + node_unit.Name() + ": re-encoding SFIXED_POINT_16 scale `" + scale_name +
+                 "` as UFIXED_POINT_16, offset " + std::to_string(scale_quant_offset) + " -> " +
+                 std::to_string(unsigned_quant_offset) + ", because HTP rejects a shifted INT16 gamma.")
+                    .c_str());
+
     const std::string unsigned_scale_name = utils::UniqueNameGenerator().New(scale_name, "_u16");
     QnnTensorWrapper unsigned_scale(unsigned_scale_name,
                                     QNN_TENSOR_TYPE_STATIC,
                                     QNN_DATATYPE_UFIXED_POINT_16,
-                                    QnnQuantParamsWrapper::PerTensor(scale_quant_scale, scale_quant_offset - 32768),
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
```diff
@@ -144,11 +145,51 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
...
+
+  if (reencode_scale) {
+    RETURN_IF_NOT(scale_info.is_initializer,
+                  "QNN RMSNorm requires a static SFIXED_POINT_16 scale when its offset is nonzero");
+    RETURN_IF(scale_quant_offset < -32767 || scale_quant_offset > 32768,
+              "QNN RMSNorm scale offset is outside the valid INT16 range");
+
+    std::vector<uint8_t> signed_bytes;
+    RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(scale_info.initializer_tensor, signed_bytes));
+    RETURN_IF(signed_bytes.size() % sizeof(int16_t) != 0,
```
qti-shubham: instead should we do element count check ?

qti-shubham: may use `utils::GetQnnTensorDataSizeInBytes`

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
```diff
@@ -144,11 +145,51 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
...
+      scale_info.qnn_data_type == QNN_DATATYPE_SFIXED_POINT_16 &&
+      scale_info.quant_param.IsPerTensor()) {
+    RETURN_IF_ERROR(scale_info.quant_param.GetPerTensorScaleOffset(scale_quant_scale, scale_quant_offset));
+    reencode_scale = scale_quant_offset != 0;
+  }
+
+  if (reencode_scale) {
+    RETURN_IF_NOT(scale_info.is_initializer,
+                  "QNN RMSNorm requires a static SFIXED_POINT_16 scale when its offset is nonzero");
+    RETURN_IF(scale_quant_offset < -32767 || scale_quant_offset > 32768,
```
qti-shubham: is it worth to have a const kInt16ZeroPointShift ?

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
```diff
@@ -144,11 +145,51 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
...
+    QnnTensorWrapper unsigned_scale(unsigned_scale_name,
+                                    QNN_TENSOR_TYPE_STATIC,
+                                    QNN_DATATYPE_UFIXED_POINT_16,
+                                    QnnQuantParamsWrapper::PerTensor(scale_quant_scale, scale_quant_offset - 32768),
+                                    std::vector<uint32_t>(*squeezed_shape),
+                                    std::move(unsigned_bytes));
+    RETURN_IF_NOT(qnn_model_wrapper.AddTensorWrapper(std::move(unsigned_scale)),
+                  "Failed to add converted RMSNorm scale tensor.");
+    input_names.push_back(unsigned_scale_name);
+    scale_info.shape = *squeezed_shape;
```
qti-shubham, line 212: should we also update `scale_info.qnn_data_type to UFIXED_POINT_16 ` to avoid stale dtype propagation ?

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
```diff
@@ -144,11 +145,51 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
...
+
+    std::vector<uint8_t> unsigned_bytes(signed_bytes.size());
+    for (size_t i = 0; i < signed_bytes.size(); i += sizeof(int16_t)) {
+      int16_t signed_value = 0;
+      std::memcpy(&signed_value, signed_bytes.data() + i, sizeof(signed_value));
+      const uint16_t unsigned_value = static_cast<uint16_t>(static_cast<int32_t>(signed_value) + 32768);
+      std::memcpy(unsigned_bytes.data() + i, &unsigned_value, sizeof(unsigned_value));
+    }
+
+    // q_s + offset_s == (q_s + 32768) + (offset_s - 32768).
```
qti-shubham, line 189: a VERBOSE log to mention SYM -> ASYM transition

onnxruntime/test/providers/qnn/rmsnormalization_test.cc
```diff
@@ -154,6 +154,14 @@ TEST_F(QnnHTPBackendTests, RMSNorm1D_LastAxis_StaticScale_AU16_WU8) {
...
 }
 
+TEST_F(QnnHTPBackendTests, RMSNorm1D_LastAxis_StaticScale_AU16_WS16Shifted) {
+  RunRMSNormQDQTest<uint16_t, int16_t>(TestInputDef<float>({1, 2, 3}, false, GetFloatDataInRange(0.0f, 10.0f, 6)),
+                                       TestInputDef<float>({3}, true, GetFloatDataInRange(0.0f, 1.0f, 3)),
+                                       {test::MakeAttribute("axis", static_cast<int64_t>(-1))},
+                                       ExpectedEPNodeAssignment::All,
+                                       true);
+}
+
```
qti-shubham, line 166: two more tests to validate :
1. Scale with nonzero offset (testing condition at line 160 - 161)
2. Dynamic scale

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
```diff
@@ -144,11 +145,51 @@ Ort::Status RMSNormalizationOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_
...
                 "non-1 leading dimensions and cannot be squeezed to match.");
 
-  if (squeezed_shape->size() == scale_info.shape.size()) {
+  const std::string& scale_name = inputs[SCALE_IDX].name;
+  bool reencode_scale = false;
+  float scale_quant_scale = 0.0f;
+  int32_t scale_quant_offset = 0;
+  if (IsNpuBackend(qnn_model_wrapper.GetQnnBackendType()) &&
+      scale_info.qnn_data_type == QNN_DATATYPE_SFIXED_POINT_16 &&
+      scale_info.quant_param.IsPerTensor()) {
```
qti-shubham: should we pass `IsPerTensor(/*include_bw*/ true)` ?

qti-shubham: possibility of 4 bit quant

qti-shubham (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
1duo: Done

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
1duo: Done — named `kInt16ToUint16Shift`.

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
1duo: Done

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
1duo: Done

onnxruntime/test/providers/qnn/rmsnormalization_test.cc
1duo: Done — added `RMSNorm1D_LastAxis_DynamicScale_AU16_WS16Shifted`.

One test covers both bullets, since they are the same scenario: the guard at 160-161 only fires for a nonzero-offset scale that is *not* an initializer, i.e. a dynamic one. Verified against the run log that it takes that path rather than passing vacuously:

```
REASON : QNN RMSNorm requires a static SFIXED_POINT_16 scale when its offset is nonzero
```

RMSNorm then falls back to CPU while the surrounding QDQ nodes stay on QNN, so the expectation is `Some`, not `None`.

Note the assertion is weaker than I would like: `Some` only checks `num_ep_nodes > 0`, so it would still pass if a regression made QNN claim the RMSNorm node. What it does lock in is that the rejection stays soft — session init succeeds and the results stay numerically correct with RMSNorm on CPU. Asserting the specific node placement would need a harness the QNN test utils do not currently expose; happy to add one if you think it is worth it.

Also worth flagging: the offset range check just below (162-163) is not reachable from a valid model. An ONNX INT16 zero point is inherently in [-32768, 32767], so the negated QNN offset is always within [-32767, 32768]. It is purely defensive, so I did not add a test for it.

onnxruntime/core/providers/qnn/builder/opbuilder/rmsnormalization_op_builder.cc
qti-yuduo: don;t have concrete thing to verify and feeling this behavior is safer than enable... doing more tests with real-models, likely leaving this as-is.

PR #775, opened 2026-08-27
[QNN EP] Keep quantized TopK inside its QDQ node group

1duo:
Fixes the QDQ node-group check so a quantized `TopK` stays quantized.

`CanCreateNodeGroup` required **every** output of a node to be `QuantizeLinear`-wrapped and to have exactly one consumer. `TopK`'s second output is integer indices — never quantized, and usually consumed by more than one node — so `OrtTopKNodeGroupSelector::Check` could never return `true` for a real `TopK`. The group was rejected, `TopK` stayed a standalone float node, and the EP kept a real `Dequantize` in front of it, running `TopK` on float32 data inside an otherwise fully quantized graph.

Both old constraints exist for the same reason: fusing the group deletes the float tensor between the target node and its `Q`. That only applies to outputs actually being absorbed, so both are now applied per output — an output is absorbed when a `Q` in `q_nodes` is its single consumer, and only then must it not be a graph output. Outputs with no `Q` consumer stay real tensors and constrain nothing. Every collected `Q` must end up absorbed, otherwise a `Q` would be folded away while the tensor it reads is still live.

### DIFF
--- a/onnxruntime/core/providers/qnn/qnn_ep_utils.cc
+++ b/onnxruntime/core/providers/qnn/qnn_ep_utils.cc
@@ -543,42 +543,44 @@ bool CanCreateNodeGroup(const OrtGraph* graph, const OrtApi& ort_api, const OrtN
     return false;
   }
 
-  // Check if the number of Q outputs matches the number of outputs that exist
   size_t num_outputs = 0;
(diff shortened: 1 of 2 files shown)


qti-ashwshan (approved):
LGTM

tirupath-qti (approved):

minfhong-qti:
Since the account used to create this PR is not contributor, the CI is in fact not run but merely bypassed. The new added test is now failing on mainline.

PR #776, opened 2026-08-27
Guard noexcept functions with QNN_EP_API_IMPL_BEGIN/END

qti-mattsinc:
* Implementations of ORT API functions marked noexcept should not throw. The previous strategy has been to attempt to isolate potentially-throwing parts of these functions and wrap them in try/catch. But this is brittle and other projects advocate for function-level exception guards instead ([0], [1]).
* This change adds such exception guards and and places them on all noexcept API entrypoints.

[0]: https://github.com/microsoft/onnxruntime/blob/v1.29.0/docs/C_API_Guidelines.md
[1]: https://github.com/microsoft/wil/wiki/Error-handling-helpers#using-exception-based-code-in-a-routine-that-cannot-throw

### DIFF
--- a/onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
+++ b/onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
@@ -24,6 +24,7 @@ GenieNodeComputeInfo::GenieNodeComputeInfo(QnnEp& ep,
 OrtStatus* GenieNodeComputeInfo::CreateStateImpl(OrtNodeComputeInfo* this_ptr,
                                                  OrtNodeComputeContext* compute_context,
                                                  void** compute_state) {
+  QNN_EP_API_IMPL_BEGIN
   auto* node_compute_info = static_cast<GenieNodeComputeInfo*>(this_ptr);
   auto& ep = node_compute_info->ep;
   auto& builder = node_compute_info->builder;
@@ -61,21 +62,16 @@ OrtStatus* GenieNodeComputeInfo::CreateStateImpl(OrtNodeComputeInfo* this_ptr,
                                     std::filesystem::is_directory(parent_folder_path, dir_ec) &&
                                     !dir_ec;
   if (parent_is_searchable) {
-    try {
-      for (const auto& entry : std::filesystem::directory_iterator(parent_folder_path)) {
-        if (entry.is_regular_file()) {
-          const std::string filename = entry.path().filename().string();
-          if (filename.rfind("tmp", 0) == 0 &&
-              filename.size() >= 5 &&
-              filename.compare(filename.size() - 5, 5, ".json") == 0) {
-            extension_path = entry.path().string();
-            break;
-          }
+    for (const auto& entry : std::filesystem::directory_iterator(parent_folder_path)) {
+      if (entry.is_regular_file()) {
+        const std::string filename = entry.path().filename().string();
+        if (filename.rfind("tmp", 0) == 0 &&
+            filename.size() >= 5 &&
+            filename.compare(filename.size() - 5, 5, ".json") == 0) {
+          extension_path = entry.path().string();
+          break;
         }
       }
-    } catch (const std::filesystem::filesystem_error& e) {
-      std::string error_msg = std::string("Error searching for extension file: ") + e.what();
-      return ep.ort_api.CreateStatus(ORT_EP_FAIL, error_msg.c_str());
     }
   } else if (!parent_folder_path.empty()) {
     ORT_CXX_LOG(ep.logger_, ORT_LOGGING_LEVEL_WARNING,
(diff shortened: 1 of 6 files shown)


onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
qti-mattsinc, line 1: note: the function pointers in `OrtNodeComputeInfo` are not marked noexcept in the ORT headers (I'm not really sure why). Still, they are API entrypoints that return `OrtStatus*`, though.  So I think it makes sense to include the guards; the errors should propogate upwards through the OrtStatus and not exceptions. This applies to the `GenieNodeComputeInfo` in this file and the `QnnNodeComputeInfo` in qnn_execution_provider.cc.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1575,8 +1575,10 @@ QnnEp::~QnnEp() {
 }
 
 const char* ORT_API_CALL QnnEp::GetNameImpl(const OrtEp* this_ptr) noexcept {
```
qti-mattsinc, line 1577: note: simple functions like `QnnEp::GetNameImpl`, `QnnEpFactory::CreateDataTransferImpl`, and `QnnEpFactory::IsStreamAwareImpl` are trivially non-throwing. So we arguably don't need the guards here. But for uniformity I've added them.

qti-mattsinc:
@minfhong-qti, since you were involved with the original port to the ABI EP, I am wondering if you agree with the proposed use of the QNN_EP_API_IMPL_BEGIN/END macros in this PR. I think that following this pattern would help ensure that our error handling is correct at the plugin ep boundary, and this topic came up in review of separate PR (#692). Just curious to hear your thoughts before I move this PR out of a draft for review.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mattsinc: Actually, the convention in ORT core seems to be leaving out the guards from trivial functions, e.g. [OrtApis::MemoryInfoGetId](https://github.com/microsoft/onnxruntime/blob/2e2543fbe9fae542f921d47a72d21d5a4ef0b710/onnxruntime/core/framework/allocator.cc#L343-L346). So I'll probably remove them from these simple functions.

onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
johnpaultaken: This is a good change, but please let me know before merge to avoid merge conflicts.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2003,6 +2005,7 @@ static void GetContextOnnxModelFilePath(const std::string& user_context_cache_pa
 OrtStatus* ORT_API_CALL QnnEp::GetGenieCapability(OrtEp* this_ptr,
                                                   const OrtGraph* graph,
                                                   OrtEpGraphSupportInfo* graph_support_info) {
+  QNN_EP_API_IMPL_BEGIN
```
minfhong-qti: `GetGenieCapability` isn't really an ORT API. I think we may remove the `ORT_API_CALL` here, and let `GetCapabilityImpl` to catch it.
In addition, probably adjust the place and order of this function in the header.

onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
minfhong-qti: We were referring to this example https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/test/autoep/library/example_plugin_ep during migration, and in their implementation, there is no `noexcept` for `NodeComputeInfo`.
Not sure whether there is a better way to unify for `QnnNodeComputeInfo`, `GenieNodeComputeInfo` and possibly any info defined later on.

minfhong-qti (commented):
Really appreciate for the effort.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mattsinc: Good point. Made it into a regular member function of `QnnEp` and removed the guards. Also, made sure to call `ReleaseStatus` in a couple places that weren't doing so before.

onnxruntime/core/providers/qnn/genie/genie_node_compute_info.cc
qti-mattsinc: These function signatures in the headers in ORT core don't have `noexcept` (because they don't use the `ORT_API2_STATUS` macro), so that probably explains the signature in the example ep. I think it's possible that these functions predate the new style of `ORT_API2_STATUS` which declares `noexcept`. Because they return `OrtStatus`, having the guards still seem beneficial to me, because otherwise we would be comingling exceptions and `OrtStatus` as ways to report errors from this function. But I'm also OK with leaving the guards out of these functions.

onnxruntime/core/providers/qnn/qnn_external_resource_importer.cc
qti-mattsinc, line 272: Leaving the guard off the functions that are not implemented because they are trivially non-throwing in their current state. They'll need to be added when they get implemented.

PR #777, opened 2026-08-27
[QNN EP] Add support for Swish

qti-yuduo:
### Description

Add QNN EP support for the ONNX opset-24 `Swish` operator.

`Swish` and `QuickGelu` use the same `x * sigmoid(alpha * x)` formulation, so this reuses the existing decomposition into QNN `Sigmoid` and `ElementWiseMultiply` nodes. The shared builder now selects the schema-specific default alpha (`1.0` for `Swish`, `1.702` for `QuickGelu`) and continues to support explicit alpha values.

`Swish` is also added to QDQ unary-op selection.

### Motivation and Context

ONNX opset-24 models that retain `Swish` currently fall back at every activation and split otherwise QNN-compatible graphs. This was observed in an autoregressive Qwen2.5-0.5B export: replacing `Swish` with `Sigmoid` + `Mul` removed the partition breaks and enabled full QNN capture. Supporting the equivalent lowering directly in QNN EP avoids requiring an external model rewrite.

### DIFF
--- a/onnxruntime/test/providers/qnn/simple_op_test.cc
+++ b/onnxruntime/test/providers/qnn/simple_op_test.cc
@@ -211,6 +211,22 @@ TEST_F(QnnCPUBackendTests, UnaryOp_HardSwish) {
                  ExpectedEPNodeAssignment::All);
 }
 
+TEST_F(QnnCPUBackendTests, UnaryOp_Swish_DefaultAlpha) {
+  RunOpTestOnCPU("Swish",
+                 {TestInputDef<float>({7}, false, {-10.0f, -2.0f, -1.0f, 0.0f, 1.0f, 2.0f, 10.0f})},
+                 {},
+                 24,
+                 ExpectedEPNodeAssignment::All);
+}
+
+TEST_F(QnnCPUBackendTests, UnaryOp_Swish_CustomAlpha) {
+  RunOpTestOnCPU("Swish",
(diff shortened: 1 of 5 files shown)


onnxruntime/test/providers/qnn/simple_op_test.cc
```diff
@@ -211,6 +211,22 @@ TEST_F(QnnCPUBackendTests, UnaryOp_HardSwish) {
                  ExpectedEPNodeAssignment::All);
 }
 
+TEST_F(QnnCPUBackendTests, UnaryOp_Swish_DefaultAlpha) {
```
qti-mbadnara: Do we even need QnnCPU tests anymore?

qti-mbadnara (approved):
Minor nit regarding CPU tests but LGTM!

onnxruntime/test/providers/qnn/simple_op_test.cc
qti-yuduo: why not? we're not yet removing support right?

qti-mbadnara: I thought #500 stop advertising it by default?

qti-yuduo: okay.. I can remove CPU and stop adding them in the future.. But I think it's not enforced anywhere, any plans to clean up and disable entirely? (so less confusing)

qti-mbadnara:
Can you please update the docs as well to indicate Swish as a supported Op @qti-yuduo

qti-shubham (approved):

tirupath-qti (approved):
LGTM

PR #778, opened 2026-08-28
[QNN EP] Fix graph splitting config incorrectly applied to VTCM binary-load path

qti-mbadnara:
### Description

Follow-up on #645 

CreateContextVtcmBackupBufferSharingEnabled always terminates via QnnContext_createFromBinaryListAsync. Graph splitting is a context-creation feature, it is applied at QnnContext_create time and baked into the context binary.

### Changes

-   Remove enable_htp_graph_splitting from CreateContextVtcmBackupBufferSharingEnabled signature and call sites
-   Remove graph splitting config block from that function; keep it only in CreateContext
-   Emit a warning when enable_htp_graph_splitting=1 and htp_share_resource_optimization=1 are both set, since the VTCM path always routes through the binary-load API and graph splitting silently has no effect

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -2129,13 +2129,6 @@ Ort::Status QnnBackendManager::SetupBackend(
   }
 
   if (status.IsOK() && (htp_share_resource_optimization_ == 1 || !load_from_cached_context)) {
-    if (htp_share_resource_optimization_ == 1 && enable_htp_graph_splitting) {
-      ORT_CXX_LOG_PTR(logger_ptr_,
-                      ORT_LOGGING_LEVEL_WARNING,
-                      "enable_htp_graph_splitting=1 has no effect when htp_share_resource_optimization=1. "
-                      "Graph splitting is a context-creation feature; the VTCM sharing path uses "
-                      "QnnContext_createFromBinaryListAsync (a binary-load API) and does not support it.");
-    }
     status = htp_share_resource_optimization_ == 1
                  ? CreateContextVtcmBackupBufferSharingEnabled(context_bin_map,
                                                                io_dispatch)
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2149,10 +2129,16 @@ Ort::Status QnnBackendManager::SetupBackend(
   }
 
   if (status.IsOK() && (htp_share_resource_optimization_ == 1 || !load_from_cached_context)) {
+    if (htp_share_resource_optimization_ == 1 && enable_htp_graph_splitting) {
```
minfhong-qti: I think the check should be performed when EP parsing the option and verified it is provided along with context enable.

minfhong-qti (commented):
Thanks for the quick follow-up.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Done!

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1368,6 +1368,12 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     enable_htp_graph_splitting_ = false;
   }
 #endif
+  if (enable_htp_graph_splitting_ && htp_share_resource_optimization_ == 1) {
```
```suggestion
  if (enable_htp_graph_splitting_ && !context_cache_enabled_) {
    ORT_CXX_LOG(...);
    enable_htp_graph_splitting_ = false;
  }
```
Prefer checking against `context_cache_enabled_` directly.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1369,6 +1369,12 @@ QnnEp::QnnEp(QnnEpFactory& factory,
   }
 #endif
 
+  if (enable_htp_graph_splitting_ && !context_cache_enabled_) {
```
minfhong-qti: Apologize I misunderstand how this option works. I think the warning should be around https://github.com/onnxruntime/onnxruntime-qnn/blob/main/onnxruntime/core/providers/qnn/qnn_execution_provider.cc#L2981 like existing warning for `prepare_and_load`.

qti-mbadnara: Done

minfhong-qti (approved):

PR #781, opened 2026-08-30
[QNN EP] Fix burst mode per-inference power overhead regression

qti-mbadnara:
### Description

Fixes a throughput regression introduced by PR #38 where sessions configured with `htp_performance_mode|burst` incur unnecessary per-inference RPC overhead and periodic DSP cold-wakeup penalties due to `kHtpBurst` being inadvertently routed through the sustained high-performance timer path. Also restores the `bool` return type of `GetPerThreadHtpPowerConfigs()`so that `OnRunStart` skips map operations entirely when no per-inference config is needed, and fixes a latent UB in `SetPerThreadHtpPowerConfigs()` when only a post-run override is active.

---

## Motivation & Context

PR #38 ("Added sustained high performance mode implementation") added `kHtpSustainedHighPerformance` — a new mode that keeps the DSP active between inferences via a per-inference `SetState(RUN_START/RUN_DONE)` state machine and a 300 ms release timer. 

The implementation changed `GetPerThreadHtpPowerConfigs()` from a `bool`-returning function (that read only
per-`Run()` options) to a `void` function that unconditionally copies the session-level `default_htp_performance_mode_` into `per_thread_htp_power_configs.default_perf_mode` for **all** non-default modes, and unconditionally called `AddPerThreadHtpPowerConfigMapping()` on every inference regardless of whether any config was actually needed. This caused `kHtpBurst` — set via the existing `htp_performance_mode|burst` session option — to go
through the same per-inference timer machinery that was designed exclusively for `kHtpSustainedHighPerformance`.

### Behavior comparison

| | Pre-PR #38 | Post-PR #38 (regression) | After this fix |
|---|---|---|---|
| Burst applied | Once at session init via `SetHtpPowerConfigs()` | Per inference via `SetState(RUN_START/RUN_DONE)` | Once at session init (restored) |
| Per-inference RPC calls | 0 | 2+ per inference | 0 |
| Per-inference map insert/erase | 0 | Every inference | 0 |
| 300 ms release timer | Never armed | Armed on every `RUN_DONE` | Never armed |
| DSP cold-wakeup | None | Periodic (timer fires when idle) | None |

### Regression mechanism

When `default_perf_mode = kHtpBurst` was set on the per-thread config, `SetPerThreadHtpPowerConfigs()` dispatched both `RUN_START` and `RUN_DONE` through `SetSustainedPerformance()`. On `RUN_DONE`, `SetSustainedPerformance()` registers the power-config ID in `boosted_config_ids_` and re-arms the 300 ms release timer (`kDefaultTimerTimeoutUs = 300000 µs`). When the timer fires and `inflight_run_count_ == 0` (the window between inferences), the DSP is relaxed to SVS+. The next inference then pays a cold-wakeup cost in the RPC/CPU↔DSP handoff layer.

This manifests as elevated P99.5–P99.9 tail latency while P50 remains flat — consistent with a periodic, low-probability spike rather than a uniform slowdown.


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -3336,10 +3343,11 @@ OrtStatus* ORT_API_CALL QnnEp::OnRunStartImpl(_In_ OrtEp* this_ptr, _In_ const :
...
     qnn::PerThreadHtpPowerConfigs_t per_thread_htp_power_configs;
-    ep->GetPerThreadHtpPowerConfigs(per_thread_htp_power_configs, run_options);
-    per_thread_htp_power_configs.power_config_id = htp_power_config_id;
-    RETURN_IF_ERROR(ep->qnn_backend_manager_->AddPerThreadHtpPowerConfigMapping(thread_id,
-                                                                                per_thread_htp_power_configs));
+    if (ep->GetPerThreadHtpPowerConfigs(per_thread_htp_power_configs, run_options)) {
+      per_thread_htp_power_configs.power_config_id = htp_power_config_id;
+      RETURN_IF_ERROR(ep->qnn_backend_manager_->AddPerThreadHtpPowerConfigMapping(thread_id,
+                                                                                  per_thread_htp_power_configs));
+    }
```
qti-kromero, line 3350: should something happen if false is returned here or is this truly a pass through?

qti-mbadnara: It is truly a pass-through, and that is the correct behavior. The session-level power state applied once at init via SetHtpPowerConfigs() is already in effect.

PR #790, opened 2026-09-01
[QNN EP] Multi Core NPU Support for QNN EP

qti-ashwshan:
### Description
Add htp_num_cores provider option to distribute HTP inference across multiple NPU cores on supported SoCs.

  - Parse and validate htp_num_cores in provider options (uint32, 0 = default single-core).
  - Query available HTP core IDs from the backend at device setup — no hardcoded core IDs.
  - Pass core count to both QNN graph config (QNN_HTP_GRAPH_CONFIG_OPTION_NUM_HTP_CORES) and device config.
  - Gate multi-core to AOT-only flows: context binary load (context_enable) and enable_htp_prepare_and_load. JIT / regular ONNX execution with htp_num_cores > 0
  returns a clear error directing the user to switch to AOT.
  - Unit tests for option validation, AOT-only policy enforcement, and backend manager core configuration.
  - Documentation update in QNN-ExecutionProvider.md

### Motivation and Context
- Certain Qualcomm SoCs expose multiple NPU cores that can execute HTP workloads in parallel, reducing inference latency for large models.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
@@ -3096,6 +3096,7 @@ Ort::Status QnnBackendManager::GetPlatformInfo() {
       if (htp_ext && htp_ext->devType == QNN_HTP_DEVICE_TYPE_ON_CHIP) {
         htp_arch_internal_ = htp_ext->onChipDevice.arch;
         vtcm_size_internal_ = static_cast<uint32_t>(htp_ext->onChipDevice.vtcmSize);
+        num_cores_internal_ = hw_info.v1.numCores;
         break;
       }
     }
--- a/onnxruntime/core/providers/qnn/builder/qnn_def.h
+++ b/onnxruntime/core/providers/qnn/builder/qnn_def.h
@@ -159,6 +159,7 @@ typedef struct HtpGraphConfigs {
   bool enable_htp_fp16_precision = false;
   bool enable_htp_monolithic_lstm = false;
   bool enable_htp_fp16_clamp_overflow = false;  // Intentionally undocumented; for internal/diagnostic use only.
+  uint32_t num_cores = 0;
 } HtpGraphConfigs_t;
 
 enum class QnnBackendType : uint8_t {
--- a/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
+++ b/onnxruntime/core/providers/qnn/qnn_execution_provider.cc
@@ -484,7 +484,8 @@ void QnnEp::ParsePerSocHtpConfigs() {
                                   htp_graph_configs_.htp_graph_finalization_opt_mode,
                                   htp_graph_configs_.enable_htp_fp16_precision,
                                   htp_graph_configs_.enable_htp_monolithic_lstm,
-                                  htp_graph_configs_.enable_htp_fp16_clamp_overflow};
+                                  htp_graph_configs_.enable_htp_fp16_clamp_overflow,
+                                  htp_graph_configs_.num_cores};
     htp_graph_configs_per_soc_.push_back(std::move(config));
   }
 
@@ -1035,6 +1036,20 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     GetSessionConfigEntryOrDefault(ort_api, session_options_, FormatEPConfigKey("vtcm_mb"), "0", vtcm_mb_str);
     if (!vtcm_mb_str.empty() && vtcm_mb_str != "0") {
       ParseVtcmSize(vtcm_mb_str, htp_graph_configs_.vtcm_size_in_mb, logger_);
+    }
+
+    // HTP num cores
+    std::string htp_num_cores_str;
+    GetSessionConfigEntryOrDefault(ort_api, session_options_, FormatEPConfigKey("htp_num_cores"), "0", htp_num_cores_str);
+    if (!htp_num_cores_str.empty() && htp_num_cores_str != "0") {
+      try {
+        htp_graph_configs_.num_cores = static_cast<uint32_t>(std::stoul(htp_num_cores_str));
+        ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_VERBOSE,
+                     ("User specified htp_num_cores: " + htp_num_cores_str).c_str());
+      } catch (...) {
+        ORT_CXX_LOG(logger_, ORT_LOGGING_LEVEL_WARNING,
+                     ("Invalid htp_num_cores: " + htp_num_cores_str + " will be skipped").c_str());
+      }
     }
   }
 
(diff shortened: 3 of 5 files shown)


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1036,6 +1037,20 @@ QnnEp::QnnEp(QnnEpFactory& factory,
     if (!vtcm_mb_str.empty() && vtcm_mb_str != "0") {
       ParseVtcmSize(vtcm_mb_str, htp_graph_configs_.vtcm_size_in_mb, logger_);
     }
+
+    // HTP num cores
```
minfhong-qti: This should be parsed before `ParsePerSocHtpConfigs`.

qti-ashwshan: thanks! will take a look.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1057,6 +1058,24 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                                              logger_);
 #endif
 
+  // HTP num cores — parsed before ParsePerSocHtpConfigs so the value is available for per-SoC config construction.
+  std::string htp_num_cores_str;
```
minfhong-qti: May leverage newly introduced `ParseIntegerOption`.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -886,9 +923,19 @@ Ort::Status QnnBackendManager::CreateDevice() {
     ORT_CXX_LOG_PTR(logger_ptr_, ORT_LOGGING_LEVEL_INFO, "Device initialized already.");
     return Ort::Status();
   }
+
+  // Offline HTP preparation on x86 targets a SoC/architecture but does not have
+  // physical device cores to enumerate. The requested core count is still sent
+  // as a graph config so it is compiled into the generated context binary.
+#if defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC)
```
```suggestion
#if QNN_ARCH_ARM64
```

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2538,6 +2568,14 @@ OrtStatus* QnnEp::CompileOnnxModel(const OrtGraph** graphs,
         QNN_GRAPH_CONFIG_INIT, QNN_HTP_GRAPH_CUSTOM_CONFIG_INIT);
     InitQnnHtpGraphConfigs(htp_graph_configs, htp_graph_configs_builder);
 
+    if (htp_graph_configs.num_cores > 0) {
```
minfhong-qti: Considering this config will not be modified once parsed from option, I think the logging during option parsing is enough.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -3179,6 +3217,16 @@ OrtStatus* ORT_API_CALL QnnEp::CompileImpl(_In_ OrtEp* this_ptr,
   // ScopedPerSocQnnBackendSetup::Init), not at this point.
   OrtStatus* compile_status = nullptr;
   if (!ep->enable_multi_soc_ep_context_) {
+    if (ep->htp_graph_configs_.num_cores > 0 &&
+        !SupportsHtpNumCoresForGraphConfigs(ep->context_cache_enabled_, ep->prepare_and_load_)) {
+      return ep->ort_api.CreateStatus(
```
minfhong-qti: I think it is a bit excessive to return failure. Prefer to move the check right after option parsing, set `num_cores` to 0 if not supported with warning message logged.
In addition, I think defining the function `SupportsHtpNumCoresForGraphConfigs` is redundant. Prefer to explicitly list the condition which is more straightforward.

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -185,6 +185,7 @@ typedef struct HtpGraphConfigs {
   bool enable_htp_monolithic_lstm = false;
   bool enable_htp_fp16_clamp_overflow = false;  // Intentionally undocumented; for internal/diagnostic use only.
   bool enable_htp_matmul_lut = ORT_QNN_HTP_MATMUL_LUT_SUPPORTED;
+  uint32_t num_cores = 0;
```
minfhong-qti: Should it be `htp_num_cores` to align with option key?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -2184,6 +2231,18 @@ Ort::Status QnnBackendManager::SetupBackend(
   }
 
   if (status.IsOK()) {
+    // Multi-core HTP device selection is only supported for AOT context
+    // generation/load flows. Leave regular ONNX/JIT execution on the default
+    // path by rejecting htp_num_cores for unsupported flows.
+    const bool allow_htp_num_cores = SupportsHtpNumCoresForSetup(load_from_cached_context, need_load_system_lib);
```
minfhong-qti: Following the comment in `qnn_execution_provider.cc`, I think let `QnnEp` to guard the correct usage is enough.

onnxruntime/core/providers/qnn/builder/qnn_def.h
qti-ashwshan: Yes, you are right.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-ashwshan: Honestly that might not align with the user's expectation as he wants multiple cores to be used so would it not be better to error out saying that it is not supported and suggest alternate route i.e. prepare only or prepare and load ?

minfhong-qti: If it is intentional to return failure status, how about `LOG_AND_THROW_ERROR` during option parsing which will be caught in `QnnEpFactory::CreateEp` and return failure status gracefully? My main opinion is that this kind of option combination or usage scenario should be reported as early as possible and in a common place.

qti-ashwshan: Yeah, let me see if we can return that sooner actually.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
```diff
@@ -811,7 +813,7 @@ Ort::Status QnnBackendManager::CreateDeviceCommon(const QNN_INTERFACE_VER_TYPE&
 
     QnnDevice_HardwareDeviceInfo_t* selected_device = nullptr;
 
-    if (allow_hw_device_enumeration && device_id_ != 0) {
+    if (allow_hw_device_enumeration && (device_id_ != 0 || htp_num_cores_ > 0)) {
```
minfhong-qti: We should have at least one E2E testcase executing the logic here.

minfhong-qti (commented):
Implementation LGTM. Final testing related comment.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-ashwshan: Discussed offline.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License
 
 #include "core/providers/qnn/qnn_execution_provider.h"
+#include "core/providers/qnn/builder/qnn_htp_graph_configs.h"
```
minfhong-qti: This include should be placed at below section.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1938,74 +1953,7 @@ OrtStatus* QnnEp::GetMultiSocSupportedNodes(const OrtGraph* graph,
 void QnnEp::InitQnnHtpGraphConfigs(
```
minfhong-qti: Why not remove this function from QnnEp and directly call `PopulateHtpGraphConfigs`?

onnxruntime/core/providers/qnn/builder/qnn_htp_graph_configs.h
```diff
@@ -0,0 +1,101 @@
...
+#pragma once
+
+#include "HTP/QnnHtpGraph.h"
+
+#include "core/providers/qnn/builder/qnn_configs_helper.h"
+#include "core/providers/qnn/builder/qnn_def.h"
+
+#include <gsl/gsl>
+
+namespace onnxruntime {
```
minfhong-qti: Prefer in `namespace qnn` and remember to remove the `qnn::` usage in this file.

PR #791, opened 2026-09-01
[QNN EP] Speed up 2D initializer transpose

1duo:
### Description

Speed up ONNX-to-QNN lowering for 2-D initializer weights.

- Transpose common 1/2/4/8-byte element types in 32x32 cache tiles.
- Skip the initializer read and the transpose itself when only QNN op validation is needed. The output buffer is still allocated and zero-filled, because callers size their QNN static tensor from it.
- Keep the existing generic transpose path for uncommon element sizes.

### Motivation and Context

Large language-model weights made the scalar transpose loop a dominant part of QNN EP session creation. Op validation also read and transposed the same initializer data even though QNN only checks shape, dtype and quantization params there.

Measured results:

- LFM initializer transpose: 3.630 s -> 0.465 s (87.2% faster).
- LFM ComposeGraph: 73.2% faster.
- LFM session creation: 10.704 s -> 7.203 s (32.7% faster).
- Qwen validation optimization, measured incrementally on tetra18-linux: 9.98 s -> 9.13 s total (8.5% faster), with combined capability time reduced 20.8%.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
@@ -1607,17 +1607,17 @@ Ort::Status TwoDimensionTranspose(const QnnModelWrapper& qnn_model_wrapper,
   const size_t elem_byte_size = qnn::utils::GetElementSizeByType(onnx_type);
   RETURN_IF_NOT(elem_byte_size != 0, "Can't get element byte size from given ONNX type");
 
+  if (skip_output_data_copy) {
+    ORT_CXX_LOG(logger,
+                ORT_LOGGING_LEVEL_VERBOSE,
+                "Skipping initializer data transpose during op validation.");
+    data_shape = std::move(output_shape);
+    return Ort::Status();
+  }
+
   std::vector<uint8_t> input_buffer;
   RETURN_IF_ERROR(qnn_model_wrapper.UnpackInitializerData(initializer, input_buffer));
   transposed_data.resize(input_buffer.size(), 0);
-
-  if (skip_output_data_copy) {  // Only shape & dtype validation are needed, no need for real tensor
-    ORT_CXX_LOG(logger,
-                ORT_LOGGING_LEVEL_VERBOSE,
-                "Only shape and dtype validation are required, so we can use dummy tensor to avoid heavy memcpy.");
-    data_shape = std::move(output_shape);  // Update parameter with final transposed shape
-    return Ort::Status();
-  }
 
   // Actual tensor content is required.
   const size_t rows = data_shape[0];
--- a/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_utils.cc
@@ -1557,6 +1557,30 @@ static Ort::Status TransposeDataRank5(gsl::span<const int64_t> input_shape,
   return Ort::Status();
 }
 
+template <size_t ElementSize>
+static void TransposeDataRank2(size_t rows,
+                               size_t cols,
+                               gsl::span<const uint8_t> input_buffer,
+                               gsl::span<uint8_t> output_buffer) {
+  constexpr size_t tile_size = 32;
+  for (size_t row_start = 0; row_start < rows; row_start += tile_size) {
+    const size_t row_end = std::min(row_start + tile_size, rows);
+    for (size_t col_start = 0; col_start < cols; col_start += tile_size) {
+      const size_t col_end = std::min(col_start + tile_size, cols);
+      for (size_t col = col_start; col < col_end; ++col) {
+        size_t dst_byte_index = (col * rows + row_start) * ElementSize;
+        for (size_t row = row_start; row < row_end; ++row) {
+          const size_t src_byte_index = (row * cols + col) * ElementSize;
+          assert(src_byte_index < input_buffer.size());
+          assert(dst_byte_index < output_buffer.size());
+          std::memcpy(&output_buffer[dst_byte_index], &input_buffer[src_byte_index], ElementSize);
+          dst_byte_index += ElementSize;
+        }
+      }
+    }
+  }
+}
+
 Ort::Status TwoDimensionTranspose(const QnnModelWrapper& qnn_model_wrapper,
                                   std::vector<uint32_t>& data_shape,
                                   const OrtValueInfo* initializer,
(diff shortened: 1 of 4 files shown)


onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1557,6 +1557,30 @@ static Ort::Status TransposeDataRank5(gsl::span<const int64_t> input_shape,
   return Ort::Status();
 }
 
+template <size_t ElementSize>
+static void TransposeDataRank2(size_t rows,
+                               size_t cols,
+                               gsl::span<const uint8_t> input_buffer,
+                               gsl::span<uint8_t> output_buffer) {
+  constexpr size_t tile_size = 32;
```
qti-shubham, line 1570: minor: small comment on why tile_size = 32 (L1 / cache-line / heuristic reasoning)

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1583,33 +1607,46 @@ Ort::Status TwoDimensionTranspose(const QnnModelWrapper& qnn_model_wrapper,
...
-    for (size_t col = 0; col < cols; col++) {
-      const size_t src_elem_index = (row * cols + col);
-      const size_t dst_elem_index = (col * output_cols + row);
-      const size_t src_byte_index = src_elem_index * elem_byte_size;
-      const size_t dst_byte_index = dst_elem_index * elem_byte_size;
-      assert(src_byte_index < input_buffer.size());
-      assert(dst_byte_index < transposed_data.size());
-
-      std::memcpy(&transposed_data[dst_byte_index], &input_buffer[src_byte_index], elem_byte_size);
+  switch (elem_byte_size) {
```
qti-shubham: minor: a small test case over the shapes that break tiled loops would be convenient.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1557,6 +1557,30 @@ static Ort::Status TransposeDataRank5(gsl::span<const int64_t> input_shape,
   return Ort::Status();
 }
 
+template <size_t ElementSize>
+static void TransposeDataRank2(size_t rows,
```
qti-shubham: renaming to TransposeTiled2D may be more meaningful.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -1583,33 +1607,46 @@ Ort::Status TwoDimensionTranspose(const QnnModelWrapper& qnn_model_wrapper,
...
-  transposed_data.resize(input_buffer.size(), 0);
-
-  if (skip_output_data_copy) {  // Only shape & dtype validation are needed, no need for real tensor
+  if (skip_output_data_copy) {
     ORT_CXX_LOG(logger,
                 ORT_LOGGING_LEVEL_VERBOSE,
-                "Only shape and dtype validation are required, so we can use dummy tensor to avoid heavy memcpy.");
-    data_shape = std::move(output_shape);  // Update parameter with final transposed shape
+                "Skipping initializer data transpose during op validation.");
+    data_shape = std::move(output_shape);
```
qti-shubham: the earlier implementation uses dummy tensor and new one is creating zero-size static buffer. is it safe/fine with QNN Validation as it gets an empty buffer — size 0, pointer null ?

qti-shubham (approved):

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
1duo: Added, with the concrete invariant rather than a heuristic: a 32x32 tile of elements up to 8 bytes wide touches at most 32 cache lines on each of the source and destination sides (2 KB per side), so both stay L1-resident for the whole tile and every source line is loaded once and fully consumed instead of once per element.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
1duo: Renamed to `TransposeTiled2D`.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
1duo: Done — 4 tests in `onnxruntime/test/providers/qnn/unit/qnn_utils_test.cc`.

The tiled code was not reachable from a test because it only existed inside the overload that takes a `QnnModelWrapper`, so I split out a raw-buffer overload that owns the element-size dispatch:

```cpp
Ort::Status TwoDimensionTranspose(size_t rows, size_t cols, size_t elem_byte_size,
                                  gsl::span<const uint8_t> input_buffer,
                                  gsl::span<uint8_t> output_buffer);
```

The initializer overload now just calls it, and the sizes that used to be assumed are checked there (`elem_byte_size != 0`, both buffers exactly `rows * cols * elem_byte_size` bytes) — previously a mismatch was only an `assert`, i.e. an out-of-bounds write in release builds.

Tests: an explicit `[2, 3]` x 2-byte expectation; a sweep of `{0,0} {0,5} {5,0} {1,1} {1,70} {70,1} {31,31} {32,32} {33,31} {31,33} {65,97}` x element sizes 1/2/4/8 against a straightforward reference transpose (empty, degenerate rank-2, just under/exactly one tile, partial tile in one dimension, partial tiles in both); element sizes 3/5/6/7/16 to cover the untiled fallback; and the size-mismatch failures.

Verified on tetra18-linux (GCC 11.4, QNN 2.48.0). The file is gated on `QNN_EP_INTERNAL_SYMBOL_ACCESS`, which needs `ENABLE_COVERAGE`, so I compiled that TU with the macro forced on under the real provider-test flags (`-Wall -Wextra -Werror`) and linked it against the EP objects to actually run it: 139/139 pass in that file (135 pre-existing + 4 new).

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
1duo: Good catch — it is not safe, and I have reverted to a correctly sized dummy buffer.

Callers size their QNN tensor from `transposed_data`, so an empty vector is not just a null/zero-size `clientBuf`:

- `lpbqgemm_fusion.cc` calls `UnpackWeightTensorData(..., validate)` (weight `axis == 0` path) and then `LowPowerBlockQuantizeData`, which fails `RETURN_IF_NOT(num_elems == data.size(), "Shape mismatch with data to quantize")` on 0 elements. A validation failure only logs at WARNING, so the LPBQ Gemm fusion would be silently rejected during `GetCapability` and the subgraph would fall back to CPU.
- Everywhere else the buffer backs a `QNN_TENSOR_TYPE_STATIC` tensor (reshape-gemm fusion, MatMulNBits, Gemm-BQ, and the new MatMul->FullyConnected path), so QNN would see `clientBuf.dataSize == 0` for a static weight.

The skip path now keeps the byte count and skips only the work that actually costs time:

```cpp
transposed_data.assign(rows * cols * elem_byte_size, 0);
```

That gives up the allocation/zero-fill saving, but `UnpackInitializerData` and the transpose — the expensive parts — are still skipped, so the measured validation win holds. I also updated the PR description, which claimed the output allocation was skipped.

1duo:
Pushed `bdf3105` addressing all four review comments (details in the threads):

- 32x32 tile choice is now explained in a comment.
- Internal helper renamed to `TransposeTiled2D`.
- New raw-buffer `TwoDimensionTranspose` overload makes the tiled path unit-testable and turns the previously assumed buffer sizes into a checked precondition; 4 tests added in `qnn_utils_test.cc`.
- The validation-only path keeps a correctly sized zero-filled dummy buffer again — an empty one gives QNN a zero-byte `clientBuf` for static weights and makes LPBQ Gemm fusion validation fail silently. Only the initializer read and the transpose are skipped now, which is where the time was, and the PR description has been corrected accordingly.

Also reverted the incidental churn in the templated `TwoDimensionTranspose` overload in `qnn_utils.h`, so that overload is unchanged from main.

Verified on tetra18-linux (GCC 11.4, QNN SDK 2.48.0): EP builds and links warning-free; `qnn_utils_test.cc` compiled with `QNN_EP_INTERNAL_SYMBOL_ACCESS=1` under the provider-test flags (`-Wall -Wextra -Werror`) and linked against the EP objects to run — 139/139 pass. All touched files are `clang-format` clean.

PR #792, opened 2026-09-01
[QNN EP] Uplevel ORT Core to 1.29.0

qti-mbadnara:
As per title

Disable two flaky Unit Tests will be taken up in a follow up PRs

- `Encryption_WithShareEpContexts_RoundTrip`
- `QnnContextShareAcrossSessions`


huaychou (commented):
Do we need to update the version of `ort_prebuilt_windows_x64` in `qcom/packages.yml`

qti-mbadnara:
> Do we need to update the version of `ort_prebuilt_windows_x64` in `qcom/packages.yml`

Yep, will fix it

tirupath-qti (approved):

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
```diff
@@ -774,9 +786,14 @@ TEST_F(QnnHTPBackendTests, Encryption_ReadCallback_ReturnsError_SessionCtorSurfa
 // and the decrypting read callback loads it back and runs. It does NOT exercise the
 // multi-model merged-binary flow (that path compiles via Ort::Session create, which cannot
 // carry a write callback in ORT 1.28); it isolates the interaction of the two feature flags.
-TEST_F(QnnHTPBackendTests, Encryption_WithShareEpContexts_RoundTrip) {
+TEST_F(QnnHTPBackendTests, DISABLED_Encryption_WithShareEpContexts_RoundTrip) {
```
qti-kromero: why is this DISABLED but also needs edits below?

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
```diff
@@ -452,6 +452,12 @@ TEST_F(QnnHTPBackendTests, Encryption_NewReadWriteCallback_RoundTrip) {
 // Baseline: htp_share_resource_optimization=1 WITHOUT encryption; hangs here → pre-existing QAIRT issue, unrelated to this PR.
 TEST_F(QnnHTPBackendTests, Encryption_VtcmSharing_Baseline_NoCallback) {
   SKIP_HTP_TEST_ON_ARCH_LESS_THAN_OR_EQUAL_TO(QNN_HTP_DEVICE_ARCH_V68);
+#if defined(__linux__) && !defined(__aarch64__)
+  // The x86 HTP CPU emulator does not support shared-resource context-binary reload
+  // (QNN_COMMON_ERROR_NOT_SUPPORTED from QnnBackendManager::SetupBackend on reload).
+  // This is a pre-existing QAIRT limitation unrelated to the ORT Core uplevel.
+  GTEST_SKIP() << "htp_share_resource_optimization context reload not supported on x86 HTP emulator.";
```
qti-kromero: a little confused why this and below would start being an issue on an ORT Core upgrade

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
```diff
@@ -197,17 +197,19 @@ index 4b1ef26fc..0108b08ad 100644
...
 --- a/onnxruntime/test/onnx/TestCase.cc
 +++ b/onnxruntime/test/onnx/TestCase.cc
-@@ -1522,6 +1522,26 @@ std::unique_ptr<std::set<BrokenTest>> GetBrokenTests(const std::string& provider
-     broken_tests->insert({"dft_rfft", "unknown version"});
-     broken_tests->insert({"dft_rfft_opset19", "unknown version"});
+@@ -1588,6 +1588,28 @@ std::unique_ptr<std::set<BrokenTest>> GetBrokenTests(const std::string& provider
+     // the CPU-materialized node runner (onnx#7959) exposes the same marginal FFT precision failure
+     // that onnx_backend_test_series_filters.jsonc already excludes globally.
      broken_tests->insert({"bitcast_bool_to_uint8", "ORT BitCast kernel does not register bool type"});
 +    // TODO: [AISW-167042] [ORT ABI] Model test failures on ABI path
```
qti-kromero, line 207: why are these getting added here? to me it feels like over time we should be reenabling these tests rather than disabling them

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
huaychou: Since it first failed on Linux and then also failed on WoS

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
huaychou: It is not tested until ort 1.29.0

cmake/patches/ort_core/0001-cpp-model-test-runner-uses-plugin-EP.patch
huaychou: I think it is not added by this PR but a github bug

qti-yuduo (approved):

PR #793, opened 2026-09-02
[QNN EP] Conditional Preserve dynamic MatMul UINT16

tzuhsuanwei:
## Motivation

Using symmetric U16 for MatMul can improve accuracy on some models. However, QA regression testing showed that applying this conversion to every dynamic MatMul can also cause regressions.

This change keeps the existing conversion as the default and uses a simple quantization-error metric to identify high-risk dynamic asymmetric-U16 MatMuls. Only those MatMuls are converted to symmetric U16; the remaining cases keep the existing path.

## Conversion policy

The existing dynamic-U16 eligibility checks are evaluated first. For an eligible asymmetric-U16 input1, estimate the maximum value-domain error from converting U16 to U8:

```text
max_rounding_error = input1_scale * (65535 / 255) / 2
```

If this value is greater than `0.1`, use the symmetric-U16 conversion. Otherwise, keep the existing asymmetric-U8 conversion.

The `0.1` threshold is the selected 10% acceptable value-domain quantization-error budget. Static weights, already-symmetric U16 inputs, and non-eligible MatMuls keep their existing behavior.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
@@ -18,6 +18,16 @@ namespace onnxruntime {
 namespace qnn {
 
 namespace {
+bool ShouldUseSymmetricU16ForMatMul(float input1_scale) {
+  constexpr float kMaximumValueQuantError = 0.1f;
+
+  // Re-encoding U16 as U8 increases the value-domain quantization step from
+  // input1_scale to input1_scale * 65535 / 255. A value can move
+  // by at most half of that U8 step when it is rounded. Keep symmetric U16
+  // when that error exceeds the 10% value-domain budget.
+  const float u8_half_step = input1_scale * 65535.0f / 255.0f / 2.0f;
+  return u8_half_step > kMaximumValueQuantError;
+}
 // Detects a block-quantized MatMul weight (ONNX MatMul input[1]).
 // Accepts weight rank 2–4: shape [..., K, N] where any leading dims beyond K/N must equal 1
 // (i.e. reshapeable to [1, 1, K, N]). Per ONNX opset 21 the scale has the same rank as the
@@ -358,11 +368,12 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnMatMul(QnnModelWrapper& qnn_mode
   //                         |
   //     input_1_uint16 -----+
   //
-  // For dynamic weights, QNN graph that passes validation:
-  //     input_0_uint16 ---------------------------> MatMul ---> output_uint16
-  //                                                   ^
-  //                                                   |
-  //     input_1_uint16_asym --> Convert(uint16_sym) --+
+  // Dynamic asymmetric U16 input[1] uses the value-domain U16-to-U8 error gate:
+  //     input_0_uint16 ----------------------------------> MatMul ---> output_uint16
+  //                                                        ^
+  //                                                        |
+  //     input_1_uint16_low_error --> Convert(uint8_asym) -+
+  //     input_1_uint16_high_error --> Convert(uint16_sym) -+
   //
   // For static weights, QNN graph that passes validation:
   //     input_0_uint16 ---------------------> MatMul ---> output_uint16
@@ -388,15 +399,18 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnMatMul(QnnModelWrapper& qnn_mode
       // QNN offsets negate ONNX zero points, so symmetric uint16 uses -32768.
       constexpr int32_t kSymmetricU16Offset = -32768;
       if (quant_param.scaleOffsetEncoding.offset != kSymmetricU16Offset) {
+        const bool use_symmetric_u16 =
+            ShouldUseSymmetricU16ForMatMul(quant_param.scaleOffsetEncoding.scale);
         RETURN_IF_ERROR(utils::InsertConvertOp(qnn_model_wrapper,
                                                convert_input_name,
                                                convert_output_name,
                                                input_info_1.qnn_data_type,
-                                               QNN_DATATYPE_UFIXED_POINT_16,
+                                               use_symmetric_u16 ? QNN_DATATYPE_UFIXED_POINT_16
+                                                                 : QNN_DATATYPE_UFIXED_POINT_8,
                                                quant_param.scaleOffsetEncoding.offset,
                                                quant_param.scaleOffsetEncoding.scale,
                                                input_1_shape,
-                                               true,  // symmetric
+                                               use_symmetric_u16,
                                                do_op_validation));
         input_names.push_back(convert_output_name);
       } else {
--- a/onnxruntime/test/providers/qnn/matmul_test.cc
+++ b/onnxruntime/test/providers/qnn/matmul_test.cc
@@ -714,47 +714,82 @@ TEST_F(QnnHTPBackendTests, MatMulOp_QDQ_Regression_uint16_dynamic_inputs) {
   }
 }
 
-TEST_F(QnnHTPBackendTests, MatMulOp_QDQ_U16DynamicInput1_ConvertOnlyIfAsymmetric) {
-  namespace fs = std::filesystem;
-  struct TestCase {
-    const char* name;
-    float input1_min;
-    size_t expected_convert_count;
-  };
-
-  for (const TestCase& test_case : {TestCase{"symmetric", -0.1f, 0},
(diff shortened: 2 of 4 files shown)


onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -378,22 +410,24 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnMatMul(QnnModelWrapper& qnn_mode
       input_1_shape = {input_info_1.shape[0], 1};
     }
     if (!input_info_1.is_initializer) {
-      // QNN offsets negate ONNX zero points, so symmetric uint16 uses -32768.
```
minfhong-qti, line 388: Why remove this comment?

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -378,22 +410,24 @@ Ort::Status MatMulOpBuilder::ProcessInputsForQnnMatMul(QnnModelWrapper& qnn_mode
...
       constexpr int32_t kSymmetricU16Offset = -32768;
-      if (quant_param.scaleOffsetEncoding.offset != kSymmetricU16Offset) {
+      const bool use_symmetric_u16 =
+          ShouldUseSymmetricU16ForMatMul(qnn_model_wrapper, node_unit, input_info_0, input_info_1);
+      if (use_symmetric_u16 && quant_param.scaleOffsetEncoding.offset == kSymmetricU16Offset) {
```
minfhong-qti: I don't think the condition here is correct.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -18,6 +18,37 @@ namespace onnxruntime {
 namespace qnn {
 
 namespace {
+bool ShouldUseSymmetricU16ForMatMul(const QnnModelWrapper& qnn_model_wrapper,
```
minfhong-qti: Keep quant error calculation only. No need other checks (e.g., data type).

minfhong-qti (commented):
There should be commit message.
Update description and commit message to describe why not always convert to sym u16.

minfhong-qti (approved):

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -18,6 +18,15 @@ namespace onnxruntime {
 namespace qnn {
 
 namespace {
+bool ShouldUseSymmetricU16ForMatMul(const TensorInfo& input1_info) {
+  constexpr float kMaximumValueQuantError = 0.1f;
+  const auto& input1_quant = input1_info.quant_param.Get().scaleOffsetEncoding;
+
+  // U8 combines about 257 U16 codes into one step. Keep symmetric U16 when the
```
qti-chuteng: What's 257 U16 codes?

onnxruntime/test/providers/qnn/qnn_node_group/qnn_graph_checker.cc
```diff
@@ -45,6 +45,47 @@ void AssertOpInQnnGraph(const std::filesystem::path& dump_dir,
       << " occurrence(s), found " << actual_count << " in " << json_path;
 }
 
+void AssertConvertOutputDataType(const std::filesystem::path& dump_dir,
+                                 uint32_t expected_data_type) {
+  std::filesystem::path json_path;
+  for (const auto& entry : std::filesystem::directory_iterator{dump_dir}) {
```
qti-chuteng: Can we refactor this logic into a helper function?

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
tzuhsuanwei: Updated the quantization-error comment to explain the value-domain calculation directly.

onnxruntime/test/providers/qnn/qnn_node_group/qnn_graph_checker.cc
tzuhsuanwei: Added LoadQnnGraph() and reused it in the graph checks to remove the duplicated loading logic.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -18,6 +18,17 @@ namespace onnxruntime {
 namespace qnn {
 
 namespace {
+bool ShouldUseSymmetricU16ForMatMul(const TensorInfo& input1_info) {
```
qti-yuduo: 0.1 is too large as a “safe for U8” bound: Gate is range/510 > 0.1 ⇔ range > 51. Only huge-range input1 gets precise sym-U16; everything else falls back to 128-257x coarser U8. 0.1 is input-referred, before fan-in. err_out ≈ sum_k |a_k|*e_k, |e_k|≤0.1. With K=512-4096 this amplifies 10-1000x at output.
Default QDQTolerance is 0.004*output_range (qnn_test_utils.h:919-923). Unless output_range>25, 0.1 alone already exceeds budget — before amplification. So it permits large absolute input error where output tolerance is tightest. No K/input0/output-range normalization to justify it.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
```diff
@@ -18,6 +18,17 @@ namespace onnxruntime {
...
+  constexpr float kMaximumValueQuantError = 0.1f;
+  const auto& input1_quant = input1_info.quant_param.Get().scaleOffsetEncoding;
```
qti-yuduo: direct .scaleOffsetEncoding access hides IsPerTensor() precondition. Take float scale or assert.

onnxruntime/test/providers/qnn/matmul_test.cc
```diff
@@ -714,47 +714,88 @@ TEST_F(QnnHTPBackendTests, MatMulOp_QDQ_Regression_uint16_dynamic_inputs) {
...
-    }
-    AssertOpInQnnGraph(graph_dir, "Convert", test_case.expected_convert_count);
+  if (::testing::Test::IsSkipped()) {
+    return;
   }
+  AssertOpInQnnGraph(graph_dir, "Convert");
+  AssertConvertOutputDataType(graph_dir, expected_convert_type);
+}
+
+TEST_F(QnnHTPBackendTests, MatMulOp_QDQ_U16DynamicInput1_LowQuantErrorUsesAsymmetricU8) {
```
qti-yuduo, line 760: symmetric-input case deleted. Symmetric path (0 Converts) now untested.

tzuhsuanwei: Added the symmetric dynamic-U16 test back. It verifies the symmetric path creates zero Convert nodes.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
tzuhsuanwei: Updated the helper to take the scale as a float. IsPerTensor() is checked by the caller before accessing the scale.

onnxruntime/core/providers/qnn/builder/opbuilder/matmul_op_builder.cc
tzuhsuanwei: Thanks for the feedback.

Originally, this path always converted the input to asym U8. After #300, a sym U16 path was introduced. While sym U16 is theoretically more accurate, experiments showed that always preserving sym U16 introduced regressions on a number of models, suggesting that not all models are sensitive to the U16→U8 conversion.

For this reason, we use a stricter gate for sym U16 preservation. The current 0.1 threshold is primarily empirical and was chosen based on results across our model set. It improves models that are genuinely impacted by the conversion while avoiding regressions on models that are otherwise unaffected. Although not derived from a strict analytical bound, it has provided a reasonable trade-off so far, and can be revisited in a follow-up PR if future data suggests a better threshold.

We also evaluated metrics that incorporate output scale. However, very small output scales can dominate the decision and make it overly dependent on output quantization parameters. After weighing the trade-offs, we found that input-side scale/error is the more relevant signal for this conversion decision, so the current heuristic intentionally focuses on the input side.

qti-yuduo (approved):

qti-chuteng (approved):

PR #795, opened 2026-09-02
Support constant-only QNN graphs on the GPU

qti-mattsinc:
* Allow QNN models with zero runtime inputs when using the GPU backend.
* This fixes QNN EP rejecting simple graphs when the ORT Core `ConstantFolding` graph transformer does not activate. The GPU backend is able to handle these graphs, but this does not seem to be supported on HTP or QNN CPU currently.
* Fixes multiple WebNN conformance failures on the GPU.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/qnn_model.cc
+++ b/onnxruntime/core/providers/qnn/builder/qnn_model.cc
@@ -879,7 +879,14 @@ Ort::Status QnnModel::SetupTensors(std::vector<QnnTensorInfo>& qnn_tensor_infos,
                                    const std::vector<QnnTensorWrapper>& tensor_wrappers,
                                    bool is_input) {
   size_t tensor_count = tensor_wrappers.size();
-  RETURN_IF(0 == tensor_count, "Zero tensor size!");
+  if (tensor_count == 0) {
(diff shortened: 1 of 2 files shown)


onnxruntime/core/providers/qnn/builder/qnn_model.cc
```diff
@@ -879,7 +879,14 @@ Ort::Status QnnModel::SetupTensors(std::vector<QnnTensorInfo>& qnn_tensor_infos,
                                    const std::vector<QnnTensorWrapper>& tensor_wrappers,
                                    bool is_input) {
   size_t tensor_count = tensor_wrappers.size();
-  RETURN_IF(0 == tensor_count, "Zero tensor size!");
+  if (tensor_count == 0) {
```
qti-yuduo, line 882: Are we gating on GPU only then? Given you mentioned other backend may not have support.

qti-mattsinc: I tested with CPU and HTP and both returned errors. For HTP, I had a look at the QAIRT internals and saw that is indeed explicitly unsupported. So, I'm gating it to GPU only here

qti-yuduo (approved):

PR #796, opened 2026-09-02
[QNN EP] Uplevel QAIRT to 2.50.0

qti-mbadnara:
As per title

### DIFF
--- a/java/src/test/android/app/build.gradle
+++ b/java/src/test/android/app/build.gradle
@@ -95,7 +95,7 @@ dependencies {
 	// Fallback versions (1.24.3, 2.45.0) are the previous known-good releases used when the
 	// requested version is not yet available in the Maven repository (e.g., same-day CI runs).
 	// See docs/execution_providers/QNN-ExecutionProvider.md for the validated version matrix.
-	def resolvedOrtVersion = getAvailableVersion('com.microsoft.onnxruntime:onnxruntime-android', ortVersion, '1.24.4', '1.24.3')
+	def resolvedOrtVersion = getAvailableVersion('com.microsoft.onnxruntime:onnxruntime-android', ortVersion, '1.27.0', '1.26.0')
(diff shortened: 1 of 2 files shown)


tirupath-qti (approved):

java/src/test/android/app/build.gradle
```diff
@@ -95,7 +95,7 @@ dependencies {
 	// Fallback versions (1.24.3, 2.45.0) are the previous known-good releases used when the
 	// requested version is not yet available in the Maven repository (e.g., same-day CI runs).
 	// See docs/execution_providers/QNN-ExecutionProvider.md for the validated version matrix.
-	def resolvedOrtVersion = getAvailableVersion('com.microsoft.onnxruntime:onnxruntime-android', ortVersion, '1.24.4', '1.24.3')
+	def resolvedOrtVersion = getAvailableVersion('com.microsoft.onnxruntime:onnxruntime-android', ortVersion, '1.27.0', '1.26.0')
```
kuanyul-qti: I don't think this modification is required.

qti-mbadnara: I was under the impression that it should reflect the android-version used to build the artifacts?

kuanyul-qti: This part does not reflect the ORT version we build with. We only need to update it when the target ORT version isn’t available on Maven Central, in which case we add a check and a fallback for it.

qti-mbadnara: Fixed

huaychou:
Just wondering, are we going to release QNN-EP-2.6.0 with QAIRT-2.50.0 or QAIRT-2.50.40?

qti-mbadnara:
> Just wondering, are we going to release QNN-EP-2.6.0 with QAIRT-2.50.0 or QAIRT-2.50.40?

Mainline will be QAIRT 2.50.0, rel-2.6.0 will be QAIRT 2.50.40

PR #799, opened 2026-09-03
[QNN EP] Wire prepare_app/run_app encryption sample apps into build

tzuhsuanwei:
## Summary
- Adds `prepare_app`/`run_app` build targets (gated on `onnxruntime_USE_QNN`) so the EPContext binary encryption sample apps build automatically as part of the normal build, landing alongside `onnxruntime_provider_test.exe`.
- Adds `encryption_test.py`, which drives the full `prepare_app` -> `run_app` round trip and reports PASS/FAIL, replacing the golden-comparison logic that used to be embedded in `run_app`.

Goal: after a normal build, `prepare_app.exe`/`run_app.exe` are produced as testing artifacts, and the encryption round trip can be verified end-to-end via `encryption_test.py`.

Test:
Artifact already verify on v81 and v73 with script encryption_test.py

### DIFF
--- a/cmake/onnxruntime_unittests.cmake
+++ b/cmake/onnxruntime_unittests.cmake
@@ -469,6 +469,32 @@ endif()
     endif()
 
     set_target_properties(ep_weight_sharing_ctx_gen PROPERTIES FOLDER "ONNXRuntimeTest")
+
+    # Compiled-model-encryption prepare/run sample apps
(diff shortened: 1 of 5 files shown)


qti-chuteng:
I’d prefer to keep the test files from onnxruntime/test/ in qcom/scripts or qcom/. The application files in onnxruntime/test/ can be used for standalone execution, while prepare_app and run_app are intended only for our internal QA use.

qti-chuteng (approved):

minfhong-qti (approved):

PR #800, opened 2026-09-04
[QNN EP] Fix operator and node-group fusion conformance issues

qti-mbadnara:
## Description

Fixes a set of correctness issues across QNN EP node-group fusions and op builders: missing QDQ-start exception list entry for Gemm, three fusions that returned a group object without first validating it on the QNN backend, and multiple op builders that used a globally-stateful `UniqueNameGenerator` for intermediate tensor names (producing counter-drift between the validate and create passes) or hardcoded `do_op_validation=false` in decomposition helper calls.

Building on top of PR #540 from @tirupath-qti. This PR is expected to solve the functional failure caused by PR #200 

---

### Node-group fusions

The QNN EP's `TryFusion` path is called during Phase 1 (`GetCapability`) with a temporary `QnnModelWrapper`. The contract requires that the fusion group validates its QNN node chain on the backend **before** returning a non-null group object. If a fusion returns a group without first calling `CreateOrValidateOnQnn(do_op_validation=true)`, any backend rejection (unsupported op, bad dtype, shape constraint) is silently swallowed during `GetCapability` and only surfaces later as a crash or garbage output during `Compile`.

Three fusions violated this contract:

- **`ReshapeGemmFusionGroup` (`TryFusion2/3/4`)** — all three overloads constructed and returned the group object without calling `CreateOrValidateOnQnn` in validate mode.

- **`ReshapeEinsumReshapeNodeGroup` (`TryFusion`)** — same issue. Additionally, the `CreateOrValidateOnQnn` function had no dedicated validate branch; both validate and create paths fell through to the same `AddTensorWrapper` + `CreateQnnNode` code, meaning a `TryFusion`-triggered validate call would have mutated the model wrapper state by adding tensors — a side-effect that must not happen during Phase 1 validation.

- **`qnn_node_group.cc` QDQ-start filter** — the early-return guard that allows QDQ-wrapped starting nodes through to `TryFusion` listed `MatMul` but not `Gemm`. Because `ReshapeGemmFusionGroup` operates on QDQ-wrapped Gemm nodes, it was never reached for QDQ models.

### Op builders — `UniqueNameGenerator` cross-pass drift

`utils::UniqueNameGenerator` maintains a global, process-wide counter. Every call to `.New()` increments the counter and appends the new value to the base string to produce a unique name. This is intentional for QNN node names (which need only be unique, not stable), but is incorrect for **intermediate tensor names** in multi-node decompositions.

When an op builder decomposes one ONNX op into a chain of QNN nodes (e.g. Sign → Abs → Cast, or a pre-Transpose → FC → Reshape), intermediate tensors are:
1. Registered in the `QnnModelWrapper` as output of node A.
2. Referenced by node B as its input.

Both A and B must use **the same string** for that tensor within a single pass. With `UniqueNameGenerator`, A and B each call `.New()` at different counter values, so the names already differ within the same pass if any other `.New()` calls intervene between them. More critically, the entire validate pass (Phase 1, `IsOpSupported`) and create pass (Phase 2, `AddToModelBuilder`) can have different counter states, so the same decomposition produces different intermediate names in each pass.

The fix is to replace all `UniqueNameGenerator().New(base, suffix)` calls that produce intermediate **tensor** names with deterministic string concatenation: `base + suffix`, where `base` is already a unique ONNX tensor or node name.

`UniqueNameGenerator` is still used (and is correct) for QNN node names, where uniqueness is sufficient and stability is not required.

### Op builders — hardcoded `do_op_validation=false`

`AddCastNode`, `AddTransposeNode`, and similar helpers accept a `do_op_validation` flag that controls whether they call `ValidateQnnNode` or `CreateQnnNode`. Hardcoding `false` silently skips backend validation during `IsOpSupported`, allowing unsupported configurations through `GetCapability` that will fail or produce incorrect results at runtime.

### Op builders — missing null guards after `GetConstantTensor`

`QnnModelWrapper::GetConstantTensor` returns `nullptr` when the named tensor is not a registered constant initializer. Several `MatMulNBits` code paths passed the result directly to `UnpackInitializerData` without a null check, which would crash or produce undefined behaviour if the weight, scale, or zero_points inputs were not constant (e.g. in a malformed or unexpected graph).

---

## Files Changed

| File | Change |
|------|--------|
| `onnxruntime/core/providers/qnn/builder/qnn_node_group/qnn_node_group.cc` | Added `"Gemm"` to the QDQ-start exception list so QDQ-wrapped Gemm nodes are forwarded to `TryFusion` instead of being rejected by the early-return guard. Updated comment. |
| `onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_gemm_fusion.cc` | `TryFusion2`, `TryFusion3`, `TryFusion4`: after constructing the candidate `ReshapeGemmFusionGroup`, call `CreateOrValidateOnQnn(do_op_validation=true)` and return `nullptr` on failure before returning the group. Removed unused `ORT_UNUSED_PARAMETER(logger)` in each overload. |
| `onnxruntime/core/providers/qnn/builder/qnn_node_group/reshape_einsum_reshape.cc` | Added a pure-validate branch at the top of `CreateOrValidateOnQnn`: when `do_op_validation=true`, builds all tensors and params on the stack and calls `ValidateQnnNode` for the three QNN nodes (Reshape, DepthToSpace, Transpose) without touching model wrapper state. Updated `TryFusion` to call this validate path and return `nullptr` on failure before returning the group. |
| `onnxruntime/core/providers/qnn/builder/opbuilder/matmulnbits_op_builder.cc` | Added `RETURN_IF_NOT(ptr != nullptr, ...)` guards after each of the three `GetConstantTensor` calls (weight at line 376, scales at line 384, zero_points at line 502) before passing the pointer to `UnpackInitializerData`. |
| `onnxruntime/core/providers/qnn/builder/opbuilder/simple_op_builder.cc` | `ProcessVariadicToBinaryChain`: replaced `UniqueNameGenerator` with deterministic names for DQ intermediate tensors (`input_names[i] + "_to_f32"`) and fold-chain output tensors (`node_unit.Name() + "_fold_" + i`). Fixed `AddCastNode` call for int64 graph-output cast: changed hardcoded `false` to `do_op_validation`. |
| `onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc` | `ProcessInputs`: replaced `UniqueNameGenerator` with `input_tensor_name + "_transpose"` for the non-constant transposed-weight intermediate. `ProcessInputsForBQGemm`: replaced `UniqueNameGenerator` with `fp16_name + "_transpose"` (transA path) and `inputs[2].name + "_fp16"` (bias dequantize). `ProcessAttributesAndOutputs`: replaced `UniqueNameGenerator` with `final_output_name + "_fc"` (absorbed-Reshape path) and `org_output_name + "_fc"` (FC+Add decomposition path). |
| `onnxruntime/core/providers/qnn/builder/opbuilder/resize_op_builder.cc` | `ProcessTfHalfPixelForNN`: replaced `UniqueNameGenerator` with `node_unit.Name() + "_tfhp_resize_out"` for the intermediate tensor connecting the 2× Resize node to the StridedSlice node. (The resize node name and all param names were and remain UniqueNameGenerator since they are not referenced cross-node.) |
| `onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc` | `ProcessExtraInputForNotEqual`: replaced `const std::string&` binding to temporary with `const std::string`, and replaced `UniqueNameGenerator` with `node_unit.Name() + "_notequal_zero"`. `EmitSignAbsCastDecomposition`: replaced `UniqueNameGenerator` with `node_unit.Name() + "_signabs_sign"` and `node_unit.Name() + "_signabs_abs"` for the Sign and Abs intermediate tensors. |
| `onnxruntime/core/providers/qnn/builder/opbuilder/topk_op_builder.cc` | `ProcessInputs`: replaced `UniqueNameGenerator` with `input_names[0] + "_transpose"` for the pre-TopK axis-permutation output. `ProcessAttributesAndOutputs`: replaced `UniqueNameGenerator` with `output_name + "_transpose"` for post-TopK output intermediates and `output_name + "_cast"` for the INT64 cast-back input tensor. |
| `onnxruntime/core/providers/qnn/builder/opbuilder/scatternd_op_builder.cc` | `ProcessScatterNDIndices`: replaced `UniqueNameGenerator` with `indices_tensor_name + "_qnn_idx"` for the renamed rewritten-indices initializer tensor. |

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/cast_op_builder.cc
@@ -91,7 +91,7 @@ Ort::Status CastOpBuilder::ProcessExtraInputForNotEqual(QnnModelWrapper& qnn_mod
   }
 
   // Build additional static input with value 0.
-  const std::string& input_name = utils::UniqueNameGenerator().New(node_unit, "_notequal_zero");
+  const std::string input_name = node_unit.Name() + "_notequal_zero";
 
   Qnn_DataType_t qnn_data_type = QNN_DATATYPE_UNDEFINED;
   ONNXTensorElementDataType input_type = input.type;
@@ -193,8 +193,8 @@ Ort::Status CastOpBuilder::EmitSignAbsCastDecomposition(QnnModelWrapper& qnn_mod
                 "Cannot get shape for FP-to-Bool Cast input.");
 
   // 2. Reserve deterministic intermediate names so DLC inspection and context cache lookups stay stable.
-  const std::string sign_out_name = utils::UniqueNameGenerator().New(node_unit, "_signabs_sign");
-  const std::string abs_out_name = utils::UniqueNameGenerator().New(node_unit, "_signabs_abs");
+  const std::string sign_out_name = node_unit.Name() + "_signabs_sign";
+  const std::string abs_out_name = node_unit.Name() + "_signabs_abs";
 
   // 3. Register the two NATIVE intermediates with the same shape/dtype as the Cast input.
   QnnTensorWrapper sign_out_wrapper(sign_out_name,
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/gemm_op_builder.cc
@@ -204,7 +204,7 @@ Ort::Status GemmOpBuilder::ProcessInputs(QnnModelWrapper& qnn_model_wrapper,
       input_shape[0] = old_input_shape[1];
       input_shape[1] = old_input_shape[0];
       const std::string& node_input_name(input_name);
-      input_tensor_name = utils::UniqueNameGenerator().New(input_tensor_name, "_transpose");
+      input_tensor_name = input_tensor_name + "_transpose";
       std::vector<uint32_t> perm{1, 0};
       RETURN_IF_ERROR(qnn_model_wrapper.AddTransposeNode(node_unit.Index(), node_input_name, input_tensor_name,
                                                          old_input_shape, perm, input_shape,
@@ -291,7 +291,7 @@ Ort::Status GemmOpBuilder::ProcessInputsForBQGemm(QnnModelWrapper& qnn_model_wra
     if (trans_a != 0) {
       RETURN_IF_NOT(act_shape_2d.size() == 2, "QNN EP: BQ Gemm transA=1 requires a rank-2 activation");
       const std::vector<uint32_t> transposed_shape = {act_shape_2d[1], act_shape_2d[0]};
-      const std::string transposed_name = utils::UniqueNameGenerator().New(fp16_name, "_transpose");
+      const std::string transposed_name = fp16_name + "_transpose";
       RETURN_IF_ERROR(qnn_model_wrapper.AddTransposeNode(node_unit.Index(), fp16_name, transposed_name,
                                                          act_shape_2d, /*transpose_perm=*/{1u, 0u},
                                                          transposed_shape, QNN_DATATYPE_FLOAT_16,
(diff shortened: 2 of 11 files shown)


qti-mbadnara:
Closing this PR and will get #540 merged

PR #801, opened 2026-09-05
[QNN EP] Add htp_graph_splitting_num_prepare_threads provider option

qti-mbadnara:
### Overview

QAIRT SDK 2.51 introduced a new context config option `QNN_HTP_CONTEXT_CONFIG_OPTION_GRAPH_SPLITTING_NUM_PREPARE_THREADS` (enum value 26) that controls how many threads are used to prepare split subgraphs in parallel when HTP Graph Splitting is enabled.

Prior to 2.51, there was no way for the EP to tune this knob — the SDK defaulted to single-threaded prepare regardless of how many splits were created. On multi-core SoCs with a large number of splits (`GPE_KWAY_PARTITIONS`), single-threaded preparation serialises what could otherwise be parallel work, increasing context creation latency.

In 2.51 the option was added as a **separate** `QnnHtpContext_CustomConfig_t` entry
(union field `numPrepareThreads`, option `QNN_HTP_CONTEXT_CONFIG_OPTION_GRAPH_SPLITTING_NUM_PREPARE_THREADS`) rather than a field inside `QnnHtpContext_GraphSplit_t`, making it independent of the existing graph splitting enabled config. The SDK-defined sentinel
`QNN_HTP_CONTEXT_NUM_PREPARE_THREADS_AUTO` (`UINT32_MAX`) instructs the runtime to auto-select: `min(max(1, hardware_concurrency), num_splits)`.

This PR exposes the knob as the new provider option `htp_graph_splitting_num_prepare_threads` and plumbs it end-to-end from session option parsing through to `QnnContext_create`.

### New provider option: `htp_graph_splitting_num_prepare_threads`

| Value | Behaviour |
|-------|-----------|
| *(not set)* | `UINT32_MAX` — auto-select: `min(max(1, hardware_concurrency), num_splits)` |
| `0` or `1` | Single-threaded prepare |
| `N > 1` | Up to N threads; capped internally at `num_splits` |

**Constraints:**
- Only meaningful when `enable_htp_graph_splitting=1`. Setting it without graph splitting enabled is accepted but has no effect on context creation.
- Requires QAIRT SDK 2.51+ at build time. On builds against SDK < 2.51 (`QNN_HTP_GRAPH_SPLITTING_NUM_THREADS_AVAILABLE` not defined), the option is parsed but immediately reset to `UINT32_MAX` and a `WARNING` log is emitted.


onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -37,6 +37,12 @@ namespace qnn {
 #define QNN_HTP_GRAPH_SPLITTING_AVAILABLE
 #endif
 
+// QNN_HTP_CONTEXT_CONFIG_OPTION_GRAPH_SPLITTING_NUM_PREPARE_THREADS is available from QAIRT SDK 2.51+.
+#if defined(QNN_SDK_VERSION_MAJOR) && QNN_SDK_VERSION_MAJOR == 2 && \
```
quic-calvnguy: This does not cover SDK versions with major versions >=3, only 2.XX >= 2.51

quic-calvnguy (changes requested):
Just a change needed for the version check for now. Will double-check if GPE config is required for AOT cases where the context bin is loaded.

onnxruntime/core/providers/qnn/builder/qnn_def.h
qti-mbadnara: Done, thanks for catching

quic-calvnguy (approved):
LGTM. Would prefer a unit test that also ensure execution for end-to-end coverage too

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1367,8 +1367,35 @@ QnnEp::QnnEp(QnnEpFactory& factory,
...
+  // Only meaningful when enable_htp_graph_splitting=1.
+  {
+    std::string num_threads_str;
+    GetSessionConfigEntryOrDefault(ort_api,
+                                   session_options_,
+                                   FormatEPConfigKey("htp_graph_splitting_num_prepare_threads"),
+                                   "",
+                                   num_threads_str);
+    if (!num_threads_str.empty()) {
+      htp_graph_splitting_num_prepare_threads_ = static_cast<uint32_t>(std::stoul(num_threads_str));
```
qti-shubham: is this safe ?
If a user sets `htp_graph_splitting_num_prepare_threads=UINT32_MAX`, or passes any malformed/out-of-range value, whether it will be handle gracefully ?

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-shubham, line 2226: Here is a dilemma:

`enable_htp_graph_splitting` and `htp_graph_splitting_num_prepare_threads` never reach this context creation path.

Please confirm If graph splitting is supported for  this workflow, thread count is ignored; 
if unsupported, the provider should reject or warn instead of accepting the option.

qti-shubham:
LTGM! Some minor comments to validate.

qti-shubham (approved):

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mbadnara: Thanks for pointing this

1) UINT32_MAX ("4294967295") — valid and intentional; std::stoul parses it correctly and it maps to auto-select (min(max(1, hardware_concurrency), num_splits)). No special handling needed.
 
2) Malformed input (e.g., "abc", "-1") — std::stoul throws std::invalid_argument. Now caught, logs a WARNING with the bad value and exception message, falls back to UINT32_MAX.

3) Out-of-range on 64-bit Linux (e.g., "99999999999") — on 64-bit Linux unsigned long is 64-bit, so std::stoul succeeds but the subsequent static_cast<uint32_t> would silently truncate. Now explicitly checked: if parsed > UINT32_MAX, logs a WARNING and falls back to UINT32_MAX.

All three error paths degrade gracefully to auto-select rather than crashing or silently producing a wrong value.

onnxruntime/core/providers/qnn/builder/qnn_backend_manager.cc
qti-mbadnara: Graph splitting is not supported when htp_share_resource_optimization=1.

Fixed by adding an explicit WARNING when the two options are used together: enable_htp_graph_splitting is not compatible with htp_share_resource_optimization=1

tirupath-qti (approved):

PR #802, opened 2026-09-06
[QNN EP] Fix DLC size inflation caused by unbounded constant DequantizeLinear folding

qti-mbadnara:
### Description

Bound constant `DequantizeLinear` folding by output size so it stops embedding oversized FP32 static blobs in the DLC.

`FoldConstantDequantizeLinear` pre-computed every constant DQ at compile time regardless of tensor size, storing the result as an FP32 STATIC tensor 4x the size of the original quantized weight. This adds a 1 MiB / 256K-element budget: a constant DQ above it is declined and left as a runtime QNN `Dequantize`, so the weight stays in the DLC in its compact quantized form.

**Scope — which weights this actually helps.** The budget only applies to inputs where a runtime QNN `Dequantize` is a faithful substitute for the folded constant, i.e. **per-tensor int8/uint8/int16/uint16 (and uint4) constants**. Two categories are exempt and still fold at any size:

- **Per-channel quantized constants.** QNN has no standalone per-channel `Dequantize`. `ExplicitOpCheck` admits a standalone per-channel Q/DQ only on the premise that folding will remove it, and declining also stops constant-ness from propagating to later Q/DQ hops in a chain — those hops then lose QNN support and the consumer's weight resurfaces as an `APP_WRITE` graph input. That is exactly the regression #339 fixed (224 leaked `v_proj` weight inputs on Qwen), so per-channel must keep folding unconditionally.
- **INT4/INT2 constants.** These reach QNN as `SFIXED_POINT_8` with their high bits masked off, to work around a QNN INT4 accuracy bug (see `UnpackInt4ToInt8`). Only the fold path undoes that mask, so a runtime `Dequantize` would decode negative values as `q + 16`.

Because transformer weights are commonly per-channel quantized, **this PR does not reduce DLC size for per-channel models** — including the Qwen models from #339, whose weights continue to fold exactly as before. The win is on per-tensor weights (e.g. the psx0 1536x6144 FC, 36 MB as FP32).

The 1 MiB budget is a fixed compile-time constant (`kQdqFoldMaxFp32Bytes`), not a provider option. It is a deliberate trade: declining a fold moves the dequantized tensor from DLC storage into HTP activation memory at graph-prepare time, so raising it is not free in either direction.

---

### Motivation and Context

The QNN EP constant-folds `DequantizeLinear` nodes fed by constant inputs: it pre-computes the FP32 output at graph-compile time and stores it as a STATIC tensor in the DLC. For small constants (bias vectors, scales, lookup tables) this is beneficial — it avoids repeated dequantization overhead at runtime.

For large per-tensor weight matrices, though, the same optimization produces FP32 static blobs 4x the size of the original INT8 weights. Across many large layers this inflates the DLC to the point where the quantized model is larger than its FP32 equivalent, defeating the purpose of quantization.

---

### Notes for reviewers

- `CanSubstituteRuntimeDequantize()` is the single place the two exemptions are decided, and it gates the size check — a "must fold" input never consults the budget. Its decision table is pinned by unit tests in `qdq_constant_folding_test.cc` (no device needed).
- The decline is only observable in the composed QNN graph, not in EP node assignment: the ONNX DQ node is marked supported either way. The e2e tests therefore assert on the dumped QNN graph JSON via `AssertOpInQnnGraph`.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/qdq_constant_folding.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/qdq_constant_folding.cc
@@ -114,6 +114,20 @@ Ort::Status FoldConstantDequantizeLinear(QnnModelWrapper& qnn_model_wrapper,
   RETURN_IF(output_info.qnn_data_type != QNN_DATATYPE_FLOAT_32,
             "Folded DequantizeLinear only supports float32 output.");
 
+  // Skip folding for large tensors: storing the dequantized FP32 blob in the DLC costs 4× the
+  // space of the original quantized bytes. Bias/scale tensors are small (< 1 MB) and benefit from
+  // folding because it eliminates a runtime op. Large weight tensors (e.g. 1536×6144 = 36 MB FP32)
+  // should remain as a QNN DequantizeLinear op so the DLC only stores the compact quantized data.
+  // A threshold of 1 MB (256 K float32 elements) keeps folding for bias-sized tensors.
+  constexpr size_t kMaxFoldElemCount = 256 * 1024;  // 256 K × 4 B = 1 MB FP32
+  size_t num_elems_check = 1;
+  for (uint32_t d : output_info.shape) {
+    num_elems_check *= d;
+    if (num_elems_check > kMaxFoldElemCount) {
+      return MAKE_EP_FAIL("DequantizeLinear output too large to fold; leaving as a runtime op to avoid FP32 DLC bloat.");
+    }
+  }
+
   std::vector<uint8_t> quant_bytes;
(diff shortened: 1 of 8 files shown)


qti-yuduo (approved):

minfhong-qti:
Hi @qti-yuduo, just curious whether it is intentional not squashing those commits during merge. IMO, squash would make the commit tree cleaner.

qti-yuduo:
> Hi @qti-yuduo, just curious whether it is intentional not squashing those commits during merge. IMO, squash would make the commit tree cleaner.

Accidentally of course. I requested admin to disable that option already, should not happen again.

PR #803, opened 2026-09-06
[CI] Migrate QDC runner to V2 API and upgrade SDK to 0.4.1

qti-mbadnara:
### Description

Migrate the QDC CI runner from the deprecated V1 Apigee proxy endpoint to the V2 endpoint, and upgrade the Qualcomm Device Cloud SDK from 0.3.0 to 0.4.1 which targets the V2 API.

---

### Motivation & Context

QDC (Qualcomm Device Cloud) is deprecating its V1 endpoints:

| V1 (deprecated) | V2 (replacement) |
|---|---|
| `https://apigwx-aws.qualcomm.com/saga/v1/public/qdc` | `https://api.qualcomm.com/devicecloud/v2` |
| `https://api.qualcomm.com/deviceloud/v1` | `https://api.qualcomm.com/devicecloud/v2` |

The SDK upgrade from 0.3.0 to 0.4.1 is required because 0.3.0 internally calls the V1 API paths, upgrading to 0.4.1 is the supported migration path.

The download URL template in `packages.yml` was also corrected from `/Windows/{version}/` to `/All/{version}/`, matching the distribution path used for 0.4.1 onward.

### DIFF
--- a/qcom/scripts/all/qdc_runner.py
+++ b/qcom/scripts/all/qdc_runner.py
@@ -269,7 +269,7 @@ class TestRuns:
 
 
 class QdcClient:
-    _QDC_PROD_API_URL = "https://apigwx-aws.qualcomm.com/saga/v1/public/qdc"
+    _QDC_PROD_API_URL = "https://api.qualcomm.com/devicecloud/v2"
 
     def __init__(self, api_url: str | None, on_behalf_of: str | None):
         api_token = os.environ["QDC_API_TOKEN"]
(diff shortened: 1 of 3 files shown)


qcom/scripts/all/qdc_runner.py
qti-kromero, line 310: it seems weird to me that this isn't an API URL - is this maybe something they added in the v2 version of the API?

qti-mbadnara: As far as I checked, this doesnt seem to exist. Can be taken up in a follow up

qti-yuduo:
Thanks for taking on this!

qti-kromero (approved):

PR #807, opened 2026-09-08
[QNN EP] Enable support for matmul lut

qti-niscmami:
### Description

  Add support for the `enable_htp_matmul_lut` QNN EP session option.

  For QAIRT 2.51 and later, MatMul LUT is enabled by default. The QNN EP passes `enable_matmul_lut=1` as an HTP graph-finalization configuration unless `enable_htp_matmul_lut=0` explicitly disables it. The option is propagated to per-SoC HTP configurations.

  For QAIRT versions earlier than 2.51, the option is ignored.

  Add unit coverage for default-enabled behavior, explicit opt-out, valid and invalid values, and unsupported QAIRT versions.

  ### Motivation and Context

  Enable the HTP MatMul LUT kernel optimization by default on QAIRT 2.51 and later, while preserving an opt-out through the `enable_htp_matmul_lut` QNN EP session option.


qti-mbadnara:
@qti-niscmami Gate the flag to the QNN API version / QNN SDK Version of QAIRT 2.51

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -991,6 +992,12 @@ QnnEp::QnnEp(QnnEpFactory& factory,
   }
 #endif
 
+  htp_graph_configs_.enable_htp_matmul_lut = ParseBoolOption(ort_api,
+                                                             session_options_,
+                                                             FormatEPConfigKey("enable_htp_matmul_lut"),
+                                                             true,
```
qti-ashwshan, line 999: the default behavior should be true ?

qti-niscmami: Yes. @tirupath-qti wanted it to be true default.

onnxruntime/core/providers/qnn/builder/qnn_def.h
```diff
@@ -164,6 +164,7 @@ typedef struct HtpGraphConfigs {
   bool enable_htp_fp16_precision = false;
   bool enable_htp_monolithic_lstm = false;
   bool enable_htp_fp16_clamp_overflow = false;  // Intentionally undocumented; for internal/diagnostic use only.
+  bool enable_htp_matmul_lut = false;
```
qti-ashwshan: If the default behavior is true, do we want to keep this false?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-mbadnara: Enable it to be True for QNN SDK >= 2.51
Gate it and dont make the provider option available for QNN SDK < 2.51

qti-mbadnara: Please update the documentation for the same

onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
```diff
@@ -684,6 +684,23 @@ TEST_F(QnnUnit_ExecutionProviderTest, Ctor_EnableHtpMonolithicLstmInvalid_LogsVe
...
+TEST_F(QnnUnit_ExecutionProviderTest, Ctor_EnableHtpMatmulLutInvalid_LogsVerbose) {
+  EpStubContext ctx;
+  ctx.log_severity = ORT_LOGGING_LEVEL_VERBOSE;
+  ctx.session_config[EPKey("enable_htp_matmul_lut")] = "maybe";
+  auto factory = MakeFactory(ctx);
+  EXPECT_NO_THROW({ auto ep = MakeEp(*factory, ctx); });
+  ExpectLogged(ctx, ORT_LOGGING_LEVEL_VERBOSE,
+               "Invalid value for ep.qnnexecutionprovider.enable_htp_matmul_lut");
+}
+
```
qti-mbadnara, line 704: Can we add few negative tests, what will happen when user tries to set this provider option when using QAIRT < 2.51

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
qti-niscmami: Done

onnxruntime/test/providers/qnn/unit/qnn_execution_provider_test.cc
qti-niscmami: Done. The guard should prevent the option from taking effect, not prevent the negative test from running. Below QAIRT 2.51, the option should be ignored without failing session creation.

onnxruntime/core/providers/qnn/builder/qnn_def.h
qti-niscmami: Fixed it.

tirupath-qti (approved):

PR #809, opened 2026-09-10
[QNN EP] Fold static empty Slice output to a STATIC tensor

qti-yuduo:
## Problem
Valid ONNX Slices with statically-empty output (e.g. input `[1,64,128]`, `starts=[64]`, `ends=[INT64_MAX]`, `axis=1` → `[1,0,128]`) are computed correctly by `slice_op_builder.cc`, then emitted unchanged as QNN `StridedSlice`, which HTP rejects with error 3110 and kills the compile. Native QAIRT succeeds.

## Fix
An empty Slice output holds no elements — its value is fully determined by the shape. In `SliceOpBuilder::ProcessAttributesAndOutputs`, right after `PrepareForComputeHelper`, a computed zero extent now folds the output to a `QNN_TENSOR_TYPE_STATIC` tensor (marked folded-constant, so downstream Q/DQ keep folding) instead of submitting `StridedSlice`. Runs identically in validation and compile passes so capability and compile agree; re-entry guarded for the two GetCapability passes sharing one wrapper. Correct for SingleNode and QDQ-group Slice alike (group outputs are the target outputs + encodings; the Q nodes are never separately built). Graph-output empty Slice cannot elide its producing op, so it declines with a precise diagnostic instead of the opaque 3110 → generic "Failed to add node".

## Tests
- `SliceEmptyOutputFoldsToStatic` on CPU (builder logic) and HTP (residency: asserts zero `StridedSlice` ops in the dumped QNN JSON, full EP assignment, accuracy parity), mirroring the issue shape.

### DIFF
--- a/onnxruntime/core/providers/qnn/builder/opbuilder/slice_op_builder.cc
+++ b/onnxruntime/core/providers/qnn/builder/opbuilder/slice_op_builder.cc
@@ -305,6 +305,51 @@ Ort::Status SliceOpBuilder::ProcessAttributesAndOutputs(QnnModelWrapper& qnn_mod
   PrepareForComputeMetadata compute_metadata(input_dimensions);
   RETURN_IF_ERROR(PrepareForComputeHelper(raw_starts, raw_ends, raw_axes, raw_steps, compute_metadata));
 
+  // HTP rejects StridedSlice with an empty output (QNN error 3110), but a static Slice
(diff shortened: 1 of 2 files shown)


qti-yuduo:
Superseded: moving to an origin-branch PR so /ci can dispatch (workflow_dispatch requires the ref on origin). The fix content is identical.

PR #822, opened 2026-09-14
[QNN EP] Fix stale SharedContext singleton leak

huaychou:
### Description

- Hoist the terminator reset out of the reuse branch into a single unified `if (stop_share_ep_contexts_) Reset...` that fires regardless of which branch ran.
- Guard the create-branch publish with `!stop_share_ep_contexts_` to avoid a pointless publish+immediate-reset cycle for a first-and-only terminator session.
- Add a load-bearing comment on the `!stop_share_ep_contexts_` guard in `CreateEPContextNodes` to explain why it prevents re-publishing after the ctor-time reset.
- Add `ResetSharedBackendManagerViaTerminatorSession()` helper in `qnn_test_utils` — constructs a throwaway `share=1 + stop=1 + htp_share_resource_optimization=1` session that drives the EP to release the singleton across the dlopen plugin boundary on real HTP devices (direct `SharedContext` access is not available to tests in normal builds).
- Remove `DISABLED_` from `Encryption_VtcmSharing_Baseline_NoCallback`, `Encryption_VtcmSharing_MultiSession_EndToEnd`, and `VTCMBackupBufferSharing`; call the helper at the end of each test before file cleanup.  x86 `GTEST_SKIP()` guards are preserved (x86 HTP emulator cannot reload shared-resource contexts).

### Motivation and Context

- When a session used `enable_vtcm_backup_buffer_sharing` or `htp_share_resource_optimization=1`, the EP called `SetSharedQnnBackendManager()` but `ResetSharedQnnBackendManager()` was only triggered inside the *reuse* branch.  A session that was both the first sharing session (create branch) and the terminator (`stop_share_ep_contexts_=1`), or a standalone vtcm session with no terminator at all, left a stale `shared_ptr` in the process-lifetime `SharedContext` singleton, causing undefined behaviour in CI and forcing three tests to be `DISABLED_`.


onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -1514,10 +1510,18 @@ QnnEp::QnnEp(QnnEpFactory& factory,
                                      skip_backend_op_validation,
                                      reused_io_limit_mb},
         ApiPtrs{ort_api, ep_api, model_editor_api}, logger_);
-    if (htp_share_resource_optimization_ == 1) {
+    // Publish for later sessions, but NOT when this session is the terminator —
+    // publishing then immediately resetting would be a no-op and is misleading.
+    if (htp_share_resource_optimization_ == 1 && !stop_share_ep_contexts_) {
```
minfhong-qti: I think current modification is incorrect.
Shared backend manager is mandatory for `htp_share_resource_optimization_=1` in current session, not only later one. Refer to `ContextCreateAsyncCallback` in qnn_backend_manager.cc.

huaychou: Fixed.

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
```diff
@@ -2275,9 +2277,21 @@ OrtStatus* ORT_API_CALL QnnEp::GetCapabilityImpl(OrtEp* this_ptr,
   if (!rt.IsOK()) {
     const std::string message = "QNN SetupBackend failed " + rt.GetErrorMessage();
     ORT_CXX_LOG(ep->logger_, ORT_LOGGING_LEVEL_ERROR, message.c_str());
+    // Reset on failure too so the singleton is not left stale if this was a terminator session.
```
minfhong-qti: Why not before the check of returned status? Then no need of two places.

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
```diff
@@ -943,6 +954,9 @@ TEST_F(QnnHTPBackendTests, Encryption_WithShareEpContexts_RoundTrip) {
       FAIL() << "phase B exception: " << e.what();
     }
   }
+  // Release the SharedContext singleton before deleting files to avoid cross-test leakage.
+  ResetSharedBackendManagerViaTerminatorSession(phase_b_ep_device.get(), provider_options,
```
minfhong-qti: Solely loading context binary with shared_ep_context will not set shared QnnBackendManager, and thus I think no need to clean up here.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -2312,6 +2312,8 @@ TEST_F(QnnHTPBackendTests, QnnContextShareAcrossSessions) {
 
   auto ort_outputs1 = scoped1.session().Run(Ort::RunOptions{}, input_names_c.data(), ort_inputs.data(), ort_inputs.size(),
                                             output_names_c.data(), 1);
+  // Release the SharedContext singleton before deleting files to avoid cross-test leakage.
+  ResetSharedBackendManagerViaTerminatorSession(scoped1.ep_device(), provider_options, ctx_model_file1.c_str());
```
minfhong-qti: Ditto. No need to clean up in this case.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
```diff
@@ -2646,6 +2652,19 @@ TEST_F(QnnHTPBackendTests, QnnContextGenWeightSharingSessionAPI) {
   auto file_size_1 = std::filesystem::file_size(qnn_ctx_binary_file_name1);
   EXPECT_TRUE(file_size_1 > 0);
 
+  // Release the SharedContext singleton before deleting files to avoid cross-test leakage.
```
minfhong-qti: `DumpModelWithSharedCtx` already stop shared, so no need below blocks.

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
```diff
@@ -491,16 +491,22 @@ TEST_F(QnnHTPBackendTests, DISABLED_Encryption_VtcmSharing_Baseline_NoCallback)
     }
   }
 
+  ProviderOptions phase_b_provider_options;
```
minfhong-qti: The provider option is identical across three places including above. I think they can be reused.

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
```diff
@@ -674,6 +682,9 @@ TEST_F(QnnHTPBackendTests, DISABLED_Encryption_VtcmSharing_MultiSession_EndToEnd
   }
 
   std::error_code ec;
+  // Release the SharedContext singleton before deleting files to avoid cross-test leakage.
+  auto s1_cleanup_options = make_provider_options();
```
minfhong-qti: Why not directly use `s1_provider_options`?

onnxruntime/core/providers/qnn/qnn_execution_provider.cc
huaychou: Moved.

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
huaychou: Removed.

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
huaychou: Removed

onnxruntime/test/providers/qnn/qnn_ep_context_test.cc
huaychou: Removed

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
huaychou: Removed

onnxruntime/test/providers/qnn/qnn_ep_context_encryption_test.cc
huaychou: Removed

minfhong-qti (approved):

PR #828, opened 2026-09-15
[CI] Add maintainer-approved device tests for fork PRs

huaychou:
## Summary

- add a maintainer-only `/device-test` PR command for reviewed fork PRs
- dispatch the existing `qualcomm-internal-ci.yml` from `main`, pinned to the reviewed fork SHA
- add a device-tests-only mode to the existing reusable CI pipeline, which runs the existing Android and Qualcomm Linux QDC paths
- remove the fork-QDC success placeholder, so an external PR cannot satisfy device-test requirements without a real device run
- publish the real maintainer-approved QDC result under the existing required Android/Linux QDC check names on the reviewed source SHA
- carry the checked-out repository/SHA through existing archive, build, and QDC reusable workflows so Artifactory hand-off uses the reviewed source identity

## Safety

Only repository collaborators, members, and owners can invoke the command. The command records the PR head SHA before dispatching; later contributor pushes require another explicit approval. The dispatch runs the workflow definition from `main`, while the existing build and QDC jobs explicitly check out that reviewed SHA.

## Validation

- `git diff --check`
- YAML parse of all changed workflow files
- Python compile check and artifact-key assertion for `CiArtifactory`


.github/workflows/qualcomm-internal-archive-testdata.yml
```diff
@@ -25,7 +33,10 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
+          repository: ${{ inputs.source_repository || github.repository }}
+          ref: ${{ inputs.source_ref || github.sha }}
           submodules: recursive
+          persist-credentials: false
```
kuanyul-qti, line 39: The intended safety model is "trusted workflow from `main`, pinned reviewed fork SHA," but after checking out `inputs.source_repository/source_ref`, the jobs still invoke repo-local actions/scripts such as `./.github/actions/upload-ci-artifact`, `./.github/actions/download-ci-artifact`, and `qcom/scripts/artifactory/*` while authenticated Artifactory/QDC tooling is available. That lets the reviewed fork tree control credential-adjacent CI helper code, which undercuts the trust boundary this PR is trying to establish. Keep trusted CI helpers in a separate `main` checkout/tools directory and use the fork checkout only as the build/test payload source.

.github/workflows/qualcomm-internal-ci-trigger.yml
```diff
@@ -22,45 +24,82 @@ on:
...
     if: |
       github.event.issue.pull_request != null &&
       (github.event.comment.body == '/ci' ||
        startsWith(github.event.comment.body, '/ci ') ||
-       startsWith(github.event.comment.body, '/ci\n')) &&
+       startsWith(github.event.comment.body, '/ci\n') ||
+       github.event.comment.body == '/device-test' ||
+       startsWith(github.event.comment.body, '/device-test ') ||
+       startsWith(github.event.comment.body, '/device-test\n')) &&
       contains(fromJSON('["COLLABORATOR", "MEMBER", "OWNER"]'),
```
kuanyul-qti, line 38: Since the `onnxruntime-qnn` repo is part of the `onnxruntime` organization, it seems that any member or owner of that organization can still trigger the CI. Is that expected?

.github/workflows/qualcomm-internal-archive-testdata.yml
huaychou: I think we cannot run the workflow on the trusted `main`. This will introduce disparity between fork PR and CI results. I think the reviewer should review the change before triggering the `/device-test`.

.github/workflows/qualcomm-internal-ci-trigger.yml
huaychou: Updated with COLLABORATOR only

yuhuchua-qti:
Could we document the validation boundary for `/device-test`? Since the dispatch runs `qualcomm-internal-ci.yml` from `main`, fork PR changes to `.github/workflows/*` are not exercised by this path. Local actions/scripts from the checked-out source may still be exercised, but workflow-definition changes are not. If that is intentional for security reasons, it would be helpful to state that fork PR workflow changes require separate maintainer validation from a trusted branch/main and should not be considered validated by `/device-test`.

huaychou:
> Could we document the validation boundary for `/device-test`? Since the dispatch runs `qualcomm-internal-ci.yml` from `main`, fork PR changes to `.github/workflows/*` are not exercised by this path. Local actions/scripts from the checked-out source may still be exercised, but workflow-definition changes are not. If that is intentional for security reasons, it would be helpful to state that fork PR workflow changes require separate maintainer validation from a trusted branch/main and should not be considered validated by `/device-test`.

I think the comment is no longer valid. Please refer to the latest change.

.github/workflows/qualcomm-internal-ci-trigger.yml
```diff
@@ -2,65 +2,180 @@
...
        startsWith(github.event.comment.body, '/ci ') ||
-       startsWith(github.event.comment.body, '/ci\n')) &&
-      contains(fromJSON('["COLLABORATOR", "MEMBER", "OWNER"]'),
-               github.event.comment.author_association)
-    runs-on: ["self-hosted", "X64", "Linux", "Build"]
+       startsWith(github.event.comment.body, '/ci\n') ||
+       github.event.comment.body == '/device-test' ||
+       startsWith(github.event.comment.body, '/device-test ') ||
+       startsWith(github.event.comment.body, '/device-test\n')) &&
+      github.event.comment.author_association == 'COLLABORATOR'
```
kuanyul-qti: **[C-1] (Critical)** <code class="c">author_association</code> is not a maintainer permission check and can allow non-write collaborators to trigger privileged CI

The new trigger uses only `github.event.comment.author_association == 'COLLABORATOR'` to decide who may run `/ci` or `/device-test`. `author_association` is a relationship label, not the user's repository permission level; it does not encode whether the commenter has write/maintain/admin access. This workflow has `contents: write` and `actions: write`, and the device path executes reviewed code with internal runners/secrets, so the authorization check needs to query the actual repository permission.

**Impact:** A direct collaborator with read/triage-style access can still satisfy the `COLLABORATOR` relationship gate and trigger privileged device CI, while org `MEMBER`/`OWNER` handling is also disconnected from actual repo permissions. The security boundary depends on a GitHub relationship label instead of explicit write-level authorization.

**Suggested fix:**

Call `github.rest.repos.getCollaboratorPermissionLevel` inside `github-script` for `context.payload.comment.user.login`, and allow only `write`/`maintain`/`admin` or the team's chosen minimum permission. Keep the job-level `if` as a cheap command filter, but fail closed on the explicit permission check before dispatching.

.github/workflows/qualcomm-internal-build-and-test-single-os.yml
```diff
@@ -268,30 +339,6 @@ jobs:
           name: "${{ inputs.target_os }}-${{ inputs.target_arch }}-py${{ inputs.target_py_vsn }}-pdb-zip"
           src_pattern: "${{inputs.target_os}}-${{ inputs.target_arch }}/${{ env.BUILD_CONFIG_DIR }}/dist/onnxruntime-qnn-*-win-*-pdb.zip"
```
kuanyul-qti: **[M-1] (Major)** Deleted fork stubs are not replaced for every previously required test context

This PR deletes `test-fork-stub` and `test-wheel-fork-stub`. Their previous comments said they intentionally used the same `name:` as the real jobs to satisfy branch protection when fork PRs cannot access secrets. After the deletion, the real `test`/`test-wheel` jobs still skip on fork PR events through `github.event.pull_request.head.repo.fork != true` (`:345`-`418`), while the new `/device-test` path only publishes QDC and Windows ARM64 device check contexts.

**Impact:** If branch protection still requires any cross-runner test or wheel-smoke context that those stubs produced, fork PRs will have no job producing that check. The failure mode changes from fake green to a permanently pending/missing required check.

**Suggested fix:**

Confirm the branch-protection required-check list and provide a replacement for every deleted stub context. If only QDC/device contexts are moving to real device runs, remove only the QDC placeholder; if the single-OS stubs also need to go, update branch protection in lock step or have the maintainer-approved path publish real checks with the same names.

huaychou: Updated

.github/workflows/qualcomm-internal-ci-trigger.yml
huaychou: Fixed

PR #841, opened 2026-09-16
[QNN EP] Add snapshot accuracy coverage runner

yuhuchua-qti:
## Summary

Based on `main` after #840 was merged. Split from #399 so the CI/runner behavior can be reviewed separately from the test-folder reorganization.

This PR updates the coverage flow so snapshot tests are treated as an accuracy-skip signal, not as the numerical correctness gate. When snapshot cannot prove the graph is unchanged, the flow falls back to the matching accuracy tests.

## Scope

- Update the coverage workflow to use the three-phase coverage flow.
- Add `qcom/scripts/linux/run_snapshot_accuracy.sh` as a standalone developer/debug helper.
- Update `generate_coverage.sh` to run component, snapshot, and accuracy phases.
- Route unverified snapshot states to matching `QnnUnit_<Op>_Accuracy*` tests.
- Treat these snapshot states as requiring accuracy fallback:
  - `[QNN_SNAPSHOT_DRIFT]`
  - `[QNN_GOLDEN_ABSENT]`
  - missing snapshot JSON
  - other snapshot failures/errors
- Fail setup if an unverified snapshot group has no matching accuracy test.

## Notes

`run_snapshot_accuracy.sh` is a standalone developer/debug helper. It is not the CI entrypoint in this PR. CI still enters through `generate_coverage.sh`, which now owns the coverage three-phase flow.

Since this PR only adds the runner/coverage flow and does not add migrated op tests, it may run with zero snapshot/accuracy tests when reviewed alone. Full functional validation of the snapshot-to-accuracy fallback is done on the final stacked branch with #842, where Clip provides the first `QnnUnit_Clip_Snapshot*` and `QnnUnit_Clip_Accuracy*` targets.

## Out of Scope

- Test infra folder reorganization. See #840.
- Clip per-op migration. See #842.
- Strict graph-drift gating; this can be considered later after golden publish/fetch/versioning is stable.

## Review Focus

- Snapshot-to-accuracy fallback policy.
- Missing golden and missing snapshot JSON behavior.
- Whether the standalone runner behavior matches the coverage CI behavior.

## Validation

Run on this PR / current branch:

- `coverage / coverage-linux-x86_64` CI passed.
- `python qcom/build_and_test.py lint_and_fix`
- `python qcom/build_and_test.py lint`
- `bash -n qcom/scripts/linux/generate_coverage.sh`
- `bash -n qcom/scripts/linux/run_snapshot_accuracy.sh`
- `git diff --check`

Also previously passed on the equivalent combined branch before this split:

- `python qcom/build_and_test.py coverage_linux_x86_64 --target-py-version 3.12`


qti-chuteng (approved):

qcom/scripts/linux/run_snapshot_accuracy.sh
```diff
@@ -0,0 +1,404 @@
...
+fi
+
+if [ -z "${binary}" ]; then
+    die "onnxruntime_provider_test not found under ${build_dir}. Is this a coverage build?"
+fi
+
+bin_dir="$(dirname "${binary}")"
+
+# Verify this is a coverage build: probe for snapshot tests.
+snapshot_probe=$("${binary}" --gtest_list_tests --gtest_filter="QnnUnit_*_Snapshot*" 2>/dev/null || true)
```
tzuhsuanwei: This probe runs without the bin_dir / LD_LIBRARY_PATH setup used below. A library-load failure is swallowed by || true and can be misreported as no snapshot tests.
Suggest running this probe in the same (cd "${bin_dir}"; export LD_LIBRARY_PATH=...; ./onnxruntime_provider_test ...) wrapper. The accuracy probe below should use the same wrapper too.

same as line:356

qcom/scripts/linux/run_snapshot_accuracy.sh
```diff
@@ -0,0 +1,404 @@
...
+    if isinstance(value, list):
+        return any(contains_marker(v, marker) for v in value)
+    return marker in str(value)
+
+
+def suite_has_drift(suite):
+    return contains_marker(suite, "QNN_SNAPSHOT_DRIFT")
+
+
+def suite_has_absent_golden(suite):
```
tzuhsuanwei, line 159: GTEST_SKIP() is reported as result: "SKIPPED" in gtest JSON, so the [QNN_GOLDEN_ABSENT] text may not be available here. A missing golden can then look like a clean snapshot pass and skip Accuracy.

Suggest treating any skipped snapshot testcase as unverified when collecting target ops. Please apply the same handling in generate_coverage.sh.

qcom/scripts/linux/generate_coverage.sh
```diff
@@ -132,17 +213,167 @@ rm -f "${build_dir}/${config}/coverage_lcov.info" \
...
+        return any(contains_marker(v, marker) for v in value)
+    return marker in str(value)
+
+
+def suite_is_unverified(suite):
+    return (
+        suite.get("failures", 0) > 0
+        or suite.get("errors", 0) > 0
+        or contains_marker(suite, "QNN_SNAPSHOT_DRIFT")
+        or contains_marker(suite, "QNN_GOLDEN_ABSENT")
```
tzuhsuanwei, line 286: Same as run_snapshot_accuracy.sh:151

qcom/scripts/linux/run_snapshot_accuracy.sh
yuhuchua-qti: Good point. 
Updated `run_snapshot_accuracy.sh` so the snapshot probe, accuracy probe, and actual snapshot/accuracy runs all go through the same `run_provider_test()` wrapper. The wrapper changes into `bin_dir`, sets `LD_LIBRARY_PATH`, and then invokes `onnxruntime_provider_test`, so probe failures should no longer be misreported as missing tests due to a different runtime environment.

qcom/scripts/linux/run_snapshot_accuracy.sh
yuhuchua-qti: Good catch. 
Updated the JSON parsing to treat skipped snapshot testcases as unverified, even if the `[QNN_GOLDEN_ABSENT]` marker text is not present in the JSON. This ensures missing goldens still route to the matching accuracy tests instead of being treated as a clean snapshot pass.

qcom/scripts/linux/generate_coverage.sh
yuhuchua-qti: Applied the same skipped-snapshot handling in `generate_coverage.sh`. Snapshot failures/errors, drift markers, missing-golden markers, and skipped snapshot testcases are now all treated as unverified and require matching accuracy coverage.

qcom/scripts/linux/generate_coverage.sh
```diff
@@ -227,8 +466,41 @@ cp "${REPO_ROOT}/qcom/scripts/linux/coverage_artifact_README.md" \
...
+# Propagate test failure after coverage report has been generated.
+#
+# The component and accuracy phases GATE (non-zero exit fails this script). The
+# snapshot phase is not a correctness gate; it is an accuracy-skip signal. Any
+# unverified snapshot result falls back to accuracy. The only setup failure is
+# when an unverified snapshot group has no matching accuracy test.
 # ---------------------------------------------------------------------------
-if [ "${test_exit}" -ne 0 ]; then
-    die "Tests failed with exit code ${test_exit}. Coverage report was still generated at ${output_dir}."
+if [ "${snapshot_exit}" -ne 0 ]; then
```
tzuhsuanwei, line 476: A skipped snapshot test normally keeps the gtest exit code at 0. In that case this block is not reached, so an unverified [QNN_GOLDEN_ABSENT] group bypasses assert_accuracy_exists_for_groups(). Accuracy happens to run globally today, but this misses the required matching-Accuracy setup check and will matter once routing becomes selective.
Suggest parsing snapshot_json whenever the snapshot phase ran, regardless of snapshot_exit, and calling assert_accuracy_exists_for_groups() whenever it reports unverified groups

tzuhsuanwei (approved):

qti-chuteng (approved):

PR #842, opened 2026-09-16
[QNN EP] Migrate Clip tests to tier infra

yuhuchua-qti:
## Summary

Stacked on #841. Split from #399 so reviewers can look at the first op migration separately from the shared infra and coverage runner changes.

This PR migrates Clip into the tier-based QNN test infra as the first concrete per-op example.

## Scope

- Add shared Clip test specs under `infra/specs/`.
- Add Clip component-tier tests.
- Add Clip snapshot-tier tests.
- Add Clip session-snapshot-tier tests.
- Add Clip accuracy-tier tests.
- Enable the Clip tier source globs and `QNN_EP_ACCURACY_UT` for coverage/export-symbol builds.
- Extend QNN quant-param JSON dump support needed by snapshot golden comparison.

## Coverage impact

From the coverage run on the equivalent combined branch before this split, `clip_op_builder.cc` coverage improved from the legacy integration baseline to the new tier-based coverage:

| Test set | Lines | Functions | Branches |
|---|---:|---:|---:|
| Legacy integration (`Qnn*BackendTests.Clip_*`) | 62.8% | 100.0% | 52.5% |
| New component + snapshot | 98.3% | 100.0% | 64.6% |
| New component + snapshot + accuracy | 98.3% | 100.0% | 64.6% |

The component/snapshot tiers provide the builder code-path coverage. The accuracy tier is still needed for numerical correctness, but it does not add additional `clip_op_builder.cc` line/branch coverage beyond component/snapshot.

## Out of Scope

- Shared infra reorganization. See #840.
- Coverage runner behavior. See #841.
- Additional op migrations.
- Disabling legacy Clip integration tests; that policy can be applied after the new infra is stable.

## Review Focus

- Whether the Clip specs are reusable across component/snapshot/session-snapshot/accuracy tiers.
- Whether the migrated Clip coverage preserves the meaningful legacy integration coverage.
- Whether the snapshot JSON additions are sufficient and appropriately scoped for Clip.

## Validation

Run on the final stacked branch:

- `python qcom/build_and_test.py lint_and_fix`
- `python qcom/build_and_test.py lint`
- `bash -n qcom/scripts/linux/generate_coverage.sh`
- `bash -n qcom/scripts/linux/run_snapshot_accuracy.sh`
- `git diff --check`

Also previously passed on the equivalent combined branch before this split:

- `python qcom/build_and_test.py coverage_linux_x86_64 --target-py-version 3.12`


cmake/onnxruntime_unittests.cmake
```diff
@@ -196,13 +196,19 @@ set (onnxruntime_test_providers_dependencies ${onnxruntime_EXTERNAL_DEPENDENCIES
 set(onnxruntime_test_framework_src_patterns)
 if(onnxruntime_USE_QNN AND NOT onnxruntime_MINIMAL_BUILD AND NOT onnxruntime_REDUCED_OPS_BUILD)
   list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/accuracy/builder/opbuilder/*)
   list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/component/*)
+  list(APPEND onnxruntime_test_framework_src_patterns ${TEST_SRC_DIR}/providers/qnn/component/builder/*)
```
kuanyul-qti: We don't need this.

onnxruntime/test/providers/qnn/infra/specs/builder/opbuilder/clip_specs.h
```diff
@@ -0,0 +1,398 @@
...
+//   * op-builder snapshot: kClipSpecs + kClipQDQFloatOpBuilderSpecs
+//                          + kClipQDQQuantSpecs + kClipFoldedConstSpecs
+//   * session snapshot   : kClipQDQFloatSessionSpecs (Group B only)
+//   * accuracy           : kClipSpecs + kClipQDQFloatAccuracySpecs
+//                          + kClipQDQQuantSpecs + kClipFoldedConstSpecs
+// Adding a case = add one literal to the right list; every consuming tier
+// picks it up with a matching name automatically.
+// ---------------------------------------------------------------------------
+
+inline const std::vector<ClipSpec> kClipSpecs = {
```
kuanyul-qti, line 377: The migrated Clip spec source omits the legacy infinity-bound behavior class. The old HTP suite covers `Clip_NegativeInfMin` and `Clip_PositiveInfMin`, which map to the Clip builder's explicit `-inf`/`+inf` finite-bound clamp, but the new shared spec lists feeding snapshot and accuracy contain no infinity-bound cases. At least add `min=-inf,max=5` and `min=-5,max=+inf` specs to the migrated matrix, or explicitly document them as retained legacy-only coverage that must be handled before legacy Clip tests are disabled.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -737,10 +737,50 @@ static nlohmann::json GetQnnTensorJSON(const Qnn_Tensor_t& tensor, bool include_
...
 
   const Qnn_QuantizeParams_t& quant_params = GetQnnTensorQParams(tensor);
-  tensor_json["quant_params"] = {
-      {"definition", quant_params.encodingDefinition},
-      {"encoding", quant_params.quantizationEncoding},
-      {"scale_offset", {{"scale", quant_params.scaleOffsetEncoding.scale}, {"offset", quant_params.scaleOffsetEncoding.offset}}}};
+  // quant_params JSON layout matches the QAIRT official qairt-dlc-to-json schema.
+  // Each case serializes only the union member valid for that encoding, avoiding
+  // undefined-behavior reads of the wrong union field.
+  nlohmann::json quant_params_json = {{"definition", quant_params.encodingDefinition},
```
huaychou, line 743: Do we need to check if encodingDefinition is encodingDefinition first?

onnxruntime/test/providers/qnn/accuracy/builder/opbuilder/clip_test.cc
```diff
@@ -0,0 +1,452 @@
...
+class QnnUnit_Clip_Accuracy_QDQFloatTest
+    : public ::testing::TestWithParam<ClipQDQFloatSpec> {};
+class QnnUnit_Clip_Accuracy_QDQQuantTest
+    : public ::testing::TestWithParam<ClipQDQQuantSpec> {};
+class QnnUnit_Clip_Accuracy_FoldedConstTest
+    : public ::testing::TestWithParam<ClipFoldedConstSpec> {};
+
+TEST_P(QnnUnit_Clip_AccuracyTest, Case) {
+  RunClipAccuracy(GetParam());
+}
```
huaychou: **[Minor]** `BuildClipQDQQuantOnnxFromSpec` dispatches on `(qdq_dtype, shape)` to pick input data, and both `kClipU8QuantizedMinSpec` and `kClipU8QuantizedMinMaxSpec` match the `UINT8 + {200}` branch, silently sharing the same 28..227 input data with no comment explaining this is intentional. A future `{200}`-shape spec with different data requirements would be silently routed to the wrong data and might pass vacuously.

```suggestion
      // 28..227 input fits both kClipU8QuantizedMin (min=0.0, range -0.1..0.099)
      // and kClipU8QuantizedMinMax (min=-0.05, max=0.05, same range). Both specs
      // share this data intentionally -- the range straddles every clip boundary
      // under scale=0.001, zp=128. If adding a new {200}-shape spec that requires
      // a different input distribution, add a distinguishing shape or a
      // per-spec name dispatch below.
      if (spec.shape == std::vector<int64_t>{200}) {
```

cmake/onnxruntime_unittests.cmake
```diff
@@ -394,12 +400,18 @@ block()
     # QNN_EP_INTERNAL_SYMBOL_ACCESS gates test code that depends on EP-internal symbols.
     # It tracks whether the test binary is link-time bound to the SHARED EP library
     # (i.e., the cmake conditions above hold), not whether any production source is
-    # under #if. When the macro is off, component test bodies compile to empty
-    # translation units, so non-coverage builds do not see undefined references.
+    # under #if. When the macro is off, tier test bodies (component/, snapshot/,
```
huaychou: **[Minor]** The updated comment listing `QNN_EP_INTERNAL_SYMBOL_ACCESS`-gated tier directories (`component/`, `snapshot/`, `session_snapshot/`, `accuracy/`) omits `infra/`, which is also in the GLOB patterns just above. Minor inconsistency that could mislead a future maintainer into thinking `infra/specs/` headers are outside the gated infrastructure.

onnxruntime/test/providers/qnn/snapshot/builder/opbuilder/clip_test.cc
```diff
@@ -0,0 +1,419 @@
...
+//   ASSERT_NE(wrapper, nullptr) << "Failed to initialize QNN graph";
+
+// Group A: Plain dtype data, optional float min/max scalar input(s).
+//
+// min/max stored in spec as `std::optional<float>` regardless of dtype; this
+// helper casts to the actual dtype when registering the scalar initializer
+// (INT32 truncates, FP16 rounds via Ort::Float16_t). Mirrors the integration-tier
+// `Clip_*` graph structure so snapshot ↔ accuracy ↔ integration test the same
+// graph.
+void RunClipSnapshot([[maybe_unused]] SnapshotBackend backend,
```
huaychou: **[Nit]** All four `Run*Snapshot*` helpers carry a `[[maybe_unused]] SnapshotBackend backend` parameter that is never used (QnnCpu no longer ships). The intent (retain for future backend flexibility) is documented at the file level, but a brief inline note in each helper — e.g. `// currently always HTP; kept for future backend flexibility` — would save readers from hunting for the explanation.

onnxruntime/test/providers/qnn/accuracy/builder/opbuilder/clip_test.cc
```diff
@@ -0,0 +1,452 @@
...
+
+TEST_P(QnnUnit_Clip_Accuracy_QDQFloatTest, Case) {
+  RunClipQDQFloatAccuracy(GetParam());
+}
```
huaychou: **[Nit]** Manual fill loop `input_data[i] = static_cast<uint8_t>(28 + i)` (and the `0..47` loop above) can be replaced with `std::iota` — `<numeric>` is already included.

```suggestion
        std::vector<uint8_t> input_data(200);
        std::iota(input_data.begin(), input_data.end(), uint8_t{28});
```

cmake/onnxruntime_unittests.cmake
yuhuchua-qti: Removed it in the follow-up commit. The op-builder component tests are covered by the more specific `component/builder/opbuilder/*` glob, so this intermediate `component/builder/*` entry is unnecessary.

onnxruntime/test/providers/qnn/infra/specs/builder/opbuilder/clip_specs.h
yuhuchua-qti: Good catch. I added `Clip_NegativeInfMin` and `Clip_PositiveInfMin` to the shared `kClipSpecs` list, so they are now covered by both the snapshot and accuracy tiers. I kept the `Clip_` prefix to match the legacy integration test names and keep golden/  test-name mapping straightforward.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
yuhuchua-qti: Fixed. The JSON still includes definition/encoding, but now only reads the quantization union payload when encodingDefinition indicates the payload is valid. This avoids dumping invalid fields for QNN_DEFINITION_UNDEFINED.

onnxruntime/core/providers/qnn/builder/qnn_utils.cc
```diff
@@ -737,10 +737,53 @@ static nlohmann::json GetQnnTensorJSON(const Qnn_Tensor_t& tensor, bool include_
...
-      {"encoding", quant_params.quantizationEncoding},
-      {"scale_offset", {{"scale", quant_params.scaleOffsetEncoding.scale}, {"offset", quant_params.scaleOffsetEncoding.offset}}}};
+  // quant_params JSON layout matches the QAIRT official qairt-dlc-to-json schema.
+  // Always serialize the definition and encoding fields. Only serialize the
+  // encoding-specific union payload when the definition says those fields are valid.
+  nlohmann::json quant_params_json = {{"definition", quant_params.encodingDefinition},
+                                      {"encoding", quant_params.quantizationEncoding}};
+  if (quant_params.encodingDefinition == QNN_DEFINITION_IMPL_GENERATED ||
+      quant_params.encodingDefinition == QNN_DEFINITION_DEFINED) {
+    switch (quant_params.quantizationEncoding) {
```
tzuhsuanwei: GetQnnTensorJSON() currently serializes payloads for SCALE_OFFSET / BW_SCALE_OFFSET / AXIS_SCALE_OFFSET / BW_AXIS_SCALE_OFFSET only. QNN EP also creates BLOCK, BW_FLOAT_BLOCK, and BLOCKWISE_EXPANSION quant params (for example the block-quant Conv paths). Those encodings now fall through the default case and emit only definition and encoding, so two tensors with different block sizes/scales/offsets can produce the same snapshot JSON.

Could we either serialize the payload for the block encodings too, or emit an explicit unsupported payload marker/error? 
That would keep future snapshot tests from silently missing block-quant regressions.

yuhuchua-qti: Good point. 
Since this PR is focused on Clip migration, I did not expand it to fully serialize BLOCK / BW_FLOAT_BLOCK / BLOCKWISE_EXPANSION payloads here. Instead, I added an explicit payload_status: unsupported_quantization_encoding marker for valid-definition quant encodings whose payload is not serialized, so snapshots will not silently imply those payloads are covered. Full block-quant payload serialization can be handled in a follow-up PR when we migrate block-quant related coverage.

kuanyul-qti (approved):

tzuhsuanwei (approved):

onnxruntime/test/providers/qnn/accuracy/builder/opbuilder/clip_test.cc
```diff
@@ -0,0 +1,452 @@
...
+          input_data[i] = static_cast<uint8_t>(i);
+        }
+        return BuildClipU8QDQQuantHandRolled<uint8_t>(spec, std::move(input_data));
+      }
+      if (spec.shape == std::vector<int64_t>{200}) {
+        // Clip_U8_QuantizedMin / Clip_U8_QuantizedMinMax: input data = 28..227 (200 elems)
+        std::vector<uint8_t> input_data(200);
+        for (size_t i = 0; i < 200; ++i) {
+          input_data[i] = static_cast<uint8_t>(28 + i);
+        }
```
huaychou: **[Nit]** Manual fill loop `input_data[i] = static_cast<uint8_t>(28 + i)` (and the `0..47` loop above) can be replaced with `std::iota` — `<numeric>` is already included.

```suggestion
        std::vector<uint8_t> input_data(200);
        std::iota(input_data.begin(), input_data.end(), uint8_t{28});
```

yuhuchua-qti: Thanks, updated. 
I replaced the manual sequential fill loops with std::iota. I also applied the same cleanup to the adjacent uint16_t sequential input-data case for consistency.

qti-chuteng (approved):

PR #844, opened 2026-09-17
[CI] Clean up Git history after CI jobs

huaychou:
### Description

- Add post job action to clean up the git history

### Motivation and Context

- The git history occupies large space, which resulting into "no space left" issue on Linux CI machine.


qcom/ep_build/tasks/build.py
```diff
@@ -226,10 +229,20 @@ def __init__(
...
     ) -> None:
         diff_file = build_dir / config / "patch.diff"
         coverage_xml = build_dir / config / "coverage" / "coverage.xml"
 
+        if base_commit:
+            fetch_command = ["git", "fetch", "--no-tags", "--depth=1", "origin", base_commit]
+            diff_base = base_commit
+            diff_separator = ""
+        else:
+            fetch_command = ["git", "fetch", "origin", base_branch.removeprefix("origin/")]
```
kuanyul-qti, line 242: The PR makes the coverage workflow checkout shallow globally, but the non-PR diff-coverage fallback still uses `git diff origin/main...HEAD`, which needs a merge base that may not exist in a depth-1 checkout. PR events are covered by `pull_request.base.sha`, but `workflow_dispatch` / non-PR `workflow_call` paths still exist and can hit this fallback. Make that fallback shallow-safe too, or keep full-history checkout for non-PR coverage modes.

huaychou: Fixed.

.github/workflows/qualcomm-internal-build-and-test.yml
```diff
@@ -50,8 +50,11 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
-          fetch-depth: 0
-          # No submodules: the gate only needs git history, not the working tree.
+          fetch-depth: 1
+          # No submodules: the gate only needs the checked-out tree and PR base tree.
+      # Register a post-job cleanup immediately after checkout. It removes .git even
+      # when a later step fails, preventing Git data from persisting on a self-hosted runner.
+      - uses: ./.github/actions/cleanup-git-history
```
yuhuchua-qti: GitHub Actions post-job steps appear to run in reverse registration order because the runner stores them in a stack: `PostJobSteps` is a `Stack<IStep>`, `RegisterPostJobStep()` pushes onto it, and `StepsRunner` later pops them when entering the post-job
phase:
- https://github.com/actions/runner/blob/main/src/Runner.Worker/ExecutionContext.cs#L221-L222
- https://github.com/actions/runner/blob/main/src/Runner.Worker/ExecutionContext.cs#L335-L370
- https://github.com/actions/runner/blob/main/src/Runner.Worker/StepsRunner.cs#L57-L76

With the current ordering, this cleanup action is registered after `actions/checkout`, so its post step should run before checkout's own post step. Since `post.js` removes `.git` entirely, could this interfere with checkout's post cleanup, such as credential or `safe.directory` cleanup? If checkout expects the repository metadata to still exist, should this cleanup action be registered before checkout, or should we use a cleanup approach that preserves enough metadata for checkout's post step?

Same concern applies to the other workflows that register this cleanup action immediately after checkout:
- `.github/workflows/qualcomm-internal-asan.yml:23`
- `.github/workflows/qualcomm-internal-coverage.yml:45`

qcom/ep_build/tasks/build.py
```diff
@@ -226,10 +229,31 @@ def __init__(
...
+        base_commit: str | None = None,
     ) -> None:
         diff_file = build_dir / config / "patch.diff"
         coverage_xml = build_dir / config / "coverage" / "coverage.xml"
 
+        if base_commit:
+            fetch_command = ["git", "fetch", "--no-tags", "--depth=1", "origin", base_commit]
+            diff_base = base_commit
+            diff_separator = ""
+        elif is_host_github_runner():
```
kuanyul-qti: **[M-1] (Major)** Non-PR GitHub runners use a two-tree diff that includes unrelated main changes in diff coverage

`base_commit` is populated only from a pull-request event, so `workflow_dispatch` and nightly coverage on a GitHub runner take this branch. The empty separator makes the later command run `git diff origin/main HEAD` instead of the prior `origin/main...HEAD` merge-base diff.

**Impact:** When a manually dispatched branch is behind or diverged from main, the report includes changes introduced on main after the divergence that are not part of the branch. That makes both the diff-coverage numerator and denominator inaccurate.

**Suggested fix:**

Preserve merge-base semantics for non-PR runs by fetching enough history and using `origin/main...HEAD`, or explicitly restrict diff coverage to runs with a PR base SHA rather than silently producing an over-broad report.

qcom/build_and_test.py
```diff
@@ -799,11 +800,20 @@ def create_venv(self, plan: Plan) -> str:
         @depends(["coverage_linux_x86_64"])
         def diff_coverage_linux_x86_64(self, plan: Plan) -> str:
             build_dir = REPO_ROOT / "build" / "linux-x86_64"
+            base_commit = None
+            if github_event_path := os.environ.get("GITHUB_EVENT_PATH"):
+                try:
+                    github_event = json.loads(Path(github_event_path).read_text())
+                    base_commit = github_event.get("pull_request", {}).get("base", {}).get("sha")
+                except (OSError, json.JSONDecodeError) as error:
+                    logging.warning(f"Unable to read the GitHub event payload: {error}")
```
kuanyul-qti: **[N-1] (Minor)** Logging call uses eager f-string formatting

This warning formats the error string before calling `logging.warning`. The project logging convention uses the logger's lazy formatting so the message is not constructed when the level is disabled.

**Suggested fix:**

Use `logging.warning("Unable to read the GitHub event payload: %s", error)`.

qcom/ep_build/tasks/build.py
huaychou: Fixed

qcom/ep_build/tasks/build.py
```diff
@@ -240,12 +253,12 @@ def __init__(
...
                 RunExecutablesTask(
-                    "Fetching base branch",
-                    [["git", "fetch", "origin", base_branch.removeprefix("origin/")]],
+                    "Fetching diff base",
+                    [fetch_command],
                 ),
                 RunExecutablesTask(
                     "Generating git diff",
-                    [[BASH_EXECUTABLE, "-c", f"git diff {base_branch}...HEAD > {diff_file}"]],
+                    [[BASH_EXECUTABLE, "-c", f"git diff {diff_base}{diff_separator} HEAD > {diff_file}"]],
```
huaychou: **[M-1] (Major)** The local diff-coverage fallback still builds an invalid git diff command

When `base_commit` is absent, the code sets `diff_separator` to `"..."`, but the format string retains a space before `HEAD`. It therefore expands to `git diff origin/main... HEAD`. Git requires the three-dot revision to be one argument, `origin/main...HEAD`; executing the current command exits with usage error 129.

**Impact:** The CI `base_commit` path works, but a local `diff_coverage_linux_x86_64` run has no GitHub event payload, takes this fallback, and deterministically fails before generating the patch.

**Suggested fix:**

Build the complete revision explicitly, for example `diff_range = base_commit if base_commit else f"{base_branch}...HEAD"`, then run `git diff {diff_range}`. Alternatively, construct two complete command lists instead of assembling revision syntax with a separator.

qcom/scripts/linux/cleanup_runner_workspaces.sh
```diff
@@ -0,0 +1,84 @@
...
+    exit 1
+fi
+
+exec 9>"${lock_file}"
+if ! flock --nonblock 9; then
+    log "Another workspace cleanup is already running; skipping."
+    exit 0
+fi
+
+if pgrep -u "${runner_user}" -f '/bin/Runner.Worker' >/dev/null; then
```
huaychou: **[M-2] (Major)** A one-time idle check does not prevent cleanup from racing a newly started job

The script checks for `Runner.Worker` only once at line 37 and then scans and removes workspaces with `rm -rf`. The lock is private to cleanup processes; the GitHub runner never acquires it. If a runner receives a new job after `pgrep` observes the idle state, cleanup continues deleting directories, and the emergency branch removes every non-underscore repository workspace.

**Impact:** A timer/job-dispatch race can delete an active checkout during build or test, causing intermittent CI failures and potentially removing job artifacts or logs before upload.

**Suggested fix:**

Coordinate cleanup with the runner job lifecycle through a shared exclusion mechanism, such as running cleanup from GitHub runner job hooks after job completion or using a lock that both the runner service and cleanup service hold. Rechecking the worker before every deletion reduces the window but does not fully eliminate the TOCTOU race.

qcom/scripts/linux/ort-runner-workspace-cleanup.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Clean stale GitHub Actions runner workspaces
+After=local-fs.target
+
+[Service]
```
huaychou: **[M-1] (Major)** A root-owned lock file will make the runner job hooks fail

The systemd unit has no `User=`, so cleanup runs as root and creates `${runner_root}/.workspace-cleanup.lock` through `exec 9>`. The resulting file is normally root-owned with mode 0644. The job-started and job-completed hooks later open the same file with `9>` while running as `ortqnnepci`; that user cannot write the root-owned file, and `set -e` terminates the hook immediately.

**Impact:** After the timer runs cleanup once, subsequent GitHub Actions jobs can fail in the started hook before any job step executes, while the completed hook can also fail to remove the active marker.

**Suggested fix:**

Run the cleanup service under the same account as the runner hooks, for example by adding `User=ortqnnepci` after confirming that account owns the workspaces. Alternatively, pre-create the lock with explicit permissions that both processes can write. Add a test that opens the same lock first as the service identity and then as the runner identity.

PR #848, opened 2026-09-18
Disable file mapping feature by default

quic-calvnguy:
### Description
Disable file mapping feature by default. Previously was enabled by default.

### Motivation and Context
This is a quick fix to unblock the release of 10D. This will be reverted at the release of 11D.

The core issue is a race condition in QAIRT SDK code that is exposed when context creation fails due to file mapping failure.


qti-chuteng:
Would it cause some regression for some QA testcase which enable file mapping successfully?

chunghow-qti:
should it merge to rel-2.6.0? cc @qti-mbadnara

tirupath-qti (approved):

